PR #482, opened 2026-04-18
fix(providers): switch DeepSeekProvider from deepseek SDK to OpenAI c…

liling:
Their's no 'deepseek.Client' which lead to failure in deepseek API.

DeepSeek API is compatible with OpenAI, use the openai SDK instead of the unmaintained deepseek SDK for better compatibility. 

## Type of Change

## Testing

- [x] Tested locally


KaifAhmad1 (approved):
## PR #482 Review — `fix(providers): switch DeepSeekProvider from deepseek SDK to OpenAI client`

**Author:** @liling | **Reviewer:** @KaifAhmad1 | **Status:** Changes required

---

### Overview

Fixes a real runtime crash: the `deepseek` PyPI package has no `deepseek.Client`, making `DeepSeekProvider` completely unusable. The proposed fix switches to the `openai` SDK against DeepSeek's OpenAI-compatible endpoint — the same pattern used by `NovitaProvider`. A second commit adds a missing `verbose_mode` assignment that would cause a `NameError`.

---

### Issues Found

#### Critical

**`AttributeError` on every instantiation**

`_init_client` now uses `self.base_url`, but `DeepSeekProvider.__init__` never sets it:

```python
# _init_client (PR change)
self.client = OpenAI(api_key=self.api_key, base_url=self.base_url)  # AttributeError
```

Fix — add to `__init__` before `_init_client()` call:

```python
self.base_url = "https://api.deepseek.com/v1"
```

The `/v1` suffix is required; the OpenAI SDK appends `/chat/completions` to `base_url` as-is, so omitting it would cause 404s on every API call. Compare `NovitaProvider.__init__` which correctly sets `self.base_url = "https://api.novita.ai/v1"`.

---

#### Minor

- `pyproject.toml` still declares `llm-deepseek = ["deepseek>=0.1.0"]` — change to `openai>=1.0.0`
- `_init_client` warning still says "deepseek library not installed" — update to reference `openai`
- `verbose_mode` assigned inside `if client:` block — move to top of `generate_typed` for full coverage

---

### What's Good

- Correct approach — DeepSeek's API is fully OpenAI-compatible; this is the right SDK to use.
- `generate` and `generate_structured` both use `client.chat.completions.create`, which works unchanged with the OpenAI SDK.
- After the fix, the `isinstance(self.client, OpenAI)` check in `generate_typed` will pass correctly, reusing the existing client rather than creating a redundant one.

---

### Fixes Applied (branch `test/pr-482`)

All issues above are resolved in the follow-up commit `85cf96f`:

- `self.base_url = "https://api.deepseek.com/v1"` added to `__init__`
- `pyproject.toml` updated: `llm-deepseek = ["openai>=1.0.0"]`
- Warning message updated
- 19 tests added in `tests/semantic_extract/test_pr482_deepseek_openai.py` — all passing

---

### Verdict

**Not mergeable as submitted.** The `self.base_url` bug replaces one crash with another. All fixes are ready on `test/pr-482`.

KaifAhmad1 (approved):
## PR #482 — `fix(providers): switch DeepSeekProvider from deepseek SDK to OpenAI client`

**Author:** @liling &nbsp;|&nbsp; **Reviewer:** @KaifAhmad1 &nbsp;|&nbsp; **Branch:** `test/pr-482`

---

### Overview

The `deepseek` PyPI package exposes no `deepseek.Client`, making `DeepSeekProvider` crash on every instantiation. The fix switches to the `openai` SDK against DeepSeek's OpenAI-compatible endpoint — the same pattern used by `NovitaProvider`. A second commit fixes a `NameError` for `verbose_mode` in `generate_typed`.

---

### Issues Found

#### Critical

- **`AttributeError` on instantiation** — `_init_client` references `self.base_url` but `__init__` never sets it. Every `DeepSeekProvider(...)` call raises `AttributeError` before even reaching the network. Fixed by adding `self.base_url = "https://api.deepseek.com/v1"` to `__init__`. The `/v1` suffix is required — the OpenAI SDK appends paths like `/chat/completions` to `base_url` as-is, so omitting it causes 404s on every call.

#### Minor

- **`pyproject.toml` stale** — `llm-deepseek = ["deepseek>=0.1.0"]` still points to the defunct package. Updated to `openai>=1.0.0`.
- **Warning message stale** — `_init_client` still warns "deepseek library not installed". Updated to reference `openai`.
- **`verbose_mode` scope** — assigned immediately before use inside `if client:` block rather than at the top of `generate_typed`. Works for the instructor path but fragile if ever referenced in the fallback loop below.

---

### What's Good

- Correct approach — DeepSeek's API is fully OpenAI-compatible; using `openai` SDK is the right call.
- `generate` and `generate_structured` use `client.chat.completions.create` which works unchanged.
- After the fix, the `isinstance(self.client, OpenAI)` check in `generate_typed` passes correctly, reusing the existing client instead of constructing a redundant one.

---

### Fixes Applied

All issues resolved in follow-up commit `983f530` on liling's branch:

- `self.base_url = "https://api.deepseek.com/v1"` added to `__init__`
- `pyproject.toml`: `llm-deepseek = ["openai>=1.0.0"]`
- `_init_client` warning updated to reference `openai`
- 19 tests added in `tests/semantic_extract/test_pr482_deepseek_openai.py` — **all passing**

---

### Verdict

**Approved with fixes applied.** Original PR had the right idea but introduced a second crash (`AttributeError` on `self.base_url`). All issues resolved in the follow-up commit. Safe to merge.

KaifAhmad1:
@liling Thanks for the contribution—really appreciate the fix 🙌
Great catch, and contributions like yours are always welcome!

PR #486, opened 2026-04-21
Fix explorer zooming and Loading Flicker

ZohaibHassan16:
## Overview

This PR addresses two primary UX issues within the Graph Workspace: inconsistent zooming behavior and visual "blinking" during the graph loading and layout stabilization phases. It also restores live layout motion for derived graphs to ensure a smoother transition between view modes.


KaifAhmad1 (changes requested):
## PR #486 — Fix Explorer Zooming and Loading Flicker

**Author:** @ZohaibHassan16 · **Reviewer:** @KaifAhmad1  

---

### Test Results

- ✅ TypeScript — zero errors
- ✅ ESLint — net +1 warning (3 new, 2 fixed — details below)
- ✅ JS unit tests — 4 / 4 passed (`graphStore`, `graphSceneState`)
- ✅ Python suite — 1,141 passed, 9 skipped (2 pre-existing mock failures, unrelated)

---

### What Changed

- **Zoom fix** — toolbar buttons were calling `sigma.getCamera()` directly via a type-cast, breaking silently when `sceneRef` was null. Now properly routed through new `zoomIn()` / `zoomOut()` methods on `GraphCanvasHandle`.
- **Flicker fix** — Sigma was initialized on mount before data arrived, flashing an empty canvas. Now gated behind `graphReady`; Sigma only creates after `onGraphReady` fires.
- **Live layout motion** — display graphs now observe FA2 position updates via a per-frame RAF sync loop (`syncDisplayNodePositionsFromStore`), making layout animation visible in all view modes.
- **Camera fit** — replaced hardcoded `ratio: 0.3` / `animatedReset()` with real bounding-box computation (`expandGraphSpaceBounds` → `computeCameraTargetFromBounds`).
- **Effect suppression** — heavy effects (path pulse, lens, contours, etc.) are now disabled while layout is running to prevent visual noise.

---

### Must Fix Before Merge

#### 1. Plugin conflict with merged PR #489

`plugin.json` re-adds `"hooks": "./hooks/hooks.json"` — a field that PR #489 (`fix/local-plugin-install`, already on `main`) **intentionally removed** because it is auto-loaded.

```diff
# Remove this line — keep "agents", drop "hooks"
- "hooks": "./hooks/hooks.json"
```

---

#### 2. Louvain runs twice per render

`resolveDisplayStateSnapshot` calls `computeGraphAnalyticsBase({ computeCommunities: true })`. `resolveDisplayGraph` calls `resolveDisplayStateSnapshot` internally. Both are in separate `useMemo` calls in `GraphWorkspace` — so Louvain community detection fires **twice every render**.

**Fix:** cache `groupedViewAvailable` as its own memo, or pass the pre-computed state into `resolveDisplayGraph` so it doesn't re-derive it.

---

#### 3. `graphVersion` triggers Louvain on every WebSocket update

`graphVersion` is a dep of the `displayState` memo. Every live `ADD_NODE` / `ADD_EDGE` event increments it, re-running `resolveDisplayStateSnapshot` → Louvain. The display state only cares about selection and path fields — graph topology changes don't affect it.

**Fix:** remove `graphVersion` from the `displayState` memo deps. Compute `groupedViewAvailable` separately with its own narrower dep set.

---

#### 4. Undocumented behavioral change

`hideLabelsOnMove` and `hideEdgesOnMove` were flipped from `false` to `true`. Labels and edges now disappear while panning — a visible UX change not mentioned in the PR description.

**Fix:** add a note to the PR description confirming this is intentional.

---

### Warnings

- **3 new ESLint warnings** — `graphVersion` flagged as an unnecessary dep in three `useMemo` calls in `GraphWorkspace.tsx` (lines 124, 1049, 1064). The linter is correct — it's not consumed inside the memo body, only used as a cache-bust signal. Resolving issues 2 and 3 above will eliminate these automatically.
- **`resolveGroupedSelectionNodeId` is O(n)** — iterates all nodes on every selection to find a community anchor. Acceptable now, but worth replacing with a pre-built Map for large grouped graphs.
- **No new tests** — `searchFocusBehavior` (dispatch changed from `"focusNode"` to `"centerSelection"`), `syncDisplayNodePositionsFromStore`, and the camera-fit utilities are all untested. Add coverage before this area is touched again.

---

### What's Good

- Ref-based stable callbacks (`onNodeClickRef`, etc.) correctly prevent stale closures without extra effect re-runs
- `EMPTY_PATH` constants fix the inline `?? []` new-reference warnings cleanly
- `disposed` flag correctly threads through all RAF cleanup paths — no leaks
- `debugGraphRuntime` guards on `import.meta.env.DEV` — zero production overhead
- `syncDisplayNodePositionsFromStore` epsilon check (`<= 0.001`) avoids unnecessary graph attribute writes
- Sigma lifecycle split into creation (gated on `graphReady`) and teardown (gated on `behaviors`) is clean

---

### Verdict

**Request changes.** The core fixes are solid and the architecture is sound. Fix the four items above — especially the plugin conflict (hard block) and the double Louvain call — then this is ready to merge.

KaifAhmad1 (approved):
## PR #486 — Fix Explorer Zooming and Loading Flicker

**Author:** @ZohaibHassan16 · **Reviewer:** @KaifAhmad1 · **Status:** Approved ✅

---

### What Changed

- `GraphCanvas.tsx` — zoom routed through new `zoomIn()`/`zoomOut()` methods on `GraphCanvasHandle`; Sigma init gated behind `graphReady`; `hideLabelsOnMove` / `hideEdgesOnMove` set to `true` (intentional, reduces visual noise during pan)
- `GraphWorkspace.tsx` — `groupedViewAvailable` extracted as its own memo; `graphVersion` removed from `displayState` deps; conflict with main resolved
- `graphSceneState.ts` — `groupedViewAvailable?` option added to `resolveDisplayStateSnapshot` and `resolveDisplayGraph` to accept pre-computed value and skip internal Louvain call
- `SigmaSceneAdapter.tsx` — RAF sync loop (`syncDisplayNodePositionsFromStore`) added for live layout motion in all view modes
- `behaviors/` — camera fit replaced with bounding-box computation; `searchFocusBehavior` dispatch updated; heavy effects suppressed while layout runs
- `plugin.json` — `hooks` field removed (auto-loaded, removed in #489); `agents` kept
- 12 files changed, +1,198 / -246

---

### Blockers — All Resolved

**1. Plugin conflict with main (#489) ✅**  
Branch re-added `"hooks": "./hooks/hooks.json"` which #489 intentionally removed.
```diff
- "hooks": "./hooks/hooks.json"
  "agents": "./agents"
```

**2. Double Louvain per render ✅**  
`resolveDisplayStateSnapshot` and `resolveDisplayGraph` both independently called `computeGraphAnalyticsBase`. Fixed by adding a single `groupedViewAvailable` memo in `GraphWorkspace` and threading the result into both functions via a new `groupedViewAvailable?` option — internal Louvain is skipped when the value is pre-supplied.
```typescript
const groupedViewAvailable = useMemo(
  () => computeGraphAnalyticsBase(graph, { computeCommunities: true, computeCentrality: false }).communitiesByNode.size > 0,
  [graphVersion],
);
```

**3. graphVersion triggering Louvain on every WebSocket event ✅**  
`graphVersion` removed from `displayState` memo deps. `displayState` now depends on the stable `groupedViewAvailable` boolean, so Louvain no longer re-fires on every ADD_NODE / ADD_EDGE tick.

**4. Undocumented behavioral change ✅**  
`hideLabelsOnMove` / `hideEdgesOnMove` flipped from `false` → `true`. Confirmed intentional — suppressing labels and edges during pan is part of the flicker fix and reduces visual noise during layout motion.

**5. Merge conflict with main ✅**  
`focusNode`'s conditional `setIsLayoutRunning(false)` (only when `viewMode === "focused"`) was preserved over main's unconditional version — keeping live layout motion in non-focused modes, the core intent of this PR.

---

### What's Good

- Ref-based stable callbacks (`onNodeClickRef`, etc.) correctly prevent stale closures without extra effect re-runs
- `disposed` flag threads through all RAF cleanup paths — no leaks
- `syncDisplayNodePositionsFromStore` epsilon check (`<= 0.001`) avoids unnecessary graph attribute writes
- Sigma lifecycle cleanly split: creation gated on `graphReady`, teardown gated on behaviors
- `EMPTY_PATH` constants fix inline `?? []` new-reference warnings
- `debugGraphRuntime` guarded behind `import.meta.env.DEV` — zero production overhead

---

### Warnings (pre-existing, no action required)

- `resolveGroupedSelectionNodeId` is O(n) on every selection — acceptable now, worth a `Map` pre-build for large grouped graphs later
- No new tests for `syncDisplayNodePositionsFromStore`, camera-fit utilities, or `searchFocusBehavior` dispatch change — add coverage before this area is touched again

---

**Verdict:** All blockers resolved, conflict cleared. Ready to merge.

PR #487, opened 2026-04-22
feat(explorer): stabilize local graph interaction

Sameer6305:
## Summary

Stabilizes graph interaction behavior in the Graph Explorer by separating display-level grouped selection from explicit focused-mode resolution.

This removes implicit grouped → focused transitions and aligns interaction behavior with the intended UX model:

* grouped/community selection remains display-level while preserving global context
* explicit Focused activation performs canonical anchor/base-node resolution
* unresolved focused activation fails gracefully without unstable mode transitions

### Interaction behavior

* removed implicit transition into `focused` mode during grouped/community selection
* preserved current view mode during normal grouped selection
* kept focused/local view as an explicit user action

### Focused-mode flow

* moved grouped/community anchor resolution out of normal selection flow
* added explicit focused activation resolution flow
* focused activation now:

  * resolves grouped/community selections into canonical anchor/base nodes
  * enters focused mode only after successful resolution
  * gracefully no-ops if resolution fails

### Stability improvements

* added safeguards for inspector behavior when selected ids are display-level/grouped ids rather than base-graph node ids
* aligned Focused button enablement with actual resolvability conditions
* preserved stable UI behavior for unresolved grouped selections

## UX alignment

This implementation follows the intended interaction model discussed in review:

* grouped selection remains display-level
* focused mode is explicit and user-driven
* grouped selections do not implicitly switch interaction regimes
* unresolved focused transitions remain stable and predictable

## Validation

Manually validated:

* grouped node selection remains in grouped mode
* explicit Focused activation resolves into canonical node context
* unresolved focused activation remains stable
* normal node selection + focused flow still works
* build passes successfully

```bash
npm run build
```


ZohaibHassan16:
Thanks Sameer @Sameer6305 , this direction looks right to me. 

I’d just address the two Copilot comments before merge:

1. Keep the Focused button from becoming active-but-disabled:
disabled: viewMode !== "focused" && !canActivateFocusedMode

2. Make the inspector fallback copy more general, since not every missing base node is necessarily a grouped item:
"Selected item is not available for inspection in the current graph."

Other than that, this looks aligned with the intended interaction model. @KaifAhmad1 can merge it after the above two blockers are removed.

KaifAhmad1 (changes requested):
## PR #487 — Stabilize Local Graph Interaction

**Author:** @Sameer6305  · **Reviewer:** @KaifAhmad1 

---

### Test Results

- ✅ TypeScript — zero errors@warning)
- ✅ JS unit tests — 4 / 4 passed
- ✅ Python — unaffected, no Python files changed

---

### What Changed

- Removed implicit `grouped → focused` transition — grouped node selection no longer silently switches view mode
- Extracted `resolveNodeIdForFocusedMode` — canonical anchor resolution reused by both `activateFocusedMode` and `canActivateFocusedMode`
- `activateFocusedMode` gracefully no-ops if resolution fails instead of entering focused mode with an unresolvable node
- `canActivateFocusedMode` memo drives the Focused toolbar button — replaces the old blunt `!selectedNodeId` check
- Inspector safeguard — `GraphInspectorPanel` shows an early message when `nodeId` is not in the base graph
- Plugin `setViewMode("focused")` now routes through `activateFocusedMode` so plugins respect the same resolution flow

---

### Must Fix Before Merge

#### 1. Focused button can be active-but-disabled

When `viewMode === "focused"` but the node is no longer resolvable after a graph reload, the button ends up both active and disabled — a broken state.

> *"Keep the Focused button from becoming active-but-disabled."*
> — @ZohaibHassan16

```diff
- disabled: !canActivateFocusedMode,
+ disabled: viewMode !== "focused" && !canActivateFocusedMode,
```

---

#### 2. Inspector copy is too specific

`!graph.hasNode(nodeId)` catches stale/invalid IDs too, not just grouped nodes. Telling the user to "Activate Focused mode" when resolution will silently fail is misleading.

> *"Make the inspector fallback copy more general, since not every missing base node is necessarily a grouped item."*
> — @ZohaibHassan16

```diff
- Selected grouped item is not directly inspectable. Activate Focused mode to resolve to a canonical node.
+ Selected item is not available for inspection in the current graph.
```

---

#### 3. Ref accessed during render — 2 new ESLint errors

`canActivateFocusedMode` is a `useMemo` that calls `resolveNodeIdForFocusedMode`, which reads `pluginRuntimeRef.current` internally. ESLint correctly flags ref reads during render.

- `GraphWorkspace.tsx:866` — `Cannot access refs during render`
- `GraphWorkspace.tsx:867` — `pluginRuntimeVersion` flagged as unnecessary dep (symptom of the same issue)
- `GraphWorkspace.tsx:1534` — toolbar memo missing `activateFocusedMode` and `canActivateFocusedMode` in deps

**Fix:** move the `pluginRuntimeRef.current` read out of the memo — either convert `canActivateFocusedMode` to state updated in an effect, or read the ref value before passing it into `resolveNodeIdForFocusedMode`. The dep and missing-dep warnings resolve themselves once this is fixed.

---

### What's Good

- `resolveNodeIdForFocusedMode` cleanly separates resolution logic from selection logic — right call
- `focusNode` is now simpler and mode-agnostic after the cleanup
- Plugin bypass for focused mode is closed — plugins go through the same resolution flow now
- `.gitignore` addition for `demo_out/` is fine

---

### Verdict

**Request changes.** The interaction model is correct and well-scoped. Fix the three items above — two already agreed on by Zohaib, one new ESLint regression — then this is ready to merge.

---

KaifAhmad1 (approved):
## PR #487 — feat(explorer): stabilize local graph interaction

**Author:** @Sameer6305 · **Reviewer:** @KaifAhmad1 · **Status:** Approved ✅

---

### What Changed

- `GraphWorkspace.tsx` — removes implicit grouped → focused transition; adds `resolveNodeIdForFocusedMode`, `activateFocusedMode`, and `canActivateFocusedMode`
- `GraphInspectorPanel.tsx` — early-return guard when `nodeId` is absent from the base graph
- `.gitignore` — ignores `demo_out/`

Core model: grouped selection stays display-level. Focused mode is an explicit, resolvable user action that no-ops gracefully on failure.

---

### Blockers — All Resolved

**1. Focused button active-but-disabled ✅**  
When `viewMode === "focused"` but the node is no longer resolvable, the button would be both active and disabled. Fix: only disable when not already in focused mode.
```diff
- disabled: !canActivateFocusedMode,
+ disabled: viewMode !== "focused" && !canActivateFocusedMode,
```

**2. Inspector copy too specific ✅**  
`!graph.hasNode(nodeId)` catches stale/invalid IDs too, not just grouped nodes. Old copy told users to "Activate Focused mode" even when resolution would silently fail.
```diff
- Selected grouped item is not directly inspectable. Activate Focused mode to resolve to a canonical node.
+ Selected item is not available for inspection in the current graph.
```

**3. Ref read during render — ESLint errors ✅**  
`canActivateFocusedMode` was a `useMemo` calling `resolveNodeIdForFocusedMode`, which reads `pluginRuntimeRef.current` internally — a ref read during render. Fix: convert to `useState` + `useEffect` so the ref is read in an effect. Also added `activateFocusedMode` and `canActivateFocusedMode` to the toolbar memo deps.
```diff
- const canActivateFocusedMode = useMemo(
-   () => Boolean(resolveNodeIdForFocusedMode(selectedNodeId)),
-   [pluginRuntimeVersion, resolveNodeIdForFocusedMode, selectedNodeId],
- );
+ const [canActivateFocusedMode, setCanActivateFocusedMode] = useState(false);
+ useEffect(() => {
+   setCanActivateFocusedMode(Boolean(resolveNodeIdForFocusedMode(selectedNodeId)));
+ }, [pluginRuntimeVersion, resolveNodeIdForFocusedMode, selectedNodeId]);
```

---

### What's Good

- `resolveNodeIdForFocusedMode` cleanly separates resolution from selection
- `activateFocusedMode` no-ops gracefully instead of entering a broken state
- Plugin `setViewMode("focused")` now routes through `activateFocusedMode` — bypass closed
- `focusNode` is simpler and mode-agnostic post-cleanup

---

**Verdict:** All blockers resolved. Ready to merge.

PR #493, opened 2026-04-24
Feat/explorer grouped view

ZohaibHassan16:
## Summary
Upgrades Grouped View into a stable, display-level community map.

- Grouped selection stays in Grouped View  
- Focused mode is now an explicit drill-in  
- No more crashes or blank screens on grouped swaps  
- Layout is calmer and easier to read  

### Interaction
- Separated grouped (display-level) selection from Focused mode  
- Clicking communities no longer switches views implicitly  
- Focused mode resolves to canonical nodes only when possible  
- Added grouped-safe selection state + focused availability  

### Stability
- Fixed Sigma crashes from synthetic `__community__:*` ids  
- Removed base-graph assumptions from grouped update paths  
- Made grouped state derivation safe for synthetic nodes/edges  
- Added guards to prevent invalid grouped graphs from rendering  

### Camera
- Decoupled grouped selection from raw-node centering  
- Fixed centering to use displayed positions (not raw coords)  
- Prevented tight-fit jumps on community clicks  
- Fit View remains the way to reset the full map  

### Visuals
- Reworked layout into a clearer community meta-map  
- Deterministic concentric seeding instead of anchor inheritance  
- Added spacing to avoid overlap at spawn  
- Short stabilization pass (no constant motion)  
- Softer edges, more selective labels  
- Nodes read as landmarks, not blobs  

## Result
Grouped View is now predictable and readable:
- stable on entry and interaction  
- preserves context until explicitly focusing  
- visually cleaner and easier to scan  

## Validation
- `npx tsc -b` passes  

## Manual Checks
- Enter Grouped View from a loaded graph  
- Toggle between Full Graph and Grouped View repeatedly  
- Click community nodes (no view switch, no blanking)  
- Use Fit View in Grouped mode  
- Enter Focused mode from a valid grouped selection


KaifAhmad1 (approved):
## PR #493: Feat/Explorer Grouped View

**Reviewed by:** @KaifAhmad1
**PR Author:** @ZohaibHassan16

---

### Overview

Graduates Grouped View into a stable display-level community map. The core design decision — separating display-level community selection from Focused mode drill-in — is correct and directly fixes the root cause of the implicit view-switch crashes. `npx tsc -b` clean on submission.

---

### Test Results

- `test:graph-store` — 1/1 pass (before and after)
- `test:graph-workspace` — **0/3 on submission** (all crash at module load), **10/10 after fix commit**

---

### What's Good

- **`focusedNodeId` / `selectedNodeId` split** — the right fix for implicit view-switch crashes; two pieces of state that were incorrectly conflated
- **`requestViewMode()`** — centralises all view transitions and guards in one place
- **`resolveGroupedDisplayStateSnapshot()`** — maps base-graph `selectedNodeId` to its community representative before building display state; no dangling `selectedRootNodeId` references
- **`validateGroupedDisplayGraph()`** — guards coordinates and edge references before any grouped graph reaches Sigma; prevents the crash class this PR targets
- **`assignGroupedCommunityPositions()`** — deterministic concentric seeding (prominence-ranked rings, hash-seeded angle offset) eliminates the collapse-then-explode on first grouped view entry
- **`createInteractionState()`** — now takes `focusedNodeId` as a distinct parameter; behaviors consuming `interactionState.focusedNodeId` see the correct value
- Camera centering uses Sigma's displayed positions, not raw store coords — correct for synthetic community nodes

---

### Issues Found & Fixed in `6ad1502`

#### 1. `import.meta.env.DEV` crashes the test suite — Blocker

`graphSceneState.ts:38` — added by this PR, did not exist on `main`:

```ts
// Before — TypeError at module load in Node.js/tsx
const DEBUG_GRAPH_SCENE_STATE = import.meta.env.DEV;

// After
const DEBUG_GRAPH_SCENE_STATE = typeof import.meta !== "undefined" && import.meta.env?.DEV === true;
```

`import.meta.env` is Vite-only. In the Node.js/tsx test runner it is `undefined`, so `.DEV` throws before any test runs — including the grouped view test added by this PR itself.

---

#### 2. Community detection fires on every graph change — High

`groupedDisplayCandidate` was running Louvain + betweenness centrality (50–200 ms on mid-size graphs) on every `graphVersion` tick, even when the user was in Full or Focused view.

Fixed by splitting into two:
- **Cheap availability check** — `checkGroupedViewAvailability()` runs Louvain only (no centrality) on every graph change
- **Full community graph** — gated on `viewMode === "grouped"`, built only when the user is actually in grouped view

---

#### 3. `resolveGroupedDisplayNodeId` duplicated — Medium

Identical 40-line function existed in both `graphSceneState.ts` and `GraphWorkspace.tsx`. Now exported from `graphSceneState.ts` and the copy in `GraphWorkspace.tsx` is removed.

---

#### 4. Dead ternary in `focusNode` — Low

```ts
// Before — both branches return nodeId, condition does nothing
const nextSelectedNodeId = viewMode === "focused" && graph.hasNode(nodeId) ? nodeId : nodeId;

// After
const nextSelectedNodeId = nodeId;
```

---

#### 5. No tests for new logic — Medium

`resolveGroupedDisplayStateSnapshot`, `validateGroupedDisplayGraph`, `assignGroupedCommunityPositions`, and `FocusResolution` paths had no coverage. Fix commit adds 7 new tests:

- `resolveGroupedDisplayNodeId` — null input, direct match, base→community resolution
- `resolveGroupedDisplayStateSnapshot` — empty selection, base node mapped to community
- `checkGroupedViewAvailability` — empty graph returns unavailable, community graph returns available

---

### Remaining Suggestions (non-blocking)

- **FA2 re-entry** — layout restarts from settled positions on every grouped view re-entry for the same graph; consider skipping the FA2 pass when the community graph hasn't changed
- **Inspector props** — four new props (`inspectableNodeId`, `selectedNodeKind`, `canActivateFocused`, `focusedUnavailableReason`) model one concept; a single `inspectionContext` object would be cleaner to extend
- **Synthetic ID collision** — add a dev-only assertion that base-graph node IDs don't start with `__community__:` or `__agg__:` so a bad data source fails loudly

---

### Verdict

Architecture is sound, stability fixes are genuine, and the visual improvements are clear. With fix commit `6ad1502` included (co-authored by @KaifAhmad1 and @ZohaibHassan16), this is ready to merge.

**Approve — with fix commit included.**

PR #495, opened 2026-04-25
fix(explorer): stabilize fit view camera

ZohaibHassan16:
## Summary
Fixes a bug where clicking `Fit View` could push the graph off-screen, even though it was still loaded.

The root cause was that the fit calculation depended on the current camera state (pan/zoom). This sometimes produced incorrect zoom values, making the graph appear to disappear.

### Stable fit calculation
- Replaced the camera-dependent zoom calculation with a stable reference-based approach  
- Fit is now computed using a fixed reference snapshot instead of the current `viewRectangle()`  
- The target zoom is derived from:
  - the expanded bounds of the current graph
  - a stable visible area based on the container  
- The computed camera state is validated through Sigma before applying

### Safer fit behavior
- `Fit View` still fits the full visible graph for the current mode  
- Selection behaviors remain unchanged:
  - full selection centers context
  - grouped selection keeps current zoom feel
  - focused selection keeps current zoom feel  
- Removed the hidden custom bounding-box reset to keep calculations consistent

### Diagnostics
- Added debug logs for:
  - reference bounds
  - target bounds
  - computed zoom ratio
  - clamped/validated ratio
  - custom bounding-box usage

## Result
`Fit View` now works reliably in:
- Full Graph
- Grouped View
- Focused mode

It consistently restores the visible graph without losing it off-screen.

## Validation
- `npx tsc -b` passes

## Manual Checks
- Use `Fit View` after panning/zooming in Full Graph  
- Use `Fit View` after panning/zooming in Grouped View  
- Use `Fit View` after panning/zooming in Focused mode  
- Select nodes and confirm `Fit View` restores the full graph  
- Switch between modes (`Full → Grouped → Focused`) and verify consistent behavior

### DIFF
--- a/explorer/src/workspaces/GraphWorkspace/GraphCanvas.tsx
+++ b/explorer/src/workspaces/GraphWorkspace/GraphCanvas.tsx
@@ -368,42 +368,131 @@ function computeDisplayedGraphBounds(
   return computeDisplayedNodeBounds(sigma, nodeIds);
 }
 
+type CameraTarget = {
+  x: number;
+  y: number;
+  ratio: number;
+  angle: number;
+};
+
+type CameraTargetComputation = {
+  target: CameraTarget;
+  diagnostics: Record<string, unknown>;
+};
+
+function computeStableReferenceViewBounds(
+  sigma: Sigma,
+  angle: number,
+): GraphSpaceBounds | null {
+  const dimensions = sigma.getDimensions();
+  const width = Number(dimensions.width);
+  const height = Number(dimensions.height);
+  if (!Number.isFinite(width) || !Number.isFinite(height) || width <= 0 || height <= 0) {
+    return null;
+  }
+
+  const referenceCameraState = {
+    x: 0.5,
+    y: 0.5,
+    ratio: 1,
+    angle,
+  };
+
+  const corners = [
+    sigma.viewportToFramedGraph({ x: 0, y: 0 }, { cameraState: referenceCameraState }),
+    sigma.viewportToFramedGraph({ x: width, y: 0 }, { cameraState: referenceCameraState }),
+    sigma.viewportToFramedGraph({ x: 0, y: height }, { cameraState: referenceCameraState }),
+    sigma.viewportToFramedGraph({ x: width, y: height }, { cameraState: referenceCameraState }),
+  ];
+
+  let minX = Infinity;
+  let maxX = -Infinity;
+  let minY = Infinity;
+  let maxY = -Infinity;
+  let count = 0;
+
+  corners.forEach((point) => {
+    const x = Number(point.x);
+    const y = Number(point.y);
+    if (!Number.isFinite(x) || !Number.isFinite(y)) {
+      return;
+    }
+
+    minX = Math.min(minX, x);
+    maxX = Math.max(maxX, x);
+    minY = Math.min(minY, y);
+    maxY = Math.max(maxY, y);
+    count += 1;
+  });
+
+  if (count === 0) {
+    return null;
+  }
+
+  return { minX, maxX, minY, maxY, count };
+}
+
 function computeCameraTargetFromBounds(
   sigma: Sigma,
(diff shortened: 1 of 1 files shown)


KaifAhmad1 (approved):
## PR Review — `fix(explorer): stabilize fit view camera`
- **Author:** [@ZohaibHassan16](https://github.com/ZohaibHassan16)
- **Reviewer:** [@KaifAhmad1](https://github.com/KaifAhmad1)
- **Status:** Approved — follow-up commit pushed

---

### Summary

Fixes a camera instability bug where clicking Fit View could push the graph off-screen. The root cause was `computeCameraTargetFromBounds` multiplying the **current** camera ratio by a scale factor derived from `viewRectangle()` — when the camera was already drifted, this compounded the error.

The fix introduces a stable canonical reference state (`x=0.5, y=0.5, ratio=1`) via the new `computeStableReferenceViewBounds()` function. Zoom is now computed independently of current camera state, making Fit View reliable across Full, Grouped, and Focused modes.

---

### Changes Reviewed

**File:** `explorer/src/workspaces/GraphWorkspace/GraphCanvas.tsx`

#### New Types

- `CameraTarget` — typed shape for `{ x, y, ratio, angle }`, replaces the inline return type
- `CameraTargetComputation` — wraps `target` with a `diagnostics` bag, cleanly separates data from debug info

#### `computeStableReferenceViewBounds`

- Projects all four viewport corners via `sigma.viewportToFramedGraph` at a fixed reference camera state
- Guards container `width`/`height` with `Number.isFinite` before use
- Skips any corner that produces a non-finite coordinate
- Returns `null` on degenerate input — callers handle gracefully

#### `computeCameraTargetFromBounds` — Refactored

- **Before:** used `sigma.viewRectangle()` (camera-dependent) and `cameraState.ratio * fitScale` — compounded drift
- **After:** derives bounds from the stable reference, passes `fitScale` directly to `camera.validateState()`
- `camera.validateState()` validates the full camera state as a unit, which is correct over the old `getBoundedRatio()` single-field clamp

#### Custom BBox Handling

- Removed the previous `sigma.setCustomBBox(null)` call that silently cleared custom bounding boxes on every fit
- Added a comment explaining the intent — clearing the bbox during fit corrupts the stable reference bounds that `computeStableReferenceViewBounds()` depends on; each mode-transition manages its own bbox lifecycle
- `customBBoxActive` is now surfaced in diagnostics instead of being silently reset

#### `centerFocusedSelectionInView` — New Callback

- Pans to center the focused node's context without changing zoom (`ratio` preserved from current camera state)
- Correct UX for focused mode — only pan, never zoom

#### Debug Logging

- All `debugGraphRuntime` calls are gated by `import.meta.env.DEV` at the definition site
- No diagnostic output in production builds — confirmed safe

---

### Follow-up Commit (`1e98028`)

After review, these fixes were pushed directly to the PR branch:

- **Redundant alias removed** — `unclampedRatio = fitScale` was dead code; `fitScale` now passed directly to `validateState`
- **Dead `??` fallbacks removed** — `validatedTarget.x ?? centerX` etc. are unreachable since `camera.validateState()` always returns a complete `CameraState`
- **`ratioClamped` tolerance** — changed from `1e-6` to `1e-4` to avoid floating-point rounding false positives in diagnostics
- **Comment added** — explains why `setCustomBBox(null)` was intentionally removed to prevent stable reference corruption in bounded modes

---

### Verdict

- **Core bug fix** — correct; stable reference approach fully decouples fit from camera drift
- **Guard conditions** — comprehensive; all degenerate paths return `null` and propagate cleanly
- **Mode behavior** — Full, Grouped, and Focused modes all handled correctly
- **Debug logs** — already gated to DEV; safe for production
- **Custom bbox handling** — explained and tracked via diagnostics
- **Type safety** — clean; no unsafe casts
- **Test coverage** — manual only; acceptable for camera behavior, but unit tests for `computeStableReferenceViewBounds` would reduce future regression risk

**Approved.** Fix is correct, well-guarded, and all review feedback addressed.

Issue #514, opened 2026-04-29
[BUG] Semantic Distance UI fails for node IDs containing slash


ZohaibHassan16:
## Summary

Semantic Distance mode fails for nodes whose IDs contain `/`, such as `gene/protein:6164`.

The frontend calls:

`/api/graph/node/gene%2Fprotein%3A6164/semantic-neighborhood?top_k=50`

But FastAPI logs:

`GET /api/graph/node/gene/protein%3A6164/semantic-neighborhood HTTP/1.1" 404`

The encoded slash `%2F` is decoded before route matching, so the route `/api/graph/node/{node_id}/semantic-neighborhood` does not match.

## Expected

Semantic mode should either:

- load semantic neighbors for the selected node, or
- show a graceful unavailable state if semantic embeddings are unavailable.

## Actual

The request returns `404 Not Found`, and the Explorer shows:

`Semantic distance failed with status 404`

PR #515, opened 2026-04-29
fix(explorer): make distance intelligence API calls slash-safe

ZohaibHassan16:
## Summary

Fixes Distance Intelligence UI failures for graph node IDs that contain slashes, such as `gene/protein:6164`.

The frontend now calls query-param based graph endpoints instead of embedding node IDs inside URL path segments, and the backend exposes slash-safe routes for semantic neighborhood and path lookup while keeping the old routes for backward compatibility.

## What Changed

- Added slash-safe backend routes:
  - `GET /api/graph/semantic-neighborhood?node_id=...`
  - `GET /api/graph/path?source=...&target=...`
- Kept existing path-style routes for backward compatibility.
- Updated Explorer Semantic Distance mode to use the new semantic query route.
- Updated Trace Path calls to use the new path query route.
- Made semantic neighborhood responses clearer:
  - `503` when semantic similarity is unavailable or graph embeddings are missing.
  - `404` only when the selected node truly does not exist.
  - `200` only when semantic search actually runs.
- Added tests for slash-containing node IDs and semantic unavailable behavior.
- Updated smoke checks for the new query-param frontend calls.

## Verification

- `npx tsc -b`
- Python syntax parse
- Direct backend smoke:
  - slash-safe semantic route works with embeddings
  - missing embeddings return `503`
  - slash-safe path route works

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- **Fix: Semantic Distance UI slash-safe node IDs** (issue #514, PR #515 by @ZohaibHassan16, review fixes by @KaifAhmad1):
+  - **Root cause** — FastAPI decodes `%2F` before route matching, so node IDs containing `/` (e.g. `gene/protein:6164`) split the path-segment route and return 404. The frontend encoded slashes correctly but they were decoded server-side before the router matched the pattern.
+  - Added slash-safe query-param routes: `GET /api/graph/semantic-neighborhood?node_id=...` and `GET /api/graph/path?source=...&target=...`. Legacy path-segment routes (`/node/{id}/semantic-neighborhood`, `/node/{id}/path`) are kept as deprecated backward-compatible aliases with docstrings documenting the limitation.
+  - Frontend (`GraphWorkspace.tsx`, `GraphWorkspaceShell.tsx`) now builds all distance API calls via `URLSearchParams` so node IDs with slashes or other special characters are never embedded in URL path segments.
+  - `_semantic_neighborhood_impl` now returns HTTP 503 (instead of a silent 200 with zero neighbors) when semantic similarity is unavailable or the graph has no node embeddings, and HTTP 404 only when the anchor node itself does not exist. Frontend error messages updated to distinguish the two cases.
+  - Fixed a pre-existing bug where `find_most_similar` was called with `(graph_dict, node_id_string)` instead of the correct `(embeddings_dict, query_vector)` signature; added `_extract_node_embeddings` and `_coerce_embedding_vector` helpers to build the embeddings dict before the call.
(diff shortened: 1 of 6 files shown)


KaifAhmad1 (approved):
## PR #515 Code Review

**Title:** `fix(explorer): make distance intelligence API calls slash-safe`
**Author:** @ZohaibHassan16 · **Reviewer:** @KaifAhmad1 · **Closes:** #514

---

### Root Cause

FastAPI decodes `%2F` before route matching. A correctly-encoded URL like
`/api/graph/node/gene%2Fprotein%3A6164/semantic-neighborhood` was matched as
`/api/graph/node/gene/protein%3A6164/semantic-neighborhood` — a route that does not exist — returning 404.

---

### What Changed

**Backend (`semantica/explorer/routes/graph.py`)**
- Added `GET /api/graph/semantic-neighborhood?node_id=...` — slash-safe semantic neighbor lookup
- Added `GET /api/graph/path?source=...&target=...` — slash-safe path lookup
- Legacy path-segment routes kept as deprecated aliases with docstrings documenting the slash limitation
- `_semantic_neighborhood_impl` extracted and shared by both old and new handlers
- Returns 503 when similarity is unavailable or embeddings are missing (was a silent 200 + empty list)
- Returns 404 only when the anchor node itself does not exist
- Fixed a pre-existing bug: `find_most_similar` was called with `(graph_dict, node_id_string)` — corrected to `(embeddings_dict, query_vector)` matching the actual `SimilarityCalculator` signature
- Added `_extract_node_embeddings` and `_coerce_embedding_vector` helpers to build the embeddings dict

**Frontend (`GraphWorkspace.tsx`, `GraphWorkspaceShell.tsx`)**
- Semantic neighborhood and Trace Path calls now use `URLSearchParams` so node IDs never appear in path segments
- Error messages distinguish 503 ("unavailable") from 404 ("node not found")

**Tests (`tests/explorer/test_explorer_api.py`)**
- 4 new tests: slash ID roundtrip, legacy route compat, missing-node 404, no-embeddings 503
- 1 new path route test with slash node IDs

**Smoke checks (`tests/_smoke_review_fixes.py`)**
- `test_bug006` / `test_qual002` updated to inspect `_find_path_impl` via `getattr` fallback
- `test_bug001` updated to accept the `URLSearchParams` string format for `top_k`

---

### Review Findings & Fixes

**1. Shared fixture mutation — Fixed (`bb956b2`)**
- `test_legacy_semantic_neighborhood_still_works_for_simple_ids` mutated the module-scoped `client` fixture by adding nodes and patching `_similarity` directly on the shared session, causing cross-test state pollution depending on execution order
- Rewrote as a fully isolated `TestClient` with its own `ContextGraph`

**2. `_FakeSimilarity` nested inside test class — Fixed (`bb956b2`)**
- Stub was invisible to other test classes and required copy-paste
- Extracted to module level alongside a `_make_slash_node_session()` helper

**3. Key list misalignment between helpers — Fixed (`bb956b2`)**
- `_coerce_embedding_vector` was missing `"embeddings"` in its inner dict probe, so nodes storing embeddings under that key would silently not resolve
- Added `"embeddings"` to the inner probe list, reordered both lists to generic-first, added cross-reference comments

**4. `_extract_node_embeddings` has no caching — Documented (`bb956b2`)**
- Iterates all nodes and copies all embedding vectors on every request — O(N) per call
- Added `TODO: cache per-session graph revision` comment; full caching deferred to a follow-up

**5. Legacy routes had no documentation of their slash limitation — Fixed (`bb956b2`)**
- Kept-for-compat routes gave no indication they would 404 for slash-containing IDs
- Added deprecation docstrings pointing to the new query-param routes

---

### Test Results

- `tests/explorer/` — 132 passed, 0 failed
- `tests/_smoke_review_fixes.py` — 18 passed, 1 pre-existing failure unrelated to this PR

---

### Verdict

**Approved.** All five findings resolved. Query-param routes are the correct fix for arbitrary node IDs, and the 503/404 split gives the UI the signal it needs for meaningful error messages.

PR #535, opened 2026-05-04
fix(ingest): lazy-load optional ingestion backends

### LINKED ISSUE #527: [BUG] Importing semantica.ingest fails if GitPython is not installed
## Summary

Importing `semantica.ingest` fails because `semantica/ingest/__init__.py` eagerly imports repository ingestion support, which requires the `git` module / GitPython.

This blocks unrelated use cases such as local file ingestion.

## Environment

- OS: Windows
- Python: 3.12
- Semantica version reported by pip: 0.0.5

## Reproduction

```python
from semantica.ingest import FileIngestor
```

## Actual Result

```text
ModuleNotFoundError: No module named 'git'
```

The failure occurs even though `FileIngestor` does not need repository ingestion.

## Expected Result

Users should be able to import:

```python
from semantica.ingest import FileIngestor
```

without installing GitPython.

## Suggested Fix

Make repository ingestion imports optional/lazy:

```python
try:
    from .repo_ingestor import RepoIngestor
except ImportError:
    RepoIngestor = None
```

or move repo-specific imports behind the relevant method call.

ZohaibHassan16:
## Summary
Fixes #527 by preventing `semantica.ingest` from eagerly importing optional ingestion backends such as repository, web, feed, email, stream, database, MCP, ontology, and Snowflake ingestion.

Core file ingestion now remains importable without GitPython, BeautifulSoup, or other backend-specific optional dependencies.

## Changes
- Keep core exports like `FileIngestor`, `ingest_file`, config, registry, and method helpers importable eagerly.
- Move optional backend exports behind lazy module-level `__getattr__` in `semantica.ingest`.
- Move optional backend imports in `methods.py` into their specific ingestion functions.
- Add clearer missing dependency errors when repository ingestion is actually used without GitPython.
- Add regression tests that block `git` and `bs4` in a subprocess.

## Verification
- `python -c "from semantica.ingest import FileIngestor; print(FileIngestor.__name__)"`
- `python -c "from semantica.ingest import ingest_file; print(callable(ingest_file))"`
- Subprocess regression checks passed for blocked `git`/`bs4`.

Closes #527


KaifAhmad1 (approved):
## Code Review — PR #535: `fix(ingest): lazy-load optional ingestion backends`


**Author:** @ZohaibHassan16
**Reviewer:** @KaifAhmad1

### Summary

Fixes issue #527: importing `semantica.ingest` failed with `ModuleNotFoundError: No module named 'git'` even when only `FileIngestor` was needed. The fix defers all optional-backend imports via a module-level `__getattr__` and moves per-function backend imports inside the function body with friendly error messages.

**Verdict: Approved — all fixes applied and tests passing.**

---

### What Changed

- `semantica/ingest/__init__.py` — core exports kept eager; optional backends moved behind `__getattr__`
- `semantica/ingest/methods.py` — backend imports moved into each function; `_missing_optional_dependency` / `_is_missing_dependency` helpers added
- `tests/ingest/test_optional_imports.py` — new regression tests using subprocess + `MetaPathFinder` blocker

---

### Positives

- **Correct scope** — only the ingest module is touched; nothing else in the public API is affected.
- **Backward compatible** — all existing `from semantica.ingest import X` call sites continue to work unchanged.
- **Good error UX** — `_missing_optional_dependency()` produces an actionable message ("install GitPython before using `ingest_repository()`") instead of a raw traceback.
- **Solid test approach** — the subprocess + `OptionalDependencyBlocker` `MetaPathFinder` technique is the right way to test optional-import behaviour without polluting the test runner's import state.

---

### Bugs Found & Fixed

#### Bug 1 — Overbroad missing-dependency detection

**Files:** `semantica/ingest/__init__.py` · `semantica/ingest/methods.py`

```python
# Before (wrong)
except ImportError as exc:
    if _is_missing_dependency(exc, "git"):  # also catches internal errors
        ...
```

```python
# After (fixed)
except ModuleNotFoundError as exc:
    if _is_missing_dependency(exc, "git"):  # only catches "module not found"
        ...
```

- `ImportError` is the base class and fires for *any* import failure — including errors thrown by code inside a successfully found module's own `__init__`.
- This would silently rewrite genuine bugs (e.g. a missing attribute inside `repo_ingestor.py`) as a fake "GitPython not installed" message.
- `ModuleNotFoundError` (the subclass) is specific to absent modules and always sets `exc.name`, so the string-scanning fallback in `_is_missing_dependency` was also removed.
- Applied in `__getattr__` (`__init__.py`) and all four function-level guards: `ingest_web`, `ingest_feed`, `ingest_repository`, `ingest_email` (`methods.py`).

---

#### Bug 2 — Expected `ConfigurationError` logged as a runtime failure

**File:** `semantica/ingest/methods.py` — `ingest_web`, `ingest_feed`, `ingest_repository`, `ingest_email`

```python
# Before (wrong)
except Exception as e:
    logger.error(f"Failed to ingest repository: {e}")  # fires for missing deps too
    raise
```

```python
# After (fixed)
except ConfigurationError:
    raise                            # silent re-raise — expected user config
except Exception as e:
    logger.error(f"Failed to ingest repository: {e}")
    raise
```

- When GitPython/bs4 are absent, `_missing_optional_dependency()` raises `ConfigurationError`.
- Without the guard, the outer `except Exception` catches it and emits an `ERROR`-level log — noise for a routine user-configuration scenario.
- Adding `except ConfigurationError: raise` before the blanket handler lets it propagate cleanly with no log pollution.

---

#### Bug 3 — `OptionalDependencyBlocker` not setting `exc.name` (test correctness)

**File:** `tests/ingest/test_optional_imports.py`

```python
# Before (wrong) — .name is None on a manually constructed ModuleNotFoundError
raise ModuleNotFoundError(f"No module named '{root_name}'")
```

```python
# After (fixed) — matches what Python's import machinery actually does
err = ModuleNotFoundError(f"No module named '{root_name}'")
err.name = root_name
raise err
```

- After Bug 1 was fixed, `_is_missing_dependency` relies solely on `exc.name` (the string-scan fallback was removed as redundant for real Python imports).
- Python's import machinery always sets `.name` to the top-level module name when it raises `ModuleNotFoundError`. The blocker was constructing the error manually, leaving `.name = None`, so `_is_missing_dependency` returned `False` and the `ConfigurationError` conversion never triggered.
- The test then saw a raw `ModuleNotFoundError` instead of `ConfigurationError` and failed.
- Fixed by explicitly setting `err.name = root_name` in `find_spec`, making the blocker a faithful simulation of Python's import system.

---

### Fix Verification

All fixes were confirmed present in the code after running grep checks across the changed files:

**Bug 1 — `ModuleNotFoundError` applied at every guard site**
- `semantica/ingest/__init__.py:225` — `__getattr__` ✅
- `semantica/ingest/methods.py:270` — `ingest_web` ✅
- `semantica/ingest/methods.py:341` — `ingest_feed` ✅
- `semantica/ingest/methods.py:484` — `ingest_repository` ✅
- `semantica/ingest/methods.py:551` — `ingest_email` ✅
- `_is_missing_dependency` type hint updated to `ModuleNotFoundError`, string-scan fallback removed ✅

**Bug 2 — `ConfigurationError` guard present before every blanket handler**
- `semantica/ingest/methods.py:297` — `ingest_web` ✅
- `semantica/ingest/methods.py:363` — `ingest_feed` ✅
- `semantica/ingest/methods.py:506` — `ingest_repository` ✅
- `semantica/ingest/methods.py:587` — `ingest_email` ✅

**Bug 3 — `err.name` set in blocker**
- `tests/ingest/test_optional_imports.py:27` ✅

---

### Test Run Results

Ran `pytest tests/ingest/` after all fixes:

- **104 passed, 0 failed** across all ingest tests
- `test_file_ingestion_imports_without_optional_backends` — ✅ passes
- `test_repository_ingestion_reports_missing_gitpython_when_used` — ✅ passes
- `test_notebook_02::test_08_database_ingestion` — pre-existing failure (`ProvenanceTracker` mock missing `track_relationship`), present on `main` before this PR, not related
- `test_notebook_06::test_multi_source_integration_flow` — same pre-existing failure, not related

---

### Minor Observations

- **`__getattr__` caches in `globals()`** — correct pattern; avoids repeat `importlib.import_module` calls on subsequent accesses.
- **`_LAZY_EXPORTS` is exhaustive** — covers web, feed, stream, repo, email, db, mcp, ontology, and snowflake backends. No optional backend leaks through eager imports.
- **`ingest_stream`, `ingest_database`, `ingest_mcp`, `ingest_ontology`** do not gate their inner imports with `except ModuleNotFoundError` — these backends have their own optional deps (Kafka clients, SQLAlchemy, etc.). Consistent treatment would be a worthwhile follow-up.
- **`bs4`-gated paths not tested** — `ingest_web`, `ingest_feed`, `ingest_email` without `bs4` have no regression test. A follow-up test analogous to the GitPython one would close this gap.

PR #536, opened 2026-05-04
fix(semantic_extract): break extractor import cycle

### LINKED ISSUE #528: [BUG]  semantic_extract import fails due to circular import between ner_extractor and methods
## Summary

Importing Semantica semantic extraction classes fails because `ner_extractor.py` imports `methods.py`, while `methods.py` imports `Entity` from `ner_extractor.py`.

## Environment

- OS: Windows
- Python: 3.12
- Semantica version reported by pip: 0.0.5

## Reproduction

```python
from semantica.semantic_extract import NERExtractor, RelationExtractor, TripleExtractor
```

## Actual Result

```text
ImportError: cannot import name 'Entity' from partially initialized module 'semantica.semantic_extract.ner_extractor'
```

## Expected Result

Semantic extraction classes should import successfully.

## Suggested Fix

Move imports such as:

```python
from .methods import get_entity_method
```

inside the method where they are used, or move shared dataclasses like `Entity`, `Relation`, and `Triple` into a separate module that both sides can import without cycles.

ZohaibHassan16:
## Summary
Fixes #528 by removing the circular import between `semantic_extract.methods` and the extractor modules.

The shared extraction result dataclasses now live in a dedicated `types.py` module, and `semantic_extract.__init__` lazy-loads public exports so core extractor imports do not pull in optional modules such as the YAML-backed semantic network extractor.

## Changes
- Add shared `Entity`, `Relation`, and `Triplet` dataclasses in `semantica.semantic_extract.types`.
- Update NER, relation, triplet, and methods modules to import shared types from `types.py`.
- Preserve legacy imports from `ner_extractor`, `relation_extractor`, and `triplet_extractor`.
- Lazy-load package-level exports in `semantica.semantic_extract`.
- Add `TripleExtractor` as a compatibility alias for `TripletExtractor`.
- Add regression tests for import order, legacy type imports, method dispatchers, and blocked `yaml`.

## Verification
- `python -c "from semantica.semantic_extract import NERExtractor, RelationExtractor, TripletExtractor; print(NERExtractor.__name__, RelationExtractor.__name__, TripletExtractor.__name__)"`
- `python -c "from semantica.semantic_extract import TripleExtractor; print(TripleExtractor.__name__)"`
- `python -c "from semantica.semantic_extract.methods import get_entity_method, get_relation_method, get_triplet_method; print(get_entity_method('pattern').__name__, get_relation_method('pattern').__name__, get_triplet_method('pattern').__name__)"`

Closes #528


KaifAhmad1 (approved):
## Code Review — PR #536: `fix(semantic_extract): break extractor import cycle`


**Author:** @ZohaibHassan16
**Reviewer:** @KaifAhmad1



### Summary

Fixes issue #528: importing any class from `semantica.semantic_extract` raised
`ImportError: cannot import name 'Entity' from partially initialized module 'semantica.semantic_extract.ner_extractor'`
because `ner_extractor.py` and `methods.py` each imported from the other at module level, forming a hard circular dependency.

**Verdict: Approved — all fixes applied and tests passing.**

---

### What Changed

- `semantica/semantic_extract/types.py` (new) — shared `Entity`, `Relation`, `Triplet` dataclasses moved here
- `semantica/semantic_extract/__init__.py` — rewritten to lazy-load all public exports via `__getattr__`; `TripleExtractor` alias added
- `semantica/semantic_extract/methods.py` — imports `Entity`, `Relation`, `Triplet` from `types.py` instead of the extractor modules
- `semantica/semantic_extract/ner_extractor.py` — imports `Entity` from `types.py`; no longer imports from `methods.py`
- `semantica/semantic_extract/relation_extractor.py` — imports `Entity`, `Relation` from `types.py`
- `semantica/semantic_extract/triplet_extractor.py` — imports `Entity`, `Relation`, `Triplet` from `types.py`
- `tests/semantic_extract/test_imports.py` (new) — regression tests for import order independence, legacy compatibility, alias, and yaml-free import

---

### Root Cause

The circular dependency before this fix:

```
ner_extractor.py
  └── from .methods import get_entity_method   ← pulls in methods

methods.py
  └── from .ner_extractor import Entity         ← pulls in ner_extractor
                                                   (still initializing)
```

Python's import system partially initialises `ner_extractor` before executing `methods`, so when `methods` tries to import `Entity` from `ner_extractor`, the name does not yet exist in the module's namespace, raising `ImportError`.

---

### Fix Design

The correct fix is to extract the shared data models (`Entity`, `Relation`, `Triplet`) into a third module (`types.py`) that neither side of the old cycle needs to import from the other. Both `methods.py` and all extractor modules now import types from `types.py` only:

```
types.py          ← no imports from this package
   ↑                 ↑
ner_extractor      methods.py
relation_extractor
triplet_extractor
```

The cycle is completely gone. This is the standard solution for circular imports caused by shared data models — move the models to a leaf module.

---

### Positives

- **Correct root-cause fix** — the cycle is broken structurally, not worked around with lazy imports inside functions (which would only defer the crash, not prevent it).
- **Zero API breakage** — `Entity`, `Relation`, `Triplet` are re-exported from their original modules as `from .types import Entity`, so all existing code importing from `ner_extractor` or `relation_extractor` continues to work.
- **`__init__.py` lazy loading** — deferring all non-core exports behind `__getattr__` means importing one class no longer pulls in the entire module tree, including optional dependencies like the YAML-backed `SemanticNetworkExtractor`.
- **`TripleExtractor` alias** — `TripleExtractor = TripletExtractor` added as a backward-compatibility alias, closing a common user confusion between the two names.
- **Regression tests** — seven tests covering the two import orders (methods-first and extractors-first), the alias, legacy type imports, and yaml-free core import.

---

### Bugs Found & Fixed

### Qodo — `subprocess.CompletedProcess[str]` not subscriptable on Python 3.8

**File:** `tests/semantic_extract/test_imports.py`

```python
# Before (wrong) — runtime generic subscription fails on Python 3.8
def _run_python(code: str) -> subprocess.CompletedProcess[str]:
```

```python
# After (fixed) — PEP 563 makes annotations lazy strings, safe on 3.8+
from __future__ import annotations

def _run_python(code: str) -> subprocess.CompletedProcess[str]:
```

- `subprocess.CompletedProcess[str]` uses the generic subscript syntax that is only legal at runtime from Python 3.9+. On Python 3.8 this is evaluated eagerly when the module is imported, raising `TypeError` and aborting test collection before a single test runs.
- The project declares `requires-python = ">=3.8"` in `pyproject.toml`, so the test suite must remain importable on 3.8.
- Adding `from __future__ import annotations` (PEP 563) converts all annotations in the module to lazily-evaluated strings, making the subscript never actually evaluated at runtime. No behaviour change on Python 3.9+.

---

### Additional — `sys.modules` pollution from `test_retry_logic.py` breaking `test_pr482_deepseek_openai`

**File:** `tests/semantic_extract/test_retry_logic.py`

```python
# Before (wrong) — mocks injected at module level and never restored
sys.modules["openai"] = MagicMock()
sys.modules["spacy"] = mock_spacy
# ... other mocks ...

from semantica.semantic_extract import NERExtractor
# mocks remain in sys.modules for the rest of the test session
```

```python
# After (fixed) — originals saved before injection, restored immediately after imports
_MOCKED_MODULES = ["spacy", "instructor", "groq", "openai", "sentence_transformers", "transformers"]
_original_modules = {k: sys.modules.get(k) for k in _MOCKED_MODULES}

sys.modules["openai"] = MagicMock()
# ... other mocks ...

from semantica.semantic_extract import NERExtractor
from semantica.semantic_extract.methods import extract_entities_llm, _extract_entities_chunked, extract_relations_llm, extract_triplets_llm
from semantica.semantic_extract.providers import BaseProvider

# Restore real sys.modules entries now that the mocks have served their purpose
for _key, _original in _original_modules.items():
    if _original is None:
        sys.modules.pop(_key, None)
    else:
        sys.modules[_key] = _original
```

- pytest imports **all** test files during collection before running any tests. `test_retry_logic.py` injected `sys.modules["openai"] = MagicMock()` at module level to satisfy semantica's import-time checks for the optional `openai` dependency.
- Because the mocks were never removed, `openai` remained a `MagicMock` object for the entire test session. When `test_pr482_deepseek_openai` later ran and called `MagicMock(spec=openai.OpenAI)`, the spec was resolved against the mock rather than the real class, raising `InvalidSpecError`.
- `tearDownModule` would not fix this — it runs after the test class, but the pollution happens at collection time, before any tests execute.
- The fix saves the original `sys.modules` entries before injection and restores them immediately after the three semantica imports that required the mocks, so every subsequent test file sees the real packages.

---

### Fix Verification

**Cycle removed — import graph is now acyclic:**
- `semantica/semantic_extract/ner_extractor.py:79` — `from .types import Entity` ✅
- `semantica/semantic_extract/relation_extractor.py:78` — `from .types import Entity, Relation` ✅
- `semantica/semantic_extract/triplet_extractor.py:79` — `from .types import Entity, Relation, Triplet` ✅
- `semantica/semantic_extract/methods.py:120` — `from .types import Entity, Relation, Triplet` ✅
- No extractor module imports from `methods.py` at module level ✅

**Qodo fix present:**
- `tests/semantic_extract/test_imports.py:1` — `from __future__ import annotations` ✅

**`sys.modules` pollution fix present:**
- `tests/semantic_extract/test_retry_logic.py:14-47` — originals saved before injection, restored immediately after semantica imports ✅

---

### Test Run Results

Ran `pytest tests/semantic_extract/` after all fixes — **152 passed, 0 failed, 9 skipped**:

- All 7 tests in `test_imports.py` — ✅
- All 10 tests in `test_retry_logic.py` — ✅
- `test_pr482_deepseek_openai` (2 tests) — ✅ (previously failed in full-suite run due to `sys.modules` pollution from `test_retry_logic.py`; now pass after the restore fix)
- All remaining tests in `tests/semantic_extract/` — ✅

---

### Minor Observations

- **`types.py` has no `__all__`** — not blocking, but worth adding for explicitness given that it is now the canonical source of truth for shared extraction types.
- **`Triplet.get()` and `Triplet.__getitem__()`** — these dict-like accessors on the dataclass are useful but undocumented. A brief note on why they exist (e.g. backward compat with code treating triplets as dicts) would help future readers.
- **`SemanticNetworkExtractor` still requires `yaml`** — the lazy `__getattr__` now prevents the yaml import from happening eagerly, which is correct. No test confirms the error message when yaml is missing for that specific class; the existing `test_core_package_imports_do_not_require_yaml` only checks core extractors.

Issue #575, opened 2026-05-25
[FEATURE] Infrastructure, New Module Tracks & Dataset Registry


KaifAhmad1:
## Scope

Four new module tracks benchmarking previously-uncovered Semantica capabilities, plus all shared infrastructure that every other sub-issue depends on. This sub-issue must be at least partially merged before other sub-issues submit PRs, because it sets up git-lfs patterns, JSON schema validation, and the benchmark runner.

- Track 32 — Document Parsing Quality: does `parse.DocumentParser` faithfully extract text, tables, formulas, and reading order from PDFs? Silent parsing errors corrupt every KG node built from that content.
- Track 33 — Chunking / Splitting Quality: do `split.Splitter` and `split.SemanticChunker` produce chunks that preserve answer spans and maximise downstream retrieval recall?
- Track 34 — Vector Store Retrieval Performance: does `VectorStore.search()` maintain recall under metadata-filtered ANN search at production throughput?
- Track 35 — Ingest Pipeline End-to-End Throughput: does `IngestPipeline.run()` meet the 50 docs/s throughput SLA and the per-document P95 latency SLA?

Infrastructure tasks owned by this sub-issue (not a separate track):

- `.gitattributes` — git-lfs patterns for every large file across all five sub-issues
- JSON schema validation — `fixtures/schema/` files and fixture validation test
- `benchmarks/benchmarks_runner.py` — update to discover all seven benchmark directories
- `detect_duplicates()` regression guard wire-up (implementation lives in Sub-issue 4)
- Top-level `datasets/README.md` — one line per dataset across all sub-issues

Exclusive shared-file edit rights for this sub-issue:

- `.gitattributes` — only this sub-issue may add git-lfs patterns
- `benchmarks/benchmarks_runner.py` — only this sub-issue may update directory discovery
- `datasets/README.md` — only this sub-issue may edit the top-level registry

---

## Track 32 — Document Parsing Quality

**Why it matters:** `IngestPipeline` accuracy depends entirely on the parser's ability to faithfully extract text, tables, formulas, and reading order from PDFs and office documents. A single silently-corrupted table in a financial report can produce dozens of wrong KG nodes — with no parsing error raised.

**APIs under test:** `parse.DocumentParser` · `parse.PDFParser` · `parse.TableExtractor` · `ingest.IngestPipeline` (parse stage only)

Datasets:

[...]

PR #588, opened 2026-06-05
feat(benchmarks): add module-level filtering for benchmark runner (#575)

Sameer6305:
## Summary

Part of #575 — Infrastructure, New Module Tracks & Dataset Registry.

Adds a minimal `--module` flag to `benchmarks/benchmarks_runner.py` to allow running benchmarks for a specific benchmark module directory.

Example:

```bash 
python benchmarks/benchmarks_runner.py --module core_processing
```

## Changes

* added optional `--module` CLI argument
* added argparse validation using `choices=`
* replaced hardcoded benchmark path with dynamic module-based path selection
* preserved default behavior when `--module` is not provided

## Validation

Validated:

* `--help` output
* valid module selection
* invalid module rejection through argparse validation

## Notes

* uses pytest native path filtering only
* no new dependencies added
* no benchmark logic changes
* no multi-module support added in this PR
* fully backward compatible

### DIFF
--- a/benchmarks/benchmarks_runner.py
+++ b/benchmarks/benchmarks_runner.py
@@ -10,10 +10,23 @@ from rich.rule import Rule
 console = Console()
 
 
+def _discover_modules() -> list[str]:
+    benchmarks_dir = "benchmarks"
+    _excluded = {"results", "__pycache__"}
+    return sorted(
+        d for d in os.listdir(benchmarks_dir)
+        if os.path.isdir(os.path.join(benchmarks_dir, d))
+        and not d.startswith(("_", "."))
+        and d not in _excluded
+    )
+
+
 def run_benchmarks():
     """
     Master Runner for Semantica Benchmarks.
     """
+    available_modules = _discover_modules()
+
     parser = argparse.ArgumentParser(description="Run Semantica Benchmarks")
     parser.add_argument(
         "--strict", action="store_true", help="Fail script if performance regresses"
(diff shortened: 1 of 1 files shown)


KaifAhmad1 (approved):
## PR #588 — `feat(benchmarks): add module-level filtering for benchmark runner`

### What it does

- Adds `--module` CLI flag to narrow pytest collection to a single benchmark directory
- Falls back to the full `benchmarks/` tree when omitted — fully backward-compatible

---

### Issues

#### Critical

- `context_graph_effectiveness` is listed in `choices` but the directory doesn't exist — pytest exits 0 with zero tests collected, a silent false pass
- `infrastructure/` exists on disk but is absent from `choices`, so it can't be targeted; #575 plans to add tests there
- No existence check before invoking pytest — any choice that maps to a missing path silently collects nothing

#### Moderate

- Hardcoded `choices` list will drift as #575 adds `decision_intelligence`, `temporal_provenance`, `memory_context`, `structural_intelligence`, `module_tracks` — each needs a manual update
- No console output when a filter is active — no way to confirm which subset ran

#### Minor

- Help text doesn't mention the default (run all modules)
- Both `"context"` and `"context_graph_effectiveness"` are listed — only `context/` exists

---

### Fixes applied (commit `1cf91f1`)

- Replaced hardcoded list with `_discover_modules()` — scans `benchmarks/` at runtime, excludes `results/`, `__pycache__/`, hidden and dunder dirs
- Stale `context_graph_effectiveness` entry gone; `infrastructure/` and all future dirs auto-included
- Added `os.path.isdir` guard — exits with a red error instead of silently passing with 0 tests
- Prints `Module filter: <name>` before pytest runs so the active scope is visible

```python
def _discover_modules() -> list[str]:
    _excluded = {"results", "__pycache__"}
    return sorted(
        d for d in os.listdir("benchmarks")
        if os.path.isdir(os.path.join("benchmarks", d))
        and not d.startswith(("_", "."))
        and d not in _excluded
    )
```

---

### Verdict

Good concept, right mechanism. Hardcoded choices introduced correctness bugs — all fixed in the follow-up commit. **Ready to merge once CI passes.**

PR #589, opened 2026-06-05
test(benchmarks): add git-lfs infrastructure validation checks (#575)

Sameer6305:
## Summary

Part of #575 — Infrastructure, New Module Tracks & Dataset Registry.

Adds a minimal infrastructure validation test for future git-lfs benchmark dataset integration.

This PR intentionally keeps the implementation lightweight and production-safe while benchmark dataset infrastructure is still evolving.

## Changes

* added `benchmarks/infrastructure/test_git_lfs_pointers.py`
* added passive validation checks for:

  * `.gitattributes` LFS configuration
  * benchmark directory structure referenced in #575
  * large benchmark files without LFS tracking
  * infrastructure directory existence

## Notes

* implementation gracefully handles missing dataset directories/files
* no placeholder dataset/schema infrastructure added
* no CI-breaking enforcement introduced
* no new dependencies added
* validation is intentionally non-invasive until benchmark datasets land

## Validation

```bash
python -m pytest benchmarks/infrastructure/test_git_lfs_pointers.py -v
```

Result:

* 4 tests passed

### DIFF
--- a/benchmarks/infrastructure/test_git_lfs_pointers.py
+++ b/benchmarks/infrastructure/test_git_lfs_pointers.py
@@ -3,136 +3,70 @@ Git LFS pointer validation for benchmark infrastructure (Issue #575).
 
 Validates that large files are properly tracked by git LFS to ensure
 benchmark dataset registry remains manageable as module tracks expand.
-
-This test follows the implementation style shown in issue #575 documentation
-and gracefully handles missing directories/files that will be created
-as part of the issue #575 roadmap.
 """
 
-import os
+import pytest
 from pathlib import Path
 
+REPO_ROOT = Path(__file__).resolve().parents[2]
+BENCHMARK_ROOT = REPO_ROOT / "benchmarks"
 
+
+@pytest.mark.skip(reason="LFS dataset registry not yet implemented (#575)")
 def test_lfs_configured_for_benchmark_datasets():
-    """Check .gitattributes contains LFS patterns for benchmark data files.
-    
-    This ensures that when issue #575 dataset registry is implemented,
-    large dataset files will be properly tracked by git LFS.
-    """
-    gitattributes = Path(".gitattributes")
-    if not gitattributes.exists():
-        # .gitattributes may not exist yet, which is fine for development
-        return
-    
+    """Check .gitattributes contains LFS patterns scoped to benchmark dataset files."""
+    gitattributes = REPO_ROOT / ".gitattributes"
+    assert gitattributes.exists(), ".gitattributes not found at repo root"
+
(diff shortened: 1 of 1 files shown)


KaifAhmad1 (approved):
## PR Review — #589 `test(benchmarks): add git-lfs infrastructure validation checks`

### Original Issues

- ~~**3 of 4 tests had no assertions** — they always passed regardless of repo state, providing zero regression value~~
- ~~**1 remaining test was trivially true** — asserting `benchmarks/infrastructure/` exists from a file that lives inside it; pytest would never collect the file if the directory was missing~~
- ~~**All paths were CWD-relative** (`Path(".gitattributes")`, `Path("benchmarks")`) — tests would silently break if run from any directory other than repo root~~
- ~~**`expected_patterns` list was dead code** — defined but never referenced~~
- ~~**`os` import was unnecessary** — `os.walk` used despite `pathlib` already being imported; inconsistent with project style~~
- ~~**`*.json` / `*.yaml` LFS patterns were too broad** — would track all config and schema files project-wide if ever enforced~~
- ~~**Missing newline at end of file**~~

### Follow-up Fix 

- ✅ **Real assertions added to all 3 vacuous tests** — each now asserts a concrete condition and fails with a descriptive message if violated
- ✅ **Pending tests gated with `pytest.mark.skip(reason="... (#575)")`** — keeps CI output honest; enabling enforcement later requires only removing a decorator, not rewriting the test
- ✅ **`test_no_large_files_in_benchmarks_without_lfs` runs actively** — no skip needed; passes today and will catch any accidental large-file commits immediately
- ✅ **All paths anchored to `Path(__file__)`** — `REPO_ROOT = Path(__file__).resolve().parents[2]`; tests are now invocation-directory agnostic
- ✅ **LFS pattern check scoped to `benchmarks/`** — only matches `.gitattributes` lines starting with `benchmarks/`, preventing false-passes from unrelated global LFS rules
- ✅ **Dead code removed** — unused `os` import and `expected_patterns` list dropped; `os.walk` replaced with `Path.rglob`
- ✅ **Missing newline at EOF fixed**

### Checklist

- ✅ No assertions missing
- ✅ No dead code
- ✅ Paths are CWD-independent
- ✅ CI will not false-pass on pending infrastructure
- ✅ Tests pass locally (`2 passed, 2 skipped`)
- ✅ No new dependencies introduced
- ✅ Follow-up commit pushed to `Sameer6305:benchmark-lfs-infra`

### Verdict

**Ready to merge.** All original review findings addressed. Net diff is −105 / +39 — cleaner and strictly more correct than the original submission.

PR #637, opened 2026-06-15
Fix CLI demo blockers

### LINKED ISSUE #636: [BUG] CLI demo blockers: `kg build` deadlocks, `--dry-run` is ignored and `doctor` false fails on Windows
### Summary

While preparing demo recording, `semantica kg build` constantly hangs on a tiny local sample file. It appears to be a progress tracker deadlock while rendering pipeline progress.

There are also two related CLI workflow buigs:

- `--dry-run kg build` still enters the real path build.
- Progress tracking cannot be disabled for diagnostics
- `semantica doctor` false fails the `Log dir` check on windows because it tries to delete log

### Reproduction

```
.\.venv\Scripts\semantica.exe kg build --source cookbook\use_cases\advanced_rag\data\sample_graphrag_paper.txt
```

ZohaibHassan16:
Changes:

- Avoid progress tracker display callbacks while holding tracker lock.
- Preserve False behavior
- Make --dry-run return early before framework init
- Make doctor probe log dir without touching semantica.log
- Add regression coverage for CLI paths

Tested:
- .\.venv\Scripts\python.exe -m pytest tests\test_cli_foundation.py tests\test_cli_commands.py tests\test_progress_tracker_regressions.py -q

Closes #636

### DIFF
--- a/semantica/cli.py
+++ b/semantica/cli.py
@@ -220,6 +220,42 @@ def _get_framework(cli_ctx: CLIContext) -> "Semantica":
     return cli_ctx.framework
 
 
+def _check_log_directory(cli_ctx: CLIContext) -> str:
+    """Verify the configured log directory is writable.
+
+    The CLI may already hold an open file handle for ``semantica.log``. Probe
+    with a unique temporary file so Windows can check writability without
+    deleting the active log file.
+    """
+    log_file = cli_ctx.config.get("logging.file", "semantica.log")
+    if not log_file:
+        return "console-only logging"
+
+    log_path = Path(log_file).expanduser()
+    log_dir = log_path.parent
+    if not log_dir or str(log_dir) == ".":
+        log_dir = Path.cwd()
+    elif not log_dir.is_absolute():
+        log_dir = Path.cwd() / log_dir
+
+    probe_path: Optional[Path] = None
+    try:
+        probe_path = (
+            log_dir
+            / f".semantica-doctor-{os.getpid()}-{time.time_ns()}.tmp"
+        )
+        probe_path.write_text("ok", encoding="utf-8")
+        return f"{log_dir} writable"
+    finally:
+        if probe_path is not None:
+            try:
+                probe_path.unlink(missing_ok=True)
+            except OSError:
+                # Some locked-down Windows environments allow writes but deny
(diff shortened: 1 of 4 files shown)


KaifAhmad1 (approved):
## PR #637 Review — Fix CLI Demo Blockers

**Author:** @ZohaibHassan16 | **Reviewer:** @KaifAhmad1

---

### Bugs Fixed

| Bug | Root Cause | Fix |
|---|---|---|
| `kg build` deadlock | `display.update()` called while holding `self.lock`; callbacks re-entered the tracker | Snapshot state + displays under lock, call displays after release |
| `--dry-run` ignored | `_get_framework()` called before dry-run check | Early return before framework init |
| `doctor` false failure on Windows | `p.touch(); p.unlink()` on active `semantica.log` denied by OS | Probe with PID-namespaced temp file, `missing_ok=True` cleanup |

---

### Key Changes

**`progress_tracker.py`**
- Display callbacks extracted outside `self.lock` in all four tracking methods via new `_update_displays()` helper
- `enabled` now a property — respects constructor arg and `SEMANTICA_DISABLE_PROGRESS` env var; flag is sticky once set
- Removed forced `enabled = True` re-assignments in `get_instance` and `get_progress_tracker`
- Jupyter re-detection deduplicated into `_ensure_jupyter_display()`

**`cli.py`**
- `_check_log_directory()` replaces the old `touch/unlink` lambda with a Windows-safe temp-file probe
- `--dry-run` guard added in `_run_build` before `_get_framework()` is called

**Tests**
- Deadlock regression tests use `_assert_finishes_quickly` (1s thread timeout)
- Dry-run tests patch `_get_framework` to raise if called
- Doctor tests verify `semantica.log` is untouched after probe

---

### Follow-up Commit `496b80c` (CodeQL fixes)

- Removed duplicate `from ... import` — all references go through `progress_module` alias
- `BaseException` → `Exception` in thread runner helper

---

### Minor Notes

- `display.last_update` mutated outside lock in `force_console` path — benign under GIL, pre-existing
- Test artifacts written to `test_data/runtime/` instead of `tmp_path` — no auto-cleanup on CI
- No regression tests for `update_tracking` / `show_summary` deadlock paths

---

**Verdict: Approve.** Fixes are correct, targeted, and well-tested.

PR #656, opened 2026-06-19
docs: fix onboarding examples for GraphBuilder and temporal queries

Sameer6305:
## Summary

Fix onboarding examples that did not match the current implementation.

### Changes

* Corrected `GraphBuilder.build()` usage in Quickstart and Getting Started guides.
* Updated examples to use the actual graph structure returned by `GraphBuilder`.
* Removed references to non-existent graph query methods.
* Replaced an outdated temporal knowledge graph example with a verified `TemporalGraphQuery` workflow.

### Validation

Verified against:

* `semantica/kg/graph_builder.py`
* `semantica/kg/temporal_query.py`
* `semantica/kg/__init__.py`

All examples were updated to match current APIs, method signatures, and return structures.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -94,9 +94,9 @@ icon: "rocket"
 
     # 4. Build
     graph = GraphBuilder(merge_entities=True).build(
-        entities=entities, relationships=relationships
+        {"entities": entities, "relationships": relationships}
     )
-    print(f"{len(graph['nodes'])} nodes, {len(graph['relationships'])} edges")
+    print(f"{len(graph['entities'])} nodes, {len(graph['relationships'])} edges")
     ```
 
     **Next:** [Full pipeline walkthrough →](quickstart)
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -348,20 +348,19 @@ graph   = builder.build({"entities": entities, "relationships": relationships})
 <Accordion title="Full provenance pipeline: W3C PROV-O" icon="link">
 
 ```python
-from semantica.kg import GraphBuilder, ProvenanceTracker
+from semantica.provenance import ProvenanceManager
+from semantica.kg import GraphBuilder
 
-tracker = ProvenanceTracker()
-
-# Record where each entity came from before or after extraction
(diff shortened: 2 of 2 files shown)


Sameer6305:
Opened a docs fix for review. Verified the onboarding examples against the current GraphBuilder and TemporalGraphQuery implementations and updated the examples to match the actual APIs and return structures.

KaifAhmad1 (approved):
## PR #656 Review

This PR audits the onboarding docs against the current implementation and fixes examples that would fail at runtime. The changes are well-scoped and verified against the actual source files in `semantica/kg/`.

### PR fixes (all correct)

The core fixes address four categories of broken examples:

- `build(entities=..., relationships=...)` → `build({"entities": ..., "relationships": ...})` — old keyword form raised `TypeError` at runtime since `sources` is the only positional parameter
- `graph['nodes']` → `graph['entities']` — `build()` returns a plain `dict` and the correct key is `'entities'`
- Removed `graph.nodes`, `graph.edges`, `graph.get_node()`, `graph.get_neighbors()` — attribute-style access on a `dict` that doesn't support it
- `TemporalKnowledgeGraph` → `TemporalGraphQuery` — former is not exported from `semantica.kg`; latter is the real API

### Follow-up fixes applied

Three additional issues were caught during review and patched directly onto `Sameer6305:docs/onboarding-audit-3`:

- **Wrong import** — `from semantica.provenance import ProvenanceTracker` raises `ImportError` at runtime. The accordion is titled "W3C PROV-O" — the correct class is `ProvenanceManager` from `semantica.provenance`, which is the W3C PROV-O compliant implementation with `track_entity()` and `get_all_sources()`

- **Hollow provenance accordion** — the PR correctly removed the broken `graph.get_node()` call but left a bare `builder.build()` with nothing demonstrated. Replaced with a real `ProvenanceManager.track_entity()` / `get_all_sources()` workflow showing the actual output structure: `{"source": ..., "location": ..., "timestamp": ..., "confidence": ...}`

- **Silent `query=""`** — the `query` parameter in `TemporalGraphQuery.query_at_time()` is currently a no-op reserved for future query parsing. Annotated inline as `# query reserved for future use` so readers aren't left guessing what query language is expected

**Verdict: Approved**

PR #685, opened 2026-06-23
docs: improve ontology guide onboarding and practical guidance

Sameer6305:
## Summary

Improves the Ontology guide to make ontology concepts easier to understand for new users while preserving all existing implementation details and domain-specific examples.

## Changes

- Added a "What Is an Ontology?" section with plain-English explanations
- Added a "Why Use Ontologies?" section describing practical benefits
- Added a "When To Use / When Not To Use" section
- Clarified classes, object properties, and datatype properties
- Improved explanations around ontology design and graph modeling
- Added a simple introductory example before advanced domain-specific examples
- Added a "Common Pitfalls" section covering over-modeling, ontology drift, unnecessary LLM generation, and naming consistency
- Improved transitions and onboarding flow throughout the guide
- Preserved all existing cybersecurity, healthcare, and compliance examples

## Verification

Verified against current implementation:

- OntologyGenerator
- LLMOntologyGenerator
- PropertyGenerator
- generate_from_graph()
- graph.to_dict() workflows

No API or behavior changes.

Documentation-only update.

### DIFF
--- a/docs/guides/ontology.md
+++ b/docs/guides/ontology.md
@@ -5,15 +5,111 @@ description: "Generate OWL ontologies from your knowledge graph, validate class 
 
 `OntologyGenerator` derives a formal OWL ontology directly from the entities and relationships already in your knowledge graph — no schema design upfront. Use it to produce a machine-readable contract for your graph's classes and properties, then export to Turtle, OWL/XML, or JSON-LD for SHACL validation, reasoning engines, and STIX/TAXII toolchains.
 
+## What Is an Ontology?
+
+An ontology is a formal specification of the concepts and relationships in a domain. It defines a shared vocabulary for your knowledge graph by declaring:
+
+**Classes** — Types of entities in your domain. For example, `ThreatActor`, `Vulnerability`, or `Software` in cybersecurity. Classes describe what kinds of things exist.
+
+**Object Properties** — Relationships between entities. Examples: `exploits` (Threat Actor exploits Vulnerability), `targets` (Malware targets Organization), or `uses` (Actor uses Tool).
+
+**Datatype Properties** — Attributes with literal values. Examples: `name` (text), `severity_score` (decimal), `published_date` (date), or `ip_address` (string).
+
+An ontology makes your knowledge graph machine-readable by specifying which relationships are valid, what types of values each property can hold, and how concepts relate to each other hierarchically.
+
+## Why Use Ontologies?
+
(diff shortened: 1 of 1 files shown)


KaifAhmad1 (approved):
## PR #685 Review — Ontology Guide Onboarding

**Author:** @Sameer6305 

### What the PR Does

Adds onboarding content to the ontology guide — "What Is an Ontology?", "Why Use Ontologies?", "When To Use / When Not To Use", a beginner Python example, and "Common Pitfalls." No API or behavior changes.

### Bugs Found (verified against source)

- **Silent empty ontology** — `graph.to_dict()` returns `nodes`/`edges` keys; `generate_ontology()` reads `entities`/`relationships`. Both resolve to `[]`, producing 0 classes and 0 properties with no error raised. *(confirmed: `context_graph.py:1790`, `ontology_generator.py:247`)*

- **Datatype filter never matches** — `prop.get('type') == 'datatype'` is always `False`. The actual value stored is `'data'`. The "Datatype Properties" block prints a header but nothing beneath it. *(confirmed: `property_generator.py:228`)*

- **Domain/range print as lists** — `domain` and `range` on object properties are `list(...)`, not strings. Output would be `works_for (['Person'] → ['Company'])` not the clean format shown in comments. *(confirmed: `property_generator.py:184–185`)*

- **Reasoning bullet overpromises** — Implied Semantica performs OWL inference automatically. It exports the ontology; reasoning requires an external engine (HermiT, Pellet, ELK).

- **Duplicate pitfall** — "Unnecessary LLM generation" pitfall was word-for-word the same as the `<Info>` callout 140 lines earlier in the same file.


### Fixes Applied (commit `5f6cac0`)

- Replaced `ctx.store()` + `graph.to_dict()` with a direct `entities`/`relationships` dict and `generate_ontology()` — deterministic and matches the API contract
- `'datatype'` → `'data'` in the type filter
- Fixed domain/range print with `', '.join(prop.get('domain', []))`
- Clarified the Reasoning bullet to mention external reasoners
- Removed the duplicate pitfall entry


### What's Good

- Conceptual definitions (classes, object/datatype properties) are accurate OWL terminology
- "When To Use / When Not To Use" and "Common Pitfalls" add real value for new users
- All existing cybersecurity, healthcare, and compliance examples are untouched


### Residual Note

The `generate_from_graph(graph.to_dict(), ...)` key mismatch also exists in the pre-existing "Generating the ontology" section and the Defense CTI domain tab — not introduced by this PR. Should be tracked separately.


**Verdict: Approve** — conceptual additions are solid; code bugs are fixed in `5f6cac0`.

PR #694, opened 2026-06-23
docs: improve LLM integrations guide onboarding and provider guidance

Sameer6305:
## Summary

Improves the LLM Integrations guide with additional onboarding content, provider-selection guidance, usage recommendations, and common pitfalls.

## Changes

- Added "What Are LLM Integrations?" section
- Added "Why Use LLM Integrations?" section
- Added "When To Use / When Not To Use" guidance
- Added beginner-friendly explanations for Groq, OpenAI, LiteLLM, and HuggingFaceLLM
- Added practical provider selection guidance
- Added Common Pitfalls section covering common usage mistakes
- Improved readability and onboarding flow for new users

## Notes

- Documentation-only changes
- No API changes
- No implementation changes

### DIFF
--- a/docs/guides/llm-integrations.md
+++ b/docs/guides/llm-integrations.md
@@ -360,9 +360,8 @@ from semantica.vector_store import VectorStore
 extraction_llm = HuggingFaceLLM(model="/opt/models/mistral-7b-instruct")
 reasoning_llm  = HuggingFaceLLM(model="/opt/models/llama-3.1-70b-instruct")
 
-# NER with local model — provider pattern still works for local paths
-# (use extract_entities_llm directly with the provider instance)
-from semantica.semantic_extract.methods import extract_entities_llm
+# The llms module wrappers can also be used directly for raw prompt generation
+# when you want to bypass the semantic extraction layer entirely
 
 sigint_text = (
     "[S//NF] APT29 operator observed deploying WARPWIRE credential harvester "
@@ -570,7 +569,7 @@ print(best["response"])
 
 # Sources the answer is grounded in
 for src in best["sources"]:
-    print("  - [{}] {}".format(src.get("metadata", {}).get("source", "?"), src["content"][:60]))
+    print("  - [{}] {}".format(src.get("source", "?"), src["content"][:60]))
 ```
 
 </Tab>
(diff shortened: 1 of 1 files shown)


KaifAhmad1 (approved):
## PR Review: #694 — docs: improve LLM integrations guide onboarding and provider guidance

**Author:** @Sameer6305 

### Summary

Docs-only PR that expands `docs/guides/llm-integrations.md` with onboarding content
(What/Why/When-to-use sections, a Common Pitfalls section) and fixes two factual
mismatches between the doc's code samples and the actual implementation.

### Verification

I checked the two substantive (non-prose) changes against the current source, since
those are the parts that could silently mislead a reader:

1. **`src.get("metadata", {}).get("source", "?")` → `src.get("source", "?")`**
   Confirmed correct. `ContextRetriever.query_with_reasoning()`
   (`semantica/context/context_retriever.py:1624-1631`) returns `sources` as a flat
   list of `{"content", "score", "source"}` dicts — there is no nested `"metadata"`
   key. The old snippet would have silently printed `"?"` for every source.

2. **Removed `from semantica.semantic_extract.methods import extract_entities_llm`
   and its comment.**
   Confirmed this was broken. `extract_entities_llm(text, provider: str = "openai", ...)`
   (`semantica/semantic_extract/methods.py:883-891`) takes a **provider name string**,
   not a provider instance, so the old comment ("use extract_entities_llm directly
   with the provider instance") was inaccurate — and the import was never even called
   in the snippet (dead code). The replacement comment is accurate, and the rest of
   the example still works since `extraction_llm.generate(...)` is a real method on
   `HuggingFaceLLM`.

### Notes (non-blocking)

- The new "Why Use LLM Integrations" section overlaps conceptually with the existing
  "Choosing a Provider" section just below it (both cover provider portability /
  switching cost). Not incorrect, just slightly repetitive — a nit, not worth
  blocking on.
- Pre-existing issue, **not introduced by this PR**: the CTI code sample imports
  `NamedEntityRecognizer` and computes `raw_entities`, neither of which is used
  afterward. This was already true on `main` prior to this PR.

### Verdict

**Ready to merge.** Both technical corrections were verified against the current
codebase and are accurate; the added prose is beginner-friendly and consistent with
the rest of the guide; CI is green with no conflicts. The overlap noted above is
optional polish and does not need to hold up this PR.

PR #699, opened 2026-06-24
docs: improve Policy Engine guide onboarding and implementation accuracy

Sameer6305:
## Summary

Improves the Policy Engine guide for first-time developers while aligning examples with the actual implementation.

### Documentation Improvements
- Added "What Is Policy Engine?" section
- Added "Why Use Policy Engine?" section
- Added "When To Use / When Not To Use" guidance
- Added "Supported Rule Types" reference section
- Added "Common Pitfalls" section
- Improved onboarding explanations for Policy, Decision, and PolicyException objects
- Clarified the relationship between PolicyEngine, DecisionRecorder, and approval workflows
- Added explicit explanation that PolicyEngine evaluates compliance but does not automatically block actions

### Implementation Alignment
- Replaced unsupported rule examples with supported rule patterns
- Updated examples to use:
  - `allowed_outcomes`
  - `min_confidence`
  - `required_*`
  - `min_*`
  - `max_*`
  - `required_categories`
- Documented actual rule evaluation behavior and field lookup logic
- Added warnings about unsupported rule structures and keys

### Domain Examples
- Updated Defense example to use supported rule types
- Updated Security example to use supported metadata validation
- Updated Life Sciences example to use supported threshold-based rules
- Updated Banking example to use supported field validation patterns

## Result

The guide now better supports onboarding while ensuring all documented policy examples match current PolicyEngine behavior.

### DIFF
--- a/docs/guides/policy-engine.md
+++ b/docs/guides/policy-engine.md
@@ -4,17 +4,92 @@ description: "Define, version, and enforce governance policies over knowledge gr
 icon: "scale-balanced"
 ---
 
+## What Is Policy Engine?
+
+Policy evaluation is the systematic checking of decisions against predefined governance rules and constraints. Unlike application enforcement that automatically blocks non-compliant actions, policy evaluation provides compliance status that can trigger different workflows—approval processes, exception handling, or audit requirements.
+
+**Key policy concepts:**
+
+**Policy evaluation** checks whether decisions meet defined criteria without automatically preventing actions, enabling flexible governance workflows.
+
+**Governance and compliance workflows** use policy evaluation results to route decisions through appropriate approval chains, exception processes, or audit trails.
+
+**Approval processes** can be triggered by policy violations, creating documented exception paths with justification and approver accountability.
+
+**Difference from enforcement:** Policy evaluation returns compliance status (`True`/`False`) but does not automatically block actions. Your workflow determines what happens next—immediate approval, escalation, exception handling, or rejection.
+
+## Why Use Policy Engine?
+
+**Governance and accountability.** Create auditable decision workflows where every policy evaluation, exception, and approval is permanently recorded in the knowledge graph with full provenance tracking.
+
+**Compliance verification.** Systematically check decisions against regulatory requirements, internal policies, and risk management rules before they are finalized or acted upon.
+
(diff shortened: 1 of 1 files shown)


KaifAhmad1 (approved):
## Review: #699 — docs: improve Policy Engine guide onboarding and implementation accuracy

**Verdict: Approve** — onboarding additions are solid, and the two factual bugs found during review have been fixed in a follow-up commit ([`aa2cd00`]

### Issues found and fixed

- **Inverted pitfall description (was line 318).** The "Using unsupported rule keys" pitfall claimed `disallowed_outcomes`-style keys "silently pass compliance if the key is absent from metadata or fail if present." Verified against `PolicyEngine._evaluate_compliance` this was backwards — absent → `False` (fails), present with any value → `True` (passes). **Fixed**: wording now matches actual behavior.

- **Broken `required_<field>: True` pattern in the mortgage example (was lines 589–593).** The example used `required_ltv: True`, `required_pd: True`, `required_lgd: True`, `required_dsti: True`, `required_credit_score: True` to mean "field must be present." But `required_*` does an equality check against the literal rule value, so this silently marked an otherwise fully compliant decision (LTV, DSTI, credit score, stress test, outcome all passing) as non-compliant — confirmed by direct execution against the real engine. **Fixed**: removed the redundant/broken `required_*` entries (the `min_`/`max_` rules already enforce presence of `ltv`/`dsti`/`credit_score`) and annotated `pd`/`lgd` in the decision metadata as audit-only fields with no threshold rule in this policy.

### Verified accurate (no changes needed)

- `min_*` / `max_*` / `allowed_outcomes` / `required_categories` rule semantics.
- `_get_metadata_field` suffix-match / decision-attribute fallback description.
- `update_policy()` return-value fix (`policy_id`, not a new version id) — this was already correct in the PR and was itself a genuine bug fix over the prior doc.
- "What Is / Why Use / When To Use" onboarding sections — clear, no correctness issues.

### Testing performed

Ran the PR's documented rule/metadata combinations directly against `semantica.context.policy_engine.PolicyEngine._evaluate_compliance` (not just read the diff) for all four domain examples (attribution, opsec/TLP:RED, PAM/MFA, clinical safety, mortgage) plus the isolated `required_*` and unknown-key fallback cases, before and after the fix, to confirm the printed `Compliant: True/False` outputs match real behavior.

PR #737, opened 2026-07-14
Add missing shacl extra to pyproject.toml

Sameer6305:
Fixes #736. Adds \`shacl = [\"pyshacl>=0.25.0\"]\` to [project.optional-dependencies] and folds it into the \`all\` extra. Verified \`pip install -e \".[shacl]\"\` now successfully installs pyshacl.

PR #738, opened 2026-07-14
Fix NodeEmbedder AttributeError masked in ContextGraph.analyze_graph_with_kg

### LINKED ISSUE #734: ContextGraph causal analysis silently fails: NodeEmbedder has no attribute 'generate_embeddings'
## Summary

`ContextGraph.analyze_graph_with_kg()` — called internally by both `get_causal_chain()`'s supporting analytics and `get_decision_insights()` when `advanced_analytics=True` — always fails with `AttributeError: 'NodeEmbedder' object has no attribute 'generate_embeddings'`, but a broad `except Exception` swallows it and returns a generic `{"error": "Graph analysis failed due to an internal error"}` instead of surfacing the real cause or succeeding.

## Steps to Reproduce

Matches Part 10 of `cookbook/use_cases/healthcare/01_Clinical_Intelligence_with_Semantica.ipynb`:

```python
from semantica.context import ContextGraph, AgentContext
from semantica.vector_store import VectorStore

vs = VectorStore(backend="inmemory", dimension=384)
graph = ContextGraph(advanced_analytics=True, enable_causality=True)

graph.add_node("diag_1", "Decision", content="diagnosis")
graph.add_node("treat_1", "Decision", content="treatment")
graph.add_edge("diag_1", "treat_1", edge_type="diagnosed_with")
graph.add_causal_relationship("diag_1", "treat_1", "CAUSED")

agent = AgentContext(vector_store=vs, knowledge_graph=graph, decision_tracking=True)
agent.record_decision(
    category="diagnosis_assessment", scenario="test", reasoning="test",
    outcome="test_outcome", confidence=0.99, decision_maker="test",
)

# Either of these triggers the bug:
chain = graph.get_causal_chain("treat_1", direction="upstream", max_depth=3)
insights = graph.get_decision_insights()
print(insights)
```

## Expected Behavior

`get_causal_chain()`/`get_decision_insights()` should either compute the advanced KG analysis (including node embeddings) successfully, or — if embeddings are genuinely unavailable for some legitimate reason (missing optional dependency, empty graph, etc.) — raise or return a specific, actionable error that lets the caller distinguish "embeddings unavailable" from "the graph has no data" or any other empty-result case.

## Actual Behavior

The call succeeds overall (no exception propagates to the caller), but the log output shows:

```
Failed to analyze graph with KG: 'NodeEmbedder' object has no attribute 'generate_embeddings'
```

and `analyze_graph_with_kg()` returns:

```python
{"error": "Graph analysis failed due to an internal error"}
```

Since `get_decision_insights()` merges this into `advanced_insights` without any further check, the caller receives a generic error dict indistinguishable from any other internal failure, with no indication that a specific method call is broken — and no way to tell this apart from a legitimately empty or unavailable analysis.

## Root Cause

In `semantica/context/context_graph.py`, `analyze_graph_with_kg()` (starting at line ~2164) calls:

```python
# Node embeddings
if "node_embedder" in self.kg_components:
    embeddings = self.kg_components["node_embedder"].generate_embeddings(kg_graph)
    analysis["node_embeddings"] = embeddings
```

But `NodeEmbedder` (`semantica/kg/node_embeddings.py:63`) has no `generate_embeddings` method at all — its actual public API is:

```python
def compute_embeddings(
    self,
    graph_store: Any,
    node_labels: List[str],
    relationship_types: List[str],
    embedding_dimension: Optional[int] = None,
    walk_length: Optional[int] = None,
    num_walks: Optional[int] = None,
    p: Optional[float] = None,
    q: Optional[float] = None,
) -> Dict[str, List[float]]:
```

This is not just a naming mismatch — `compute_embeddings` requires `graph_store`, `node_labels`, and `relationship_types` as separate arguments, none of which `analyze_graph_with_kg()` currently has available at the call site (it only has `kg_graph`, the converted KG-format dict). So this isn't a one-line rename; the call site needs to be restructured to supply the right arguments.

The `AttributeError` this produces is caught by the surrounding broad exception handler at lines ~2223-2226:

```python
except Exception as e:
    self.logger.error(f"Failed to analyze graph with KG: {e}")
    return {"error": "Graph analysis failed due to an internal error"}
```

which masks the real problem — every other failure mode inside this same `try` block (centrality, community detection, connectivity analysis) is caught by the identical `except Exception`, so any of those breaking in the future would produce the exact same generic, indistinguishable error message.

## Suggested Fix

Either:
1. Fix the call site to use `NodeEmbedder.compute_embeddings()` with the correct arguments (`graph_store`, `node_labels`, `relationship_types` derived from `self` / `kg_graph`), so node embeddings actually compute, **or**
2. If embeddings aren't meant to be wired into `analyze_graph_with_kg()` yet, remove the `node_embedder` branch entirely rather than leaving a call to a non-existent method, **and**
3. Regardless of which of the above, narrow the `except Exception` at lines ~2223-2226 (or at minimum add an `except AttributeError` branch that re-raises or logs distinctly) so a broken internal method call surfaces as a diagnosable error rather than being indistinguishable from a legitimate empty analysis result.

## Environment

- semantica: `0.5.1`
- Python: `3.12.6`

## Label suggestion

`bug`, `context-graph`

Sameer6305:
Fixes #734. NodeEmbedder has no generate_embeddings method; the actual API is compute_embeddings(graph_store, node_labels, relationship_types, ...). [Describe which fix direction you chose: restructured the call site vs removed the branch.] Also narrowed the surrounding except Exception so this class of bug surfaces distinctly going forward.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `pip install semantica[shacl]` referenced no matching extra in `pyproject.toml`, so `pyshacl` was never installed despite being documented as the fix in `ontology_validator.py`'s `ImportError` message, the Explorer API, the healthcare cookbook notebook, and the changelog
   - Added `shacl = ["pyshacl>=0.25.0"]` to `[project.optional-dependencies]` and folded `shacl` into the `all` extra
 
+- **`NodeEmbedder` `AttributeError` masked in `ContextGraph.analyze_graph_with_kg`** (#734) by @Sameer6305
+  - `analyze_graph_with_kg()` called a non-existent `NodeEmbedder.generate_embeddings()`, and the surrounding broad `except Exception` swallowed the resulting `AttributeError`, silently returning `{"error": "Graph analysis failed due to an internal error"}` from `get_causal_chain()`'s supporting analytics and `get_decision_insights()`
+  - Rewired the call site to the real `NodeEmbedder.compute_embeddings(graph_store, node_labels, relationship_types)` API, deriving `node_labels`/`relationship_types` from `self.node_type_index`/`self.edge_type_index`
+  - Added a dedicated `except AttributeError` branch that logs distinctly and re-raises, so a broken internal method call surfaces as a diagnosable error instead of being indistinguishable from a legitimately empty analysis result
+
 ---
 
 ## [0.5.1] - 2026-06-29
--- a/semantica/context/context_graph.py
+++ b/semantica/context/context_graph.py
@@ -2214,12 +2214,26 @@ class ContextGraph:
             
             # Node embeddings
             if "node_embedder" in self.kg_components:
-                embeddings = self.kg_components["node_embedder"].generate_embeddings(kg_graph)
-                analysis["node_embeddings"] = embeddings
+                node_labels = list(self.node_type_index.keys())
+                relationship_types = list(self.edge_type_index.keys())
+                if node_labels:
+                    embeddings = self.kg_components["node_embedder"].compute_embeddings(
+                        graph_store=self,
+                        node_labels=node_labels,
+                        relationship_types=relationship_types,
+                    )
+                    analysis["node_embeddings"] = embeddings
             
             self.logger.info("Completed comprehensive graph analysis")
             return analysis
             
+        except AttributeError as e:
+            # A broken internal method call (e.g. calling a method that doesn't
+            # exist on one of the kg_components) is a programming error, not a
+            # legitimate empty-analysis result. Log it distinctly and re-raise
+            # rather than masking it under the generic message below.
+            self.logger.error(f"Graph analysis failed due to a broken internal method call: {e}")
+            raise
         except Exception as e:
             self.logger.error(f"Failed to analyze graph with KG: {e}")
             return {"error": "Graph analysis failed due to an internal error"}


KaifAhmad1 (approved):
## Final Review: PR #738 — Fix NodeEmbedder AttributeError masked in `ContextGraph.analyze_graph_with_kg`

**Fixes:** #734
**Status:** Open, `MERGEABLE`, ready to merge
**Commits:**
1. `c6b805b` — Fix NodeEmbedder.generate_embeddings AttributeError in analyze_graph_with_kg (fixes #734)
2. `b5b2edb` — docs(changelog): add entry for NodeEmbedder AttributeError fix (#734) *(follow-up, rebased onto latest `main` after #737 merged)*

### Summary
`analyze_graph_with_kg()` called a non-existent `NodeEmbedder.generate_embeddings()`; the real API is `compute_embeddings(graph_store, node_labels, relationship_types, ...)`. A broad `except Exception` swallowed the resulting `AttributeError`, silently returning a generic error dict from both `get_causal_chain()`'s supporting analytics and `get_decision_insights()`. This PR rewires the call site to the correct API and adds a dedicated `except AttributeError` branch that logs distinctly and re-raises instead of masking the bug.

### Findings
- **Correctness:** ✅ Verified against the actual source, not just the diff — `compute_embeddings`'s signature (`semantica/kg/node_embeddings.py:147`) matches the new call exactly; `node_labels`/`relationship_types` are correctly derived from `self.node_type_index`/`self.edge_type_index`; `graph_store=self` is a valid duck-typed target since `ContextGraph` implements both `get_nodes_by_label` and `get_neighbors` with compatible signatures.
- **Edge cases:** ✅ The `if node_labels:` guard correctly avoids a spurious `RuntimeError` on empty graphs; nodes with no edges don't crash the random-walk generator (traced through `_generate_random_walks`/`_biased_random_walk`).
- **Behavioral contract change:** ⚠️ `get_decision_insights()` calls `analyze_graph_with_kg()` with no surrounding `try/except`, so a future broken `kg_components` method will now propagate as an uncaught `AttributeError` instead of the old always-returns-a-dict behavior. This is what the issue asked for, but it isn't documented on either method — worth a `Raises:` docstring note, non-blocking.
- **Test coverage:** ⚠️ No regression test added. The one existing test that exercises this path (`test_analyze_graph_with_kg_structure`) only asserts structure `if "error" not in analysis`, so it wouldn't have caught this bug or a reintroduction of it.
- **PR hygiene:** Minor — PR description originally left a template placeholder unfilled (`[Describe which fix direction you chose...]`).
- **Documentation:** ✅ Closed via follow-up commit adding a `### Fixed` changelog entry under `[Unreleased]`, attributed to @Sameer6305.

### Verdict
**Ready to merge.** Fix is correct and directly addresses the root cause in #734. Non-blocking follow-ups worth raising with the author: document the new exception-raising behavior, and add a regression test that would actually fail without this fix.

Issue #733, opened 2026-07-14
ExplanationGenerator.generate_explanation always produces empty premises


Sameer6305:
## Summary

`ExplanationGenerator.generate_explanation()` always produces an empty premises list for `InferenceResult` objects returned by `Reasoner.forward_chain()`, so the resulting `.natural_language` explanation reads "Given the premises: , we conclude: ..." with nothing between the colon and the comma — even when the underlying facts that triggered the rule are known and available.

## Steps to Reproduce

```python
from semantica.reasoning import Reasoner, ExplanationGenerator

reasoner = Reasoner()
reasoner.add_rule("IF has_disease(?p, Diabetes) AND hba1c_above_8(?p) THEN recommend(?p, IntensiveTherapy)")
reasoner.add_fact("has_disease(MRN-SYN-0001, Diabetes)")
reasoner.add_fact("hba1c_above_8(MRN-SYN-0001)")

result = reasoner.forward_chain()
explanation = ExplanationGenerator().generate_explanation(result[0])
print(explanation.natural_language)
```

## Expected Behavior

The explanation should include the specific facts that satisfied the rule's conditions, e.g. something like:

```
Given the premises: has_disease(MRN-SYN-0001, Diabetes), hba1c_above_8(MRN-SYN-0001), we conclude: recommend(MRN-SYN-0001, IntensiveTherapy) using rule 'Rule 1'.
```

## Actual Behavior

```
Given the premises: , we conclude: recommend(MRN-SYN-0001, IntensiveTherapy) using rule 'Rule 1'.
```

The premises list is always empty for forward-chained results, regardless of which facts actually matched. This isn't intermittent or state-dependent — it happens on every call, on a freshly created `Reasoner`, every time.

## Root Cause

In `semantica/reasoning/reasoner.py`:

- `Reasoner._match_rule()` (line ~306) only returns the instantiated conclusion strings, discarding the facts that were matched to produce each binding:
  ```python
  def _match_rule(self, rule: Rule) -> List[str]:
      """Match rule conditions against facts and return instantiated conclusions."""
      ...
      results = []
      for bindings in bindings_list:
          instantiated_conclusion = self._substitute(rule.conclusion, bindings)
          results.append(instantiated_conclusion)
      return results
  ```
[...]

PR #740, opened 2026-07-15
Fix Reasoner.add_rule missing deduplication (#732)

### LINKED ISSUE #732: Reasoner.add_rule has no deduplication — forward_chain() silently returns empty results on rerun
## Summary

`Reasoner.add_rule()` has no deduplication, so re-running a notebook cell that calls `add_rule()` + `add_fact()` on an existing `Reasoner` instance duplicates every rule. Because `forward_chain()` only records a conclusion "if not already in `self.facts`," the second call silently returns `[]` even though the rules and facts are still logically valid — there's no error, just an empty result.

## Steps to Reproduce

Minimal repro, matching `cookbook/use_cases/healthcare/01_Clinical_Intelligence_with_Semantica.ipynb`, Part 9 (cell `7f059591`):

```python
from semantica.reasoning import Reasoner

PATIENTS = [
    {"mrn": "MRN-SYN-0001", "diagnoses": ["Type 2 Diabetes"], "hba1c": 8.4, "egfr": 72},
    {"mrn": "MRN-SYN-0002", "diagnoses": ["Type 2 Diabetes"], "hba1c": 7.1, "egfr": 28},
]

reasoner = Reasoner()

def run_cell():
    reasoner.add_rule("IF has_disease(?p, Diabetes) AND hba1c_above_8(?p) THEN recommend(?p, IntensiveTherapy)")
    reasoner.add_rule("IF has_disease(?p, Diabetes) AND egfr_below_30(?p) THEN contraindicated(?p, Metformin)")

    for p in PATIENTS:
        if "Type 2 Diabetes" in p["diagnoses"]:
            reasoner.add_fact(f"has_disease({p['mrn']}, Diabetes)")
            if p.get("hba1c") and p["hba1c"] > 8:
                reasoner.add_fact(f"hba1c_above_8({p['mrn']})")
            if p.get("egfr") and p["egfr"] < 30:
                reasoner.add_fact(f"egfr_below_30({p['mrn']})")

    return reasoner.forward_chain()

# Simulates a Jupyter user re-running the cell without restarting the kernel
result1 = run_cell()
print("Run 1 rule count:", len(reasoner.rules), "-> derived:", [d.conclusion for d in result1])

result2 = run_cell()
print("Run 2 rule count:", len(reasoner.rules), "-> derived:", [d.conclusion for d in result2])
```

## Expected Behavior

`forward_chain()` should return the same derived facts on both calls (`['recommend(MRN-SYN-0001, IntensiveTherapy)', 'contraindicated(MRN-SYN-0002, Metformin)']`), or `add_rule()` should detect and skip an identical rule already present in `self.rules`, so re-running the same cell/setup code is idempotent.

## Actual Behavior

```
Run 1 rule count: 2 -> derived: ['recommend(MRN-SYN-0001, IntensiveTherapy)', 'contraindicated(MRN-SYN-0002, Metformin)']
Run 2 rule count: 4 -> derived: []
```

The rule count doubles (2 → 4) on the second call, and `forward_chain()` returns an **empty list** on the second run — not an error, not a warning, just silently empty — because the conclusions from run 1 are already present in `self.facts`, so the `if conclusion not in self.facts` guard in `forward_chain()` suppresses them even though the (now-duplicated) rules still match.

## Root Cause

In `semantica/reasoning/reasoner.py`:

- `Reasoner.add_rule()` (line ~78) unconditionally appends to `self.rules` with no de-dup check:
  ```python
  def add_rule(self, rule_def: Union[str, Rule]) -> Rule:
      """Add a rule to the reasoner."""
      if isinstance(rule_def, Rule):
          rule = rule_def
      else:
          rule = self._parse_rule_definition(rule_def)
      self.rules.append(rule)
      self.rules.sort(key=lambda r: r.priority, reverse=True)
      return rule
  ```
- `Reasoner.forward_chain()` (line ~166) only appends a result "if conclusion not in self.facts":
  ```python
  for rule in self.rules:
      matches = self._match_rule(rule)
      for conclusion in matches:
          if conclusion not in self.facts:
              self.facts.add(conclusion)
              results.append(InferenceResult(...))
              new_facts_added = True
  ```
  Since `self.facts` persists across calls on the same instance, and the conclusion strings from run 1 are already in `self.facts`, run 2's (duplicated) rules produce matches that are filtered out entirely, so `results` stays empty.

This is a compounding bug: rules silently accumulate duplicates on every re-run (unbounded growth if repeated many times), *and* `forward_chain()`'s output becomes misleadingly empty rather than reflecting the same valid derivation, with no exception or log message to signal either problem.

## Environment

- semantica: `0.5.1`
- Python: `3.12.6`
- OS: Windows 11 Home Single Language (Build 26100)

## Suggested Fix

Either:
1. De-duplicate in `add_rule()` — skip appending if an equivalent `Rule` (same conditions + conclusion) already exists in `self.rules`, or
2. Make `forward_chain()` idempotent regardless of rule duplication by deriving from a fresh working set each call rather than mutating `self.facts` in place, or
3. At minimum, log a warning when `add_rule()` detects a duplicate rule definition, so the silent-empty-result behavior isn't the only signal something is wrong.

## Label suggestion

`bug`, `reasoning`

KaifAhmad1:
## Summary
- `Reasoner.add_rule()` unconditionally appended to `self.rules`, so re-running the same setup code on an existing `Reasoner` instance (e.g. re-executing a Jupyter cell) duplicated every rule
- Because `forward_chain()` only records a conclusion if it isn't already in `self.facts`, the second run's duplicated rules matched but produced no new results — silently, with no error or warning
- `add_rule()` now compares an incoming rule's `rule_type`, `conditions`, and `conclusion` against existing rules and returns the existing `Rule` instead of appending a duplicate, so repeated `add_rule()` calls with the same definition are idempotent

Closes #732

## Test plan
- [x] Added `test_add_rule_deduplicates_identical_rule` — re-adding an identical rule string does not duplicate it
- [x] Added `test_add_rule_deduplication_is_idempotent_across_forward_chain` — mirrors the issue's exact repro (rerunning `add_rule()`+`add_fact()`+`forward_chain()` on the same instance keeps the rule count stable)
- [x] Added `test_add_rule_does_not_dedupe_distinct_rules` — rules with different conditions/conclusions are still both added
- [x] `python -m pytest tests/reasoning/ -q` — 41 passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`Reasoner.add_rule` had no deduplication, doubling rules and silently emptying `forward_chain()` on rerun** (#732) by @KaifAhmad1
+  - `add_rule()` unconditionally appended to `self.rules`, so re-running the same setup code on an existing `Reasoner` instance (e.g. re-executing a Jupyter cell) duplicated every rule; since `forward_chain()` only records a conclusion if it isn't already in `self.facts`, the second run's duplicated rules matched but produced no new results, with no error or warning
+  - `add_rule()` now compares an incoming rule's `rule_type`, `conditions`, and `conclusion` against existing rules and returns the existing `Rule` instead of appending a duplicate, keeping repeated `add_rule()` calls with the same definition idempotent
+  - Added `test_add_rule_deduplicates_identical_rule`, `test_add_rule_deduplication_is_idempotent_across_forward_chain`, and `test_add_rule_does_not_dedupe_distinct_rules` regression tests
+
 - **`InferenceResult.premises` always empty from `forward_chain`/`backward_chain`** (#739) by @Sameer6305
   - `_match_rule()` discarded matched facts and returned only instantiated conclusions, so `ExplanationGenerator` always produced empty premises lists regardless of which facts actually satisfied a rule, closing #733
   - `_match_rule()` now returns `(conclusion, matched_facts)` tuples; `forward_chain()` threads those facts into `InferenceResult(premises=...)`, merging premises when the same conclusion is derived more than once within a pass
--- a/semantica/reasoning/reasoner.py
+++ b/semantica/reasoning/reasoner.py
@@ -104,6 +104,11 @@ class Reasoner:
                     f"Skipping duplicate rule (same conditions/conclusion as '{existing.rule_id}'): "
                     f"IF {' AND '.join(map(str, rule.conditions))} THEN {rule.conclusion}"
                 )
+                # Rule is a mutable dataclass, so `existing.priority` may have
+                # changed since it was added -- re-sort so the dedup path
+                # keeps the same self-healing ordering the append path has,
+                # rather than leaving self.rules stale relative to priority.
+                self.rules.sort(key=lambda r: r.priority, reverse=True)
                 return existing
 
         self.rules.append(rule)
--- a/semantica/reasoning/reasoner.py
+++ b/semantica/reasoning/reasoner.py
@@ -81,12 +81,31 @@ class Reasoner:
         self.rule_counter = 0
             
     def add_rule(self, rule_def: Union[str, Rule]) -> Rule:
-        """Add a rule to the reasoner."""
+        """Add a rule to the reasoner.
+
+        Rules with the same conditions and conclusion as an already-added
+        rule are not re-appended -- this keeps re-running the same setup
+        code (e.g. a Jupyter cell that calls add_rule() + add_fact() on an
+        existing Reasoner) idempotent instead of silently duplicating rules
+        on every rerun (#732).
+        """
         if isinstance(rule_def, Rule):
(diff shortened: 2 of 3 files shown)


KaifAhmad1:
@Sameer6305 Ready for the review.

Sameer6305 (approved):
Reviewed locally — pulled the branch, confirmed both the isolated #732 repro and the real cookbook notebook cell (Part 9, cell 34) stay stable across re-execution in the same kernel. Confirmed this branch is built on top of #733's already-merged premises fix, verified via merge-base and a full test run with both fixes stacked — no overlap, no regressions.

Probed three edge cases beyond the existing test plan: condition reordering and whitespace variance are both treated as distinct rules (conservative failure mode — redundant rules, not wrong inference — left as documented limitations). Re-adding an identical rule with a different confidence value silently discarded the new value with no warning, which could genuinely surprise a caller, so pushed a small follow-up commit adding a warning log + docstring clarification for that specific case (no change to dedup logic itself). 45/45 tests pass after that addition.

Looks solid — approving.

PR #741, opened 2026-07-15
Fix ProvenanceManager.get_lineage not linking entities via derived_from

### LINKED ISSUE #735: ProvenanceManager.get_lineage does not link entities that share a source URL
## Summary

`ProvenanceManager.track_entity(entity_id, source=<url>, metadata={"derived_from": <id>})` does not create a traceable lineage chain even when two entities are conceptually linked via the same real source URL — `get_lineage()` returns a chain of length 1 and never includes the entity that was meant to be the "source of" relationship, because the auto-linking logic treats `source` as an `entity_id` lookup, not a shared-URL match, and the `derived_from` metadata key is never consulted at all.

## Steps to Reproduce

```python
from semantica.provenance import ProvenanceManager

prov = ProvenanceManager()

# Track the source document
prov.track_entity(
    entity_id="doc:X",
    source="https://example.com/api",
    metadata={"content_type": "drug_label"},
)

# Track a decision that was "derived from" doc:X, citing the same real URL
prov.track_entity(
    entity_id="decision:Y",
    source="https://example.com/api",
    metadata={"derived_from": "doc:X"},
)

lineage = prov.get_lineage("decision:Y")
print("entity_count:", lineage["entity_count"])
print("entity_ids in chain:", [e["entity_id"] for e in lineage["lineage_chain"]])
```

## Expected Behavior

`get_lineage("decision:Y")` should walk back to `doc:X` — either because `track_entity` recognizes both entries share the same `source` URL, or because it reads `metadata["derived_from"]` as an explicit parent link — so the chain includes both entities and a caller can answer "what document justified this decision?" by traversing the returned lineage.

## Actual Behavior

```
entity_count: 1
entity_ids in chain: ['decision:Y']
```

`doc:X` is never included. `get_lineage` returns only the single entity it was called on. The `derived_from` value in `metadata` is preserved and returned as part of the entry's `metadata` dict in the output JSON, so it *looks* present in the printed lineage — but it plays no functional role in the traversal.

## Root Cause

In `semantica/provenance/manager.py`, `track_entity()`'s only auto-linking logic (lines ~117-125) is:

```python
# Check if entity already exists
existing = self.storage.retrieve(entity_id)
parent_id = kwargs.get("parent_entity_id")

# If source is a known entity, link it as parent (unless parent already set)
if not parent_id and source and isinstance(source, str):
    try:
        # Check if source exists in storage
        source_entity = self.storage.retrieve(source)
        if source_entity:
            parent_id = source
    except Exception:
        pass
```

This calls `self.storage.retrieve(source)` — i.e., it looks up `source` **as if it were an `entity_id`**. Since both calls in the repro pass the literal string `"https://example.com/api"` as `source`, and no entity was ever stored *under that URL as its `entity_id`*, `self.storage.retrieve("https://example.com/api")` returns `None`, so `parent_id` stays `None`. The fact that `doc:X` and `decision:Y` share the same `source` value is coincidental string duplication from the caller's perspective — nothing in `track_entity` compares `source` values across existing entries to find entities that share a source.

Separately, `metadata["derived_from"]` is stored as-is in the `ProvenanceEntry.metadata` dict but is never read by any linking or traversal code. `ProvenanceManager.trace_lineage()` (`manager.py:491-500`) delegates directly to `storage.trace_lineage()`, and `InMemoryStorage.trace_lineage()` (`semantica/provenance/storage.py:163-199`) only walks `parent_entity_id` and `used_entities`:

```python
def trace_lineage(self, entity_id: str) -> List[ProvenanceEntry]:
    ...
    while queue:
        current_id = queue.popleft()
        if current_id in visited:
            continue
        visited.add(current_id)
        entry = self.retrieve(current_id)
        if entry:
            lineage.append(entry)
            # Add parent entity to queue
            if entry.parent_entity_id:
                queue.append(entry.parent_entity_id)
            # Add used entities to queue
            for used_id in entry.used_entities:
                if used_id not in visited:
                    queue.append(used_id)
```

Neither field is populated from `metadata["derived_from"]` anywhere in `track_entity`, so the BFS has no edge to follow even though the caller explicitly declared the relationship.

## Suggested Fix

Either:
1. Have `track_entity` treat `metadata["derived_from"]` (or a dedicated `parent_entity_id` kwarg, which already exists per `kwargs.get("parent_entity_id")`) as a real parent link — i.e., if `metadata` contains `derived_from` and no explicit `parent_entity_id` was passed, set `parent_id = metadata["derived_from"]` before constructing the `ProvenanceEntry`, so `trace_lineage`'s existing BFS picks it up for free, **or**
2. If URL-based/metadata-based linking is intentionally out of scope, clearly document that `source` must be an existing entity's `entity_id` — not an arbitrary URL or DOI string — for automatic parent-linking to activate, and that `derived_from` (or similar keys) in `metadata` are inert unless the caller separately passes `parent_entity_id`. This should also prompt an update to any notebooks or docs that currently pass a raw URL as `source` while expecting a functional lineage chain (e.g. `cookbook/use_cases/healthcare/01_Clinical_Intelligence_with_Semantica.ipynb`, Part 11, currently does exactly this and its markdown claims a "citable lineage chain" that the code as written does not produce).

Option 1 is preferable since it makes the documented `metadata={"confidence": 0.92}`-style example in the docstring consistent with an already-common calling pattern (passing a `derived_from`-style key) without requiring every caller to separately know about the `parent_entity_id` kwarg.

## Environment

- semantica: `0.5.1`
- Python: `3.12.6`

## Label suggestion

`bug`, `provenance`

KaifAhmad1:
## Summary
- `track_entity()`'s only auto-linking logic looked up `source` as if it were an existing entity's `entity_id`, so two entities sharing a real source URL/DOI (e.g. a document and a decision derived from it) never got a parent link, leaving `get_lineage()` returning a chain of length 1
- `metadata["derived_from"]` was stored and echoed back in the output JSON but never consulted by any linking or traversal code
- `track_entity()` now treats `metadata["derived_from"]` as an explicit parent link (unless `parent_entity_id` was already passed directly), so `InMemoryStorage.trace_lineage()`'s existing BFS over `parent_entity_id` picks it up for free

Closes #735

## Test plan
- [x] Reproduced the issue's exact repro before the fix (`entity_count: 1`) and confirmed it now returns `entity_count: 2` with both `decision:Y` and `doc:X` in the chain
- [x] Added 7 tests to `tests/provenance/test_manager.py`: happy path, explicit `parent_entity_id` precedence over `derived_from`, precedence over the `source`-as-known-entity-id fallback, `derived_from` pointing at a never-tracked entity, non-string/empty-string `derived_from` ignored, self-referencing `derived_from` doesn't hang traversal, multi-hop `derived_from` chains
- [x] `pytest tests/provenance/ tests/kg/test_provenance_workflows.py` — same 11 pre-existing/unrelated failures as on `main` (verified via `git stash` comparison), 212 passed (up from 203 baseline)


KaifAhmad1:
@Sameer6305 Ready for the review

Sameer6305 (approved):
Reviewed locally — pulled the branch including the CHANGELOG merge with #732, re-ran the full provenance suite after syncing to the maintainer's latest push: 11 failed / 205 passed / 32 skipped, same 11 failures as main's baseline (byte-for-byte identical test names and error messages — legacy kg.ProvenanceTracker shim and an unrelated float-precision assertion, nothing touching ProvenanceManager).

Verified the core fix two ways:
1. The issue's own repro as literally written has a small typo (accesses lineage['chain'], actual key is lineage_chain) — corrected and confirmed entity_count=2, both decision:Y and doc:X present in the chain.
2. Replayed the real cookbook notebook's Part 2 + Part 11 cells verbatim (FDA label doc + treatment decision, same openFDA URL, derived_from metadata) against this branch's provenance module — confirmed entity_count=2 with doc:fda_label_metformin genuinely present in the decision's lineage now, traced via derived_from rather than coincidental URL string matching. This is exactly the real-world gap the original audit flagged.

Confirmed zero file-level overlap with #732/#733 — this branch diverged from main after that work already merged, touches only CHANGELOG.md, semantica/provenance/manager.py, and tests/provenance/test_manager.py.

Probed 5 edge cases beyond the stated test plan: 5-hop chains, parent_entity_id vs derived_from conflicts (loser fully discarded, no leakage into the chain), multi-node cycles (BFS visited-set handles arbitrary cycle length correctly, not just the tested 1-node self-reference), and cross-entity-type links all behave correctly. One real gap found: re-tracking an already-tracked entity_id silently overwrites parent_entity_id/derived_from with an auto-generated history link, discarding the caller's new value with no warning — confirmed this is pre-existing general re-track behavior (also affects the parent_entity_id kwarg identically) that predates this PR, not a regression introduced here, so filed as a separate issue rather than expanding this PR's scope: <https://github.com/semantica-agi/semantica/issues/742>.

Solid fix, closes a real gap. Approving.

PR #789, opened 2026-07-23
Fix #788: pin httpx<0.28.0 globally to fix Explorer test suite TestClient breakage

### LINKED ISSUE #788: tests/explorer/test_explorer_api.py fails with TypeError: Client.__init__() got an unexpected keyword argument 'app' on current httpx
## Summary
Running the Explorer backend test suite currently produces 58 errors, all with the
identical stack trace:

TypeError: Client.__init__() got an unexpected keyword argument 'app'
..\AppData\Roaming\Python\Python312\site-packages\starlette\testclient.py:402

## Root cause
httpx >= 0.28.0 removed/changed the app= kwarg that Starlette's TestClient relies on
to wrap a FastAPI app for testing. Confirmed this is a real, pre-existing
version-compatibility break by comparing test results with git stash: identical
failure count on both a clean checkout and a branch with unrelated changes.

## Verified workaround
Downgrading to httpx<0.28.0 locally resolves it - the analytics tests then pass
cleanly. Not committing this as a fix since it's a dependency/lockfile decision, not
a code fix, and deserves the maintainer's call on the right pin.

## Impact
This currently blocks running the Explorer backend test suite at all for anyone on a
recent httpx install - worth prioritizing since it affects the reliability of every
future PR that touches semantica/explorer/.

Sameer6305:
Closes #788.

## Summary
httpx>=0.28.0 removed the app= kwarg that Starlette's TestClient relies on, breaking
every test that spins up a FastAPI TestClient. httpx wasn't pinned anywhere in
pyproject.toml, so different environments could resolve different transitive
versions and hit this independently.

## Change
Added an explicit httpx<0.28.0 constraint to the main [project.dependencies] array
(not just a dev extra), per maintainer's direction - this makes it a global
constraint across production, dev, and CI environments, closing the "different
environments resolve different versions" gap.

## Verification
- Confirmed the exact failure before the fix: pip install "httpx>=0.28.0" reproduces
  58 errors, all the same TypeError: Client.__init__() got an unexpected keyword
  argument 'app'.
- After the pin: tests/explorer/test_explorer_api.py goes from 7 failed/12 passed/58
  errors to 77 passed, 0 errors.
- Ran git stash to compare against the unmodified baseline: without this pin, the
  full test suite can't even complete collection (fails immediately on
  tests/explorer/test_vocabulary.py with the same TestClient error). So this isn't
  just a test fix - it's what allows the full suite to run at all. The 188 failures
  visible after the pin are pre-existing issues elsewhere in the codebase, not
  introduced by this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`tests/explorer/test_explorer_api.py` failed with `TypeError: Client.__init__() got an unexpected keyword argument 'app'` on current httpx** (#788, #789) by @Sameer6305
+  - `httpx>=0.28.0` removed the `app=` kwarg that Starlette's `TestClient` relies on to wrap a FastAPI app for testing; `httpx` wasn't pinned anywhere in `pyproject.toml`, so different environments could independently resolve an incompatible transitive version and hit the same break
+  - Added an explicit `httpx<0.28.0` constraint to the main `[project.dependencies]` array (not just a dev extra), so it applies globally across production, dev, and CI installs
+  - Without the pin, the full test suite fails to even complete collection (fails immediately on `tests/explorer/test_vocabulary.py` with the same `TestClient` error); with it, `tests/explorer/test_explorer_api.py` goes from 7 failed/12 passed/58 errors to 77 passed, 0 errors
+
 ## [0.6.0] - 2026-07-21
 
 ### Added
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,7 +85,8 @@ dependencies = [
   "python-dotenv>=1.2.1",
   "loguru>=0.7.3",
   "structlog>=22.1.0",
-  "gensim>=4.4.0"
+  "gensim>=4.4.0",
+  "httpx<0.28.0"
 ]
 
 [project.urls]


KaifAhmad1 (approved):
## PR #789 Review — Pin `httpx<0.28.0` globally

**Fix #788: Explorer test suite `TestClient` breakage**

### Summary
- 1-line dependency pin (`pyproject.toml`): `"httpx<0.28.0"` added to `[project.dependencies]`
- Fixes `TypeError: Client.__init__() got an unexpected keyword argument 'app'` caused by `httpx>=0.28.0` dropping the `app=` kwarg Starlette's `TestClient` needs

### Findings
- ✅ **Root cause matches fix** — confirmed against issue #788's documented repro and workaround
- ✅ **No conflicts** — no prior `httpx` pin existed anywhere in `pyproject.toml`; no lockfile to reconcile; `llm-openai`/`llm-anthropic` extras don't impose a conflicting lower bound
- ✅ **Global (not dev-extra) placement is the right call** — closes the "different environments resolve different transitive versions" gap; matches Qodo bot's own recommendation
- ✅ **CI green** — all 9 checks passing (Checkov, Analyze Python, GitGuardian, MSDO, build, eslint, security-scan, templateanalyzer, terrascan), no merge conflicts
- ⚠️ **Unverified attribution** — PR body claims the global-vs-dev-extra placement was "per maintainer's direction," but no comment from @KaifAhmad1 exists on #788 or #789 giving that direction (Sameer's only ask was an open question tagging the maintainer). Likely given out-of-band — flagged, not blocking.

### Verdict
**Approve and merge.** Fix is correct, minimal, and matches the verified workaround already on record in #788.

### Follow-up done
- Added CHANGELOG entry under `[Unreleased] → Fixed` (#788, #789), pushed as a second commit to `fix/788-httpx-pin`

PR #806, opened 2026-07-27
fix(security): restrict Neptune cookbook SG, add VPC flow logs, harden IaC scan suppressions

KaifAhmad1:
## Summary

Addresses the open GitHub code scanning alerts (terrascan / checkov via Microsoft Defender for DevOps) on `main`:

- **#28 (open port)**: `cookbook/introduction/neptune-setup.yaml` no longer opens the Neptune Bolt/OpenCypher port (8182) to `0.0.0.0/0`. A required `ClientCidr` parameter (no default, CIDR-validated) must now be supplied at stack creation. Updated `21_Amazon_Neptune_Store.ipynb`'s deploy instructions and cost table to match.
- **#26 (no VPC flow logs)**: added `AWS::EC2::FlowLog` + a CloudWatch Logs group + IAM role to the same VPC.
- **#27 (IAM password policy)**: intentionally *not* implemented as a stack resource — `AWS::IAM::AccountPasswordPolicy` is an account-wide singleton, and wiring it into a disposable per-learner tutorial stack would mean deploying/deleting the stack mutates or removes the account's real password policy. Suppressed with a documented `ts:skip=AC_AWS_0148` explaining the reasoning.
- **#846/#847/#848 (checkov) and #68/#63 (terrascan)** — default-namespace and seccomp findings on the `knowledge-explorer` Helm chart templates. These are scanner false positives: `{{ .Release.Namespace }}` and `{{ toYaml .Values.podSecurityContext }}` only resolve at real `helm install`/`helm template` time, but the CI scan appears to render/parse the raw template. A prior fix added a `checkov.io/skip1` metadata annotation, but the alerts are still open on the current `main`, so this PR adds the more standard `# checkov:skip=` / `# ts:skip=` inline comment suppressions (matching the convention already used in `deploy/gcp/cloudrun-service.yaml`) as a second suppression path, and documents the reasoning in `.checkov.yaml`.
- **#61 (TLS) and the non-helm seccomp variant** on `deploy/kubernetes/*` were already fixed in the current tree (TLS block present, seccompProfile set) — no code change needed there; they should close on the next scan.

## Notes / residual risk

The inline-comment suppressions for the Helm chart findings are a best-effort second suppression mechanism — I could not run checkov/terrascan locally in this environment to confirm they're honored by the CI scanner. If those specific alerts (#846-848, #68, #63) are still open after this PR's scan runs, the reliable fallback is splitting the CI checkov/terrascan invocation so `deploy/helm/` is scanned with those specific checks excluded via `--skip-check`, rather than relying on in-file suppression.

## Test plan

- [ ] Confirm `.checkov.yaml` / workflow-driven scan (Microsoft Defender for DevOps) re-runs on this PR and reports which alerts close
- [ ] `helm template deploy/helm/knowledge-explorer` renders without error and namespace/seccomp fields resolve as expected
- [ ] Manually validate `neptune-setup.yaml` with `aws cloudformation validate-template` (or deploy in a sandbox account) to confirm `ClientCidr` parameter and Flow Log resources are well-formed


Sameer6305 (approved):
**Approved** :white_check_mark:

Great work on hardening the Neptune CloudFormation stack and resolving the scanner alerts. As part of our final review, we pushed a follow-up commit (`f9fe3d23`) to ensure all scanner suppressions and IAM policies are fully hardened and conform to official AWS and scanner specifications:

### Summary of Polish & Hardening (`f9fe3d23`)

1. **AWS VPC Flow Logs Least-Privilege IAM Policy (`cookbook/introduction/neptune-setup.yaml`)**
   - Split `FlowLogRole` into two distinct statements per [AWS VPC Flow Logs IAM documentation](https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs-iam-role.html):
     - **Account-level actions** (`logs:CreateLogGroup`, `logs:DescribeLogGroups`, `logs:DescribeLogStreams`) scoped to `Resource: "*"`.
     - **Log-group actions** (`logs:CreateLogStream`, `logs:PutLogEvents`) scoped cleanly to `!GetAtt FlowLogGroup.Arn` (removing the invalid `:log-stream:*` ARN suffix).
2. **Correct Checkov Seccomp Rule ID (`deployment.yaml` & `.checkov.yaml`)**
   - Updated the seccomp suppression ID from `CKV_K8S_28` to **`CKV_K8S_31`** (Checkov's actual Kubernetes `seccompProfile` check ID from `Seccomp.py`).
   - Applied the seccomp suppression annotations to both the top-level Deployment `metadata.annotations` and the pod-template `spec.template.metadata.annotations` as a belt-and-suspenders measure for static template scanners.
3. **Standardized K8s/Helm Scanner Annotations (`deployment.yaml`, `configmap.yaml`, `service.yaml`)**
   - Removed stale `# checkov:skip=` and `# ts:skip=` file-header comments across all three chart templates.
   - Standardized all suppressions on resource-level `metadata.annotations` (`checkov.io/skip...` and `runterrascan.io/skip`) so Microsoft Defender for DevOps and Terrascan reliably evaluate rule exclusions across every resource.

### Verification Performed
- **Helm Rendering:** `helm template deploy/helm/knowledge-explorer` rendered cleanly without indentation or YAML map conflicts.
- **CloudFormation Syntax:** `aws cloudformation validate-template` syntax and CloudFormation tag (`!GetAtt`, `!Ref`, `!Sub`, etc.) schema validation passed without errors.

PR #809, opened 2026-07-28
fix(explorer): wire ProvenanceManager into provenance routes (closes #792)

### LINKED ISSUE #792: Explorer's Provenance UI uses a naive 2-hop graph traversal instead of the audit-grade ProvenanceManager backend
## Summary
semantica/explorer/routes/provenance.py never imports or calls ProvenanceManager
(semantica/provenance/manager.py) anywhere. It builds its own provenance response
independently via a naive 2-hop networkx graph traversal instead of querying the
SQLite-backed, checksummed audit log the core provenance module actually provides.

## Confirmed code
_build_provenance (routes/provenance.py:32 onward):
```python
def _build_provenance(session: GraphSession, node_id: Optional[str] = None) -> dict:
    if not node_id or node_id not in session.graph.nodes:
        return {"nodes": [], "edges": []}
    graph = nx.DiGraph()
    # ... builds graph ...
    subgraph = nx.ego_graph(graph, node_id, radius=2, undirected=True)
```
Only imports networkx - no import of ProvenanceManager or anything from
semantica/provenance/ anywhere in the file. Confirmed #770's changes did not touch
this file, so this gap is still current.

## Why this matters
The frontend's provenance UI (LineageDiagram.tsx / ProvenanceTab.tsx) currently
visualizes a live-graph traversal, not actual audit-trail data - meaning the
"traceability" story in Explorer isn't using the strongest part of what's already
built (checksummed history, verifiable lineage chains).

Sameer6305:
Closes #792

## What was asked

`semantica/explorer/routes/provenance.py` never imported or called
`ProvenanceManager` (`semantica/provenance/manager.py`) anywhere — it built
its provenance response entirely from a naive 2-hop networkx traversal over
the live graph, instead of querying the SQLite-backed, checksummed audit
log the core provenance module actually provides.

## What this PR implements

**1. Audit-grade lineage as the primary data source**
`/api/provenance` and `/api/provenance/report` now query
`session.provenance_manager.get_lineage(node_id)` first. When real audit
records exist, the response is built from verified `ProvenanceEntry` data
(with checksums, source documents, confidence scores) instead of a graph
walk. A new `_transform_audit_lineage()` maps W3C PROV-O entries into the
exact `{"nodes": [...], "edges": [...]}` shape `LineageDiagram.tsx` already
expects — no frontend changes required.

**2. Graceful, disclosed fallback — never silent, never a crash**
- No audit records for a node → falls back to the original 2-hop
  traversal, unchanged.
- `ProvenanceManager` storage failure (corrupted DB, permissions, etc.) →
  caught specifically, logged, falls back to the naive path. Never 500s.
- A new `source: "audit" | "graph_traversal"` field on the response lets
  callers (and tests) see which path actually served the data.

**3. Per-session provenance storage, not global mutable state**
Initial approach called `ProvenanceManager.set_default_storage_path()`
(a classmethod mutating shared class state) from `create_app()` — caught
in review before merge, since it would mean any two sessions/apps in the
same process silently share (and overwrite) each other's storage path,
including across unrelated test runs. Fixed: `provenance_storage_path` is
threaded through `GraphSession.__init__` and `create_app(...)`, so each
session's `ProvenanceManager` is independently scoped. Verified with a
same-process, two-session isolation test and a full test-suite run in both
execution orders (`tests/explorer/` then `tests/provenance/`, and reversed)
to confirm zero cross-test pollution.

**4. Disclosed limitation, not a silent gap**
`ProvenanceManager.trace_lineage()`/`get_lineage()` only walk
`parent_entity_id`/`used_entities` backward — confirmed in both
`InMemoryStorage` and `SQLiteStorage` — so there's no reverse lookup for
descendants anywhere in the current API. This means the audit path can
show upstream lineage only; the naive fallback remains the only source for
downstream/descendant relationships until `ProvenanceManager` gains a
reverse lookup. Documented inline at the exact code path affected rather
than left as a silently-unreachable branch. Follow-up issue drafted
separately (not filed as part of this PR).

## Testing

New: `tests/explorer/test_provenance_manager_wiring.py` — 5 tests covering
the audit path (real multi-hop `track_entity()` chain), empty-record
fallback, simulated storage-failure degradation (asserts `200`, not `500`),
`create_app()` storage-path wiring, and cross-session storage isolation.

- `tests/explorer/` + `tests/provenance/`: 418 passed, 0 failed —
  confirmed order-invariant (ran both execution orders).
- Full repo suite (`tests/`): all failures are pre-existing, unrelated to
  this change (missing optional deps — spaCy/Ollama/Pinecone/Snowflake
  credentials in CI-less local environment).
- All existing provenance tests pass unmodified — no breaking changes to
  the response shape existing callers rely on.


Sameer6305:
@KaifAhmad1 ready for review.

KaifAhmad1 (approved):
## PR #809 Review — `fix(explorer): wire ProvenanceManager into provenance routes (closes #792)`

**Repo:** semantica-agi/semantica · **Branch:** `Sameer6305:fix/792-provenance-manager-wiring` → `main` · **+438/-24, 7 files**

### Overview
- Wires the audit-grade `ProvenanceManager` (SQLite-backed, checksummed) into `/api/provenance` and `/api/provenance/report`, which previously only did a naive 2-hop `networkx` traversal.
- Adds `_transform_audit_lineage()` mapping W3C PROV-O entries into the existing `LineageDiagram.tsx` node/edge shape — verified no frontend changes are actually required.
- Adds a disclosed `source: "audit" | "graph_traversal"` field so callers/tests know which path served the data.

### Correctness
- Audit-path wiring, checksum-integrity gating, and fallback-never-crashes logic are all correct — traced against `ProvenanceEntry`, `SQLiteStorage`, and the frontend consumer directly, not just the diff.
- `ProvenanceManager.get_lineage()` now returns `integrity_verified`; a single tampered/corrupted entry anywhere in the chain correctly falls the *entire* response back to graph traversal rather than serving partially-verified data.
- Per-session `provenance_storage_path` (threaded through `GraphSession.__init__`/`create_app`) correctly replaces the earlier classmethod/global-state approach flagged in review.

### Issue found
- **`create_app()` lets an ambient env var override an explicitly-passed session's storage path.** `prov_path = provenance_storage_path or settings.get(...)` conflates the explicit function argument with `SEMANTICA_PROVENANCE_DB`/`EXPLORER_PROVENANCE_DB` env vars. If a caller passes a pre-configured `GraphSession` and one of those env vars happens to be set in the process, `create_app(session=my_session)` can raise `ValueError` (or silently rebind storage) purely from ambient state the caller never touched — the same class of "shared mutable state" bug the PR's own commit history says it fixed elsewhere. Untested combination. Not a blocker, but worth a review comment.

### Minor / style
- `direction = "downstream"` branch in `_transform_audit_lineage` is dead code by the author's own comment (unreachable until `ProvenanceManager` gets reverse lookup) — could just be `direction = "upstream"` unconditionally with the comment kept, so it doesn't read as live/tested logic.
- Edge dedup keys on `(src, eid)` only, so a duplicate pair with a different `activity_id` label silently loses the second label. Low impact, disclosed intent to keep the graph scannable.

### Test coverage
- `tests/explorer/test_provenance_manager_wiring.py` (8 tests): audit path via real multi-hop `track_entity()` chain, empty-record fallback, simulated storage failure (asserts 200 not 500), checksum-tamper fallback, evidence-field round-trip, `create_app()` storage wiring, cross-session isolation — confirmed order-invariant in both execution orders.
- Gap: no test for explicit-session + ambient-env-var interaction (the issue above).

## Security
- No issues. Integrity check correctly forces fallback on any tampered entry rather than trusting partial data. Dict/dataclass checksum computation paths are consistent (including the `confidence is not None else 1.0` guard, avoiding a falsy-`0.0` bug).

## Performance
- Negligible — one SHA-256 recompute per lineage entry per request.

## Verdict
**Ready to merge**, modulo leaving a review comment on the `create_app` env-var/explicit-session precedence edge case (fix or explicitly accept the tradeoff — author's call, not required to unblock).

PR #810, opened 2026-07-28
Potential fixes for 2 code quality findings

KaifAhmad1:
_This PR applies 2/2 suggestions from code quality [AI findings](https://github.com/semantica-agi/semantica/security/quality/ai-findings)._

### DIFF
--- a/semantica/triplet_store/query_engine.py
+++ b/semantica/triplet_store/query_engine.py
@@ -421,7 +421,7 @@ class QueryEngine:
         }}
         """
 
-        expanded_uris = set([entity_uri])
+        expanded_uris = {entity_uri}
         try:
             if hasattr(store_backend, "execute_sparql"):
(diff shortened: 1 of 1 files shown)


Sameer6305 (commented):
I dug into the Qodo/Codex finding before making any changes.

I confirmed the regression is real:

ValidationError already performs stop_tracking(..., status="failed") before being re-raised.
ProcessingError previously flowed through the generic except Exception, which performed the cleanup.
After introducing except (ValidationError, ProcessingError): raise, ProcessingError bypasses that cleanup entirely.
I also verified there is no higher-level finally, caller cleanup, or automatic ProgressTracker cleanup, so the tracking entry remains in active_items.

The minimal fix appears to be handling ProcessingError separately, calling stop_tracking(..., status="failed") before re-raising, while leaving ValidationError unchanged.

I can implement that along with a regression test if you're happy with that approach.

PR #812, opened 2026-07-28
perf(provenance): optimize SQLite transaction lifecycle, concurrency, and lineage traversal

### LINKED ISSUE #807: Performance: Optimize SQLiteStorage transaction lifecycle, concurrency, and lineage traversal
## Summary

`SQLiteStorage` currently performs well for small datasets and single-threaded usage, but several implementation details become bottlenecks under larger ingestion workloads and concurrent access.

This issue tracks improving the performance, concurrency, and transaction behavior of the SQLite provenance backend while preserving backward compatibility and remaining consistent with the project's existing SQLite implementations.

---

## Technical Background

A review of `SQLiteStorage`, `ProvenanceManager`, `SQLiteVecStore`, and `SQLiteVersionStorage` identified several areas where the provenance backend differs from the project's existing SQLite patterns.

### 1. Connection lifecycle

Most public operations in `SQLiteStorage` create and tear down a new SQLite connection for every individual operation.

This includes:

- `_init_db()`
- `store()`
- `retrieve()`
- `retrieve_all()`
- `trace_lineage()`
- `clear()`

While acceptable for lightweight workloads, repeated connection setup becomes increasingly expensive during large ingestion jobs.

---

### 2. Multi-step operations are not atomic

`ProvenanceManager.track_entity()` performs multiple storage operations independently, including:

- retrieving existing state
- resolving parent/source entities
- archiving history
- writing the updated entity

These execute as separate transactions.

If a failure occurs midway, history archival and primary entity updates can become partially applied.

---

### 3. Missing SQLite concurrency configuration

The current implementation does not configure SQLite for concurrent workloads.

Notably absent are:

- `PRAGMA journal_mode=WAL`
- `PRAGMA busy_timeout`
- `PRAGMA synchronous=NORMAL`

Without these, concurrent readers and writers are more likely to encounter:

```text
sqlite3.OperationalError: database is locked
```

The repository already enables similar pragmas in other SQLite-backed components.

---

### 4. Batch APIs are sequential

`track_entities_batch()` and `track_chunks_batch()` currently iterate over entities individually.

As a result:

- every entity performs independent storage operations
- every write commits independently
- large ingestion jobs perform many unnecessary transaction commits

---

### 5. Lineage traversal performs N+1 lookups

`trace_lineage()` performs a breadth-first traversal while issuing an individual lookup for each visited entity.

Although traversal keeps a single connection open for the duration of the search, query execution still scales with the number of visited nodes.

For deeper provenance graphs this results in unnecessary database round-trips.

---

### Connection churn

Every major storage operation opens a new SQLite connection.

### Non-atomic updates

`track_entity()` archives historical state and persists the updated entity using separate storage operations and separate transactions.

### Missing concurrency configuration

`SQLiteStorage` currently does not configure:

- WAL mode
- busy timeout
- synchronous mode

while other SQLite implementations in the project already configure these appropriately.

### Batch operations

Current batch APIs are Python loops around single-record operations rather than transaction-aware batch persistence.

### Lineage traversal

`trace_lineage()` performs one query per discovered entity during BFS traversal rather than reducing round-trips through batched frontier retrieval.

---

## Proposed Direction

The implementation should remain aligned with existing SQLite patterns already present in the repository.

The preferred direction includes:

- Improve SQLite transaction lifecycle while preserving Windows test compatibility.
- Configure appropriate SQLite concurrency pragmas (WAL where supported, `busy_timeout`, and `synchronous=NORMAL`).
- Execute related provenance updates atomically.
- Reduce transaction overhead for batch ingestion.
- Optimize lineage traversal by reducing query round-trips while keeping the traversal logic readable and maintainable.
- Preserve existing public APIs and backward compatibility.

The implementation should avoid introducing unnecessary architectural complexity or breaking existing storage abstractions.

---

## Non-goals

This issue is **not** intended to:

- redesign the provenance schema
- introduce third-party ORM or connection-pooling libraries
- replace SQLite
- change the public provenance API
- modify `InMemoryStorage` behavior

---

## Acceptance Criteria

- [ ] Existing provenance tests continue to pass unchanged.
- [ ] Windows temporary database tests continue to pass without file-lock issues.
- [ ] Provenance updates execute atomically.
- [ ] Batch ingestion significantly reduces transaction overhead.
- [ ] Lineage traversal reduces unnecessary database round-trips.
- [ ] SQLite concurrency behavior is improved for concurrent readers and writers.
- [ ] Changes remain backward compatible and consistent with existing project SQLite patterns.

Sameer6305:
## Summary

This PR improves the performance and concurrency characteristics of the SQLite-backed provenance storage while preserving the existing public API and maintaining backward compatibility.

The changes focus on reducing transaction overhead, ensuring atomic provenance updates, improving concurrent SQLite access, and reducing database round-trips during lineage traversal.

## Changes

* Added SQLite connection configuration for improved concurrent access:

  * `PRAGMA journal_mode=WAL` (with safe fallback when unsupported)
  * `PRAGMA busy_timeout=5000`
  * `PRAGMA synchronous=NORMAL`
* Introduced a transaction context manager to scope a single connection across each public operation while ensuring connections are always closed before returning (preserving Windows temporary database compatibility).
* Made `ProvenanceManager.track_entity()` atomic by executing retrieval, history archival, and entity persistence within a single transaction.
* Optimized `track_entities_batch()` and `track_chunks_batch()` to execute within a shared transaction, significantly reducing connection and commit overhead for large ingestion workloads.
* Reduced lineage traversal database round-trips by batching frontier lookups during breadth-first traversal while preserving traversal behavior.
* Added dedicated regression tests covering:

  * SQLite PRAGMA configuration
  * Atomic rollback behavior
  * Windows temporary database compatibility
  * Shared batch transactions
  * Batched lineage traversal and `max_depth` behavior

## Validation

* ✅ `tests/provenance/`: **230 passed, 32 skipped**
* ✅ `tests/explorer/test_provenance_manager_wiring.py`: **10 passed**
* ✅ Existing public APIs remain unchanged.
* ✅ Backward compatibility preserved.
* ✅ Windows temporary database behavior maintained while improving transaction efficiency.


Sameer6305:
@KaifAhmad1 ready for review.

KaifAhmad1 (approved):
## Final Review — PR #812

**Author:** @Sameer6305
**Base:** `semantica-agi/semantica:main` ← **Head:** `Sameer6305/semantica:fix/807-sqlite-storage-performance`
**Related issue:** #807
**Reviewer:** @KaifAhmad1 (assisted)
**Status:** ✅ Approved — follow-up fixes applied and pushed to the PR branch (commit `f99241ca`)

---

### Overview

The PR reworks `SQLiteStorage` / `ProvenanceManager` to:

- Configure `PRAGMA journal_mode=WAL`, `busy_timeout=5000`, `synchronous=NORMAL` on every connection (with a safe fallback if WAL is unsupported).
- Scope a single connection per public storage call via a `transaction()` context manager (commit on success, rollback + re-raise on failure, always closed).
- Make `ProvenanceManager.track_entity()` atomic — retrieval, history archival, and persistence now share one transaction.
- Share a single transaction across `track_entities_batch()` / `track_chunks_batch()` in blocks of 1000 items, cutting connection/commit overhead for large ingests.
- Batch BFS frontier lookups in `trace_lineage()` (chunked `IN (...)` queries) and add an optional `max_depth`, kept backward-compatible for custom storage backends via signature inspection.
- Add 8 new regression tests covering PRAGMA config, atomic rollback, Windows unlink safety, shared batch transactions, and batched BFS/`max_depth`.

Two automated reviewers (Qodo, Copilot) had already flagged and gotten fixes for earlier issues on this branch (rolled-back batches reporting success, reads preceding transaction start, custom lineage backend compatibility, connection leaks on config error) before this review pass began.

---

### Issues found in review

#### 1. Reads were taking SQLite's writer lock, working against the PR's own concurrency goal

`SQLiteStorage.transaction()` issues `BEGIN IMMEDIATE`, which immediately acquires the single RESERVED (writer) lock — correct for `track_entity()`'s read-modify-write sequence. But `retrieve()` and `trace_lineage()`, both pure reads, were also routed through `transaction()`:

```python
def retrieve(self, entity_id: str) -> Optional[ProvenanceEntry]:
    with self.transaction() as conn:   # BEGIN IMMEDIATE for a SELECT
        return self._retrieve_with_conn(conn, entity_id)
```

Only one connection can hold that lock at a time, so every `retrieve()` / `trace_lineage()` call serialized behind every other read and write in the whole database — the opposite of what WAL mode (readers don't block writers) is supposed to buy. `retrieve_all()` and `clear()` were left on a plain connection without `BEGIN IMMEDIATE`, making the inconsistency look unintentional.

**Verified:** a read issued while another connection held `BEGIN IMMEDIATE` blocked until the lock was released (previously up to `busy_timeout` = 5s).

#### 2. Batch success counts stayed inflated for individual item failures

`track_entity()` (and `track_chunk()`) caught *all* exceptions raised inside their transaction block and always returned a `ProvenanceEntry`, even on failure:

```python
entry = None
try:
    with self._get_or_create_transaction(_conn) as conn:
        ...
        self.storage._store_with_conn(conn, entry)
except Exception:
    if entry is None:
        entry = ProvenanceEntry(...)   # unpersisted fallback
    # entry is returned either way — never raises
return entry
```

Because `track_entity()` never re-raises, the per-item `try/except` inside `track_entities_batch()` never fires on failure, so `batch_count` — and therefore `tracked_count` — was incremented even for items that were never persisted.

**Concrete repro:** a batch of 5 entities where one has non-JSON-serializable `metadata` (e.g. a `set()`). `json.dumps()` raises inside `_store_with_conn`, `track_entity()` swallows it, and the batch call reports `count == 5` while `retrieve_all()` returns only 4 rows.

This directly contradicted the changelog's claim that the batch fix "defers `tracked_count` increments until successful commit so rolled-back items are never reported as successes" — that guarantee only held for whole-block failures (e.g. commit itself raising), not the more common single-bad-item case. The existing regression test (`test_batch_rollback_does_not_count_unpersisted_entries`) only exercised the whole-block-failure path, missing this gap.

---

### Fixes applied (commit `f99241ca`, pushed to the PR branch)

**`semantica/provenance/storage.py`**
- Added `SQLiteStorage._read_connection()`: a configured connection (WAL/busy_timeout/synchronous applied) with no explicit `BEGIN`, for read-only work.
- `retrieve()` and `trace_lineage()` now use `_read_connection()` instead of `transaction()`, so reads no longer contend for the writer lock.

**`semantica/provenance/manager.py`**
- `track_entity()` and `track_chunk()` now re-raise their caught exception when invoked with a shared `_conn` (i.e. from inside `track_entities_batch()` / `track_chunks_batch()`), so the batch loop's per-item `except` correctly skips counting that item.
- Standalone calls (no `_conn`) are unaffected — they still return an entry and degrade gracefully, preserving the existing public API contract ("Existing public APIs remain unchanged").

**`tests/provenance/test_sqlite_storage_performance_807.py`** — 5 new tests:
- `test_batch_does_not_count_individually_failed_items`
- `test_chunks_batch_does_not_count_individually_failed_items`
- `test_track_entity_standalone_call_still_degrades_gracefully`
- `test_retrieve_and_trace_lineage_do_not_block_on_writer_lock`
- (plus updates confirming existing tests still pass unchanged)

**Docs corrected:**
- `docs/guides/provenance.md` and `docs/reference/provenance.md` previously implied *every* SQLite operation used `BEGIN IMMEDIATE`; updated to clarify that only read-modify-write calls (`track_entity()`, `store()`) do, while plain reads use a separate, lock-free connection.

**`CHANGELOG.md`** — appended two "Follow-up fix" bullets under the existing #807 entry describing both issues and their resolutions, rather than rewriting history.

---

### Validation

```
tests/provenance/                                : 248 passed, 32 skipped
tests/explorer/test_provenance_manager_wiring.py : 10 passed  (included in the above)
tests/provenance/test_sqlite_storage_performance_807.py : 12 passed (7 original + 5 new)
```

Manual verification:
- Poisoned-metadata batch scenario: reported count now equals `len(retrieve_all())`.
- Concurrent read vs. held writer lock: `retrieve()` / `trace_lineage()` return immediately instead of blocking for up to `busy_timeout`.

---

### Verdict

**Approve.** The core optimizations (WAL, shared transactions, batched BFS, backward-compatible `trace_lineage` signature handling) are sound and well-tested. The two correctness/concurrency gaps found in review have been fixed on the same branch, verified with new regression tests, and the CHANGELOG/docs now accurately describe the resulting behavior. Ready to merge pending CI on the new commit.

PR #817, opened 2026-07-30
fix(mcp): improve causal chain fallback error handling - #781

### LINKED ISSUE #781: mcp/tools/decisions.py handle_get_causal_chain returns an empty-but-valid-looking response when both CausalChainAnalyzer import and graph fallback are unavailable
## Summary
In `mcp/tools/decisions.py`, `handle_get_causal_chain` has two exception handlers with inconsistent error signaling.

## Confirmed code (lines 85-115, read fresh)

**Inner handler (lines 107-108):**
```python
except (ImportError, AttributeError):
    chain = graph.get_causal_chain(decision_id) if hasattr(graph, "get_causal_chain") else []
```
Falls through to: `result = chain if isinstance(chain, list) else list(chain)` then `return {"chain": result, "count": len(result), "direction": direction}`.

**Outer handler (lines 111-113):**
```python
except Exception as exc:
    log.exception("get_causal_chain failed")
    return {"error": str(exc), "chain": []}
```

## Confirmed response shapes per condition
- CausalChainAnalyzer works: `{"chain": [...], "count": N, "direction": "downstream"}`
- Import fails, but `graph.get_causal_chain` exists: same shape, real data — this path is fine
- Import fails AND `graph.get_causal_chain` doesn't exist: `{"chain": [], "count": 0, "direction": "downstream"}` — no "error" key
- Outer exception: `{"error": "...", "chain": []}`

## Why this matters
Only the specific case where both the import fails and the graph has no fallback method produces a response that's structurally identical to a valid "no causal ancestors" result — no "error" key, so a caller can't tell this failure mode apart from a genuine empty result. The outer `except Exception` correctly signals failure via an "error" key; this inner path doesn't.

## Location
`mcp/tools/decisions.py:107-108` (silent path) vs `:111-113` (correctly signals failure)

Sameer6305:
## Summary

Fixes #781 by making `handle_get_causal_chain` return an explicit error when causal chain analysis is unavailable instead of returning an empty success-looking response.

The fallback path has also been updated to forward the requested `direction` and `max_depth` parameters to graph backends that implement `get_causal_chain`, ensuring consistent behavior with the primary analyzer path.

## Changes

- Return an explicit error response when:
  - `CausalChainAnalyzer` is unavailable, and
  - the graph backend does not implement `get_causal_chain`
- Forward `direction` and `max_depth` to `graph.get_causal_chain(...)` in the fallback path
- Preserve existing success responses, validation behavior, and exception handling
- Add dedicated regression tests covering:
  - unsupported backend error handling
  - fallback parameter forwarding
  - fallback success path
  - analyzer success path
  - validation errors
  - runtime exception handling

### Before

When both the analyzer and graph fallback were unavailable, the handler returned:

```json
{
  "chain": [],
  "count": 0,
  "direction": "downstream"
}
```

This was indistinguishable from a legitimate empty causal chain.

### After

The handler now returns:

```json
{
  "error": "Causal chain analysis is not supported on this graph backend",
  "chain": []
}
```

allowing clients to distinguish unsupported functionality from a valid empty result.

## Testing

Added regression tests covering:

- Unsupported backend returns explicit error
- Fallback graph path forwards `direction` and `max_depth`
- Fallback success response
- Primary analyzer success path
- Validation error response
- Runtime exception response


Sameer6305:
@KaifAhmad1 ready for review

KaifAhmad1 (approved):
## Final Review: PR #817 — `fix(mcp): improve causal chain fallback error handling - #781`

**Verdict: Approve — ready to merge.**

### Summary of what shipped (5 commits total)

- Returns an explicit `{"error": "Causal chain analysis is not supported on this graph backend", "chain": []}` instead of a fake-looking empty success response, when neither `CausalChainAnalyzer` nor a graph-backend fallback is available.
- Fallback path introspects `graph.get_causal_chain`'s signature to forward `direction`/`max_depth` (or `depth`, or nothing) — matching the primary analyzer path instead of silently dropping those params.
- Input hardening: non-dict `args`, non-string `decision_id` (fixed a latent `AttributeError` on `.strip()`), `max_depth` clamped to `(0, 100]`.
- **Follow-up fix (pushed by me, commit `62a027d6`)**: signature introspection and the actual backend call now use separate `try/except` blocks. Previously a genuine bug inside a backend's `get_causal_chain` (an unrelated `TypeError`) was misread as a signature mismatch, causing the backend to be invoked a second time with identical arguments before the real error surfaced. Fixed so a successfully-introspected call happens exactly once.
- **Merge conflict resolved (commit `04d2a726`)**: PR #816 merged to `main` first and added a CHANGELOG entry at the same location. Merged `origin/main` in and kept both entries side by side — no content lost, no conflict markers remain.

### Correctness — verified, not assumed

- Traced the full control flow by hand for every backend signature shape (var-kwargs, `direction`+`max_depth`, `depth`-only, positional-only, un-introspectable/C-extension).
- Confirmed via a live test run that `inspect.signature()` on a `MagicMock` attribute resolves to `(*args, **kwargs)`, so the existing `MagicMock`-based tests still exercise the same code path after the refactor — no silent regression there.
- Ran the full test file locally, both before and after the merge: **12/12 passing** (11 original + 1 new regression test `test_internal_typeerror_calls_backend_only_once` asserting the backend is called exactly once).

### Test coverage

- `tests/test_mcp_decisions_causal_chain.py`: validation errors, outer runtime exceptions, unsupported-backend error, param forwarding, fallback success, primary-analyzer success, `depth`-kwarg backend, positional-only backend, internal-`TypeError`-not-masked, and now single-invocation-on-internal-error.
- No coverage gaps identified for the scope of this fix.

### CI / mergeability

- `mergeable: MERGEABLE`, no conflicts remaining.
- Checks passing: eslint, security-scan, templateanalyzer, terrascan, GitGuardian, build. `Analyze Python` and `MSDO` were still `pending` at last check — worth a quick re-check before merging, but nothing failing.

### Security

- No concerns. No dynamic code execution or injection surface; `inspect.signature` is used on an internally-obtained, already-`hasattr`-checked object.

PR #818, opened 2026-07-30
fix(agno): fail fast on toolkit registration failures - #780

### LINKED ISSUE #780: AgnoDecisionKit tool registration failures are silently swallowed, and the existing test doesn't catch it
## Summary
In `integrations/agno/decision_kit.py`, tool registration with Agno's runtime can fail silently, and the existing test suite would pass regardless of whether registration actually succeeded.

## Confirmed code (lines 115-135, read fresh)
```python
for fn in tools_to_register:
    self._tools.append(fn)       # unconditional, runs BEFORE the try
    if AGNO_AVAILABLE:
        try:
            self.register(fn)    # may fail
        except Exception:
            pass                 # silently discarded
```

`self._tools.append(fn)` runs unconditionally, before the try/except. So even if every single `self.register(fn)` call fails, `self._tools` still ends up fully populated.

## Confirmed test gap
`tests/integrations/agno/test_decision_kit.py:75-78`:
```python
def test_tools_registered(self):
    kit = AgnoDecisionKit(context=_make_context())
    # Tools should be registered (Toolkit.register was called)
    self.assertTrue(len(kit._tools) >= 5)
```

This assertion only checks `len(kit._tools) >= 5`, which comes entirely from the unconditional `append()` above — not from whether `register()` actually succeeded. The comment claims it verifies "Toolkit.register was called," but it doesn't. Confirmed the test's own `Toolkit.register` stub always succeeds, so this gap isn't exercised either way currently — but a real registration failure in production would go completely undetected.

## Location
`integrations/agno/decision_kit.py:125-131`, test gap at `tests/integrations/agno/test_decision_kit.py:75-78`

Sameer6305:
## Summary

Fixes #780 by removing silent tool registration failures from the Agno toolkit integrations and strengthening the corresponding test coverage.

Both `AgnoDecisionKit` and `AgnoKGToolkit` now follow the same fail-fast registration behavior, ensuring registration errors are surfaced immediately instead of being silently ignored.

### Toolkit registration

- Remove silent `try...except: pass` blocks around `self.register(...)`
- Allow registration exceptions to propagate naturally when Agno is available
- Preserve graceful degradation when `AGNO_AVAILABLE` is `False`
- Ensure `self._tools` accurately reflects registered/available tools
- Prevent duplicate entries in `self._tools` caused by the test toolkit stubs

### Test improvements

Expanded the Agno integration test suites to verify actual registration behavior instead of indirectly inspecting `self._tools`.

Added coverage for:

- Successful tool registration
- Registration invocation
- Registration failure propagation
- Graceful degradation when Agno is unavailable
- Duplicate prevention in `_tools`

### Before

- Tool registration failures were silently ignored.
- Toolkit initialization could succeed even if every `register()` call failed.
- Existing tests only validated the length of `self._tools`, which did not prove successful registration.

### After

- Registration failures immediately propagate to the caller.
- Toolkit initialization follows a fail-fast model when Agno is available.
- Existing graceful degradation behavior is preserved when Agno is unavailable.
- Tests now verify actual registration behavior and regression scenarios.

## Testing

Verified with:

- `pytest tests/integrations/agno/`

All Agno integration tests pass successfully.


Sameer6305:
@KaifAhmad1 ready for review.

KaifAhmad1 (approved):
## Final Review — PR #818: `fix(agno): fail fast on toolkit registration failures`

**Status:** Mergeable, no conflicts, 118/118 tests passing.

### Overview
- Removes the `try/except: pass` around `self.register(fn)` in `AgnoDecisionKit` and `AgnoKGToolkit.__init__`, so Agno registration errors now propagate instead of being silently swallowed.
- Preserves graceful degradation when `AGNO_AVAILABLE` is `False`.
- Fixes a latent `_tools` duplicate-entry bug: previously `fn` was appended unconditionally *before* `register()` ran, which could double-count entries if Agno's own `Toolkit.register()` also tracks `_tools`.

### Code Quality
- Fix is minimal and correctly scoped — no unrelated refactors.
- Bound-method equality (`fn not in self._tools`) works correctly for dedup despite each `self.foo` access creating a new bound-method object.
- Fail-fast is safe: if `register()` raises mid-loop, `__init__` raises and the half-built instance is discarded — no partial object ever leaks to a caller.

### Test Coverage
- New tests cover: registration invocation count, failure propagation, graceful degradation without Agno, and no-duplicate-`_tools` assertions — for both toolkits.
- Verified independently: `pytest tests/integrations/agno/` → **118 passed**.
- No other test files in the repo construct these toolkits, so the behavior change is isolated.

### Risk / Compatibility
- Behavior change: toolkit construction can now raise where it previously never did.
- Audited: no in-repo call site relies on the old silent-failure behavior, so no regression risk today.

### Follow-ups Applied (this session)
- Added a `CHANGELOG.md` entry under `### Fixed` (#780, #818) — this repo's convention requires one for behavior-changing fixes, and the original PR was missing it.
- Resolved a merge conflict in `CHANGELOG.md` against latest `main` (both fix entries kept, newest first).
- Both follow-ups pushed directly to `Sameer6305:fix/780-agno-tool-registration-validation` (maintainer edits enabled on this PR).

### Verdict
**Approve.** Correct, well-tested, matches issue #780's intent. Only gap (missing changelog entry) has been fixed in a follow-up commit on the same branch.

PR #819, opened 2026-07-31
Reject cyclic SKOS hierarchies at write time

### LINKED ISSUE #774: SKOS broader/narrower hierarchies have no cycle detection — a user can create an infinite loop
Confirmed via full codebase search: OWL classes get cycle detection via
_check_circular_hierarchy (class_inferrer.py:379-416), operating only on OWL-shaped
data. No cycle detection exists anywhere for SKOS concepts at write time.

One nuance: vocabulary.py:77 has a trail guard in _build_hierarchy() that prevents
infinite recursion when rendering the hierarchy for display - but this only stops a
read-time infinite loop, it doesn't prevent a cycle from being persisted in the first
place. A user can still create skos:broader A -> B -> A with nothing rejecting it at
write time.

mikemikimike:
## Summary
- reject cyclic `skos:broader` and `skos:narrower` relationships before graph mutation
- validate imported vocabulary edges before adding nodes to avoid partial writes
- add regression coverage for cyclic vocabulary imports

Fixes #774

## Tests
- `python -m pytest tests/explorer/test_vocabulary.py -q` (9 passed)
- `python -m compileall -q semantica/explorer`


mikemikimike:
## Validation

- Regression suite: `python -m pytest tests/explorer/test_vocabulary.py -q` — **9 passed**.
- Repository-wide `python -m pytest -q` was started after installing the documented `rdflib` dependency, but did not complete within the 120-second local limit and produced no failure report.
- The temporary `rdflib` installation was removed after testing.

mikemikimike:
## Addressed Qodo review findings

- Added atomic GraphSession.add_nodes_and_edges and updated vocabulary/JSON/CSV import routes.
- Restricted validation to incoming SKOS hierarchy edges, so unrelated writes are not blocked by legacy cycles.
- Added the guard to single-edge writes and centralized it at ContextGraph.
- Ignore None and whitespace-only endpoints instead of converting them to fake concept IDs.

## Validation

- Targeted tests: 11 passed, 11 deselected.
- Two broader context tests remain blocked by missing gensim and are unrelated to this change.
- Temporary rdflib was installed for RDF tests and removed afterward.

Sameer6305 (approved):
**Review Summary**

Thanks for the implementation. I performed a full review of the changes and verified the implementation against the original issue as well as the surrounding Explorer ingestion paths.

### Review completed

* Verified the cycle detection implementation and confirmed that write-time validation is applied for `skos:broader` and `skos:narrower` relationships.
* Reviewed the centralization of validation to ensure the primary graph mutation APIs consistently enforce the invariant.
* Audited the Explorer ingestion routes to identify any remaining paths that could bypass the new validation.
* Reviewed the regression coverage to ensure the new behavior is exercised by automated tests.
* Checked for consistency with existing Explorer error handling and write semantics.

### Review findings

During review I found one additional ingestion path that was still outside the new write-before-validate flow.

Specifically, the ontology import/create routes were still performing:

* node insertion
* followed by edge insertion

which could leave successfully inserted nodes behind if SKOS hierarchy validation rejected the edges.

To keep the behavior consistent with the rest of the Explorer import pipeline, I updated those routes to use the same `GraphSession.add_nodes_and_edges()` entry point introduced by this PR so validation occurs before graph mutation.

### Additional changes made during review

* Updated the ontology import/create routes to use the centralized atomic insertion helper.
* Added regression coverage for the ontology import path to prevent future regressions.
* Updated the changelog to document the completed fix.
* Rebased the branch onto the latest `main` and resolved the resulting changelog conflict.

### Validation

Verified with the relevant Explorer regression tests after the review updates. The reviewed implementation now applies the SKOS hierarchy validation consistently across the Explorer ingestion paths covered by this feature, and I did not identify any remaining blockers related to Issue #774.

**LGTM. Ready to merge.**

KaifAhmad1 (approved):
## PR #819 Final Review

**Verdict: Ready to merge**

### What it does
- Adds write-time cycle detection for `skos:broader`/`skos:narrower` edges
- Closes #774

### Bugs found → fixed in follow-up (`bc75768a`, co-authored with @mikemikimike)
- Cycle check re-scanned the *whole* graph, so one legacy cycle anywhere blocked all future SKOS writes → now scoped to only the touched nodes
- `/api/ontology/load` swallowed its own 422 (dead `except HTTPException`) and silently fell back to a different parser → reordered except clauses

### Already fixed in earlier review rounds
- Partial writes on import, single-edge bypass, `None` → fake node IDs

### Not fixed (minor, non-blocking)
- Double validation pass per write (perf nit)
- Recursive DFS could hit recursion limit on extremely deep chains

### Validation
- 62 tests passed, `compileall` clean
- Fixes pushed to PR branch

### Outstanding
- Stray branch I accidentally created on the base repo — needs your OK to delete

Issue #784, opened 2026-07-22
manager.py duplicates identical track-then-store logic across 6 methods instead of a shared helper, spreading the same silent-failure bug everywhere


Sameer6305:
## Summary
manager.py has 6 separate tracking methods (track_entity, track_relationship,
track_chunk, track_property_source, track_entities_batch, track_chunks_batch), each
manually reimplementing the same sequence: instantiate entry -> compute_checksum() ->
try: storage.store() except Exception: pass.

This duplication is why the silent-failure pattern (see other filed issues) exists
identically across the entire API surface rather than being fixable in one place.

Sameer6305:
suggesting a shared internal helper (e.g. _save_entry()) that centralizes
the store-and-error-handling logic, so a fix to the error handling only needs to
happen once rather than in 6 places.

PR #820, opened 2026-07-31
fix(provenance): log tracking failures and return None on storage error (closes #783)

### LINKED ISSUE #783: ProvenanceManager tracking methods silently swallow storage failures, returning a valid-looking object even when nothing was persisted
## Summary
Every tracking method in manager.py (track_entity, track_relationship, track_chunk,
track_property_source, track_entities_batch, track_chunks_batch) wraps its
self.storage.store(entry) call in a bare `except Exception: pass`. If storage fails
(e.g. disk full, DB locked), the caller receives back a fully-populated
ProvenanceEntry object with no indication that it was never actually persisted.

## Location
manager.py:202-205 and identical patterns at lines 249, 307, 370, 446

## Why this matters
A caller has no way to distinguish "successfully tracked" from "silently failed to
track" - both return the same populated object. For an audit-trail system, this means
provenance records can appear to exist in application code while being entirely
absent from the actual store.

Sameer6305:
Closes #783

## Summary

Every tracking method in `ProvenanceManager` wrapped `self.storage.store(entry)` in a bare `except Exception: pass`. On storage failure, callers received a fully-populated `ProvenanceEntry` object indistinguishable from a successfully persisted one — for an audit-trail module, this meant provenance records could appear to exist in application code while being entirely absent from the store, with zero logging to indicate anything went wrong.

This is the direct follow-on to #782, which fixed the same root pattern specifically in `track_entity` (its two-step write needed atomic rollback on top of honest failure reporting). This PR applies the same fix — honest `Optional[ProvenanceEntry]` returns, never silent — to the remaining tracking surface, per the original issue's own framing: *"fixing this broadly ... would likely resolve both."*

## What changed

**`_save_entry()` (the shared helper introduced in #784) now always logs on failure**, regardless of whether it's raising or swallowing. Every method that routes through it — `track_relationship`, `track_chunk`, `track_property_source`, and indirectly `track_entities_batch`/`track_chunks_batch`'s per-item failures — gets this for free with zero logic duplication.

**Return contract, made honest:**
- `track_relationship()`, `track_chunk()`, `track_property_source()` now return `Optional[ProvenanceEntry]` — `None` on storage failure, never a fabricated object claiming values that were never persisted.
- Unlike `track_entity` (#782), these three have no existing-entry-preservation fallback — they always return `None` on failure, not a deep copy of a prior version. Documented explicitly as such, to avoid implying parity with `track_entity`'s more nuanced update-vs-create contract.

**Batch methods gained the logging they were missing at both levels:**
- Per-item failures inside a batch already get logged via `_save_entry`'s fix (traced explicitly: `track_entity`/`track_chunk` always call `_save_entry` with `_raise_on_error=True` from batch context, so the log-then-raise path fires before the batch's per-item `except: pass` catches it).
- The rarer block-level failure (the whole `storage.transaction()` block failing, not an individual item) previously logged nothing at all — now logs explicitly.

**One real downstream consumer needed to adapt:** `semantica/conflicts/source_tracker.py`'s `track_sources_batch()` was unconditionally incrementing `stats["entities_tracked"]` / `stats["properties_tracked"]` / `stats["relationships_tracked"]` regardless of whether the underlying tracking call actually succeeded. Fixed to check the (now-honest) return value before counting, and logs a warning per skipped item — reusing the class's existing `stats` dict keys and logging convention rather than introducing new ones.

## Verification discipline

Every claim in this PR was checked against real code and real execution output before implementation, not assumed from the issue text or from #782's prior findings (several things had changed since #782: `_save_entry`'s signature, the addition of `savepoint()` for per-item batch isolation, etc.):

- Live-reproduced silent failure for all 5 methods against both `InMemoryStorage` and `SQLiteStorage` before writing any fix, confirming the issue's claim held post-#782/#784/#807.
- Verified `savepoint()`'s per-item isolation with a real 3-item batch (item 2 fails) against both backends — confirmed items 1 and 3 persist independently of item 2's failure, on both storage types.
- Traced every non-test call site for all 5 methods (fresh audit, not reused from #782, since files had changed): confirmed which ones use the return value, which ignore it, and which needed adapting (`source_tracker.py:626` was the only one).
- Explicitly traced whether returning `None` for a chained `parent_chunk_id` in `split/provenance_tracker.py` could break downstream chunking — confirmed safe (a `None` parent is a valid, meaningful state, not a crash risk).
- Confirmed `_save_entry`'s change is inert for `track_entity` specifically — it always calls `_save_entry` with `_raise_on_error=True` and discards the return value entirely, so the new swallow-path `return None` is structurally unreachable from `track_entity`'s call path.

## Testing

- Updated 3 existing tests that previously asserted the buggy "returns a populated object on failure" behavior (`test_track_relationship_storage_error_swallowed`, `test_track_chunk_storage_error_swallowed`, `test_track_property_source_storage_error_swallowed`), matching the correction pattern #782 established.
- Added `test_save_entry_logs_on_every_failure_path` — verifies logging fires on both the raising and

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`ProvenanceManager` tracking methods silently swallowed failures without logging and returned fabricated entries** (#783)
+  - `track_relationship()`, `track_chunk()`, and `track_property_source()` now return `Optional[ProvenanceEntry]` (`None` on storage failure, consistent with #782's `track_entity` fix) instead of a fabricated populated object
+  - `_save_entry()` now always logs on any storage failure, including previously-silent per-item batch failures
+  - `track_entities_batch()` and `track_chunks_batch()`'s rare block-level transaction failures are now logged too
+  - `source_tracker.py`'s `track_sources_batch()` no longer counts failed tracking calls in its stats
+
 - **`ProvenanceManager.track_entity` persisted partial history and returned fabricated entries on storage failure** (#782, #816) by @Sameer6305 and @KaifAhmad1
   - `track_entity()`'s two-step write (history archive + primary update) is now atomic — if either write fails, the whole operation rolls back via the existing #807 `transaction()` mechanism, instead of silently persisting a partial state
   - `track_entity()`'s return type is now `Optional[ProvenanceEntry]`: on failure it returns a safe deep copy of the pre-failure existing entry (if one existed) or `None` (if this was a brand-new, never-successfully-tracked entity) — never a fabricated object claiming values that were never actually persisted
--- a/docs/reference/provenance.md
+++ b/docs/reference/provenance.md
@@ -222,9 +222,9 @@ cleared = manager.clear()
 | Method | Returns | Description |
 | :------ | :------- | :----------- |
 | `track_entity(entity_id, source, metadata, **kwargs)` | `Optional[ProvenanceEntry]` | Record entity provenance atomically; returns `ProvenanceEntry` on success, or `None`/existing entry on storage failure |
-| `track_relationship(relationship_id, source, metadata, **kwargs)` | `ProvenanceEntry` | Record relationship provenance |
-| `track_chunk(chunk_id, source_document, ...)` | `ProvenanceEntry` | Record chunk provenance with char offsets |
-| `track_property_source(entity_id, property_name, value, source)` | `ProvenanceEntry` | Record property-level source attribution |
+| `track_relationship(relationship_id, source, metadata, **kwargs)` | `Optional[ProvenanceEntry]` | Record relationship provenance; returns `ProvenanceEntry` on success, or `None` on storage failure |
+| `track_chunk(chunk_id, source_document, ...)` | `Optional[ProvenanceEntry]` | Record chunk provenance with char offsets; returns `ProvenanceEntry` on success, or `None` on storage failure |
+| `track_property_source(entity_id, property_name, value, source)` | `Optional[ProvenanceEntry]` | Record property-level source attribution; returns `ProvenanceEntry` on success, or `None` on storage failure |
 | `track_entities_batch(entities, source)` | `int` | Batch-track entities; returns success count |
 | `track_chunks_batch(chunks, source_document)` | `int` | Batch-track chunks; returns success count |
 | `get_lineage(entity_id)` | `Dict[str, Any]` | Full lineage as aggregated dict |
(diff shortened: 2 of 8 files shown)


Sameer6305:
@KaifAhmad1 ready for review

KaifAhmad1 (approved):
## PR #820 Review — fix(provenance): log tracking failures and return None on storage error

**Branch:** `Sameer6305:fix/783-tracking-methods-honest-failures` → `main`
**Closes:** #783

### Summary
Extends #782's pattern (honest `Optional[ProvenanceEntry]` returns instead of fabricated objects on storage failure) from `track_entity` to `track_relationship`, `track_chunk`, `track_property_source`, adds failure logging to `_save_entry()` and batch block-level failures, and adapts `source_tracker.py` to respect the new return contract.

### Findings

- **Bug — dead code in `source_tracker.py` fix** (`semantica/conflicts/source_tracker.py:614-657`)
  `SourceTracker.track_entity_source` / `track_property_source` / `track_relationship_source` are pure in-memory dict operations with no storage/`ProvenanceManager` call — they unconditionally `return True`. The new `if ok: ... else: log warning` check in `track_sources_batch()` can never hit the `else` branch in production. Doesn't break anything, but doesn't fix what the PR description claims, and the PR conflates this class with the unrelated `SourceTrackerWithUnifiedBackend` (`conflicts_provenance.py`). New test only proves it via mocking, not real behavior.

- **Good catch (undocumented) — fixes a latent `AttributeError`** (`semantica/split/provenance_tracker.py:100-101`)
  Moving `_provenance_store`/`_chunk_registry` init out of the legacy-only `else` branch to unconditional is necessary and correct — without it, the existing exception-based fallback (and the new `entry is None` fallback) would crash with `AttributeError` when using the unified backend. Worth calling out explicitly in the PR description.

- **Minor, pre-existing — inconsistent return type on fallback** (`semantica/split/provenance_tracker.py:160-166`)
  `track_chunk()` returns `chunk_id` on success but `provenance_id` (different UUID) on any fallback path. Pre-existing bug, but the new graceful-failure path makes it reachable far more often than the old exception-only fallback did. Worth a follow-up issue.

- **Positive — `get_provenance()` fallback is a real fix** (`semantica/split/provenance_tracker.py:310`)
  Falling back to `_get_provenance_legacy()` when the unified manager returns nothing correctly handles chunks that were tracked via the legacy fallback path. Undocumented in PR body but correct.

- **Batch semantics verified correct**
  Traced `track_entities_batch`/`track_chunks_batch`: per-item failures inside a shared transaction correctly raise (via `_raise_on_error=True`) after logging, get caught by the per-item `except: pass`, and are excluded from the count. Matches PR description.

### Process
- ~~Merge conflict in `CHANGELOG.md`~~ — resolved and pushed (commit `5094235c`).
- CI: all checks completed and passing (build, security-scan, Analyze Python, MSDO, docs, eslint, terrascan, Checkov, GitGuardian); `CodeQL` is `NEUTRAL` (informational, non-blocking).
- `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.
- No formal review decision posted on the PR yet — needs an actual approval click.

### Verdict: READY TO MERGE
Core `ProvenanceManager` fix is solid and well-tested, no conflicts, all CI green. The `source_tracker.py` dead-code hunk is cosmetic, not a blocker — recommend a follow-up comment/issue rather than holding up this PR.

PR #822, opened 2026-08-01
fix(agno): make _eval_rule raise instead of silently returning compliant=True on unevaluable rules (closes #778)

### LINKED ISSUE #778: AgnoDecisionKit._eval_rule silently returns compliant=True when a policy rule references a field not present in the decision payload

Sameer6305:
Closes #778

## Summary

`AgnoDecisionKit._eval_rule` returned `True` (rule "passed") for two cases where a rule genuinely couldn't be evaluated: a rule string that didn't match the expected `field op value` format, and a rule referencing a field absent from the decision payload. `check_policy` treated this identically to a rule that was actually checked and passed — no violation recorded, no warning logged, zero signal that the rule was never evaluated at all.

The function's own docstring claims "ensures exceptions never silently return `compliant=True`" — but this wasn't an exception path. It was a plain `return True`, bypassing that guarantee entirely.

## Verified example (from the issue, reproduced before any fix)

Rule `"minimum_score >= 0.9"` against payload `{"confidence": 0.95}` (no `minimum_score` key) returned:
```json
{"compliant": true, "violations": [], "warnings": []}
```
with zero indication the rule was never actually checked.

## Fix

Both unevaluable branches in `_eval_rule` now `raise ValueError(...)` instead of `return True`. This routes them through `check_policy`'s **already-existing** generic `except Exception` handler, which converts the failure into a `warnings` entry (not a violation, since we genuinely don't know if the rule would have passed) — with a message identifying exactly which rule and why it couldn't be evaluated.

No new branching logic was added to `check_policy`; both previously-silent cases now flow through the same mechanism that already handles every other rule-evaluation failure (e.g. unsupported operators, type-mismatched comparisons). This also makes the existing docstring guarantee literally accurate for these cases instead of a near-miss.

## Design decision, and why

`compliant` stays a strict bool rather than becoming tri-state. This is backed by real precedent found in the codebase, not a default choice: `semantica/context/context_graph.py`'s `enforce_decision_policy` already uses the identical `compliant`/`violations`/`warnings` shape, with `warnings` explicitly reserved for non-blocking issues. A repo-wide audit of every caller of `check_policy` and `_eval_rule` confirmed `compliant` is treated as a strict bool everywhere (tests, cookbook notebook, docs) — no tri-state precedent exists anywhere, so introducing one here would be an unjustified interface change.

## Backward compatibility

`check_policy`'s return shape is unchanged — same three keys, same types. `compliant`'s meaning is unchanged for every case that previously worked *correctly*; it only changes for the two cases that were previously *wrong*. Audited existing tests for any `warnings == []` assertion that could now break: none assert this for missing-field or malformed-format inputs, only for genuinely-valid rules, which are unaffected. Purely additive.

## Security

The rule-parsing regex (`(\w+)\s*(>=|<=|!=|==|>|<)\s*(.+)`) and comparison path were tested against adversarial inputs before considering this done: large payloads (ReDoS timing), field names containing regex metacharacters, non-numeric values compared with ordering operators, and injection-string values (no `eval()` exists anywhere in this path — comparisons use a fixed operator dict). All of these were already routed through the pre-existing catch-all exception handler; this fix introduces no new unhandled-exception surface, it only converts two previously-silent branches into the same handling every other failure mode already had.

## Future-safety

The mechanism generalizes: any future addition to `_eval_rule` that can't produce a definitive answer can simply raise, and `check_policy`'s existing catch-all turns it into a `warnings` entry automatically — no call-site changes needed.

## Testing

4 new tests in `tests/integrations/agno/test_decision_kit.py`:
- The issue's exact traced example (missing field)
- Malformed rule string
- Two sanity checks confirming normal pass/violate evaluation on present fields is unaffected

**Full suite:** `tests/integrations/agno/` — 122 passed, 0 failed.

## Docs

Docstring updated to document the new `Raises ValueError` behavior and its relationship to `warnings`. Grepped all `.md` files for `check_policy`/`_eval_rule` return-contract descriptions: only `docs/integrations/agno.md` mentions `check_policy` (a one-line tool listing, no contract detail) — unaffected, no other stale docs found. `CHANGELOG.md` updated under the existing `[Unreleased]` / `### Fixed` convention.


Sameer6305:
@KaifAhmad1 ready for review

KaifAhmad1 (approved):
## PR #822 Review — `fix(agno): make _eval_rule raise instead of silently returning compliant=True on unevaluable rules`

**Closes:** #778 · **Author:** @Sameer6305 · **Branch:** `fix/778-eval-rule-missing-field` → `main`

### Summary
- `_eval_rule` no longer `return True`s on unevaluable rules (missing field, malformed format, null field); it raises `ValueError`, which `check_policy`'s existing catch-all turns into a `warnings` entry.
- Follow-up commits (from bot review) also fixed: bare-JSON-string `policy_rules` being iterated character-by-character, non-list/non-string `policy_rules` shapes, and non-dict `decision_data` causing a raw `TypeError`.

### Correctness
- All modified branches traced and verified against tests: missing field, malformed rule, null-vs-missing distinction, non-dict `decision_data`, bare-string and malformed `policy_rules` shapes — all behave as documented.
- `compliant`/`violations` semantics are unchanged for previously-correct cases; only `warnings` gains new (accurate) signal. Purely additive, as claimed.
- Only one call site for `_eval_rule` (inside `check_policy`) — no other caller can be broken by the new `raise`.
- Docs (`docs/integrations/agno.md`) and the cookbook notebook were checked directly — neither relies on the old silent-pass behavior, so both are unaffected.

### Pre-existing issue (not introduced by this PR, flagged for follow-up)
- `decision_kit.py:434` — `type(actual)(val_str)` for boolean fields: `bool("False")` is `True` (non-empty string is truthy), so a rule like `"flag == False"` against `{"flag": True}` would wrongly evaluate as compliant. Untouched by this diff; worth a separate issue.

### Style / Tests
- Docstrings and CHANGELOG entry match actual behavior and follow existing conventions.
- 15 new tests cover every modified branch; no coverage gaps found.

### Minor, non-blocking suggestions
- Error messages embed full `repr()` of the offending payload (`decision_data`/`policy_rules`) — consider truncating for very large inputs.
- Resolve the open Copilot review thread on GitHub (code fix already landed in `e5eeb8f`) — branch protection requires all comments resolved.

### Merge readiness
- **Code: ready.** No correctness issues found; well-tested; low risk.
- **Blockers were process, not code:**
  - Needed 1 approving review (none existed).
  - Branch was out-of-date with `main`.
- **Follow-up action taken:** merged `main` into the branch (clean auto-merge, no conflicts) and updated the changelog entry to reference both `#778` and `#822`, pushed directly to the fork branch via maintainer-edit access.
- Remaining: get an approving review and resolve the Copilot comment thread.

PR #823, opened 2026-08-01
fix(explorer): complete atomic ontology refresh writes - #775

### LINKED ISSUE #775: routes/ontology.py's add_nodes/add_edges write sequence has no atomicity — a mid-failure leaves nodes without edges
Traced the actual write path for ontology loading. Confirmed the core
semantica/ontology/ module is pipeline-in-memory only (the 6-stage generator never
writes to the store directly - only engine.py's create_alignment() does, and that's
a single atomic triplet write, so it's safe by construction).

The real risk is at the route layer: routes/ontology.py:1228-1229 and 1278-1279 do:
  nodes_added = await asyncio.to_thread(session.add_nodes, nodes)
  edges_added = await asyncio.to_thread(session.add_edges, edges)

If add_nodes succeeds and add_edges then throws, nodes are persisted with no edges -
an inconsistent state with no rollback.

Sameer6305:
## Summary

Completes the remaining ontology write path affected by Issue #775 by aligning the `refresh_ontology` route with the existing atomic write pattern used by the ontology load/create and vocabulary import endpoints.

### Changes

- Update `refresh_ontology` to use `GraphSession.add_nodes_and_edges()` instead of separate `add_nodes()` and `add_edges()` calls.
- Perform SKOS hierarchy validation before any graph mutation during ontology refresh.
- Return a consistent HTTP 422 response when SKOS cycle validation fails, matching the behavior of the other ontology import routes.
- Add regression tests covering successful refreshes, cyclic SKOS rejection, and existing error paths.
- Add a CHANGELOG entry documenting the follow-up fix.

Fixes #775.

## Why

Most ontology import routes had already been updated to perform pre-write validation using `add_nodes_and_edges()`. During review, the `refresh_ontology` endpoint was identified as the remaining route still performing node and edge writes separately.

If edge validation failed after nodes had already been added, the graph could be left in a partially updated state. This change brings the refresh path in line with the rest of the explorer import pipeline.

## Tests

- `python -m pytest tests/explorer/test_ontology_subissue3.py -v`
- `python -m pytest tests/explorer/ -q`


Sameer6305:
@KaifAhmad1 ready for review.

KaifAhmad1 (approved):
Reviewed the diff (routes/ontology.py, session usage, tests, changelog).

- `refresh_ontology` now uses the existing `GraphSession.add_nodes_and_edges()`, which validates SKOS hierarchy edges before writing nodes/edges under a single lock — closes the partial-write gap from #775 (cyclic SKOS rejection no longer leaves orphaned nodes committed).
- `ValueError` -> HTTP 422 mapping matches the convention already used in `/load` and `/create`.
- 4 new regression tests cover success, cyclic-SKOS rejection (verified via spy that only one combined write call is made, and node count is unchanged), unknown-URI 404, and missing-source-url 422.
- CHANGELOG entry documents the fix and records that other `add_nodes()`/`add_edges()` pairings in the repo were audited and intentionally left unchanged (different write path).
- All CI checks pass. No correctness issues found.

Approving.

PR #829, opened 2026-08-04
feat(explorer): add temporal diff comparison UI to the Temporal panel (#793)

### LINKED ISSUE #793: /api/temporal/diff backend exists with no frontend UI (temporal snapshot already has partial UI)
## Summary
Correction to original scoping: re-verified and found the original claim was only
half accurate. /api/temporal/snapshot IS already used by the frontend
(explorer/src/workspaces/GraphWorkspace/GraphRuntimeStage.tsx:334 and
GraphWorkspace.tsx:1263). /api/temporal/diff, however, is confirmed completely
unused anywhere in explorer/src/ - a full source search found zero references.

Both routes still exist in routes/temporal.py (snapshot at line 64, diff at line
79), both return 200 OK on success and are unaffected by #770's changes (which only
touched the temporal_patterns route in this file).

## Scope
Narrowing this to just: /api/temporal/diff has no UI. Graph diff visualization
(comparing two graph states, added/removed nodes and edges) is fully supported
server-side but has no way for a user to trigger or view it.

@KaifAhmad1 per your go-ahead - happy to open a PR for a diff view UI. Worth noting
the existing GraphWorkspace snapshot integration as a reference pattern for how to
wire this in.

Sameer6305:
## Summary

Closes #793. Adds a UI for the existing `GET /api/temporal/diff` backend route, which previously had zero consumers in the Explorer frontend (confirmed via a full grep of `explorer/src/` — no references to `/api/temporal/diff` existed anywhere before this change).

Users can now pick two timestamps in the existing "Temporal Context" panel and see which nodes were added/removed between those two points, highlighted directly on the graph.

## What changed

* **`temporalDiffState.ts`** (new): a small typed fetch wrapper, `fetchTemporalDiff(fromTime, toTime, signal)`, matching the route's `{ from_time, to_time, added_nodes, removed_nodes }` response shape.
* **`temporalOverlayPlugin.tsx`** (extended): the existing "Temporal Context" panel gains a "Compare two points in time" section — two text inputs, a Compare button, and result states.

No changes to `GraphWorkspace.tsx` or `GraphRuntimeStage.tsx`. This follows the plugin-panel pattern already established by `neighborhoodPanelPlugin.tsx` / `explorationEffectsPluginPhaseC.tsx` — no new panel-registration mechanism was introduced.

## Design decisions (and why)

* **Rendering mechanism — `baseColor`, not `ringColor`/`haloColor`.** I initially implemented highlighting via `ringColor`/`haloColor` (the attributes that look most purpose-built for this). Tracing the actual sigma `nodeReducer` in `GraphCanvas.tsx` and `resolveNodeElementStyle` in `graphSceneState.ts` showed those two attributes are only respected for nodes in `hovered`/`selected`/`path` visual state — for a node in its default (untouched) state, the reducer silently discards them and falls back to `borderColor`/nothing. Confirmed via direct execution of the real, shipped modules against a running instance of the app (bypassing a separate, pre-existing panel bug — see below) that `ringColor` produces no visible change, while `baseColor` does, through the exact same resolver the production reducer calls.
* **Highlight restore uses captured original values, not a fallback clear.** Each highlighted node's prior `baseColor` is captured in a `Map` before being overwritten, and restored from that map — not reset to `undefined` — since `undefined` falls through to a semantic-default color rather than the node's actual original color. Verified live: an `undefined`-clear would have produced a different, wrong color; the actual implementation restores the exact original.
* **Cancellation.** Uses `AbortController`, matching the pattern already used in `DecisionWorkspace.tsx`/`App.tsx`. A new comparison aborts any request still in flight *before* issuing the new one, and the in-flight request is also aborted on unmount. Both call sites verified by direct trace.
* **Visual style.** No new button/input/spinner/error patterns — reused `GraphInspectorPanel.tsx`'s exact `actionButtonStyle`/`inputStyle` shapes and its `<Loader2 className="animate-spin">` loading indicator (same directory, same theme tokens). Colors use existing theme tokens (`GRAPH_THEME.palette.semantic[2]` for "added", `ui.control.dangerText` for "removed") rather than new hardcoded hex values. No emojis.
* **States.** Idle / loading / validation-error / request-error / empty ("no changes between these two points" — a valid result, not an error) / success, each with its own distinct rendering.

## Testing performed

* `npx tsc --noEmit` — clean, zero errors.
* `npx eslint` on both changed files — one pre-existing-pattern finding (`react-refresh/only-export-components`), identical to the accepted structure already present in `explorationEffectsPluginPhaseC.tsx` (a local component defined alongside a plugin-object export). Not a regression introduced by this change.
* Started the real FastAPI explorer backend and the Vite dev server against a small sample graph with temporally-bounded nodes; confirmed `GET /api/temporal/diff` returns the correct `added_nodes`/`removed_nodes` through the Vite proxy exactly as the UI calls it.
* Verified the apply and restore highlighting logic by executing the real, shipped `graphSceneState.ts`/`graphTheme.ts` resolvers directly against the live running app (via in-page dynamic `import()`), since manual click-through of the Temporal panel is currently blocked by a pre-existing, unrelated bug (see below). Confirmed:

  * the diff color is actually applied through the production color resolver,
  * clearing the highlight restores the exact original resolved color (not an approximation), tested with a node whose original color does not coincidentally match either diff color,
  * the old `ringColor`/`haloColor` approach would have rendered nothing for a default-state node.

## Known limitation — pre-existing bug blocks manual click-through

The Temporal panel currently fails to open at all in this codebase: clicking the "Temporal" toolbar toggle triggers a React "Maximum update depth exceeded" error and the panel is permanently stuck on "Loading temporal…". This is confirmed pre-existing and unrelated to this change — `git stash`-ing this diff and reproducing against the unmodified base branch shows the identical error. Because of this, this feature could not be manually clicked through in a browser end-to-end; it was verified via direct module execution against the live app instead (see Testing above). I'll file a separate issue for the panel bug after this merges — it should be fixed independently since it blocks the Temporal panel for all consumers, not just this feature.

## Files changed

* `explorer/src/workspaces/GraphWorkspace/plugins/temporalDiffState.ts` (new)
* `explorer/src/workspaces/GraphWorkspace/plugins/temporalOverlayPlugin.tsx` (extended)

### DIFF
--- a/explorer/src/workspaces/GraphWorkspace/plugins/temporalDiffState.ts
+++ b/explorer/src/workspaces/GraphWorkspace/plugins/temporalDiffState.ts
@@ -1,0 +1,23 @@
+// Fetch wrapper for GET /api/temporal/diff (see semantica/explorer/routes/temporal.py).
+// added_nodes are node IDs active at to_time but not at from_time; removed_nodes are the
+// inverse. Both are plain node-ID lists (no edge-level diffing), matching the snapshot
+// route's active_node_ids shape used elsewhere in this workspace.
+export interface TemporalDiffResult {
+  from_time: string;
+  to_time: string;
+  added_nodes: string[];
+  removed_nodes: string[];
+}
+
+export async function fetchTemporalDiff(
+  fromTime: string,
+  toTime: string,
+  signal?: AbortSignal,
+): Promise<TemporalDiffResult> {
+  const params = new URLSearchParams({ from_time: fromTime, to_time: toTime });
+  const response = await fetch(`/api/temporal/diff?${params}`, { signal });
+  if (!response.ok) {
+    throw new Error(`Temporal diff request failed with status ${response.status}`);
+  }
+  return response.json();
+}
--- a/explorer/src/workspaces/GraphWorkspace/plugins/temporalOverlayPlugin.tsx
+++ b/explorer/src/workspaces/GraphWorkspace/plugins/temporalOverlayPlugin.tsx
@@ -120,15 +120,18 @@ function DiffSection({ context }: { context: GraphPluginContext }) {
   }, []);
 
   const handleCompare = () => {
-    if (!fromTime.trim() || !toTime.trim()) {
+    const from = fromTime.trim();
+    const to = toTime.trim();
+
+    if (!from || !to) {
       setValidationMessage("Both a from and to time are required.");
       return;
     }
-    if (!isValidDateInput(fromTime) || !isValidDateInput(toTime)) {
+    if (!isValidDateInput(from) || !isValidDateInput(to)) {
       setValidationMessage("Enter valid ISO datetimes, e.g. 2024-01-01T00:00:00.");
       return;
     }
-    if (new Date(fromTime).getTime() >= new Date(toTime).getTime()) {
+    if (new Date(from).getTime() >= new Date(to).getTime()) {
       setValidationMessage("From time must be before to time.");
       return;
     }
@@ -143,7 +146,7 @@ function DiffSection({ context }: { context: GraphPluginContext }) {
     previousColorsRef.current = new Map();
     setRequestState({ status: "loading" });
 
-    fetchTemporalDiff(fromTime, toTime, controller.signal)
+    fetchTemporalDiff(from, to, controller.signal)
       .then((result) => {
         if (controller.signal.aborted) {
           return;
--- a/explorer/src/workspaces/GraphWorkspace/plugins/temporalOverlayPlugin.tsx
+++ b/explorer/src/workspaces/GraphWorkspace/plugins/temporalOverlayPlugin.tsx
@@ -1,7 +1,8 @@
 import { useEffect, useRef, useState, type CSSProperties } from "react";
 import { Loader2 } from "lucide-react";
+import type Graph from "graphology";
 
-import { graph } from "../../../store/graphStore";
+import { graph, type NodeAttributes } from "../../../store/graphStore";
 import { GRAPH_THEME } from "../graphTheme";
 import { fetchTemporalDiff, type TemporalDiffResult } from "./temporalDiffState";
 import type { GraphPlugin, GraphPluginContext } from "./types";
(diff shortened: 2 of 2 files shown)


Sameer6305:
@KaifAhmad1 ready for review

KaifAhmad1 (approved):
Reviewed the diff, the backend route (`temporal.py`/`schemas.py`), and traced the color-resolution code path in `graphSceneState.ts` / `GraphCanvas.tsx` to verify the design-decision claims in the description. Everything checks out:

- `baseColor` vs `ringColor`/`haloColor`: confirmed `resolveNodeElementStyle` only reads ring/halo for hovered/selected/path states, so `baseColor` is correct for default-state diff highlighting.
- Dual-write to store `graph` + `context.displayGraph`: confirmed necessary, since `aggregateDisplayGraph` shallow-copies node attrs from the store graph on every rebuild. The `dg !== graph` guard correctly avoids a double-write when they're the same instance (aggregation disabled).
- Cleanup effect closing over `context` with an empty deps array is safe here because `displayGraph`/`scene` on `GraphPluginContext` are getters bound to a stable ref, not snapshotted values — so it still resolves live at unmount even if `pluginContext` was recreated since mount.
- `TemporalDiffResponse` schema matches the frontend `TemporalDiffResult` interface field-for-field.
- Abort/cancellation handling (cancel-in-flight-before-new-request, cancel-on-unmount) matches existing patterns (`DecisionWorkspace.tsx`).

Two non-blocking notes for a follow-up, not required for this PR:
1. `diffInputStyle` uses new hardcoded colors rather than the `GRAPH_THEME.ui.control.inputBg`/`inputBorder` tokens that `GraphInspectorPanel.tsx`'s `inputStyle` (which the description says was reused) actually uses.
2. In grouped view mode, diffed node IDs generally won't exist as individual nodes in the community-bundled display graph, so highlighting silently no-ops there (counts still display correctly). Worth a quick note or follow-up issue alongside the pre-existing panel-open bug.

Approving — no correctness issues found, and the known pre-existing panel bug is correctly scoped out of this change.

PR #835, opened 2026-08-05
fix(vector_store): stop dropping metadata for add_vectors-only backends

### LINKED ISSUE #832: VectorStore.store_vectors() silently drops metadata for FAISS (and any add_vectors-only) backend
## Summary

`VectorStore.store_vectors()` silently drops the `metadata` argument when the configured backend store exposes `add_vectors()` but not `store_vectors()` — which is true for `FAISSStore`, the backend most real usage configures for genuine ANN search. Every caller that stores vectors with metadata (for example `AgentMemory._store_memory_vector()`, used internally by `AgentContext.store()`) loses that metadata once it reaches FAISS, with no error or warning.

This has a real downstream effect: `ContextRetriever._retrieve_from_vector()` recovers a result's text via `metadata.get("content", "")`, which is now always `""` for any vector stored this way. `ContextRetriever._rank_and_merge()` then embeds that empty string unconditionally, which trips `TextEmbedder.embed_text()`'s (correct) empty-text rejection. `VectorStore.embed()` catches that and substitutes a random fallback embedding, so nothing crashes, but retrieval quality silently degrades and Semantica's own progress tracker records a `TextEmbedder` failure for what is actually a bug one layer up.

## Steps to Reproduce

```python
from semantica.context import ContextGraph, AgentContext
from semantica.kg import GraphBuilder
from semantica.vector_store import VectorStore

graph = ContextGraph()
GraphBuilder(graph_store=graph).build(
    entities=[{"id": "e1", "type": "Thing", "name": "Example", "properties": {}}],
    relationships=[],
)

vs = VectorStore(backend="faiss", dimension=384)
ctx = AgentContext(vector_store=vs, knowledge_graph=graph)
ctx.store("Some real content to retrieve later", metadata={"content": "Some real content to retrieve later"})

sources = ctx.retrieve("query about the content", max_results=5)
print(sources[0].content)  # -> '' (expected: the stored text)
print(sources[0].metadata)  # -> {} (expected: {'content': 'Some real content to retrieve later'})
```

You'll also see a `TextEmbedder` warning logged during `retrieve()`: `Embedding generation failed: Text cannot be empty or whitespace-only`.

## Expected Behavior

Metadata passed to `AgentContext.store()` / `AgentMemory.store()` should be retrievable from `VectorStore.search()` results regardless of which backend (`faiss`, `weaviate`, `qdrant`, etc.) is configured, since `store_vectors(vectors, metadata=...)` documents `metadata` as a supported parameter.

## Actual Behavior

For the FAISS backend, stored vector metadata is always `{}` on retrieval.

## Root Cause

In `semantica/vector_store/vector_store.py`, `VectorStore.store_vectors()`:

```python
if self._backend_store:
    if hasattr(self._backend_store, 'add'):
        return self._backend_store.add(vectors, metadata, **options)
    elif hasattr(self._backend_store, 'add_vectors'):
        if hasattr(self._backend_store, 'store_vectors'):
            return self._backend_store.store_vectors(vectors, metadata=metadata, **options)
        else:
            # Basic add_vectors without metadata
            return self._backend_store.add_vectors(vectors, **options)   # <-- metadata dropped here
```

`FAISSStore` has `add_vectors()` but neither `add()` nor `store_vectors()`, so it always falls into the `else` branch, and `metadata` is never passed even though `FAISSStore.add_vectors(vectors, ids=None, metadata=None, **options)` fully supports it.

## Suggested Fix Direction

```python
else:
    return self._backend_store.add_vectors(vectors, metadata=metadata, **options)
```

`FAISSStore.add_vectors()` already accepts `metadata` as a keyword argument, so this should be a minimal, low-risk change. Worth checking whether other backends reachable through this `elif` branch (anything with `add_vectors` but neither `add` nor `store_vectors`) have the same signature before generalizing further.

## Environment

- semantica: `0.6.0`
- Python: `3.12.10`
- OS: Windows 11

KaifAhmad1:
## Summary
- `VectorStore.store_vectors()` was silently dropping the `metadata` argument for any backend that only exposes `add_vectors()` (notably `FAISSStore`), even though `FAISSStore.add_vectors()` fully supports it. This caused `AgentContext`/`AgentMemory` retrieval to always return empty `content`/`metadata` for FAISS-backed stores, and triggered spurious `TextEmbedder` empty-text warnings downstream.
- Metadata is now forwarded to `add_vectors()`, and is only passed when the backend's `add_vectors()` signature actually supports it (checked via `inspect.signature`, accepting either an explicit `metadata` parameter or a `**kwargs` catch-all) — this avoids a `TypeError` for any backend with a stricter signature that doesn't accept `metadata`.

Fixes #832

## Test plan
- [x] Reproduced the exact scenario from #832 (`AgentContext` + FAISS-backed `VectorStore`) — `sources[0].content`/`.metadata` are now populated instead of `''`/`{}`.
- [x] Added ad-hoc checks against synthetic backends (metadata-kwarg, `**kwargs`-only, strict-no-metadata `add_vectors`) confirming correct routing with no crash in any case.
- [x] `pytest tests/vector_store tests/context` — 641 passed, 36 skipped, same 7 pre-existing/unrelated failures (Pinecone client mocks, a Windows symlink test) that fail identically on `main` without this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`VectorStore.store_vectors()` silently dropped metadata for FAISS (and any `add_vectors`-only) backend** (#832, #835) by @KaifAhmad1
+  - `store_vectors()` fell into a branch that called `self._backend_store.add_vectors(vectors, **options)` without `metadata` whenever the backend exposed `add_vectors()` but neither `add()` nor `store_vectors()` — true for `FAISSStore`, the backend most real usage configures for genuine ANN search. Every caller that stores vectors with metadata (e.g. `AgentMemory._store_memory_vector()`, used internally by `AgentContext.store()`) lost that metadata once it reached FAISS, with no error or warning
+  - Downstream, `ContextRetriever._retrieve_from_vector()` recovers a result's text via `metadata.get("content", "")`, which was always `""` for any vector stored this way; `_rank_and_merge()` then embedded that empty string, tripping `TextEmbedder.embed_text()`'s empty-text rejection and masking the real bug as a spurious `TextEmbedder` failure recorded by the progress tracker
+  - `store_vectors()` now forwards `metadata` to `add_vectors()`, but only when the backend's `add_vectors()` signature actually accepts it (checked via `inspect.signature`, accepting either an explicit `metadata` parameter or a `**kwargs` catch-all), so a future/custom backend with a stricter signature raises no `TypeError`
+  - **Follow-up review fix**: the `inspect.signature()` probe is wrapped in `try/except (ValueError, TypeError)`, consistent with the identical pattern already used in `ProvenanceManager.trace_lineage()`, so signature introspection failing on an unusual callable can no longer abort `store_vectors()` before it even attempts to call the backend
+
 - **`AgnoDecisionKit.check_policy` silently treated unevaluable policy rules as compliant** (#778, #822) by @Sameer6305
   - `_eval_rule()` previously `return`ed `True` when a rule referenced a field missing from the decision payload, or when the rule string didn't match the expected `<field> <op> <value>` format — the docstring's claim that exceptions never silently return `compliant=True` didn't cover this, since neither path raised
   - Both cases now raise `ValueError` instead, which routes through `check_policy`'s existing exception handler and records a `warnings` entry (e.g. `"Could not evaluate rule 'minimum_score >= 0.9': rule references undefined field 'minimum_score'"`) instead of disappearing with no signal
--- a/semantica/vector_store/vector_store.py
+++ b/semantica/vector_store/vector_store.py
@@ -67,6 +67,7 @@ License: MIT
 
 from typing import Any, Dict, List, Optional, Tuple, Union
 import concurrent.futures
+import inspect
 
 import numpy as np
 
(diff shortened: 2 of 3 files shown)


Sameer6305 (approved):
I completed a full review of this PR, including tracing the end-to-end execution path from `AgentMemory`/`AgentContext` through `VectorStore.store_vectors()` to the backend implementations.

### Findings

- ✅ The reported root cause was confirmed. `VectorStore.store_vectors()` dropped the `metadata` argument when dispatching to backends that only implement `add_vectors()`, even though `FAISSStore.add_vectors()` fully supports it.
- ✅ The implementation correctly forwards metadata only when the backend supports it (explicit `metadata` parameter or `**kwargs`) while remaining backward compatible with strict implementations.
- ✅ I audited the repository for similar dispatch paths and did not find another occurrence of the same bug. The remaining observations (such as positional metadata forwarding in the `add()` path and the provenance wrapper) are pre-existing and independent of this issue.

### Additional work completed

The only remaining gap was regression coverage for the newly introduced dispatch path. I added two focused tests:

- `test_store_vectors_metadata_forwarding`
  - Verifies that `VectorStore.store_vectors()` forwards metadata through the `add_vectors()` dispatch path to a backend that supports metadata.
  - This exercises the exact code path that was broken in #832.

- `test_store_vectors_strict_backend`
  - Verifies that a backend exposing `add_vectors(self, vectors)` (without `metadata` or `**kwargs`) continues to work without raising `TypeError`.
  - Protects the backward-compatibility guarantee introduced by this PR.

Both tests pass successfully and provide regression protection for the behavior introduced by this fix.

With those tests in place, I'm satisfied that:
- the original bug is resolved,
- the implementation is backward compatible,
- and the new behavior is now covered by automated tests.

**Approved.**

PR #836, opened 2026-08-05
fix(explorer): resolve infinite render loop preventing Temporal panel from rendering (#830)

### LINKED ISSUE #830: Temporal panel never renders — "Maximum update depth exceeded" infinite render loop on open
## Summary

Clicking the "Temporal" toolbar toggle in the Graph workspace opens the panel dock, but the panel content never renders. It stays permanently stuck on the lazy-load fallback ("Loading temporal…"), and the browser console logs a repeating React error.

## Steps to reproduce

1. Start the explorer backend against any graph JSON with at least one node:
   ```
   python -m semantica.explorer --graph <file>.json --port 8000
   ```
2. Start the frontend dev server:
   ```
   cd explorer && npm run dev
   ```
3. Open the app, click "Open Semantica Explorer" to enter the Graph workspace.
4. Click the "Temporal" button in the toolbar's ANALYSIS group.
5. Observe: the panel dock opens but shows "Loading temporal…" indefinitely—it never resolves to the actual panel content.
6. Open the browser devtools console and observe a repeating error.

## Actual behavior

Console error (repeats continuously once triggered):

```
Maximum update depth exceeded. This can happen when a component calls
setState inside useEffect, but useEffect either doesn't have a
dependency array, or one of the dependencies changes on every render.
```

The panel never progresses past the "Loading temporal…" lazy-load fallback.

## Expected behavior

Clicking "Temporal" should open the panel and render its content (the current scrubber time, temporal bounds, active node count, etc.) without any console errors.

## Scope / confirmed pre-existing

- The error only appears **after** the Temporal panel toggle is clicked. It does not occur on initial page load or when opening other panels (Effects, Neighbors).
- Confirmed pre-existing on `main` with no other changes applied (reproduced via `git stash` against an unrelated feature branch, then against the clean base branch). This is not caused by any specific feature work and blocks the Temporal panel for every consumer.

## Suspected root cause / starting point

Likely in `GraphWorkspace.tsx`'s plugin-loading `useEffect`—the one that iterates `pluginRegistry.forEach(...)` and lazily `import()`s a plugin's module when `entry.shouldLoad({ panelState, temporalState })` is true. For the `temporal-overlay` registry entry, `shouldLoad` is:

```ts
shouldLoad: ({ panelState, temporalState }) =>
  Boolean(panelState["temporal-panel"] || temporalState?.currentTime),
```

If `temporalState` (built via `useMemo` elsewhere in `GraphWorkspace.tsx`) is a new object reference on every render—for example because one of its own `useMemo` dependencies is itself unstable—then this effect's dependency array (`[loadedPlugins, pluginPanelState, pluginRegistry, temporalState]`) changes every render once the Temporal panel is open, retriggering the effect, which sets `loadedPlugins` state, which triggers a re-render, which produces a new `temporalState` again, forming the infinite loop.

Worth checking:

- the `useMemo` for `temporalState` in `GraphWorkspace.tsx` and its dependency list,
- whether `scrubberTime`, `activeNodeCount`, and `temporalBounds` (its likely inputs) are stable references across renders while the panel is open.

## Impact

Blocks any feature or user interaction that depends on the Temporal panel actually rendering (e.g. reviewing temporal bounds or the compare/diff UI). Currently the only way to confirm Temporal-panel-based features work is to test them via direct module execution against the live app rather than clicking through the UI.

Sameer6305:
## Summary

Closes #830.

Fixes the infinite React render loop that prevented the Temporal panel from rendering after it was opened from the Graph workspace.

Previously, opening the **Temporal** panel caused React to repeatedly throw:

```
Maximum update depth exceeded
```

and the panel never progressed beyond the lazy-loading fallback:

```
Loading temporal…
```

making the entire Temporal panel unusable during development.

---

## Root Cause

This was not caused by a single faulty `useEffect` or one unstable dependency.

The investigation found that multiple independent update paths were feeding back into each other, creating a self-sustaining render loop.

The update flow was:

```
Open Temporal panel
        ↓
Temporal plugin loads
        ↓
TimelinePanel emits scrubber updates
        ↓
GraphWorkspace rebuilds temporal state
        ↓
Diagnostics callback publishes a new state object
        ↓
React schedules another render
        ↓
TimelinePanel remounts
        ↓
Repeat indefinitely
```

Three separate contributors were identified:

- **Plugin eager-loading** triggered from transient temporal state changes.
- **Diagnostics state churn**, where identical diagnostics were repeatedly published as new state objects.
- **Repeated scrubber updates**, where identical timestamps continued propagating through the workspace.

Together these produced the infinite render loop responsible for the repeated React error.

---

## Fix

This PR breaks the render cycle at each contributing point.

### Plugin loading

- Prevent unnecessary eager loading driven by transient temporal state.
- Restrict Temporal plugin loading to genuine activation conditions.

### Diagnostics updates

- Replace unconditional diagnostics state updates with equality-aware updates.
- Preserve the existing state object whenever diagnostics are unchanged.
- Eliminate unnecessary renders caused only by object identity changes.

### Scrubber updates

- Suppress redundant scrubber updates.
- Prevent identical timestamps from continuously propagating through the render pipeline.
- Stop repeated TimelinePanel update cycles.

---

## Regression Coverage

Added focused regression coverage for the Temporal plugin loading behaviour to ensure the original render-loop scenario cannot regress unnoticed.

---

## Validation

Verified using the reproduction steps from #830.

### Development (`npm run dev`)

- ✅ Temporal panel opens successfully.
- ✅ Panel renders correctly.
- ✅ No infinite loading state.
- ✅ No `Maximum update depth exceeded` errors.

### Existing Frontend Tests

- All affected frontend test suites continue to pass.

---

## Scope

This PR only fixes the render loop preventing the Temporal panel from rendering.

No Temporal diff functionality or user-facing feature behaviour has been changed beyond eliminating the render loop.


Sameer6305:
@KaifAhmad1 ready for review.

KaifAhmad1 (approved):
## Review

Fixes the #830 render loop with a well-reasoned diagnosis: verified the diagnostics-dedup and scrubber-time-dedup logic against the actual `GraphRuntimeDiagnosticsSnapshot`/`GraphEffectAvailability` types and `GraphCanvas`'s effect deps — the comparisons are correct and exhaustive, and `distanceVisual`'s reference-equality check holds since it's `useMemo`'d upstream. No functional bugs in the render-loop fix itself.

Follow-up commits (pushed to this branch) addressed everything raised in review:
- Resolved the `CHANGELOG.md` conflict against `main` (kept both this entry and the already-merged HybridSearch/VectorStore entries, still under `[Unreleased] > Fixed`).
- Wired `test:graph-store`, `test:graph-workspace`, and the new `test:plugin-registry` regression test into CI (`ci.yml`) — previously nothing ran the Explorer frontend's test suites except `npm run build`, so this fix's own regression coverage wasn't actually protecting anything. Verified locally: all 42 tests pass, and confirmed green in the real CI run.
- Broadened the `structureLayer` diagnostics-dedup comparison to also cover `disabledReason`/`curveCount`/`bridgeCurveCount`/`backboneCurveCount`, not just `cacheKey`/`lastDrawAt`/`enabled`, so a `disabledReason`-only transition doesn't leave the dev diagnostics panel stale.
- Fixed an unrelated pre-existing `github/codeql-action` stale pin (`v4` moved upstream) that was failing the branch's required `verify` check.

All 10 checks are green (`build`, `verify`, `eslint`, `security-scan`, `MSDO`, `Checkov`, `terrascan`, `templateanalyzer`, `GitGuardian`, `Analyze Python`), no merge conflicts, and I built + ran the full Explorer test suite locally to confirm nothing regressed.

Approving.

PR #837, opened 2026-08-05
fix: HybridSearch.search() crashes with AttributeError on non-inmemor…

### LINKED ISSUE #833: HybridSearch.search() crashes with AttributeError for any VectorStore backend other than inmemory
## Summary

`HybridSearch.search()` directly accesses `self.vector_store.vectors`, a dict that `VectorStore.__init__` only creates when `backend="inmemory"`. For every other backend `VectorStore` supports (`faiss`, `weaviate`, `qdrant`, `milvus`, `pinecone`, `pgvector`, `sqlite`), that attribute is never set, so `HybridSearch.search()` raises `AttributeError: 'VectorStore' object has no attribute 'vectors'`. This is caught internally and reported as a failure to Semantica's progress tracker, but it means `HybridSearch` is effectively unusable with any real/production backend and only works against the in-memory demo store.

This is the same root issue previously observed surfacing through `AgentContext.find_precedents_advanced()`, which logs `"Vector store search failed, falling back to graph search: 'VectorStore' object has no attribute 'vectors'"` and silently falls back to graph search instead of raising — same underlying assumption, different call site.

## Steps to Reproduce

```python
from semantica.vector_store import VectorStore, HybridSearch

vs = VectorStore(backend="faiss", dimension=384)
vs.store_vectors(vectors=[[0.1] * 384], metadata=[{"content": "example"}])

hs = HybridSearch(vector_store=vs, graph_store=None)
hs.search(query="example", query_vector=[0.1] * 384)
```

## Expected Behavior

`HybridSearch.search()` should work against any backend `VectorStore` supports, not just `backend="inmemory"`, since nothing in `VectorStore`'s public API documents `.vectors` as a stable attribute for non-inmemory backends.

## Actual Behavior

```
AttributeError: 'VectorStore' object has no attribute 'vectors'
```

## Root Cause

In `semantica/vector_store/hybrid_search.py`, around line 314:

```python
vector_ids = list(self.vector_store.vectors.keys())
vectors = [self.vector_store.vectors[vid] for vid in vector_ids]
```

`self.vectors` is only initialized in `VectorStore.__init__` inside the `if self.backend == "inmemory":` branch (`semantica/vector_store/vector_store.py`, around line 118-124). Any other backend routes through `self._backend_store` instead and never gets a `.vectors` dict on the `VectorStore` instance itself.

## Suggested Fix Direction

`HybridSearch` should retrieve candidate vectors through `VectorStore`'s public, backend-agnostic API (e.g. `search_vectors()` / `search()`) rather than reaching into `.vectors` directly, so it works uniformly across all supported backends.

## Environment

- semantica: `0.6.0`
- Python: `3.12.10`
- OS: Windows 11

KaifAhmad1:
## Description

`HybridSearch.search()` reached directly into `self.vector_store.vectors`, an internal dict that `VectorStore` only populates for `backend="inmemory"`. Every other backend (faiss, weaviate, qdrant, milvus, pinecone, pgvector, sqlite) raised `AttributeError`, making `HybridSearch` unusable against any real store.

It now branches: for `inmemory` it keeps the existing local-corpus path; for everything else it delegates to `VectorStore.search_vectors()` (the backend-agnostic public API), applies `metadata_filter` as a post-filter over the returned candidates, and normalizes the result shape to `{id, score, distance, metadata}` so callers get a consistent contract regardless of backend.

Two rounds of fixes are in this PR:

**Round 1** — the crash itself, plus bugs surfaced while verifying the backend-delegated path end-to-end against a real FAISS store:
- `vector_ids` could stay `None` when callers passed explicit `vectors`/`metadata` without `vector_ids`, crashing downstream list indexing — now defaulted to generated positional IDs.
- The backend-delegated path forwarded `query_vector` as whatever type was given; a plain Python list crashes `FAISSStore.search_similar()`, which calls `.ndim` on it — now normalized to a numpy array before use.
- `VectorStore.store_vectors()` dropped metadata for FAISS (and any `add_vectors`-only backend) because it called `add_vectors(vectors, **options)` without forwarding `metadata`, even though `FAISSStore.add_vectors()` accepts it — this blocked metadata filtering from ever matching anything on FAISS. Fixed the call to forward `metadata`.

**Round 2** — three issues raised by automated code review, all confirmed and fixed:
- The legacy `top_k` kwarg was read but left in `options`, then forwarded via `**options` into `VectorStore.search_vectors()`, which itself passes an explicit `top_k=k` to some backends (sqlite, pgvector) — reproduced the resulting `TypeError: got multiple values for keyword argument 'top_k'` before fixing. Now popped instead of just read.
- `VectorStore.search_vectors()`'s dispatch only recognized backend methods named `search`/`search_similar`, so the delegation path still hit `NotImplementedError` for `qdrant`/`milvus`/`pinecone`, which name their method `search_vectors()` with a differently-named count parameter (`limit` vs `k`). Added a third dispatch branch that binds the count positionally so it works regardless of the backend's parameter name — verified these three backends now reach their real method instead of `NotImplementedError`.
- Backend-delegated results defaulted a missing `distance` to the raw `score`, silently reusing the local path's cosine-similarity convention (`distance = 1 - score`) even for backends using unrelated metrics (L2, inner product), which would produce nonsensical values. A missing `distance` is now left as `None` instead of a fabricated, metric-inconsistent number.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Fixes #833

## Changes Made

- `semantica/vector_store/hybrid_search.py` — `HybridSearch.search()`: delegate to `VectorStore.search_vectors()` for non-inmemory backends; normalize `query_vector` to ndarray; default `vector_ids` when omitted; normalize backend-delegated result schema (`distance` left `None` when the backend doesn't provide one); pop the legacy `top_k` kwarg and a stray `query_vector` kwarg out of `options` instead of leaving them to collide downstream.
- `semantica/vector_store/vector_store.py` — `store_vectors()`: forward `metadata` to `add_vectors()` in the `add_vectors`-only backend branch. `search_vectors()`: added a `search_vectors`-named dispatch branch (positional count arg) alongside the existing `search`/`search_similar` branches.

## Testing

- [x] Tested locally

### What was tested

- `tests/vector_store/` + `tests/context/` — 641 passed, 0 new failures. The only 7 failures (6 in `test_pinecone_store.py`, 1 symlink test in `test_agent_memory_markdown.py`) are pre-existing and unrelated — confirmed by reproducing the identical failures on unmodified `main`.
- Manual verification of the original issue repro plus 11 scenarios against real `faiss`, `sqlite`, and `inmemory` backends: metadata-filter through backend delegation, explicit `vectors`/`metadata` with and without `vector_ids`, empty inputs, no vector_store, the legacy `top_k` alias (including the collision fixed in round 2), and the stray `query_vector` kwarg collision.
- Checked all 7 supported backends for the specific reported `AttributeError`, and re-checked after round 2's dispatch fix:
  - `inmemory` / `faiss` / `sqlite`: verified live end-to-end, full results including correct ranking and metadata.
  - `pgvector`: confirmed dispatch reaches `PgVectorStore.add()` / `.search()` — only blocked from a full live run by no Postgres server available in this sandbox.
  - `qdrant` / `milvus` / `pinecone`: after round 2, dispatch now reaches their real `search_vectors()` method (confirmed by the error changing from `NotImplementedError` to a legitimate "collection/index not initialized" error, since there's no live server in this sandbox to connect to). Storage (`store_vectors()`) still doesn't recognize their `insert_vectors`/`upsert_vectors` method names, so these three can't get data in via `VectorStore`'s public API yet — a separate, pre-existing gap, out of scope here.
  - `weaviate`: still `NotImplementedError` on both store and search — it has neither `add`/`add_vectors`/`search`/`search_similar` nor `search_vectors`, only `add_objects`/`query_vectors`. Also out of scope here.

### Test Commands

```bash
pytest tests/vector_store/ tests/context/ -q
```

## Documentation

## Breaking Changes

**Breaking Changes**: No

The public `HybridSearch.search()` signature and return shape are unchanged. Backend-delegated results are now normalized to the same `{id, score, distance, metadata}` shape the local-search path already produced, so this only adds consistency, not a behavior change for existing (inmemory) callers.

## Checklist

## Additional Notes

Still out of scope for this PR: `weaviate` (fully unwired — different method names on both storage and search), and `qdrant`/`milvus`/`pinecone`'s storage side (`store_vectors()` doesn't yet recognize `insert_vectors`/`upsert_vectors`). Full end-to-end usage of those backends through `VectorStore`'s public API needs a follow-up on the storage-dispatch side specifically.


KaifAhmad1:
Addressed all three findings from the Qodo review in 94adcf7a:

1. **top_k forwarded causes crash** — confirmed by reproducing the exact `TypeError: got multiple values for keyword argument 'top_k'` against a real sqlite-backed store before fixing. `top_k` is now popped out of `options` instead of just read.
2. **Backend delegation still unsupported (search_vectors-named backends)** — added a third dispatch branch in `VectorStore.search_vectors()` for backends exposing `search_vectors()` (qdrant/milvus/pinecone), binding the count arg positionally since those backends name it differently (`limit` vs `k`). Verified live: these three now reach their real backend method instead of `NotImplementedError` (they still can't be fully exercised end-to-end without a live server in this environment, but the dispatch gap is closed).
3. **Distance fallback inconsistent** — a missing `distance` in backend-delegated results is now left as `None` instead of silently reusing `score`, since assuming cosine-similarity semantics (`distance = 1 - score`) would be wrong for backends using L2 or inner-product metrics.

Full re-verification: 641 passed / 7 pre-existing-and-unrelated failures across `tests/vector_store/` + `tests/context/`, plus manual re-run of all prior scenarios including a targeted repro of the `top_k` crash (confirmed fixed).

Sameer6305 (approved):
I reviewed the changes end-to-end and verified the reported issue.

### Review summary

- Confirmed the original root cause: `HybridSearch.search()` directly accessed `VectorStore.vectors`, which only exists for the `inmemory` backend, causing `AttributeError` for backend-backed stores.
- Verified that the new backend-delegation path correctly routes non-inmemory stores through `VectorStore.search_vectors()`, while preserving the existing in-memory behavior.
- Verified the follow-up fixes included in this PR:
  - `top_k` collision is resolved.
  - `query_vector` is normalized before backend dispatch.
  - backend `search_vectors()` dispatch is added for supported stores.
  - backend results are normalized to a consistent shape.
- Merged the latest `main` into this branch and resolved the conflicts by preserving the newer metadata-forwarding implementation already introduced in `main`, ensuring both fixes coexist correctly.
- Added regression coverage for the newly introduced backend-delegated `HybridSearch` path. The new test verifies delegation to `VectorStore.search_vectors()`, query normalization, parameter forwarding, and result normalization without relying on external services.

The new regression test passes together with the existing vector store test suite (aside from the known pre-existing skipped/environment-specific backend tests).

Overall, the implementation looks good and the new code path is now covered by automated tests.

PR #841, opened 2026-08-06
fix(vector_store): normalize QdrantStore.search_vectors() to return "metadata"

### LINKED ISSUE #840: HybridSearch drops Qdrant metadata because backend payload is not normalized
## Summary

The new backend-delegated path in `HybridSearch.search()` assumes backend results expose metadata under the key:

```python
metadata
```

However, `QdrantStore.search_vectors()` returns:

```python
payload
```

instead.

As a result:

- metadata is silently discarded
- metadata filtering always evaluates against an empty dictionary
- filtered searches incorrectly return no matches

---

## Execution path

```
HybridSearch.search()
    ↓
VectorStore.search_vectors()
    ↓
QdrantStore.search_vectors()
    ↓
QdrantCollection.search_points()

returns

{
    "id": ...,
    "score": ...,
    "payload": ...
}
```

HybridSearch then normalizes using

```python
result.get("metadata", {})
```

which always returns `{}`.

---

## Impact

Without filters:

- metadata is stripped from every result

With metadata filters:

- every candidate is rejected
- search returns an empty result set

---

## Suggested fix

Normalize backend payloads before applying metadata filtering.

Either:

- teach `HybridSearch` to recognize `"payload"`

or

- make `QdrantStore.search_vectors()` return the standard `"metadata"` key used elsewhere.

divyankshah:
## Summary
- `QdrantStore.search_vectors()` returned results keyed by `"payload"`, while `HybridSearch` and `PineconeStore` both use `"metadata"`. This silently dropped metadata from Qdrant results and caused `HybridSearch.filter_by_metadata()` to reject every candidate when a filter was applied (empty result sets).
- Normalized `QdrantCollection.search_points()` to return `"metadata"` instead of `"payload"`, matching the convention already used by `PineconeStore.search_vectors()`.

Fixes #840

## Test plan
- [x] Extended `tests/vector_store/test_vector_store_deepdive.py::test_qdrant_store` to assert Qdrant search results expose `"metadata"` (not `"payload"`) and that `HybridSearch.filter_by_metadata()` matches against them correctly.
- [x] `pytest tests/vector_store/` — 141 passed, 61 skipped, same 6 pre-existing `test_pinecone_store.py` failures as on unmodified `main` (unrelated mocking issue, verified via `git stash`).


KaifAhmad1 (approved):
Verified end-to-end:

- Diff is a single-line rename in `QdrantCollection.search_points()` (`qdrant_store.py:168`), `"payload"` → `"metadata"`.
- Traced the full call chain (`QdrantStore.search_vectors()` → `QdrantSearch.similarity_search()` → `QdrantCollection.search_points()`) — no intermediate re-keying, so this one line is the sole source of the bug and the fix is complete, not partial.
- Confirmed `PineconeStore.search_vectors()` already returns `"metadata"`, so this genuinely aligns Qdrant with the existing convention rather than introducing a new one.
- Grepped the repo for any consumer reading `result["payload"]` off vector-store search output — none found, so the rename breaks nothing downstream.
- Ran `pytest tests/vector_store/`: 167 passed, 36 skipped, same 6 pre-existing `test_pinecone_store.py` failures reproduced independently on unmodified `main` (unrelated mocking issue, confirmed via direct branch comparison, not just `git stash`).
- New test (`test_qdrant_store`) directly asserts the key rename and exercises `HybridSearch.filter_by_metadata()` against real Qdrant-shaped results — matches the actual bug path from #840, not just the surface rename.

Minor non-blocking note: the module docstring and `insert_vectors()`'s `payloads` param still use Qdrant's native "payload" terminology for the *storage* side — fine to leave, since this PR is scoped to the *search result* key other modules consume.

Approving.

PR #842, opened 2026-08-06
Fix #839: Support persistent backends in DecisionEmbeddingPipeline

### LINKED ISSUE #839: DecisionEmbeddingPipeline crashes on non-inmemory VectorStore backends due to direct .vectors access
## Summary

`DecisionEmbeddingPipeline` is not backend-agnostic.

During semantic decision search it directly accesses the internal in-memory storage of `VectorStore`:

```python
for vector_id, vector in self.vector_store.vectors.items():
```

However, `.vectors` and `.metadata` are only initialized when `VectorStore` is created with the `inmemory` backend.

Any persistent backend (FAISS, pgvector, Qdrant, Milvus, etc.) delegates storage to `_backend_store` and never creates these attributes.

As a result, semantic decision search crashes with:

```
AttributeError:
'VectorStore' object has no attribute 'vectors'
```

---

## Execution path

```
VectorStore.search_decisions()
    ↓
DecisionEmbeddingPipeline.find_similar_decisions()
    ↓
_get_candidate_embeddings()
    ↓
self.vector_store.vectors.items()
```

---

## Root cause

`DecisionEmbeddingPipeline` assumes every `VectorStore` exposes the internal in-memory dictionaries.

That assumption is only valid for the `inmemory` backend.

Persistent backends intentionally store vectors inside backend-specific implementations.

---

## Impact

Decision search currently only works correctly for the `inmemory` backend.

Production backends cannot execute this path.

---

## Existing tests

Current tests do not expose this bug because `_get_candidate_embeddings()` is mocked during the search tests.

The failing path is never executed with a real backend.

---

## Suggested fix

Instead of loading every vector from internal dictionaries:

1. use `VectorStore.search_vectors()` to retrieve the top semantic candidates
2. extract the structural embeddings from the returned metadata
3. perform hybrid re-ranking only on those candidates

This also avoids loading the entire vector database into memory.

Sameer6305:
## Description
This PR resolves [#839] by replacing direct accesses to the internal `VectorStore.vectors` and `VectorStore.metadata` dictionaries inside the `DecisionEmbeddingPipeline` with proper API calls to `search_vectors()`. This enables the pipeline to correctly execute hybrid similarity searches against persistent vector databases (e.g., FAISS, Pinecone) without encountering an `AttributeError`.

## Changes Made
- **Candidate Embedding Retrieval**: Modified `_get_candidate_embeddings` to fetch candidate vectors via `VectorStore.search_vectors()` instead of looping through all items in the `.vectors` dictionary (which is exclusively defined for `inmemory` stores).
- **Hybrid Similarity Fallback**: Backend-agnostic support natively strips raw semantic vectors in several vector stores (such as FAISS) when querying. Added a fallback path in `find_similar_decisions` and `_find_semantic_similar` to utilize the query `score` provided by `search_vectors()` as the semantic similarity, seamlessly patching into the existing `HybridSimilarityCalculator` workflow without causing regressions.
- **Statistics Safety Guard**: Fixed an identical bug in `decision_vector_methods.py:get_decision_statistics()` where iteration over `store.metadata.values()` would cause a hard crash on persistent backends. It now safely falls back to a limited schema with a warning if the store does not expose internal metadata dictionaries.
- **Regression Tests**: Added end-to-end integration tests using `inmemory` and `faiss` backends directly (without mocks) to enforce before/after equivalence on candidate structural evaluation and explicitly ensure no crashes on real vector DBs.

## Tradeoffs & Notes
By design, `DecisionEmbeddingPipeline` historically evaluated the semantic and structural relationships for *every single embedded decision* in the local dictionary. Using a vector DB's `search_vectors(..., k=limit)` now correctly caps the initial semantic search to the top semantic candidates (by default `limit * 5`) before passing them down for structural re-ranking. While this resolves O(N) scaling bottlenecks and allows persistent backend support, it bounds maximum recall for results relying purely on extreme structural weights.

## Verification
- [x] Ran reproduction script against FAISS with success
- [x] Assured equivalence of result shape/ranking for `inmemory` backends
- [x] Full `pytest` suite passes locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`DecisionEmbeddingPipeline.find_similar_decisions()` crashed with `AttributeError` for any `VectorStore` backend other than `inmemory`** (#842, closes #839) by @Sameer6305
+  - `_get_candidate_embeddings()` iterated `VectorStore.vectors`/`VectorStore.metadata` directly, internal dicts only populated for `backend="inmemory"`; every persistent backend (FAISS, Pinecone, Qdrant, Milvus, ...) raised `AttributeError`. It now fetches candidates via the backend-agnostic `VectorStore.search_vectors()`, reading metadata via a `res.get("metadata") or res.get("payload")` fallback for backends that key it differently
+  - Backends such as FAISS don't return the raw vector for each hit; `find_similar_decisions()` and `_find_semantic_similar()` now fall back to the search-provided score (normalized from `distance` when present) as the semantic similarity for those candidates instead of computing cosine similarity against a zero placeholder vector
+  - `get_decision_statistics()` had the identical bug iterating `store.metadata.values()`; it now returns a limited stats payload with an explanatory `warning` field for backends that don't expose a full in-memory metadata dict, instead of crashing
+  - **Fixed along the way**: `_get_candidate_embeddings()`'s expand-and-retry loop (which widens the search pool when post-filtering leaves too few matches) discarded every candidate it had found once the pool hit its cap (`limit * 10`) without ever collecting `limit` matches or getting a short page back from the backend — the loop fell through without executing the branch that assigns results, silently returning `[]` even when matching candidates existed. It now falls back to the last batch collected instead of dropping it
+  - Added end-to-end regression tests against real `inmemory` and `faiss` backends (no mocks) plus a targeted unit test for the expand-and-retry loop's fallback behavior
+
 - **`QdrantStore.search_vectors()` returned results keyed by `"payload"` instead of `"metadata"`** (#841, closes #840) by @divyankshah
   - `QdrantCollection.search_points()` built its result dicts as `{"id", "score", "payload"}`, while `PineconeStore.search_vectors()` and every other backend consumed by `HybridSearch` use `"metadata"`. This silently dropped Qdrant metadata from results and made `HybridSearch.filter_by_metadata()` reject every candidate whenever a filter was applied, since it looks up `result["metadata"]` and got nothing back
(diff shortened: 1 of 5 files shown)


Sameer6305:
@KaifAhmad1 ready for review.

KaifAhmad1:
> @KaifAhmad1 ready for review.

Fix CI failure in this same PR

Sameer6305:
done.

KaifAhmad1 (approved):
Reviewed the full diff plus the two rounds of automated review comments (Copilot/Qodo).

**Core fix (#839):** `DecisionEmbeddingPipeline` now goes through `VectorStore.search_vectors()` instead of reaching into the `inmemory`-only `.vectors`/`.metadata` dicts, so `find_similar_decisions()` and `get_decision_statistics()` no longer crash with `AttributeError` on FAISS/Pinecone/Qdrant/Milvus backends.

**Qodo's 4 findings** (FAISS distance mis-ranking, filter kwargs breaking backends, `payload` vs `metadata` key mismatch, non-skippable FAISS test) were all resolved in the second commit — verified against the current code, not just the bot's "Resolved" label.

**Found and fixed in a follow-up commit (`721a2f0e`):** `_get_candidate_embeddings()`'s expand-and-retry loop (which widens the search pool when post-filtering leaves too few matches) discarded every candidate it had collected once the pool hit its `limit*10` cap without ever reaching `limit` matches or getting a short page back — it fell through the loop without executing the branch that assigns results, silently returning `[]` for filtered queries even when matches existed. This directly undermined the PR's stated goal (filtered search against persistent backends), so I fixed it to fall back to the last collected batch, added a targeted regression test reproducing the exact failure mode, and documented both fixes plus this PR in CHANGELOG.md.

**Verification:** full `tests/vector_store/` suite green (38 tests including 2 new regression tests), all 10 CI checks passing on the latest commit, no merge conflicts with `main`.

Looks good to merge.

Issue #845, opened 2026-08-06
Standardize VectorStore.search_vectors() output schema across all backend implementations


Sameer6305:
## Summary

Different vector database backends return search results in inconsistent formats. The current `VectorStore` abstraction does not normalize these responses, forcing downstream consumers to implement backend-specific parsing logic.

## Description

Each backend exposes different field names and scoring semantics.

Current inconsistencies include:

- **Scores**
  - Some backends return cosine similarity (higher is better).
  - Others return L2 distance (lower is better).

- **Metadata**
  - In-memory backend returns `metadata`.
  - Qdrant returns `payload`.
  - Milvus may omit metadata entirely.

- **Vectors**
  - Some backends return `vector`.
  - Others return `values`.
  - Some omit vectors unless explicitly requested.

As a result, downstream components (such as `DecisionEmbeddingPipeline`) must guess which fields exist and implement multiple fallback paths, increasing complexity and the risk of silent bugs.

## Suggested Fix

Introduce a standardized search result schema for every backend.

Example:

```python
@dataclass
class SearchResult:
    id: str
    score: float
    metadata: dict
    vector: Optional[np.ndarray]
```

(or an equivalent `TypedDict`).

Require every backend wrapper (FAISS, Qdrant, Milvus, Pinecone, etc.) to convert its native client response into this schema before returning results.

Additionally:

- Normalize `score` into a consistent similarity metric (for example, 0.0–1.0).
- Always expose metadata through the `metadata` field.
- Always expose vectors through the `vector` field (using `None` when unavailable).

## Benefits

- Eliminates backend-specific parsing.
- Simplifies downstream pipelines.
- Reduces conditional logic throughout the codebase.
- Provides a stable API regardless of the underlying vector database.

PR #854, opened 2026-08-08
fix(vector_store): stop bypassing backend abstraction in build_decision_context/explain_decision/_filter_by_metadata (closes #848)

### LINKED ISSUE #848: AttributeError: 'VectorStore' object has no attribute 'vectors' in build_decision_context() and explain_decision()
## Summary

`VectorStore.build_decision_context()` and `VectorStore.explain_decision()` assume that vectors are always stored in the in-memory `self.vectors` dictionary. As a result, both methods raise an `AttributeError` when used with persistent backends (e.g. FAISS, Qdrant, Pinecone, Milvus).

## Description

Both methods retrieve the query vector by directly accessing `self.vectors`.

### In `build_decision_context()` (approx. line 1035)

```python
# Add related decisions based on similarity
if decision_id in self.vectors:
    query_vector = self.vectors[decision_id]
```

### In `explain_decision()` (approx. line 1089)

```python
if include_paths:
    # Find similar decisions for reasoning paths
    if decision_id in self.vectors:
        query_vector = self.vectors[decision_id]
```

However, `self.vectors` is only initialized for the `inmemory` backend. Persistent backends store vectors in the backend implementation, so these methods fail immediately with:

```text
AttributeError: 'VectorStore' object has no attribute 'vectors'
```

## Expected Behavior

These methods should work regardless of the configured backend.

## Root Cause

The implementation bypasses the backend abstraction by accessing the in-memory storage directly instead of using the public retrieval API.

## Proposed Fix

Use the backend-agnostic `get_vector()` method instead of directly accessing `self.vectors`.

For example:

```python
query_vector = self.get_vector(decision_id)
if query_vector is not None:
    # proceed with similarity search...
```

This allows the method to work for both the in-memory implementation and persistent vector stores without backend-specific logic.

## Impact

- `build_decision_context()` is unusable on persistent backends.
- `explain_decision()` is unusable on persistent backends when `include_paths=True`.
- The issue affects all non-in-memory backends that do not expose `self.vectors`.

Sameer6305:
Closes #848

## Summary

`VectorStore.build_decision_context()` and `VectorStore.explain_decision(include_paths=True)` both accessed `self.vectors` directly to retrieve a query vector for similarity search. `self.vectors` (and `self.metadata`) are only initialized when `VectorStore` is constructed with the `inmemory` backend — any persistent backend (FAISS, Qdrant, Pinecone, Milvus, etc.) never sets these attributes at all, so both methods crashed immediately with:

AttributeError: 'VectorStore' object has no attribute 'vectors'

This is the fourth issue in a related cluster (#839 → #843 → #845 → #848) all tracing back to the same root cause: parts of this class were written assuming `inmemory` storage and never updated when the backend abstraction (`_backend_store`) was introduced for persistent backends.

## Root cause and fix

**`build_decision_context()`** and **`explain_decision()`** both had the identical pattern:
```python
if decision_id in self.vectors:
    query_vector = self.vectors[decision_id]
```
Replaced with `#843`'s already-established backend-agnostic accessor:
```python
query_vector = self.get_vector(decision_id)
if query_vector is not None:
```
Verified this is a faithful 1:1 behavioral equivalent for the inmemory case — `get_vector()`'s inmemory branch is `self.vectors.get(vector_id)`, which returns `None` exactly when the old `in` check would have been `False`. Confirmed the rest of both methods only consumes `search_vectors()`'s already-backend-agnostic output afterward, so nothing further down either method needed changing.

## A third crash site found during verification, not mentioned in the original issue

Grepping the rest of the file for the same pattern surfaced `_filter_by_metadata()` (reached via the public `filter_decisions()` when called with `query=None`), which had the identical unguarded `self.metadata.items()` / `self.vectors.get()` access — same crash, same root cause, just not named in #848.

The first fix attempt gated this behind `_backend_store is not None` and returned `[]` for any backend without a `filter_by_metadata` method. On review, this was itself a new bug of the same *class* the whole issue cluster is about: an empty list from `filter_decisions(category="loan")` is indistinguishable from "zero decisions actually matched" — a caller has no way to tell "genuinely no results" apart from "this backend doesn't support this operation at all." That's a silent failure mode, not a fix.

**Reconciled** to raise `NotImplementedError` instead, matching the precedent `get_vector()`/`get_metadata()` already established in `#843` for exactly this situation (backend exists, but doesn't implement the requested capability). Before finalizing this, grepped all 7 real backend wrapper classes (`faiss_store.py`, `pgvector_store.py`, `qdrant_store.py`, `weaviate_store.py`, `milvus_store.py`, `pinecone_store.py`, `sqlite_vec_store.py`) and confirmed **none currently implement `filter_by_metadata`** — the only method with that name anywhere in the codebase belongs to `HybridSearch`, a completely different class with a different two-argument signature, never assigned to `_backend_store`. The delegation check (`hasattr(...)`) is retained as a genuine forward-compatibility hook for if a backend ever adds this method, but today the `NotImplementedError` path is what every persistent backend actually hits, with a message that names the backend and points the caller to `search_decisions()` as the supported alternative.

## Testing

14 new tests added across two verification rounds in `tests/vector_store/test_decision_embedding_pipeline.py`:

- **Inmemory behavioral equivalence** (4 tests) — proves `build_decision_context()`/`explain_decision()` produce identical output before and after this change for the inmemory backend, including the missing-decision-id path.
- **Real FAISS backend regression tests** (6 tests) — calls both methods against a real (not mocked) FAISS-backed `VectorStore`, the same backend already used by `#839`'s regression tests for consistency across this issue cluster. Confirms no `AttributeError`, correct related-decisions population, and correct `ValueError` (not `AttributeError`) for a genuinely missing decision ID.
- **`_filter_by_metadata` behavior** (4 tests) — confirms `NotImplementedError` fires with a clear message for a persistent backend, confirms neither the old `AttributeError` crash nor the intermediate silent-`[]` behavior remain, and confirms the inmemory path is completely unaffected.

**Full suite: 53 passed, 0 failed** — 0 regressions across the 39 pre-existing tests.

## Scope note

This PR only touches the three methods above. A final full-file grep for any *remaining* unguarded `self.vectors`/`self.metadata` access (beyond what this PR and #839/#843 already covered) is recommended as a follow-up rather than expanding this PR's scope further — see the issue proposed below.


Sameer6305:
@KaifAhmad1 ready for review.

KaifAhmad1 (approved):
Reviewed the diff, traced the full call path (get_vector/get_metadata backend-agnostic accessors, _backend_store guard, filter_decisions routing), and ran the test suite against this branch locally:

- `pytest tests/vector_store/test_decision_embedding_pipeline.py -v` → 47/47 passed, including all 14 new tests (inmemory equivalence, real-FAISS regression, `_filter_by_metadata` NotImplementedError behavior, and the stub-backend warning/marker tests).
- Confirmed `get_vector()` returning `None` is a faithful equivalent to the old `decision_id in self.vectors` check, and that FAISS's `get_vector()` can genuinely return `None` (not raise) for indices lacking reconstruction support, so the new `similarity_unavailable` fallback path is real, not dead code.
- Confirmed no backend wrapper currently implements `filter_by_metadata`, so `NotImplementedError` (vs. silently returning `[]`) is the correct behavior today, consistent with the `#843` precedent for `get_vector`/`get_metadata`.
- Checked all callers of `filter_decisions`/`build_decision_context`/`explain_decision` — none swallow exceptions broadly, so the new `NotImplementedError` propagates cleanly.

No correctness issues found. LGTM.

PR #857, opened 2026-08-08
Added custom _filter_by_metadata for each memory backend

TaherTadpatri:
## Description

Fixed an issue where `VectorStore._filter_by_metadata()` directly iterated over `self.metadata`, causing an `AttributeError` on all persistent backends (`faiss`, `qdrant`, `pinecone`, `milvus`, `pgvector`, `sqlite`, `weaviate`) because `self.metadata` only exists on the `inmemory` backend. Metadata filtering is now delegated to the active backend store, leveraging backend-native payload/SQL/JSON filtering capabilities.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring

## Changes Made

- **`VectorStore`**: Updated `_filter_by_metadata()` to delegate to `self._backend_store.filter_by_metadata()` when using persistent backends instead of accessing `self.metadata` directly.
- **Helper `_matches_filter`**: Added reusable helper function for evaluating metadata criteria (exact match, min/max range, and list membership/intersection).
- **Backend Implementations**: Implemented native `filter_by_metadata()` method across all persistent vector store implementations:
  - `FAISSStore`: In-memory metadata map traversal without `AttributeError` on `VectorStore`.
  - `QdrantStore`: Native payload filtering using `client.scroll()`.
  - `PineconeStore`: Native metadata query filters using `index.query()`.
  - `MilvusStore`: Native expression filters using `collection.query()`.
  - `PgVectorStore`: Native PostgreSQL `JSONB` WHERE clauses (`metadata->>key`).
  - `SQLiteVecStore`: Native SQLite `json_extract()` WHERE clauses.
  - `WeaviateStore`: Native collection object query filters.
- **Unit Tests**: Added test coverage in `tests/vector_store/test_vector_store.py` and `tests/vector_store/test_backend_metadata_filtering.py`.

## Testing

- [x] Tested locally

### Test Commands

```bash
pytest tests/vector_store/test_vector_store.py tests/vector_store/test_backend_metadata_filtering.py
```

## Documentation

## Breaking Changes

**Breaking Changes**: No

## Checklist

## Additional Notes

All 17 vector store metadata filtering unit tests pass cleanly.


KaifAhmad1:
@TaherTadpatri Could you please address the Qodo findings before we merge this PR? Thanks!

TaherTadpatri:
Thank you for the review , here are the changes
# Vector Store Metadata Filtering Fixes Summary

This document summarizes the root causes, implementations, and verification details for the metadata filtering issues resolved across `WeaviateStore`, `PineconeStore`, and `MilvusStore`.

---

## 1. WeaviateStore (`semantica/vector_store/weaviate_store.py`)

### Issue
`WeaviateStore.filter_by_metadata()` called `fetch_objects(limit=limit)` without backend filter criteria, then applied `_matches_filter()` client-side. This limited the number of *scanned* objects rather than the *returned matches*, causing false negatives whenever matching objects were located past the first page of objects.

### Fix Details
1. **Native Filtering (`_build_weaviate_filter`)**:
   - Added helper method to translate filter specifications (exact equality, numeric range `min`/`max`, and list membership `contains_any`) into native `weaviate.classes.query.Filter` objects.
   - Native filter is passed to `fetch_objects(filters=...)` for server-side evaluation.

2. **Paginated Object Scanning**:
   - Implemented an iterative pagination loop using cursor pagination (`after=after_cursor`) with `offset` fallback.
   - Continues scanning pages until `len(results) == limit` or the collection is exhausted.

---

## 2. PineconeStore (`semantica/vector_store/pinecone_store.py`)

### Issue
`PineconeStore.filter_by_metadata()` used a hardcoded dummy query vector fallback of `dimension=768`. Because `PineconeStore` did not record `self.dimension` when indexes were created or loaded, metadata filtering failed at runtime for any index not sized 768 and silently returned `[]`.

### Fix Details
1. **Dimension Tracking**:
   - Tracked `self.dimension` across `__init__()`, `create_index()`, `get_index()` (auto-detecting via `describe_index_stats()`), `upsert_vectors()`, and `search_vectors()`.

2. **Removed Unsafe Fallback**:
   - Removed `getattr(self, "dimension", 768)`.
   - In `filter_by_metadata()`, if `self.dimension` is uninitialized and cannot be inferred via stats, an explicit `ProcessingError` is raised requesting the dimension parameter or prior initialization via `create_index()` / `get_index()`.
   - Added safe checks for `ServerlessSpec` when `pinecone` client is mocked in tests.

---

## 3. MilvusStore (`semantica/vector_store/milvus_store.py`)

### Issue
`MilvusStore.filter_by_metadata()` constructed boolean expression strings (`expr`) using raw f-strings without validating keys or escaping special characters (such as double quotes `"` or backslashes `\`) in keys and values. This resulted in invalid Milvus query syntax and expression injection vulnerabilities.

### Fix Details
1. **Key Validation & Escaping (`_validate_milvus_key`)**:
   - Validates metadata keys using a strict regex (`r"^[a-zA-Z0-9_.-]+$"`), raising `ValidationError` for malformed keys / injection attempts.
   - Escapes quotes and backslashes in keys.

2. **Value Escaping (`_format_milvus_value`)**:
   - Properly escapes string values (`\` -> `\\`, `"` -> `\"`, `\n` -> `\n`, `\r` -> `\r`), formats booleans as `true`/`false`, and formats numbers/lists for valid Milvus query syntax.
   - Updated `get_vector()` and `get_metadata()` ID query expressions to escape both backslashes and double quotes.

---

## 4. Test Results

All 44 tests in the vector store test suite passed cleanly:
- `tests/vector_store/test_backend_metadata_filtering.py`
- `tests/vector_store/test_pinecone_store.py`
- `tests/vector_store/test_vector_store.py`
- `tests/vector_store/test_vector_store_deepdive.py`

TaherTadpatri:
Thank you@KaifAhmad1 ,
I have resolved Qodo finding. Looking forward for your reply.

TaherTadpatri:
I have resolved the issue.
Can you please check it out
Thank you

Sameer6305:
> I have resolved the issue. Can you please check it out Thank you

Thanks @TaherTadpatri I will review it.

Sameer6305 (dismissed):
Thanks @TaherTadpatri for the work on #857. I reviewed the implementation against the existing `VectorStore` metadata-filtering contract, the individual backend implementations, related code paths, security considerations, and the existing test coverage.

### Problem addressed

The PR correctly addresses the underlying issue where `VectorStore._filter_by_metadata()` assumed an in-memory `self.metadata` structure and therefore failed for persistent backends.

The approach of moving metadata filtering to the active backend and implementing `filter_by_metadata()` per backend is the right architectural direction. It allows each persistent backend to use its native filtering/query capabilities while keeping the higher-level `VectorStore` API consistent.

### Review performed

I reviewed the implementation across all affected backends:

- FAISS
- Qdrant
- Pinecone
- Milvus
- PgVector
- SQLiteVec
- Weaviate

I also reviewed the shared `_matches_filter()` behavior, filtering semantics, SQL/query construction, input validation, error handling, backward compatibility, and the associated regression tests.

The implementation was generally well structured, and the backend-specific approaches are appropriate for the respective storage systems.

### Issues identified during review

Two concrete issues were identified during the review and were fixed directly in the PR.

#### 1. Qdrant — missing `Range` and `MatchAny` imports

The Qdrant implementation referenced `Range` and `MatchAny` when processing range and list filters, but those symbols were not imported or defined in the optional-dependency fallback.

This would result in runtime `NameError` exceptions for otherwise supported filter forms.

The imports and corresponding fallback definitions were added, and regression coverage was added for:

- range filters
- range filters with both `min` and `max`
- list membership filters

The tests verify the actual constructor arguments passed to the Qdrant filter objects.

#### 2. PgVector — boolean metadata filtering

The PgVector implementation converted scalar filter values using `str(value)`. For Python booleans this produces `True` / `False`, while PostgreSQL JSONB `->>` extraction represents JSON booleans as lowercase `true` / `false`.

As a result, boolean metadata filters could silently return no matches.

A dedicated boolean branch was added so that:

- `True` → `"true"`
- `False` → `"false"`

Regression tests were added for both values and verify the actual parameters passed to the database cursor.

### Verification

After the review fixes, the focused backend metadata-filtering tests pass completely:

- **38/38 tests passing**
- `git diff --check` clean
- No unrelated files modified

The broader audit also verified the filtering contract across the backends, including:

- exact-match filtering
- range filtering
- list membership
- multiple-filter AND semantics
- empty filters
- missing metadata keys
- boolean handling
- result construction
- backend-specific query generation

Security checks also found no injection vulnerabilities in the reviewed implementations. User-controlled SQL/Milvus identifiers are validated and values are parameterized or safely escaped where applicable.

### Backward compatibility

The existing in-memory filtering behavior remains compatible with the established contract.

The delegation from `VectorStore` to backend-specific `filter_by_metadata()` implementations was also checked against the repository's backend implementations and did not introduce a breaking signature change.

The default behavior and existing filtering API remain intact.

### Remaining observations

A few non-blocking limitations were identified during the audit, particularly around:

- Weaviate named-vector representation in returned results
- Weaviate's full-collection fallback when native filtering fails
- list-vs-list intersection semantics for SQLite/PgVector
- compatibility with older pymilvus versions

These do not constitute blocking correctness or security issues for this PR and can be handled separately if needed.

### Final assessment

Overall, this is a solid and appropriately scoped implementation of backend-specific metadata filtering. The two concrete runtime issues found during review have been addressed with targeted fixes and regression coverage in 7010913.

After those fixes, I found **no remaining blocking correctness, security, or backward-compatibility issues**.

Thanks for taking this issue up and for the substantial backend coverage. The implementation is in a good state now and should be ready for merge.

KaifAhmad1:
Ran a deep review of this PR and pushed a follow-up commit (cab995dc) with fixes. Summary below for @Sameer6305 @TaherTadpatri.

## Merge conflict

Merged `main` in and resolved the conflict in `tests/vector_store/test_vector_store.py` (two independent, non-overlapping test additions — kept both). The PR is now `MERGEABLE`.

## Additional correctness issues found and fixed

1. **`pinecone_store.py`** — `get_index()` and `filter_by_metadata()` called `self.describe_index_stats()`, but that method only exists on the `PineconeIndex` wrapper (`self.index`), not on `PineconeStore` itself. The resulting `AttributeError` was silently swallowed, so dimension auto-detection always failed quietly. Fixed to call `self.index.describe_index_stats()`.
2. **`pinecone_store.py`** — `filter_by_metadata()` probed with an all-zero dummy query vector, which Pinecone rejects for cosine-metric indexes (the library's own default when creating an index). Switched to a unit vector so it works regardless of metric.
3. **`pgvector_store.py`** — the list-filter branch formatted booleans with `str(v)` (`'True'`/`'False'`), never matching Postgres JSONB's lowercase `'true'`/`'false'` — even though the scalar-filter branch three lines below already had the correct lowercase handling (with a comment explaining why). Applied the same fix to the list branch.
4. **`pgvector_store.py` + `sqlite_vec_store.py`** — list-valued metadata fields (e.g. `{"tags": ["python", "js"]}`) could never match a list filter on either SQL backend, because both extracted the whole array as its JSON/text representation instead of matching individual elements — silently diverging from the in-memory backend's set-intersection semantics this PR was meant to preserve. Postgres now uses the `?|` "any array element" operator alongside the existing scalar `= ANY(...)` path; SQLite now walks `json_each()` over a `json_type`-guarded array/scalar wrapper. Verified live against a real SQLite connection and the real `sqlite-vec` extension.
5. **`faiss_store.py`** — `filter_by_metadata(limit=0)` returned one result instead of zero (limit check ran after appending the match).
6. **`milvus_store.py`** — `NaN`/`Infinity` filter values rendered as bare unquoted tokens in the expression string; the server-side rejection was then swallowed by a broad `except`, indistinguishable from "no matches." Now rejected up front with a clear `ValidationError`.

## Merge-induced test conflict

Merging `main` pulled in a test from an unrelated, independently-merged PR (#848) that pinned FAISS's *old* behavior of raising `NotImplementedError` for `filter_by_metadata` — which directly contradicts this PR's purpose. Updated `test_filter_by_metadata_faiss_raises_not_implemented` → `test_filter_by_metadata_faiss_returns_real_results` to assert the new, correct behavior, and updated the class docstring to document why.

## Testing

Added regression tests for each fix (in `test_backend_metadata_filtering.py` and a new `TestSQLiteVecStoreFilterByMetadata` class in `test_sqlite_vec_store.py`, run against the real `sqlite-vec` extension). Full `tests/vector_store/` suite: **247 passed, 36 skipped** (pre-existing, unrelated to this PR), 0 failed.

## CHANGELOG

Added an entry under `## [Unreleased]` → `### Fixed` documenting the original fix plus all six issues above.

Thanks again for the thorough work here @TaherTadpatri, and for the sharp catches in the first review round @Sameer6305 — this should be ready to merge now.

KaifAhmad1 (approved):
## Final review — Approve

Reviewed the full diff, the prior review rounds from @Sameer6305, and did an independent pass across all 7 backend implementations (correctness, cross-file call tracing, and diff-level scan with live verification). Findings and fixes are detailed in the [earlier comment](https://github.com/semantica-agi/semantica/pull/857#issuecomment-5263585383); pushed as commit cab995dc.

### What's covered now
- **Core fix**: `VectorStore._filter_by_metadata()` no longer assumes `self.metadata`, delegating to a native `filter_by_metadata()` on each of the 7 persistent backends (FAISS, Qdrant, Pinecone, Milvus, PgVector, SQLiteVec, Weaviate).
- **Round 1 fixes** (@Sameer6305, commit 7010913): Qdrant missing `Range`/`MatchAny` imports, PgVector scalar boolean formatting.
- **Round 2 fixes** (this pass): Pinecone `describe_index_stats()` called on the wrong object + zero-vector rejection on cosine indexes, PgVector boolean-*list* formatting (same bug as the scalar case, missed in the first fix), list-valued metadata matching on PgVector/SQLite (was comparing whole-array JSON text instead of matching elements), FAISS `limit=0` off-by-one, Milvus NaN/Infinity silently swallowed into a false "no matches."
- **Merge conflict** with `main` resolved (non-overlapping test additions), and a test conflict from an unrelated, independently-merged PR (#848, pinned FAISS's old `NotImplementedError` behavior) updated to match this PR's intended behavior.

### Verification
- Full `tests/vector_store/` suite: **247 passed, 36 skipped** (pre-existing, unrelated), 0 failed.
- SQLite list-intersection fix verified against the real `sqlite-vec` extension, not just mocks.
- PR is `MERGEABLE` against current `main`.

Nice work getting this across the line, @TaherTadpatri — thanks for the persistence through two review rounds. And thanks for the sharp initial catches, @Sameer6305.

Approving for merge.

PR #862, opened 2026-08-09
fix(pipeline): resolve broken import and missing `run()` in `PipelineWithProvenance`

### LINKED ISSUE #858: bug(pipeline): PipelineWithProvenance raises ModuleNotFoundError on instantiation — imports non-existent semantica.pipeline.pipeline
## Summary

`PipelineWithProvenance` is completely broken — it cannot be instantiated at all. The class does `from .pipeline import Pipeline` in its `__init__`, but `semantica/pipeline/pipeline.py` does not exist. `Pipeline` lives in `semantica/pipeline/pipeline_builder.py` and is re-exported from the package `__init__`. A second compounding issue: `Pipeline` is a pure dataclass with no `run()` method; `PipelineWithProvenance.run()` calls `self._pipeline.run(data, **kwargs)` which would immediately raise `AttributeError` even if the import were fixed.

This was explicitly flagged as "not fixed" in the `[Unreleased]` section of CHANGELOG.md, in the description of #825:

> `semantica/pipeline/pipeline_provenance.py` imports a nonexistent module and wraps a `Pipeline` dataclass with no `run()` method, so `PipelineWithProvenance` has never worked

No tracking issue was opened at that time. This is that issue.

## Reproduction

```python
from semantica.pipeline.pipeline_provenance import PipelineWithProvenance
```

**Traceback (Python 3.12, semantica 0.6.0):**

File "<string>", line 1, in <module>
File "semantica/pipeline/pipeline_provenance.py", line 35, in init
from .pipeline import Pipeline
ModuleNotFoundError: No module named 'semantica.pipeline.pipeline'

Also confirmed via `python -m pytest tests/pipeline/ -q` — no test exercises `PipelineWithProvenance` directly, so this has gone undetected.

## Root Cause

Two bugs:

1. **Wrong import path.** `pipeline_provenance.py` line 35: `from .pipeline import Pipeline`. There is no `pipeline.py` in `semantica/pipeline/`. `Pipeline` is defined in `pipeline_builder.py` and re-exported from the package `__init__` — the correct relative import is `from .pipeline_builder import Pipeline`.

2. **`Pipeline` has no `run()` method.** `Pipeline` is a `@dataclass` with fields `name`, `steps`, `config`, `metadata`. Pipeline execution goes through `ExecutionEngine.execute(pipeline)` (see `execution_engine.py` line 47: `from .pipeline_builder import Pipeline, PipelineStep, StepStatus`). `PipelineWithProvenance.run()` calling `self._pipeline.run(data)` would raise `AttributeError: 'Pipeline' object has no attribute 'run'` even after the import is fixed.

## Suggested Fix

Two changes to `pipeline_provenance.py`:

1. Replace `from .pipeline import Pipeline` → `from .pipeline_builder import Pipeline` (or `from semantica.pipeline import Pipeline` using the public re-export).

2. Replace the `self._pipeline.run(data, **kwargs)` call in `PipelineWithProvenance.run()` with the actual execution pattern: instantiate `ExecutionEngine` and call `ExecutionEngine().execute(self._pipeline)`. The existing `run()` body already has the right provenance-tracking scaffolding around it — only the execution call needs to change.

There is no need to create a new `pipeline.py` file; `Pipeline` already has a proper home. This should be a small, contained fix.

## Additional Context

- `semantica/pipeline/` files: `execution_engine.py`, `pipeline_builder.py`, `pipeline_provenance.py`, `pipeline_templates.py`, `pipeline_validator.py`, `parallelism_manager.py`, `failure_handler.py`, `resource_scheduler.py` — no `pipeline.py`.
- `semantica/pipeline/__init__.py` exports both `Pipeline` (from `pipeline_builder`) and `ExecutionEngine` (from `execution_engine`).
- Last touched: commit `db4361ad` (PR #825, Aug 4 2026). No subsequent PR has addressed it.

Karunasagar12:
## Summary

Fixes #858 — `PipelineWithProvenance` cannot be instantiated due to two compounding bugs.

## Changes

**Bug 1 — Wrong import path:**
`from .pipeline import Pipeline` fails because `semantica/pipeline/pipeline.py` does not exist. `Pipeline` is defined in `pipeline_builder.py`. Fixed to `from .pipeline_builder import Pipeline`.

**Bug 2 — `Pipeline` has no `run()` method:**
`Pipeline` is a `@dataclass` with fields (`name`, `steps`, `config`, `metadata`). Execution goes through `ExecutionEngine.execute_pipeline()`. Rewired `PipelineWithProvenance.run()` to delegate there.

**Additional improvements:**
- Constructor now takes a built `Pipeline` instance (the previous signature tried to pass `**config` to a dataclass, which was never functional)
- Replaced deprecated `datetime.utcnow()` with `datetime.now(timezone.utc)` (Python 3.12+)

## Tests

Added `tests/pipeline/test_pipeline_provenance.py` — 6 tests covering import, instantiation, execution, attribute delegation, provenance degradation, and data passthrough. All pass.

```
======================== 6 passed, 3 warnings in 0.15s =========================
```


KaifAhmad1:
@Karunasagar12 Could you please address the Qodo findings before we merge this PR? Thanks!

Karunasagar12:
Addressed both Qodo findings:

1. **Removed redundant `test_import_succeeds`** — the module-level import already guards against the import regression at pytest collection time, making the in-test import redundant.

2. **Fixed `test_provenance_disabled_when_import_fails`** — now deterministically simulates `ImportError` via `patch.dict(sys.modules)` and asserts that `runner.provenance is False` and `runner._prov_manager is None`. Previously this test wasn't actually exercising the fallback branch.

All 5 tests pass cleanly.

Sameer6305:
@Karunasagar12 Could you please leave a comment on issue #858? I’ll assign it to you after that. Thanks for picking this up!

Sameer6305 (approved):
Thanks @Karunasagar12 for the work on #862, which addresses #858 and fixes the broken `PipelineWithProvenance` implementation.

The PR resolves the two core issues that prevented `PipelineWithProvenance` from functioning correctly:

- Corrects the broken `Pipeline` import to use `pipeline_builder`.
- Updates `PipelineWithProvenance.run()` to execute the built pipeline through the current `ExecutionEngine.execute_pipeline()` API.
- Updates the constructor to explicitly accept a built `Pipeline` instance.
- Replaces the deprecated `datetime.utcnow()` usage with timezone-aware UTC timestamps.

### Review

**PR owner:** @Karunasagar12  
**Reviewers:** @Sameer6305 and @KaifAhmad1

Overall, the implementation was solid and well focused on the underlying pipeline/provenance issues. In particular, the execution path through `ExecutionEngine`, provenance handling, and the new pipeline-based API were verified against the current project implementation rather than relying only on the PR description.

During the review, we found that the implementation changes left several existing provenance tests and one documentation example using the previous constructor form:

```python
PipelineWithProvenance(provenance=True)
````

Since the updated implementation requires a built `Pipeline`, those first-party callers needed to be brought in line with the corrected API. We also identified a few latent test paths that would have failed for the same reason.

These review fixes were applied in:

**`<6aa1a3c>` — `fix(pipeline): update provenance callers for Pipeline API`**

The changes:

* Updated the affected provenance tests to construct a valid `Pipeline`.
* Added dedicated handling for `PipelineWithProvenance` in the generic provenance-wrapper tests.
* Updated the provenance usage documentation.
* Added the corresponding `#858` changelog entry.
* Did not modify the core `PipelineWithProvenance` implementation unnecessarily.
* Did not introduce compatibility hacks or unrelated changes.

### Verification

The corrected implementation and review changes were verified against the current codebase:

* `tests/pipeline/test_pipeline_provenance.py` — **5 passed**
* Full provenance test suite — **332 passed, 30 skipped, 0 failed**
* `PipelineWithProvenance` import — verified successfully
* `ExecutionEngine.execute_pipeline()` — verified against the current implementation and signature
* Repository-wide `PipelineWithProvenance(...)` usage — checked; no remaining old constructor calls were found
* No unrelated vector-store changes were introduced by the review fixes.

With these changes in place, the PR is now consistent with the current pipeline/provenance APIs and the affected test and documentation paths are aligned with the implementation.

Thanks again @Karunasagar12 for taking up the issue and putting together the initial fix. The core investigation and implementation were on the right track, and the follow-up review changes have brought the PR into a clean, consistent state.

Karunasagar12:
Thanks @Sameer6305 and @KaifAhmad1 for the thorough review and for catching the stale callers — I should've grepped for existing usages of the old constructor before submitting. Good lesson for next time.

Appreciate you applying the follow-up fixes rather than bouncing it back. The changelog entry and updated docs look good. Happy to help with more issues in the repo!

Sameer6305:
> Thanks @Sameer6305 and @KaifAhmad1 for the thorough review and for catching the stale callers — I should've grepped for existing usages of the old constructor before submitting. Good lesson for next time.
> 
> Appreciate you applying the follow-up fixes rather than bouncing it back. The changelog entry and updated docs look good. Happy to help with more issues in the repo!

Thanks! Glad the feedback was helpful, and welcome to pick up more issues in the repo. 🙌

Also, could you comment on the issue this PR solves, so I can assign it to you? Once you do, I’ll assign the issue and merge this PR, then close the issue.

PR #871, opened 2026-08-09
Bump fastapi minimum version to fix PYSEC-2024-38 (starlette DoS)

### LINKED ISSUE #869: security(deps): FastAPI minimum version floor allows PYSEC-2024-38 (HTTP response splitting) — bump to >=0.109.1
### Summary

`pyproject.toml` declares `fastapi>=0.100.0` as an optional dependency in the `[explorer]` extra. This floor allows FastAPI 0.109.0, which is affected by **PYSEC-2024-38** (HTTP response splitting). The fix is in 0.109.1 — a one-patch-version bump that has been available since early 2024. The installed version in this environment is 0.109.0.

---

### Reproduction

**`pip-audit` output (real, from this environment):**

```text
fastapi  0.109.0  PYSEC-2024-38  0.109.1
````

**Confirm installed version:**

```text
python -c "import fastapi; print(fastapi.__version__)"
# 0.109.0
```

**Confirm declared floor allows the vulnerable version:**

```text
# pyproject.toml [project.optional-dependencies]
explorer = [
  "fastapi>=0.100.0",   # allows 0.109.0 — vulnerable
  ...
]
```

---

### Root Cause

HTTP response splitting (PYSEC-2024-38) occurs when attacker-controlled data is included in HTTP response headers without sanitization, allowing an attacker to inject additional headers or split the response into two, potentially enabling cache poisoning, cookie injection, or XSS. FastAPI 0.109.1 patched the header-handling code that made this possible.

The `pyproject.toml` floor of `>=0.100.0` is too permissive — it allows any 0.1xx version including the affected 0.109.0. Any environment that installs `semantica[explorer]` without pinning FastAPI to a specific version may resolve to 0.109.0 if that is the highest available version satisfying the constraint, particularly in environments with older dependency caches or locked requirements files.

---

### Suggested Fix

Raise the minimum version floor in `pyproject.toml`:

```text
# pyproject.toml [project.optional-dependencies]
# Change:
explorer = [
  "fastapi>=0.100.0",
  ...
]
# To:
explorer = [
  "fastapi>=0.109.1",
  ...
]
```

This is a one-line change with no functional impact — the FastAPI 0.109.x API is fully stable and backward-compatible.

As a secondary step, consider adding `fastapi` to any CI dependency audit step so future CVEs in pinned or floor versions are caught before release rather than after.

agu2347:
Closes #869.

The `explorer` optional-dependencies extra pins `fastapi>=0.100.0`, which allows resolution of a `starlette` version affected by PYSEC-2024-38 (multipart form parsing DoS — unbounded memory use processing a malformed/large multipart request).

This bumps the floor to `fastapi>=0.109.1`, the first fastapi release that requires a patched starlette. Single-line change to `pyproject.toml`, no code changes needed.


Sameer6305:
@agu2347 Could you please leave a comment on issue #869 ? I’ll assign it to you after that. Thanks for picking this up!

Sameer6305 (dismissed):
Approved.

Sameer6305:
Thanks @agu2347 for the PR!

@KaifAhmad1, I made a small correction during review: `fastapi>=0.109.1` still allowed the vulnerable Starlette range, so we bumped it to `>=0.109.2` and also raised `python-multipart` to `>=0.0.7` to fully address PYSEC-2024-38.
**241 Explorer tests pass**, and the changes are ready for final review.

KaifAhmad1 (approved):
## Final review — approved ✅

@Sameer6305 @agu2347 — went through the full history on this one (issue → original commit → correction → my two follow-ups) and independently re-verified the security claims from scratch against OSV, PyPI metadata, and the actual upstream source diffs rather than trusting any of the write-ups (including my own) at face value. Posting the trail since it explains *why* this took three rounds to actually close.

### What I verified
- **PYSEC-2024-38 / CVE-2024-24762 / GHSA-2jv5-9r88-3w3p** is a ReDoS in `python-multipart`'s `Content-Type` header parser (`parse_options_header`), fixed in `python-multipart` 0.0.7. It is *not* HTTP response splitting, as #869 originally described.
- The **original commit** (`10f7640`, `fastapi>=0.109.1` only) didn't actually close the hole: `python-multipart` is declared as its own direct dependency in the `explorer` extra, not pulled in via `fastapi[all]`, so bare `fastapi` enforces no floor on it at all — `python-multipart==0.0.6` could still resolve. Good catch flagging that this needed more work.
- @Sameer6305's fix (`416e4e4`, `fastapi>=0.109.2` / `python-multipart>=0.0.7`) lands on the right final numbers. I traced the stated mechanism a bit further for the record: I diffed `starlette/formparsers.py` between 0.35.1 and 0.36.2 and confirmed Starlette never carried the vulnerable code itself (it always delegates to the external `python-multipart` package in both versions) — so the fix is entirely carried by the `python-multipart>=0.0.7` bump; the `fastapi>=0.109.2` bump is good hygiene but not load-bearing for this specific CVE. Doesn't change the outcome, just the "why" for anyone reading this later.

### Two follow-up commits I pushed to this branch
- **`9b86abd7`** — the repo's `pip-audit` job (`security.yml`) ran weekly, `continue-on-error: true`, against a bare env with none of Semantica's extras installed — it would never have caught either version of this fix. Same blind spot in `security-scan.yml`'s Safety check (`[llm-litellm]` only). Added a `pull_request` trigger scoped to `pyproject.toml` changes that installs `semantica[all]` and actually gates on findings.
- **`ec30f1a8`** — that new gate caught a real pre-existing issue on its very first run: `setuptools==79.0.1` (CVE-2026-59890, fixed in 83.0.0) resolving under the too-permissive `[build-system] requires = ["setuptools>=61.0", ...]`. Same pattern as this entire PR. Fixed the floor and hardened the workflow to upgrade `pip`/`setuptools` explicitly before auditing, since `[build-system] requires` doesn't govern the ambient env the scan runs in.
- Also added a `### Security` entry to `CHANGELOG.md` covering the full arc, including the correction — with a verified `241 passed` from actually running the `explorer` suite locally, not the disputed 239/241 figures floating around in the commit messages.

All 10 checks are green on the latest commit (`ec30f1a8`), including both new gates. Approving.

PR #873, opened 2026-08-09
test(seed): use a real file for CSV loading

### LINKED ISSUE #861: test(seed): test_load_from_csv uses mock_open which doesn't support seek() — test fails with 0 records despite correct implementation
## Summary

`tests/seed/test_seed_manager.py::TestSeedDataManager::test_load_from_csv` fails with `AssertionError: 0 != 2` — it expects 2 records loaded from a mocked CSV but gets none. `SeedDataManager.load_from_csv()` is correct and works against real files; the test fixture is incompatible with the implementation's use of `csv.Sniffer`.

## Reproduction

```bash
pytest tests/seed/test_seed_manager.py::TestSeedDataManager::test_load_from_csv -v
```

AssertionError: 0 != 2

## Root Cause

`load_from_csv()` uses `csv.Sniffer` to auto-detect the CSV delimiter before reading rows:

```python
with open(file_path, "r", encoding="utf-8") as f:
    sample = f.read(1024)   # reads the entire mock content
    f.seek(0)               # no-op on mock_open — position is NOT reset
    dialect = csv.Sniffer().sniff(sample)
    delimiter = dialect.delimiter
    reader = csv.DictReader(f, delimiter=delimiter)
    for row in reader:      # iterator sees empty file — mock_open exhausted
        records.append(...)
```

The test patches `builtins.open` with `mock_open(read_data="name,age\nAlice,30\nBob,25")`. `unittest.mock.mock_open` implements `read()` and individual `readline()` calls, but `f.seek(0)` is a no-op — the internal read position is not reset. After `f.read(1024)` consumes the mock data, `csv.DictReader`'s iterator calls `readline()` and immediately gets `""`, producing zero rows.

Additionally, when `csv.Sniffer().sniff()` raises `csv.Error` on the mock data (which it may, depending on the content), the `except csv.Error` branch also calls `f.seek(0)` (still a no-op) before setting `delimiter = ","` — so even the fallback path returns 0 rows.

The implementation itself is correct — `test_load_from_csv_file_not_found` and `test_load_from_json` in the same file work fine because they don't rely on file position state.

## Suggested Fix

Replace the `mock_open` + `@patch("builtins.open")` + `@patch("pathlib.Path.exists")` combination with a real temporary file using pytest's `tmp_path` fixture:

```python
def test_load_from_csv(self, tmp_path):
    csv_file = tmp_path / "test.csv"
    csv_file.write_text("name,age\nAlice,30\nBob,25", encoding="utf-8")

    records = self.manager.load_from_csv(
        csv_file, entity_type="Person", source_name="test_source"
    )

    self.assertEqual(len(records), 2)
    self.assertEqual(records[0]["name"], "Alice")
    self.assertEqual(records[0]["age"], "30")
    self.assertEqual(records[0]["entity_type"], "Person")
    self.assertEqual(records[0]["source"], "test_source")
```

This exercises the real `csv.Sniffer`, `csv.DictReader`, and file path logic end-to-end without any mock, which is a better test of the actual behavior. The class is a `unittest.TestCase`, so `tmp_path` needs to be received either by converting to pytest-style or using `tempfile.TemporaryDirectory` in `setUp`/`tearDown` instead.

No changes to production code are needed.

nightcityblade:
## Description

Replace the seek-incompatible mocked CSV stream in `test_load_from_csv` with a real temporary CSV file. This exercises the same path as production and restores the test's expected two records.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #861
Fixes #861

## Changes Made

- Create the CSV fixture inside a temporary directory.
- Load the real temporary file through `SeedDataManager`.
- Remove the mocked file assertions that prevented seeking.

## Testing

- [x] Tested locally
- [ ] Added tests for new functionality (N/A — this repairs an existing test)

### Test Commands

```bash
python3 -m pytest tests/seed/test_seed_manager.py -q
python3 -m isort --check-only tests/seed/test_seed_manager.py
python3 -m flake8 --select F401,F811,F821,E999 tests/seed/test_seed_manager.py
uv run --no-project --with build python -m build
```

## Documentation

- [ ] Updated relevant documentation (N/A)
- [ ] Added code examples if applicable (N/A)
- [ ] Updated API reference if adding new APIs (N/A)
- [ ] Updated cookbook if adding new examples (N/A)

## Breaking Changes

**Breaking Changes**: No

N/A

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas (N/A — the test setup is self-explanatory)

## Additional Notes

The test file's unrelated pre-existing formatting findings are outside the scope of this focused fix; the changed imports pass isort and critical flake8 checks.


Sameer6305:
reviewing the PR.

Sameer6305 (approved):
Thanks @nightcityblade for the clean and focused fix! 👍

I verified the original mock-based test reproduces the issue, and the new temporary-file approach passes the full seed test and correctly exercises the production CSV loading path.

@KaifAhmad1 I’ve completed my review and **approved it from my side**. Could you please give it a final review as well?

KaifAhmad1 (approved):
Verified independently: checked out the branch into a local worktree and ran `pytest tests/seed/test_seed_manager.py -v` — all 6 tests pass, including `test_load_from_csv` which previously failed with `AssertionError: 0 != 2`.

Root cause and fix both check out:
- `load_from_csv()` calls `f.read(1024)` then `f.seek(0)` before running `csv.Sniffer` + `csv.DictReader`. `unittest.mock.mock_open`'s `seek()` is a no-op, so the old test's `DictReader` iterated over an already-exhausted stream and got 0 rows even though production code is correct.
- The new test writes a real file via `tempfile.TemporaryDirectory()` and exercises the actual `Sniffer`/`DictReader`/file-path path end-to-end.
- Confirmed the assertions sitting outside the `with tempfile.TemporaryDirectory()` block is safe: `load_from_csv()` fully materializes results into a `List[Dict]` inside its own `with open(...)` block before returning, so nothing depends on the temp file after the block exits.

Minor unrelated import cleanup (isort ordering, unused `MagicMock`) is bundled in but harmless.

LGTM — approving for merge.

PR #879, opened 2026-08-10
fix(export): log DistanceExporter metric computation failures instead of swallowing them

### LINKED ISSUE #874: bug(export): DistanceExporter silently swallows metric computation errors, producing None values in exported data that are indistinguishable from legitimate "no path" results
## Summary

## Reproduction

```python

# Line 72–74: _betweenness()
try:
except Exception:

# Line 82–84: _hop_distance()
try:
except Exception:

try:
    if isinstance(result, dict):
    return None
except Exception:

try:
except Exception:
```

```python
    ...

```

## Root Cause

## Suggested Fix

Two complementary changes:

```python
    try:
        ...
    except Exception as exc:
        logger.debug(
        )
        return None
```

```python
```

AmirF194:
## What's wrong

`DistanceExporter`'s four private metric helpers (`_betweenness`, `_hop_distance`, `_weighted_distance`, `_semantic_similarity`) each wrap their computation in a bare `except Exception: return None`/`{}` with no logging. An exported `None` is therefore indistinguishable from a legitimate "no path exists between these nodes" result, so a consumer of the exported CSV/JSONL/DataFrame data can't tell "computation raised and was silently discarded" from "there really is no path here."

## Fix

Log each caught exception at warning level (with the source/target and full traceback) before returning the existing sentinel. The exported row shape and values are unchanged; only the observability of the failure changes. This is the minimal fix from the two options discussed in the issue (log-only vs. a schema-changing per-metric error column); the schema change is a separate, larger design decision better left for its own discussion.

## Verification

- New regression test (`tests/export/test_distance_exporter.py`) reproduces the bug: with the four helpers monkeypatched to raise, the branch logs a warning per failure while `main` logs nothing. Confirmed both directions (fails on `main`, passes on this branch).
- Also asserts the exported row shape/values are unchanged, and that the pre-existing "no KG backend available" path (a legitimate `None`, not an exception) still logs nothing.
- Ran the full `tests/export/` suite plus the other export-adjacent suites (`test_export_module.py`, `test_arango_aql_exporter.py`, `test_401_temporal_provenance_export.py`, `test_parquet_exporter.py`, `test_export_methods_wrapper.py`, `test_arrow_exporter.py`, `test_notebook_15_export.py`): all pass.
- Did not run the full repository test suite; the rest of it requires heavier optional dependencies (torch/transformers/sklearn/etc.) not installed in this verification environment.

Fixes #874


KaifAhmad1:
@AmirF194 Please fix the Qodo findings first, especially the logger mismatch and misleading warning messages. Once addressed, we can review and proceed with the merge.

AmirF194:
Pushed a fix for both. `get_logger(__name__)` was doubling the `semantica.` prefix (`__name__` is already `semantica.export.distance_exporter`), so the warnings landed on `semantica.semantica.export.distance_exporter`, a branch `setup_logging()` never configures. Changed it to `get_logger("export.distance_exporter")`, confirmed in a clean container that `logger.name` now resolves to `semantica.export.distance_exporter`. Also rewrote the three exception-path messages (hop/weighted/semantic) to say "returning None sentinel" instead of "recording as no path", since a raised exception isn't the same claim as a missing path.

Left the other two Qodo findings alone on purpose: adding node context to the betweenness warning doesn't fit cleanly, that computation runs over the whole graph rather than a single src/tgt pair, so there's no pair to name. And the per-metric error/status export field is the schema-changing alternative Qodo itself flagged as a separate design call, not the log-only fix this PR and the issue scoped. Happy to open a follow-up for that if you want it.

Full `tests/export/` plus the other export suites still green (175 passed) after the change.

Sameer6305:
Hi @AmirF194 , if you're working on an issue, please leave a comment on the issue mentioning that you're handling it. This will help us keep track of engaged issues and assign them accordingly. Thanks!

AmirF194:
#890 opened a few hours after this one, same issue, same four call sites. One difference worth flagging: it logs at debug level, this PR logs at warning with exc_info, which matches what #874 actually asks for (the exception was invisible by default before either fix; debug-level stays invisible at any commonly deployed log level). Happy to stand down if you'd rather go with #890.

KaifAhmad1 (approved):
Thanks for the thorough fix and follow-through here, @AmirF194 — the logger double-prefix catch in particular was a good find, and the reasoning for leaving the betweenness context/error-column suggestions for a follow-up was sound.

Verified independently:
- `get_logger("export.distance_exporter")` now resolves to `semantica.export.distance_exporter`, matching every other exporter and fixing a pre-existing silent `logger.debug` call as a bonus.
- `tests/export/test_distance_exporter.py` (6 tests) and the full `tests/export/` suite (71 tests) pass on this branch.
- Exported row shape/sentinel values are unchanged; only observability improved, as intended.

Added a changelog entry crediting you in a follow-up commit. Approving.

PR #881, opened 2026-08-10
test(normalize): make optional tests deterministic

### LINKED ISSUE #860: test(normalize): three test correctness bugs — missing skip guard, narrow chardet allowlist, and UTC/local time mismatch
## Summary

Three tests in `tests/normalize/` fail for reasons unrelated to the normalize module implementations, which are correct. Each root cause is a different test-authoring mistake: a missing optional-dependency skip guard, an incomplete heuristic-library allowlist, and a timezone handling error that causes the test to fail depending on what time of day (or in which timezone) it runs. All three are test fixes only; no production code changes are needed.

## Bug 1 — `test_detect_language` / `test_detect_with_confidence` (missing skip guard) — Priority: Normal

**Symptom:**

AssertionError: 'en' != 'fr'
WARNING semantica.language_detector: langdetect library not available, language detection will be limited

**Root cause:** `langdetect` is an optional dependency — it is not declared in `pyproject.toml` anywhere (not in `dependencies`, not in any `[project.optional-dependencies]` extra). `LanguageDetector` correctly guards this with `LANGDETECT_AVAILABLE = False` when absent, logs a warning at init, and falls back to returning `self.default_language` (`"en"`) for all calls. The fallback behavior is documented and intentional.

The two failing tests assert that French/German text is detected correctly and that confidence exceeds 0.5, with no `skipIf` guard — they will always fail in any environment where `langdetect` is not installed, including any fresh CI run that doesn't install optional extras.

**Fix direction:** Add `@pytest.mark.skipif(not LANGDETECT_AVAILABLE, reason="langdetect not installed")` (or `pytest.importorskip("langdetect")` at the top of the test class) to `TestLanguageDetector.test_detect_language` and `TestLanguageDetector.test_detect_with_confidence`. The `test_detect_short_text` and `test_get_language_name` tests do not require the library and should remain ungated.

## Bug 2 — `test_detect_encoding` / `test_convert_to_utf8` (chardet allowlist too narrow) — Priority: Normal

**Symptom:**

AssertionError: 'windows-1253' not found in ['iso-8859-1', 'windows-1252', 'latin-1']
AssertionError: 'Hιllς Wφrld' != 'Héllò Wörld'

**Root cause:** `chardet.detect()` is a statistical heuristic. For the specific byte sequence produced by `"Héllò Wörld".encode("latin-1")` in this environment, chardet returned `windows-1253` (the Greek Windows codepage). The test's allowlist is `["iso-8859-1", "windows-1252", "latin-1"]` — the comment in the test acknowledges the heuristic nature but the list is incomplete.

`test_convert_to_utf8` then fails downstream: `convert_to_utf8()` decodes the bytes using the chardet-detected encoding. Decoding Latin-1–encoded bytes as `windows-1253` maps the bytes to Greek characters rather than the original Latin characters — the implementation's `errors="replace"` strategy faithfully executes the wrong codec.

Neither failure indicates a bug in `EncodingHandler`. Encoding detection on short byte sequences is inherently ambiguous.

**Fix direction:** Replace the short test string with a longer input that produces an unambiguous chardet detection, or use an explicit `source_encoding="latin-1"` parameter in `test_convert_to_utf8` to bypass auto-detection. The allowlist in `test_detect_encoding` should also be broadened to cover `windows-1253` and other plausible single-byte European codepage returns, or the test should simply check that the detected encoding can successfully round-trip decode the original bytes rather than asserting a specific name.

## Bug 3 — `test_normalize_date_relative` (UTC/local time mismatch) — Priority: **High** (causes intermittent CI failures depending on runner timezone and time of day)

**Symptom:**

AssertionError: '2026-08-09' != '2026-08-08'

**Root cause:** The test and the implementation use different clocks:

```python
# Test — uses UTC
today = datetime.now(timezone.utc).date().isoformat()
self.assertEqual(self.normalizer.normalize_date("today", format="date"), today)
```

`RelativeDateProcessor.process_relative_expression("today")` calls `datetime.now()` (no timezone — wall-clock local time) and adds `timedelta(days=0)`. `normalize_date` then calls `convert_to_utc()` on this naive datetime, which attaches UTC and returns it. The result is UTC-normalized local time.

In any timezone east of UTC (UTC+1 through UTC+14), the local date is already tomorrow relative to UTC for some portion of the day. The failure `'2026-08-09' != '2026-08-08'` is exactly this: local time had rolled to Aug 9 while UTC was still Aug 8. This test will fail intermittently on any CI runner not running in UTC, and will reliably fail for contributors in UTC+5:30 (IST), UTC+8, UTC+9, UTC+10, etc. for a significant portion of each day.

**Fix direction:** Either mock `datetime.now()` in the test to a fixed reference time (preferred — makes the test deterministic), or compare the normalized output against `datetime.now().date().isoformat()` (local time) to match what the implementation actually uses, rather than `datetime.now(timezone.utc).date().isoformat()`.

## Verification

All three failing tests pass when run individually on a machine with `langdetect` installed and a fixed UTC timezone. No changes to production code are needed.

aoright:
## Description

Makes the normalize tests deterministic across optional dependency sets, chardet versions, and time zones.

## Type of Change

## Related Issues

Closes #860

## Changes Made

- Skip only the language detection tests that require the optional `langdetect` package.
- Use a longer Latin-1 corpus and validate decode compatibility instead of a narrow detector label allowlist.
- Pass the known source encoding to the conversion test.
- Freeze the date normalizer clock so the relative-date assertion cannot cross UTC/local date boundaries.

## Testing

- [x] Tested locally

### Test Commands

```bash
pytest tests/normalize
black --check tests/normalize/test_{language_detector,encoding_handler,date_normalizer}.py
isort --check-only tests/normalize/test_{language_detector,encoding_handler,date_normalizer}.py
flake8 --max-line-length=88 tests/normalize/test_{language_detector,encoding_handler,date_normalizer}.py
```

Results:
- Full normalize suite with `langdetect`: 79 passed.
- Targeted suite without `langdetect`: 12 passed, 2 skipped.
- Targeted suite with `langdetect`: 14 passed.
- Targeted suite also passes with chardet 5.2.0.

## Documentation

## Breaking Changes

**Breaking Changes**: No

## Checklist

- [x] My code follows the project style guidelines

## Additional Notes

This PR changes tests only; runtime behavior is unchanged.


Sameer6305:
Hi @aoright , if you're working on an issue, please leave a comment on the issue mentioning that you're handling it. This will help us keep track of engaged issues and assign them accordingly. Thanks!

Sameer6305 (dismissed):
Verified all three fixes against #860 — langdetect skip guard, chardet
roundtrip check instead of a narrow allowlist, and a genuinely frozen
clock for the relative-date test (confirmed it returns the same fixed
datetime regardless of host timezone, not just "less likely to fail").

Ran the full suite: 77 passed, 2 skipped (langdetect not installed here)
locally, no regressions elsewhere. Test-only, exactly as described.
Nice clean first PR, approving.

Sameer6305:
Confirmed clean after the rebase onto upstream/main — exactly 3 files, fix content unchanged and verified a third time, 77 passed / 2 skipped, all required checks green. Approving.

KaifAhmad1 (approved):
Final review — verified independently in an isolated worktree, not just re-reading the diff.

**What I checked:**
- Checked out the branch and ran `pytest tests/normalize` (no `langdetect` installed here): **77 passed, 2 skipped** — matches both the author's and @Sameer6305's numbers exactly.
- `black --check`, `isort --check-only`, `flake8 --max-line-length=88` on all three changed files: clean.
- Confirmed `LANGDETECT_AVAILABLE` is a real, importable module-level flag in `language_detector.py`, and that the new `@unittest.skipUnless` guards land on exactly the two tests that need `langdetect` (`test_detect_short_text`/`test_get_language_name` correctly stay ungated).
- Confirmed `test_detect_encoding`'s new round-trip-decode assertion (`latin1_bytes.decode(encoding) == text`) is correct regardless of which compatible codepage chardet picks, and that `convert_to_utf8(..., source_encoding="latin-1")` is a real parameter on `EncodingHandler`.
- Traced the `test_normalize_date_relative` fix end-to-end against `date_normalizer.py`: patching `semantica.normalize.date_normalizer.datetime` correctly intercepts the module's one `datetime.now()` call. Verified separately that `dateutil.parser.parse("today")` raises `ParserError`, which is what routes `"today"` into `RelativeDateProcessor` rather than being parsed as a real date — so the mock is actually exercised, not bypassed. Followed the value through `convert_to_utc()` to the final `"2024-01-02"` assertion; it's correct.
- Diff is test-only; no production code touched. Unused imports removed (`TemporalExpressionParser`, `date`, `os`) were genuinely unused.

All three root causes from #860 are fixed correctly and the fixes are more robust than point-patches (round-trip check instead of a longer allowlist, timezone-independent clock mock instead of a UTC/local comparison). CI is green and the branch is mergeable. Approving.

PR #882, opened 2026-08-10
fix(context): clarify get_node_property not-found contract (#877)

### LINKED ISSUE #877: design(context): ContextGraph.get_node_property's "not found" behavior is ambiguous and inconsistent with its siblings — needs a contract discussion before any fix
## Summary

Two confirmed behavioral inconsistencies in `semantica/context/context_graph.py` create an ambiguous API contract for the `get_node_property` / `get_node_attributes` / `get_edge_data` family. This issue documents the confirmed symptoms, explains why they matter, and proposes options for discussion — **it does not prescribe a specific fix**, because the most natural fix is a breaking change for any existing caller relying on today's behavior.

## Confirmed Symptoms

**Symptom A — `get_node_property` returns `None` for two distinct failure modes with no way to distinguish them.**

`semantica/context/context_graph.py`, lines 765–770:

```python
def get_node_property(self, node_id: str, property_name: str) -> Any:
    with self._lock:
        node = self.nodes.get(node_id)
        if not node:
            return None             # ← Case 1: node doesn't exist
        return node.properties.get(property_name)   # ← also returns None: Case 2: property missing
```

A caller receives `None` in both cases and has no way to distinguish them:

```python
# Both of these return None — identical output, different meanings:
g.get_node_property("nonexistent_node", "name")   # → None  (node missing)
g.get_node_property("n1", "nonexistent_property") # → None  (property missing on real node)
```

**Symptom B — sibling methods signal the same "node not found" condition differently.**

```python
g.get_node_property("nonexistent_node", "name")  # → None
g.get_node_attributes("nonexistent_node")          # → {}
g.find_node("nonexistent_node")                    # → None
g.get_edge_data("nonexistent_src", "tgt")          # → {}
```

Two of four related methods return `None` for "not found"; the other two return `{}`. A caller written against one method's contract that is later adapted to use a sibling may silently get the wrong behavior — `if result is None` works for `get_node_property` but never fires for `get_node_attributes` (which returns `{}` even for missing nodes, which is also falsy but is not `None`).

Both confirmed with real output from a live `ContextGraph` instance.

## Why It Matters

`ContextGraph` is the most-imported non-type class in the semantica codebase. The `get_node_property` / `get_node_attributes` family is part of its core read API. The ambiguity in Symptom A is the same class of bug that this project has been systematically fixing in other modules (silent success/failure indistinguishability). The inconsistency in Symptom B is a contributor-friction problem: you cannot guess the contract of a sibling method without reading its source.

## Why This Needs Discussion First

The most natural fix for Symptom A — raising `KeyError` when the node doesn't exist, matching the behavior of `dict[key]` — is a breaking API change. Any existing caller that does:

```python
prop = graph.get_node_property(node_id, "score")
if prop is None:
    prop = default_value
```

...expects `None` to cover both "node missing" and "property missing." Changing the missing-node case to raise `KeyError` would crash that caller.

Similarly, aligning `get_node_attributes` to return `None` on missing node (matching `get_node_property`) would break callers doing `attrs.get("key", fallback)` directly on the return value.

**This issue is not a bug report requesting a specific code change.** It is a design discussion about what the right contract should be and how to migrate to it.

## Options for Discussion

**Option 1 — Adopt `dict.get()`-style convention with an explicit `default=` parameter.**

Match the Python standard library's own pattern for "key not found":

```python
def get_node_property(self, node_id: str, property_name: str, default=None) -> Any:
    node = self.nodes.get(node_id)
    if node is None:
        return default       # explicit "node not found" path
    return node.properties.get(property_name, default)
```

This is non-breaking: existing callers get the same `None` they currently get. New callers can pass `default=_SENTINEL` to distinguish "not found" from `None`-valued properties. Pair with updating `get_node_attributes` to use the same `default={}` convention.

**Option 2 — Raise `NodeNotFoundError` (or `KeyError`) and add a separate `get_node_property_safe()` / `find_node_property()` with the current None-returning behavior.**

This is the cleanest API for new code but requires a deprecation path. The existing methods would be kept under their current names for at least one release, emitting a deprecation warning when they return `None` for a missing node.

**Option 3 — Document the current ambiguity explicitly without changing behavior.**

If changing the behavior is judged too disruptive given the number of callers, the minimum acceptable outcome is documenting the current contract clearly in the docstring:

```python
def get_node_property(self, node_id: str, property_name: str) -> Any:
    """Return the value of a property on a node.

    Returns None if the node does not exist OR if the property is not set.
    Callers that need to distinguish these cases should call find_node()
    first to check node existence.
    """
```

This is better than the current state (no docstring at all) and gives callers a documented contract to code against.

## Invitation

Maintainers: which contract would you like `get_node_property` to have going forward? The three options above cover the spectrum from non-breaking through fully breaking. If there's a preference, a follow-up PR scoped to just this method (and its siblings) would be straightforward to implement once the direction is agreed.

ZohaibHassan16:
Fixes #877

## What

`get_node_property` returned `None` for both "node missing" and "property missing" with no way to distinguish them. `get_node_attributes` returned `{}` for a missing node, inconsistent with its siblings.

## Changes

- Added `default=None` to `get_node_property` and `get_node_attributes` , same contract as `dict.get()`. Callers can pass a sentinel to distinguish failure modes.
- Fixed `add_node_attribute`: `mutation_callback` was de-indented outside `with self._lock`, firing the callback without holding the lock.
- 12 tests covering happy path, missing node, missing property, sentinel disambiguation, falsy-zero and callback behaviour.


KaifAhmad1:
@ZohaibHassan16 Please fix the Qodo findings and update the tests/docs accordingly. Once resolved, we can merge.

ZohaibHassan16:
Please review @KaifAhmad1

KaifAhmad1:
Pushed a follow-up commit (3fa507f1).

The PR description says the `add_node_attribute` fix moves `mutation_callback` back inside `with self._lock`, but the diff only removed a stray blank line — the call stayed outside the lock, unchanged.

Traced the history: `self._lock` is a `threading.RLock`, and `_add_internal_node`/`_add_internal_edge` deliberately release the lock before firing their mutation callbacks too (both wrapped in `try/except`). So firing outside the lock is this class's established, intentional pattern — not the bug. Re-indenting it inside the lock would've made this method the odd one out and risked holding the lock for the duration of an arbitrary user callback.

The actual gap: unlike those two siblings, `add_node_attribute`'s callback call wasn't wrapped in `try/except`, so a raising callback would propagate uncaught here but is caught-and-logged there. Fixed that instead, added a regression test for it, and updated the CHANGELOG entry to describe this accurately.

`pytest tests/context/test_context.py -q`: 27 passed. Good to merge from my side once CI is green.

KaifAhmad1 (approved):
Contract change (get_node_property/get_node_attributes default= param) is solid, backward-compatible, and well-tested. Follow-up commit corrects the add_node_attribute callback fix and updates the CHANGELOG accordingly — see comment above for details. Approving.

PR #885, opened 2026-08-10
fix(semantic_extract): reset vector similarity state when scoring fails

### LINKED ISSUE #875: bug(semantic_extract): _calculate_vector_similarity silently abandons similarity scoring on exception, returning a stale or incorrect match to callers
## Summary

`semantica/semantic_extract/methods.py` contains a function `_calculate_vector_similarity()` (or the equivalent internal similarity loop) that catches all exceptions during vector similarity computation and silently passes. The consequence is that a caller receives a match result — `(best_idx, best_score)` — that may reflect whatever partial state existed before the exception, not the actual best match over the full candidate pool. There is no signal to the caller that the computation failed partway through.

## Reproduction

**Exact location — `semantica/semantic_extract/methods.py`, line 384:**

```python
# context: iterating candidates and computing spaCy vector similarity
for i, candidate in enumerate(candidates):
    if not candidate:
        continue
    cand_doc = nlp(candidate)
    if cand_doc.vector_norm:
        score = doc.similarity(cand_doc)
        if score > vector_score:
            vector_score = score
            vector_idx = i
except Exception:
    pass                # ← silent; vector_idx and vector_score retain whatever partial state they had
```

**Concrete consequence:**

If the similarity loop raises (e.g., a malformed candidate string causes `nlp()` to throw, or `doc.similarity()` raises on a zero-norm vector) partway through `n` candidates:

- `vector_idx` retains the index of the best match found in candidates `0..k` where `k < n`
- Candidates `k+1..n` were never scored
- The caller receives `(vector_idx, vector_score)` as if it were the result over all candidates
- There is no log, no raised exception, no return value that indicates truncated scoring

The variable `vector_score` and `vector_idx` are defined before the loop and carry through — the function continues to `if vector_score > best_score: best_idx = vector_idx`, so the stale partial result becomes the return value of the outer function.

## Root Cause

`semantica/semantic_extract/methods.py`, line 384: `except Exception: pass` inside the vector-similarity scoring loop, with no logging and no state reset. The loop state (`vector_idx`, `vector_score`) is not cleared on exception, so partial results silently propagate.

## Suggested Fix

**Minimum fix — log the exception:**

```python
except Exception as exc:
    logger.debug(
        "Vector similarity computation failed at candidate index %d: %s — "
        "partial result from %d candidates will be used as fallback", i, exc, i
    )
```

**Better fix — reset the partial result on exception:**

If the exception indicates the similarity scores already computed are unreliable (e.g., the nlp model produced a nonsense embedding for an earlier candidate), resetting `vector_idx = -1` and `vector_score = 0.0` inside the except block and continuing with a different strategy is safer than propagating a partial result.

**Consider whether this should propagate:** For a system used in KG construction pipelines where entity matching accuracy matters, a silent wrong match may be more harmful than a raised exception that the caller can handle. Whether to log-and-continue or raise depends on whether `_calculate_vector_similarity` is documented as best-effort or as a required accuracy step — that's worth making explicit in the docstring either way.

ArmanGrewal007:
## Description

Previously, the vector similarity block could catch an exception and silently continue, leaving previously computed partial similarity state in memory. That could return a stale or misleading match.

## Type of Change

## Related Issues

Fixes #875 

## Changes Made

This change:
- Logs the vector similarity exception at debug level.
- Resets vector fallback state to neutral values after exception: `vector_score = 0.0` and `vector_idx = -1`
- Keeps fallback behaviour intact (other matching stages still run).
- Clarifies in the function docstring that embedding/vector stages are best-effort and fallback continues instead of raising.

## Testing

- [x] Tested locally

## Documentation

## Breaking Changes

**Breaking Changes**: [No]

## Checklist


KaifAhmad1:
@ArmanGrewal007 Please fix the Qodo finding by adding the candidate index and exception traceback to the debug log. Then we can merge.

KaifAhmad1 (approved):
## Final review — approved

Small, focused, correct fix. Verified directly, not just by reading the diff.

**What it does:** in `find_best_match_index`'s vector-similarity stage (`semantica/semantic_extract/methods.py:369-399`), if the scoring loop throws partway through (e.g., `nlp()` chokes on a malformed candidate), the `except` block now:
- resets `vector_score = 0.0` and `vector_idx = -1`, so a stale partial match from candidates scored before the failure can never leak into `if vector_score > best_score` and become the function's return value
- logs at debug level with the specific failing candidate index and full traceback (`exc_info=True`) — addresses the Qodo finding exactly as requested
- keeps the function best-effort per the new docstring line — execution falls through to fuzzy matching instead of raising

**Verified with a repro:** patched `get_nlp_model` to throw on the 2nd candidate after successfully scoring the 1st (which would previously have left `vector_idx=0, vector_score=0.5` as stale state). Confirmed the debug log fires with the correct candidate index and full traceback, and the vector stage's result is correctly discarded — final result falls through to fuzzy-match fallback instead of returning the stale vector match.

No tests exist for `find_best_match_index` today (pre-existing gap, not introduced by this PR) — a regression test would be a nice follow-up, but not a blocker given the fix was independently verified. All CI checks pass, the review thread is resolved.

Thanks for the clean fix, @ArmanGrewal007. Approving.

PR #887, opened 2026-08-10
fix(explorer): normalize decision timestamps to ISO 8601 in /api/decisions routes

### LINKED ISSUE #884: bug(explorer): every /api/decisions* route returns 500 — record_decision writes a float timestamp, DecisionResponse types it as str
Every `/api/decisions*` route returns HTTP 500 as soon as the graph contains a decision. `ContextGraph.record_decision()` stores `timestamp` as a float epoch, but `DecisionResponse.timestamp` is typed `Optional[str]`, so pydantic rejects the value the library itself produced.

Because decisions are the platform's flagship feature, this makes the entire Decisions workspace of the Knowledge Explorer unusable — the list, the causal chain, the precedent search, and the compliance gate all 500.

## Reproduce

```python
# build_graph.py
import json

g.record_decision(
    category="loan_underwriting",
    scenario="Underwriting review for A-7291",
    reasoning="DTI within policy; clean 36-month credit history",
    outcome="approved",
    confidence=0.94,
    entities=["applicant_A7291"],
)
open("graph.json", "w").write(json.dumps(g.to_dict(), default=str))
```

```bash
python build_graph.py
python -m semantica.server &
curl -X POST http://127.0.0.1:8000/api/import -F "file=@graph.json"
# {"status":"success","nodes_added":3,"edges_added":2}

curl http://127.0.0.1:8000/api/decisions
```

**Expected:** the decision list.

**Actual:**

```json
{"detail":"Internal server error"}
```

Server log:

```
File "semantica/explorer/routes/decisions.py", line 52, in list_decisions
  return [_node_to_decision(node) for node in nodes[skip : skip + limit]]
File "semantica/explorer/routes/decisions.py", line 19, in _node_to_decision
  return DecisionResponse(
pydantic_core._pydantic_core.ValidationError: 1 validation error for DecisionResponse
timestamp
  Input should be a valid string [type=string_type, input_value=1786357493.150935, input_type=float]
```

The same failure hits `GET /api/decisions/{id}`, `/{id}/chain`, `/{id}/precedents` and `/{id}/compliance`, since all of them route through `_node_to_decision`. `GET /api/graph/stats` is unaffected and happily reports `"decision": 3`, which makes the 500 look like a data problem rather than a schema one.

## Root cause

Producer — `semantica/context/context_graph.py:2526`, inside `record_decision`:

```python
timestamp = datetime.now().timestamp()   # float
...
"timestamp": timestamp,
```

Consumer — `semantica/explorer/schemas.py:144`:

```python
    ...
```

Passed through unconverted at `semantica/explorer/routes/decisions.py:19-27`:

```python
timestamp=properties.get("timestamp"),
```

Note that the decision records returned by the Python API (`find_similar_decisions`, `get_decision_insights`) carry both `timestamp` as a float and `recorded_at` as an ISO-8601 string, so an ISO representation is already produced elsewhere in the codebase. `recorded_at` does not survive into `to_dict()`'s node properties, though, so the route cannot simply read that instead.

## Suggested fix

Either coerce in the route:

```python
def _coerce_timestamp(value):
    from datetime import datetime, timezone
    if value is None:
        return ""
    if isinstance(value, (int, float)):
        return datetime.fromtimestamp(float(value), tz=timezone.utc).isoformat()
    return str(value)
```

or widen the schema to `Union[str, float, None]` and let clients handle both. Coercion seems preferable — it keeps a single wire representation and matches the ISO-8601 `recorded_at` already present in the record.

A regression test that imports a graph containing at least one decision and asserts `GET /api/decisions` returns 200 would catch this; the current tests appear to exercise the decision routes only against graphs with no decisions in them.

## Environment

- pydantic 2.13.4, fastapi 0.141.1

yunaremaia:
## Summary

Fixes #884 — every `/api/decisions*` route returned HTTP 500 as soon as the graph contained a decision.

### Root cause

`ContextGraph.record_decision()` stores `timestamp` as an epoch float (`datetime.now().timestamp()`), while `DecisionResponse.timestamp` is typed `Optional[str]`. `_node_to_decision()` passed the raw property through, so pydantic rejected the value the library itself produced. All routes that serialize decisions (list, get, precedents) funnel through `_node_to_decision`, so the entire Decisions workspace of the Knowledge Explorer was unusable.

### Fix

Added a `_normalize_timestamp()` helper at the API boundary (the single funnel for all decision routes):

- **float/int epoch → ISO 8601 UTC string** (`datetime.fromtimestamp(value, tz=timezone.utc).isoformat()`), matching the ISO string convention already used across the explorer API (e.g. `datetime.now(UTC).isoformat()` in ontology/temporal routes, which parse timestamps with `datetime.fromisoformat`)
- **str → passthrough** (imported graphs may already store ISO strings)
- **datetime → `.isoformat()`**
- **None/unknown → None** (never 500 the flagship feature for a malformed value)

The internal storage (`_decisions`, `_temporal_index`) keeps the epoch float — analytics/sorting rely on it — so the conversion happens only where the API contract demands a string.

### Tests

New `TestDecisions::test_decision_with_float_timestamp_returns_iso_string`:
- records a decision via `record_decision()` (the exact producer path from the issue)
- asserts `/api/decisions`, `/api/decisions/{id}` and `/api/decisions/{id}/precedents` return 200
- asserts the timestamp is a parseable ISO 8601 string with UTC tzinfo

Also verified the imported-graph path (node with raw float `timestamp` property) returns 200 with an ISO string.

Full explorer suite: **83 passed** (was 82; +1 new test).

### DIFF
--- a/semantica/explorer/routes/decisions.py
+++ b/semantica/explorer/routes/decisions.py
@@ -34,9 +34,21 @@ def _normalize_timestamp(value: Any) -> Optional[str]:
     if value is None or isinstance(value, str):
         return value
     if isinstance(value, datetime):
+        # Normalize to UTC (assume UTC for naive datetimes) so every
+        # response carries a consistent ISO 8601 offset — the helper's
+        # documented contract.
+        if value.tzinfo is None:
+            value = value.replace(tzinfo=timezone.utc)
+        else:
+            value = value.astimezone(timezone.utc)
         return value.isoformat()
     if isinstance(value, (int, float)):
-        return datetime.fromtimestamp(value, tz=timezone.utc).isoformat()
+        # Malformed imported timestamps (NaN/inf/out-of-range) must degrade
+        # to None instead of raising and 500-ing the whole endpoint.
+        try:
+            return datetime.fromtimestamp(value, tz=timezone.utc).isoformat()
+        except (ValueError, OverflowError, OSError):
+            return None
     return None
 
 
--- a/semantica/explorer/routes/decisions.py
+++ b/semantica/explorer/routes/decisions.py
@@ -3,7 +3,8 @@ Decision routes using ContextGraph-native fallbacks.
 """
 
 import asyncio
-from typing import Optional
+from datetime import datetime, timezone
+from typing import Any, Optional
 
 from fastapi import APIRouter, Depends, HTTPException, Query
 
@@ -12,6 +13,31 @@ from ..schemas import CausalChainResponse, CausalDistanceReport, ComplianceRespo
 from ..session import GraphSession
 
 router = APIRouter(prefix="/api/decisions", tags=["Decisions"])
+
+
+def _normalize_timestamp(value: Any) -> Optional[str]:
+    """Normalize a decision timestamp to an ISO 8601 string (UTC).
+
+    Decisions recorded via :meth:`ContextGraph.record_decision` store an
+    epoch float (``datetime.now().timestamp()``), while the API contract
+    (``DecisionResponse.timestamp``) requires a string. Imported graphs may
+    contain either form, so numeric epochs are converted here — this is the
+    single funnel for every ``/api/decisions*`` route.
+
+    Args:
+        value: The raw ``timestamp`` property of a decision node.
+
+    Returns:
+        An ISO 8601 string (UTC) for numeric/datetime input, the original
+        string for string input, ``None`` for missing or unrecognized values.
+    """
+    if value is None or isinstance(value, str):
+        return value
+    if isinstance(value, datetime):
+        return value.isoformat()
+    if isinstance(value, (int, float)):
+        return datetime.fromtimestamp(value, tz=timezone.utc).isoformat()
+    return None
 
 
 def _node_to_decision(node: dict) -> DecisionResponse:
@@ -23,7 +49,7 @@ def _node_to_decision(node: dict) -> DecisionResponse:
         reasoning=properties.get("reasoning", ""),
         outcome=properties.get("outcome", ""),
         confidence=float(properties.get("confidence", 0.0) or 0.0),
-        timestamp=properties.get("timestamp"),
+        timestamp=_normalize_timestamp(properties.get("timestamp")),
         metadata=properties,
     )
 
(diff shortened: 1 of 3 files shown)


KaifAhmad1:
@yunaremaia Please fix both Qodo findings around invalid epoch conversion and UTC datetime normalization, then push the update. We can merge after that.

yunaremaia:
Fixed both Qodo findings @KaifAhmad1 — pushed to the branch:

1. **Unhandled epoch conversion errors** — `datetime.fromtimestamp()` now wraps `ValueError`/`OverflowError`/`OSError`, so NaN/inf/out-of-range imported timestamps degrade to `None` instead of 500-ing the endpoint.
2. **Datetime not normalized UTC** — naive datetimes are assumed UTC (`replace(tzinfo=timezone.utc)`), aware datetimes convert via `astimezone(timezone.utc)`, so every response carries a consistent ISO 8601 UTC offset.

Added `tests/explorer/test_normalize_timestamp.py` — 11 unit tests: valid/invalid epochs, NaN/inf/out-of-range, naive/non-UTC/UTC datetimes, passthrough types. All green; the 83 explorer API tests still pass.

Sameer6305:
Hi @yunaremaia , if you're working on an issue, please leave a comment on the issue mentioning that you're handling it. This will help us keep track of engaged issues and assign them accordingly. Thanks!

yunaremaia:
Rebased onto the latest main (was behind by several commits) — the fix and 11 tests are unchanged, just re-applied on top of current HEAD. Ready for re-review.

yunaremaia:
Rebased again onto the latest main (SPARQL injection fix merged upstream) — no conflicts, 11 tests still green. Ready for review.

Sameer6305:
Thanks @yunaremaia for the contribution. This issue was already assigned to @joseedson18jc, and their PR was merged, so the issue is now closed.

This PR covers the same fix, so I’ll close it as a duplicate. Feel free to pick another unassigned issue or explore the repo for something new to work on.

yunaremaia:
Thanks @Sameer6305 for letting me know — glad the fix landed with @joseedson18jc's PR. Apologies for the duplicate; I'll check issue assignment before opening a PR from now on. No action needed on my side — closing this out clean.

PR #890, opened 2026-08-10
fix(export): log exceptions in DistanceExporter metric helpers instead of silently swallowing

### LINKED ISSUE #874: bug(export): DistanceExporter silently swallows metric computation errors, producing None values in exported data that are indistinguishable from legitimate "no path" results

Karunasagar12:
## Summary

Fixes #874 — four metric helpers in `DistanceExporter` catch all exceptions with bare `except Exception: return None/{}`, making it impossible to distinguish computation failures from legitimate "no path" results in exported data.

## Changes

Added `logger.debug()` calls to all four helpers (`_betweenness`, `_hop_distance`, `_weighted_distance`, `_semantic_similarity`) that record the node pair and exception message before returning the sentinel value.

This preserves existing behavior (callers still receive `None/{}`) while giving operators visibility into failed computations when debug logging is enabled. The debug level was chosen deliberately — these failures may occur frequently in large graphs with disconnected subgraphs, so `warning` would be too noisy.

## Tests

Added `tests/export/test_distance_exporter_logging.py` — 5 tests verifying:
- Each helper logs on failure (with node IDs and exception message)
- `compute_pairs()` still produces complete rows when all metrics fail

```
============================== 5 passed in 0.78s ===============================
```

### DIFF
--- a/semantica/export/distance_exporter.py
+++ b/semantica/export/distance_exporter.py
@@ -71,7 +71,8 @@ class DistanceExporter:
         try:
             result = self._centrality.calculate_betweenness_centrality(graph_dict)
             return result.get("betweenness", {}) if isinstance(result, dict) else {}
-        except Exception:
+        except Exception as exc:
+            logger.debug("betweenness_centrality failed: %s", exc)
             return {}
 
     def _hop_distance(self, graph_dict: Dict[str, Any], src: str, tgt: str) -> Optional[int]:
@@ -81,7 +82,12 @@ class DistanceExporter:
             result = self._path_finder.bfs_shortest_path(graph_dict, src, tgt)
             path = result.get("path", []) if isinstance(result, dict) else (result or [])
             return len(path) - 1 if path else None
-        except Exception:
+        except Exception as exc:
+            logger.debug(
+                "hop_distance(%s, %s) failed: %s — returning None "
+                "(caller cannot distinguish this from 'no path exists')",
+                src, tgt, exc,
+            )
             return None
 
     def _weighted_distance(self, graph_dict: Dict[str, Any], src: str, tgt: str) -> Optional[float]:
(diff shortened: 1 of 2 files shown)


Sameer6305:
@Karunasagar12 can you fix qodo reviews?

Sameer6305:
@Karunasagar12 also can you comment on the issue you are solving in this PR so i can assign it to you.

KaifAhmad1:
Thanks for taking this on, @Karunasagar12 — closing as a duplicate of #879, which fixes the same four helpers in `distance_exporter.py` for the same issue (#874) and has already been reviewed, tested, and approved.

A few reasons #879 was picked over this one:
- Logs at `warning` level with `exc_info=True` (full traceback). This PR uses `debug`, which stays invisible at any commonly-deployed log level — the issue's whole point was that these failures are currently silent, and `debug` doesn't meaningfully change that in practice.
- #879 also found and fixed the underlying cause of the "no logs show up" problem: `get_logger(__name__)` in this module double-prefixes to `semantica.semantica.export.distance_exporter`, a logger name `setup_logging()` never configures. Without that fix, even these new debug calls wouldn't reach the app's log handler.

If you want to pick up the follow-up work flagged in #879's review (a per-metric error/status field in the exported rows, which is a schema change and was intentionally left out of scope there), that'd be a great next contribution — feel free to open a new issue/PR for it. Appreciate the effort here either way.

Issue #883, opened 2026-08-10
bug(explorer): POST /api/enrich/extract always returns 503 — route imports extract_entities/extract_relations, which do not exist in semantic_extract.methods


joseedson18jc:
`POST /api/enrich/extract` — the entity/relation extraction endpoint advertised in the README — always fails with HTTP 503, on every install, because the route imports two names that do not exist anywhere in the package.

## Reproduce

```bash
pip install semantica
python -m semantica.server   # or: semantica server start

curl -X POST http://127.0.0.1:8000/api/enrich/extract \
  -H "Content-Type: application/json" \
  -d '{"text": "Apple CEO Tim Cook announced record earnings."}'
```

**Expected** (this is the README's own example, under "REST API"): extracted entities and relations.

**Actual:**

```json
{"detail":"semantic_extract module not available. Ensure spacy and transformers are installed."}
```

The message is misleading — spacy and transformers were installed, and `import semantica.semantic_extract` succeeds.

## Root cause

`semantica/explorer/routes/enrich.py:162-163`:

```python
from ...semantic_extract.methods import extract_entities as _extract_entities
from ...semantic_extract.methods import extract_relations as _extract_relations
```

Neither `extract_entities` nor `extract_relations` exists in `semantica/semantic_extract/methods.py`. That module defines only the per-strategy variants:

```
extract_entities_huggingface, extract_entities_llm, extract_entities_ml,
extract_entities_pattern, extract_entities_regex, extract_entities_rules,
extract_relations_cooccurrence, extract_relations_dependency, ...
```

They are not re-exported from the package namespace either:

```python
>>> import semantica.semantic_extract as se
>>> [n for n in dir(se) if n.startswith("extract")]
[]
>>> from semantica.semantic_extract.methods import extract_entities
ImportError: cannot import name 'extract_entities' from 'semantica.semantic_extract.methods'.
Did you mean: 'extract_entities_ml'?
```

The `except ImportError:` handler at `enrich.py:175-179` catches this and reports it as a missing-dependency problem, which hides the real cause.

## Suggested fix

Add the two missing facades to `semantica/semantic_extract/methods.py`, delegating to the classes the README documents:

```python
def extract_entities(text, confidence_threshold: float = 0.7, **kwargs):
    from .named_entity_recognizer import NamedEntityRecognizer
    return NamedEntityRecognizer(
        confidence_threshold=confidence_threshold
    ).extract_entities(text, **kwargs)


[...]

PR #891, opened 2026-08-10
perf(semantic_extract): cache spaCy models instead of calling spacy.load() on every extraction (132 ms → 2.1 ms)

### LINKED ISSUE #889: perf(semantic_extract): extract_entities_ml and two relation extractors call spacy.load() on every invocation — ~120 ms per call, 60x the actual work
`extract_entities_ml()`, `extract_relations_similarity()` and `extract_relations_dependency()` call `spacy.load()` on every invocation. The model is re-read from disk and re-initialized per call, which costs ~120 ms each time and dominates the actual work by roughly 60×.

The module already solves this for one code path — `get_nlp_model()` caches into a `_nlp_cache` global (`methods.py:181-204`) — but the three extraction functions bypass it.

## Measurements

Python 3.12.13, macOS 26.5 (arm64), spacy 3.8.15, `en_core_web_sm` 3.8.0, on the sentence `"Apple CEO Tim Cook announced record earnings in Cupertino."`:

| | median |
| --- | --- |
| `extract_entities_ml(text)` (loads the model each call) | **132 ms** |
| the same doc through an already-loaded `nlp` | **2.1 ms** |
| `spacy.load("en_core_web_sm")` on its own | **120 ms** |

So ~91% of each call is model loading, and successive calls never get cheaper — 119, 113, 199, 160, 132 ms across five runs, with no warm-up curve.

The same shows up end-to-end over the REST API (`POST /api/enrich/extract`, single short sentence): 0.96 s, 1.30 s, 0.50 s on three successive requests — flat, rather than the drop to tens of milliseconds you would expect once a model is resident.

## Affected call sites

```
methods.py:679   extract_entities_ml            nlp = spacy.load(model)
methods.py:683   extract_entities_ml            nlp = spacy.load("en_core_web_sm")     # fallback
methods.py:1403  extract_relations_similarity   nlp = spacy.load(model_name)
methods.py:1408  extract_relations_similarity   nlp = spacy.load("en_core_web_sm")     # fallback
methods.py:1508  extract_relations_dependency   nlp = spacy.load(model)
```

versus the cached path that already exists:

```
methods.py:181-204  get_nlp_model    global _nlp_cache; if _nlp_cache: return _nlp_cache
```

These are not obscure internals — `NamedEntityRecognizer.extract_entities()` dispatches into `extract_entities_ml()`, so this is on the path taken by the CLI (`semantica extract`), the MCP `extract_entities` tool, the pipeline `ner_extract` step, and `POST /api/enrich/extract`. Batch processing through `process_batch()` multiplies it by the number of documents.

## Suggested fix

A module-level cache keyed by model name, mirroring what `get_nlp_model()` already does but parameterized, and guarded by a lock so concurrent requests do not each start their own load:

```python
_MODEL_CACHE: Dict[str, Any] = {}
_MODEL_CACHE_LOCK = threading.Lock()


def _load_spacy_model(name: str):
    """Load a spaCy model once per process, keyed by name."""
    cached = _MODEL_CACHE.get(name)
    if cached is not None:
        return cached
    with _MODEL_CACHE_LOCK:
        cached = _MODEL_CACHE.get(name)   # re-check: another thread may have won
        if cached is None:
            cached = spacy.load(name)
            _MODEL_CACHE[name] = cached
    return cached
```

then `nlp = _load_spacy_model(model)` at the five sites above, keeping the existing `OSError` fallback behavior intact. `functools.lru_cache` would be terser, but an explicit dict keeps the fallback path readable and makes the cache inspectable in tests.

Worth deciding as part of this: whether `get_nlp_model()` should fold into the same cache. It loads with `disable=["parser", "ner", "lemmatizer"]` for similarity work, so its entry is not interchangeable with the NER one — the key would need to cover the `disable` set, not just the model name.

## Notes

Found while fixing #883 — that endpoint was returning 503 unconditionally, so this hot path was not reachable over REST and the cost never showed up there. The behavior itself is older than that fix and independent of it, which is why it is filed separately rather than folded into #886.

Not attached to the exact shape above; if you would rather cache at the `NamedEntityRecognizer` instance level, or push it down into a shared loader alongside `HuggingFaceModelLoader`, happy to send a PR in whichever direction you prefer.

joseedson18jc:
## Description

`extract_entities_ml()`, `extract_relations_similarity()` and `extract_relations_dependency()` called `spacy.load()` on every invocation. This routes those five call sites through a process-level cache, taking `extract_entities_ml()` on a short sentence from **132 ms to 2.1 ms** with identical output.

The module already had a cached loader for one code path — `get_nlp_model()` and its `_nlp_cache` global at `methods.py:181-204` — the extraction functions simply bypassed it.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #889

## Changes Made

**`load_spacy_model(name)`** — a dict keyed by model name, guarded by a lock so concurrent callers do not each start their own load. The five sites (`methods.py` 679/683, 1403/1408, 1508 pre-patch) now call it.

Three details worth reviewing:

- **Errors are not cached and propagate unchanged.** Callers rely on `OSError` to fall back to `en_core_web_sm` and then to pattern extraction; a failed load leaves the cache untouched so that path is unaffected.
- **`get_nlp_model()` keeps its own `_nlp_cache`.** It loads with `disable=["parser", "ner", "lemmatizer"]` for similarity work, so its model is not interchangeable with the NER one. Folding the two would need a key covering the `disable` set — happy to do that here if you would rather have one cache, but it seemed out of scope for a perf fix.
- **Entries record the `spacy` module object they came from.** `tests/test_ner_configurations.py` and others patch `methods.spacy` with a mock and assert on `load` call counts. A cache keyed on name alone would hand one test's mock to the next; the identity check makes a swapped module miss the cache. `clear_spacy_model_cache()` is exported for tests that want to be explicit.

## Testing

- [x] Tested locally

New `tests/semantic_extract/test_spacy_model_cache.py` (6 tests) covers: load-once across repeated calls, same object returned, distinct models cached separately, failed loads propagating and staying uncached, a replaced `spacy` module bypassing the cache, and `extract_entities_ml()` loading once across two calls.

### Test Commands

```bash
pytest tests/semantic_extract/test_spacy_model_cache.py -q
# 6 passed

pytest tests/semantic_extract tests/test_ner_configurations.py tests/test_extractors_dispatch.py -q
# 62 failed, 109 passed, 9 skipped
```

That failure count needs context: **the same 62 tests fail on `main` before this change** (62 failed, 103 passed there — the +6 are the new ones). I diffed the two `FAILED` lists; the sets are identical, nothing introduced and nothing fixed. They look like pre-existing gaps around optional providers and assorted `TypeError`/`AssertionError` cases, unrelated to model loading.

`flake8 --max-line-length 120` on the two changed files reports the same 235 findings before and after — none new. I did not run `black` over `methods.py`; it is not currently black-clean and reformatting would bury a 30-line change in a whole-file diff.

### Benchmark

Python 3.12.13, macOS 26.5 (arm64), spacy 3.8.15, `en_core_web_sm` 3.8.0, on `"Apple CEO Tim Cook announced record earnings in Cupertino."`:

| | median | runs (ms) |
| --- | --- | --- |
| `extract_entities_ml()` before | 132 ms | 119, 113, 199, 160, 132 |
| `extract_entities_ml()` after | **2.1 ms** | 2.1, 1.9, 1.9, 2.1, 2.1 |
| `spacy.load()` alone | 120 ms | |

Entities are unchanged: `Apple/ORG`, `Tim Cook/PERSON`, `Cupertino/GPE`.

Trade-off to be aware of: a loaded pipeline now stays resident for the life of the process, which is the point, but it does mean a long-running server holds the model in memory rather than releasing it between requests. `clear_spacy_model_cache()` is there if that ever needs to be managed.

Found while working on #883 — that endpoint returned 503 unconditionally, so this cost was invisible over REST. Filed and fixed separately from #886, which stays scoped to the two correctness bugs.

@KaifAhmad1 for review.

### DIFF
--- a/semantica/semantic_extract/methods.py
+++ b/semantica/semantic_extract/methods.py
@@ -108,6 +108,7 @@ License: MIT
 
 import re
 import difflib
+import threading
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from typing import Any, Dict, List, Optional, Tuple, Union
 
@@ -152,6 +153,39 @@ spacy, SPACY_AVAILABLE = safe_import("spacy")
 # Global cache for spacy model and text embedder to avoid reloading
 _nlp_cache = None
 _embedder_cache = None
+
+# Cache for models loaded by name, so extraction functions do not pay
+# spacy.load() on every call. Entries record the spacy module they were loaded
+# from: tests patch `methods.spacy` with a mock, and an entry produced by a
+# different module object must not be handed back to a later caller.
+_spacy_model_cache: Dict[str, Tuple[Any, Any]] = {}
+_spacy_model_cache_lock = threading.Lock()
+
+
+def load_spacy_model(name: str):
+    """Load a spaCy model once per process, keyed by model name.
+
+    Raises whatever ``spacy.load`` raises (``OSError`` for a missing model), so
+    callers keep their existing fallback behavior.
+    """
+    cached = _spacy_model_cache.get(name)
+    if cached is not None and cached[0] is spacy:
+        return cached[1]
+
+    with _spacy_model_cache_lock:
+        cached = _spacy_model_cache.get(name)
+        if cached is not None and cached[0] is spacy:
+            return cached[1]
+        nlp = spacy.load(name)
+        _spacy_model_cache[name] = (spacy, nlp)
+        return nlp
+
+
+def clear_spacy_model_cache() -> None:
+    """Drop every cached spaCy model. Intended for tests."""
+    with _spacy_model_cache_lock:
+        _spacy_model_cache.clear()
+
 
 def get_text_embedder():
     """
@@ -676,11 +710,11 @@ def extract_entities_ml(
         return extract_entities_pattern(text, **kwargs)
 
     try:
-        nlp = spacy.load(model)
+        nlp = load_spacy_model(model)
     except OSError:
         logger.warning(f"spaCy model {model} not found, using en_core_web_sm")
         try:
-            nlp = spacy.load("en_core_web_sm")
+            nlp = load_spacy_model("en_core_web_sm")
         except OSError:
             logger.warning(
                 "spaCy model not available, falling back to pattern extraction"
@@ -1400,12 +1434,12 @@ def extract_relations_similarity(
             # Prefer larger models for vectors
             for model_name in ["en_core_web_lg", "en_core_web_md", "en_core_web_sm"]:
                 if spacy.util.is_package(model_name):
-                    nlp = spacy.load(model_name)
+                    nlp = load_spacy_model(model_name)
                     break
             if not nlp:
                  # Try loading what we have
                  try:
-                     nlp = spacy.load("en_core_web_sm") 
+                     nlp = load_spacy_model("en_core_web_sm") 
                  except:
                      pass
         except Exception:
@@ -1505,7 +1539,7 @@ def extract_relations_dependency(
         return extract_relations_pattern(text, entities, **kwargs)
 
     try:
-        nlp = spacy.load(model)
+        nlp = load_spacy_model(model)
     except OSError:
         logger.warning(f"spaCy model {model} not found")
         return extract_relations_pattern(text, entities, **kwargs)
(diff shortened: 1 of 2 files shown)


joseedson18jc:
Superseded by #886 — @KaifAhmad1 asked for this fix to land in that PR, so the commit now sits there as `9050391` and this branch is redundant.

The issue it addresses, #889, stays open until #886 merges. Reopening this is a `git push` away if you would rather review the perf change on its own.

PR #894, opened 2026-08-10
fix(semantic_extract): add missing extract_entities/extract_relations facades (fixes 503 on /api/enrich/extract)

### LINKED ISSUE #883: bug(explorer): POST /api/enrich/extract always returns 503 — route imports extract_entities/extract_relations, which do not exist in semantic_extract.methods

Noir299:
## Summary

Fixes #883 — `POST /api/enrich/extract` always returns HTTP 503 with a misleading "semantic_extract module not available" message.

## Root cause

`semantica/explorer/routes/enrich.py` imports `extract_entities` / `extract_relations` from `semantic_extract.methods`, but that module only defines per-strategy variants (`extract_entities_ml`, `extract_relations_pattern`, …). The `ImportError` is caught by the generic `except ImportError:` handler in the route and reported as a missing dependency.

## Fix

Add the two generic facades to `semantica/semantic_extract/methods.py`, delegating to the classes documented in the README:

- `extract_entities(text, confidence_threshold=0.7, **kwargs)` → `NamedEntityRecognizer`
- `extract_relations(text, entities=None, confidence_threshold=0.6, **kwargs)` → `RelationExtractor` (auto-extracts entities when omitted, matching the route's call `_extract_relations(body.text)`)

## Tests

Added `tests/semantic_extract/test_extract_facades.py`:

- `test_facades_are_exported` — the import regression itself (always runs)
- `test_extract_entities_returns_entities` / `test_extract_relations_auto_extracts_entities` — run real extraction when spaCy is installed (`pytest.importorskip("spacy")`; spaCy is a core dependency, so these run in CI)

## Checklist

- [x] Verified both classes' constructors/method signatures match the facade delegation
- [x] Regression test for the import fix


Noir299:
Thanks for the review! Fixed in a877194: the `/api/enrich/extract` route now passes the already-extracted entities into `extract_relations()`, so NER runs once per request instead of twice (spaCy/HF model loads included).

Added a regression test `test_extract_relations_reuses_provided_entities` that asserts the facade forwards provided entities and never re-calls `extract_entities` when they are present.

Sameer6305:
Thanks @Noir299 for the contribution and for working on this issue.

However, the issue was already assigned to @joseedson18jc, and their PR landed first and has since been merged. The issue is now closed, so this PR is a duplicate and I’ll close it.

Feel free to pick another unassigned issue, or explore the repo and raise a new issue if you find something worth addressing.

PR #901, opened 2026-08-10
test(normalize): make environment-sensitive tests deterministic

### LINKED ISSUE #860: test(normalize): three test correctness bugs — missing skip guard, narrow chardet allowlist, and UTC/local time mismatch

SaurabhScripts:
## Description

Makes the normalize test suite deterministic across optional-dependency, encoding-detector, and timezone environments.

The production implementations are unchanged. The failures came from tests assuming that an optional package was installed, asserting one exact result from a statistical encoding heuristic, and comparing a local-time result with the current UTC date.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Fixes #860

## Changes Made

- Skip only the language-detection assertions that require optional `langdetect`, while keeping fallback behavior tests active.
- Verify ambiguous single-byte encoding detection through codec round-trip behavior and pass an explicit source encoding when testing conversion.
- Freeze the normalize module's clock for relative-date testing so results do not depend on timezone or time of day.
- Remove unused test imports and format the touched test files.

## Testing

- [x] Tested locally

### Test Commands

```bash
# Without optional langdetect installed
.venv/bin/pytest tests/normalize -q
# 77 passed, 2 skipped

# With optional langdetect installed
.venv/bin/pytest tests/normalize/test_language_detector.py -q
# 4 passed

.venv/bin/pytest tests/normalize -q
# 79 passed

.venv/bin/black --check tests/normalize/test_language_detector.py tests/normalize/test_encoding_handler.py tests/normalize/test_date_normalizer.py
.venv/bin/isort --check-only tests/normalize/test_language_detector.py tests/normalize/test_encoding_handler.py tests/normalize/test_date_normalizer.py
.venv/bin/flake8 tests/normalize/test_language_detector.py tests/normalize/test_encoding_handler.py tests/normalize/test_date_normalizer.py --extend-ignore=E203 --max-line-length=88
.venv/bin/python -m build
git diff --check
```

## Documentation

## Breaking Changes

**Breaking Changes**: No

Only tests are changed; production behavior and public APIs remain unchanged.

## Checklist

## Additional Notes

The failures were reproduced locally before applying the test-only fixes:

```text
5 failed, 9 passed
```

### DIFF
--- a/tests/normalize/test_date_normalizer.py
+++ b/tests/normalize/test_date_normalizer.py
@@ -1,11 +1,22 @@
 import unittest
-from datetime import datetime, date, timedelta, timezone
+from datetime import datetime, timedelta, timezone
+from unittest.mock import patch
+
 from semantica.normalize.date_normalizer import (
     DateNormalizer,
-    TimeZoneNormalizer,
     RelativeDateProcessor,
-    TemporalExpressionParser
+    TimeZoneNormalizer,
 )
+
+
+class FrozenDateTime(datetime):
+    @classmethod
+    def now(cls, tz=None):
+        value = cls(2024, 1, 15, 23, 30)
+        if tz is None:
+            return value
+        return value.replace(tzinfo=timezone.utc).astimezone(tz)
+
 
 class TestDateNormalizer(unittest.TestCase):
     def setUp(self):
@@ -14,38 +25,35 @@ class TestDateNormalizer(unittest.TestCase):
     def test_normalize_date_iso(self):
         # Test ISO8601 parsing
         self.assertEqual(
-            self.normalizer.normalize_date("2023-01-01", format="date"),
-            "2023-01-01"
+            self.normalizer.normalize_date("2023-01-01", format="date"), "2023-01-01"
         )
         self.assertEqual(
             self.normalizer.normalize_date("2023-01-01T12:00:00", format="ISO8601"),
-            "2023-01-01T12:00:00+00:00"
+            "2023-01-01T12:00:00+00:00",
         )
 
     def test_normalize_date_relative(self):
-        # Test relative date parsing (e.g., "today", "yesterday")
-        # Note: These depend on current date, so we might need to mock datetime if strictly testing logic,
-        # but for now we'll assume the relative processor uses current time.
-        # We can check if it returns a valid ISO date string.
-        today = datetime.now(timezone.utc).date().isoformat()
-        self.assertEqual(
-            self.normalizer.normalize_date("today", format="date"),
-            today
-        )
+        with patch("semantica.normalize.date_normalizer.datetime", FrozenDateTime):
+            self.assertEqual(
+                self.normalizer.normalize_date("today", format="date"),
+                "2024-01-15",
+            )
 
     def test_normalize_timezone(self):
         # Test timezone conversion
         # "2023-01-01T12:00:00+01:00" -> UTC should be "2023-01-01T11:00:00+00:00"
         normalized = self.normalizer.normalize_date(
-            "2023-01-01T12:00:00+01:00", 
-            timezone="UTC"
+            "2023-01-01T12:00:00+01:00", timezone="UTC"
         )
         self.assertEqual(normalized, "2023-01-01T11:00:00+00:00")
 
     def test_parse_temporal_expression(self):
         # Test range parsing
-        result = self.normalizer.parse_temporal_expression("from 2023-01-01 to 2023-01-31")
+        result = self.normalizer.parse_temporal_expression(
+            "from 2023-01-01 to 2023-01-31"
+        )
         self.assertIsNotNone(result.get("range"))
+
 
 class TestTimeZoneNormalizer(unittest.TestCase):
     def setUp(self):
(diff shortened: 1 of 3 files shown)


KaifAhmad1:
Thanks for the PR, @SaurabhScripts — the fix itself looks solid. This is a duplicate of #881, though: both PRs address the same three root causes from #860 with essentially the same approach (`@unittest.skipUnless(LANGDETECT_AVAILABLE, ...)` on the langdetect-dependent tests, a round-trip decode check instead of a fixed chardet allowlist, an explicit `source_encoding="latin-1"` for the conversion test, and a frozen `datetime.now()` for the relative-date test).

#881 was opened first, is already reviewed and approved, has a green CI run, and I've verified it independently (checked out the branch, ran the full `tests/normalize` suite, confirmed lint/format checks, and traced the datetime-mocking logic against the actual implementation). Closing this one in favor of #881 to avoid two PRs racing to close the same issue.

Appreciate you jumping on this — please feel free to grab another open issue.

PR #904, opened 2026-08-10
test(split): cover untested chunker classes (#864)

### LINKED ISSUE #864: test(split): 9 of 11 public chunker classes have zero test coverage
## Summary

`semantica/split/` exports 11 chunker classes and 14 method functions. The module's only test file (`tests/split/test_splitter.py`) contains 4 tests that cover only `TextSplitter.__init__`, method-list assignment, `SemanticChunker.__init__`, and the `Chunk` dataclass. No actual splitting behavior is tested, and none of the following classes have any test coverage at all:

- `EntityAwareChunker`
- `RelationAwareChunker`
- `GraphBasedChunker`
- `OntologyAwareChunker`
- `HierarchicalChunker`
- `SlidingWindowChunker`
- `StructuralChunker`
- `TableChunker`
- The 14 exported method functions (`split_entity_aware`, `split_graph_based`, etc.)

## Why This Matters

These are GraphRAG-native chunkers — the README specifically calls them out as a key differentiator for KG/GraphRAG workflows. They have real logic (entity boundary detection, triplet preservation, community detection, hierarchical multi-level chunking) and real external dependencies (spaCy, tiktoken, community detection). A bug introduced in any of these classes would go entirely undetected.

`EntityAwareChunker.chunk()` was verified to run without crashing on basic input, so this is not a confirmed crash — it's a coverage gap.

## Suggested Contribution

Add tests for the untested chunker classes. Good starting points:

1. **`EntityAwareChunker`** — test that entity boundaries are preserved across chunks (i.e. a named entity isn't split mid-token between chunks).
2. **`SlidingWindowChunker`** — test overlap invariant: the last `chunk_overlap` tokens of chunk N appear at the start of chunk N+1.
3. **`StructuralChunker`** — test heading-based splits on a Markdown/HTML document.
4. **`TableChunker`** — test that table rows aren't split mid-row.

For chunkers that require optional dependencies (spaCy `en_core_web_sm`, tiktoken, community detection), wrap with `pytest.importorskip` or `pytest.mark.skipif` rather than mocking the dependency away — these tests are most useful when they exercise real library behavior.

pravit-amp:
## Summary
- Adds tests for SlidingWindow, Structural, Table, EntityAware, and related KG chunkers
- Closes #864

## Test plan
- [ ] `pytest tests/split/test_chunkers.py -v`
- [ ] CI green

### DIFF
--- a/semantica/split/methods.py
+++ b/semantica/split/methods.py
@@ -1623,8 +1623,9 @@ def split_sliding_window(
         )
 
     try:
+        effective_stride = stride if stride is not None else (chunk_size - overlap)
         chunker = SlidingWindowChunker(
-            chunk_size=chunk_size, overlap=overlap, stride=stride, **kwargs
+            chunk_size=chunk_size, overlap=overlap, stride=effective_stride, **kwargs
         )
         return chunker.chunk(text, preserve_boundaries=preserve_boundaries, **kwargs)
     except Exception as e:
--- a/semantica/split/methods.py
+++ b/semantica/split/methods.py
@@ -1180,7 +1180,9 @@ def split_graph_based(
         )
 
         entities = ner_extractor.extract(text)
-        relations = relation_extractor.extract(text)
+        # Relation extraction requires the extracted entities; failing to pass
+        # them raises and can trigger the broad fallback to recursive splitting.
+        relations = relation_extractor.extract(text, entities)
 
         # Build graph
         G = nx.Graph()
--- a/tests/split/test_chunkers.py
+++ b/tests/split/test_chunkers.py
@@ -409,6 +409,18 @@ class TestGraphBasedChunker:
         chunks = chunker.chunk(text)
         assert len(chunks) >= 1
         assert all(isinstance(c, Chunk) for c in chunks)
+        # Ensure the graph-based path actually ran (not fallback-to-recursive).
+        assert any(
+            c.metadata.get("method") == "graph_based" for c in chunks
+        ), "Expected at least one graph_based chunk"
+        assert any(
+            c.metadata.get("strategy") == "community"
+            and c.metadata.get("algorithm") == "louvain"
+            for c in chunks
+        ), "Expected graph_based chunk metadata to include strategy/algorithm"
+        assert not any(
+            c.metadata.get("method") == "recursive" for c in chunks
+        ), "Graph-based fallback to recursive was triggered"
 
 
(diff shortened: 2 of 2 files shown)


ZohaibHassan16:
Thanks for addressing the initial Qodo feedback.

There are still some issues as Qodo pointed out: the new tests for `split_sliding_window` and `split_graph_based` are passing because they silently fall back to `split_recursive`. We need to fix the production code for both helpers and tighten the tests so they fail if a fallback occurs.

1. In `semantica/split/methods.py`, `stride` defaults to `None` and gets passed into `SlidingWindowChunker(stride=stride)`. This breaks math inside the chunker (`start += self.stride`) and triggers the exception fallback. 

  For fix I think we should  only pass `stride` to the chunker if it is not `None`.

2. In `semantica/split/methods.py`, the call to `relation_extractor.extract(text)` is missing the required second argument. 
   
   Here the fix should be changing it to `relation_extractor.extract(text, entities)`.

3. In `tests/split/test_chunkers.py`, update `test_split_sliding_window` and `test_split_graph_based` to explicitly assert that `c.metadata.get("method") != "recursive"` for the chunks. That way the test actually proves the target path ran successfully.

pravit-amp:
> Thanks for addressing the initial Qodo feedback.
> 
> There are still some issues as Qodo pointed out: the new tests for `split_sliding_window` and `split_graph_based` are passing because they silently fall back to `split_recursive`. We need to fix the production code for both helpers and tighten the tests so they fail if a fallback occurs.
> 
> 1. In `semantica/split/methods.py`, `stride` defaults to `None` and gets passed into `SlidingWindowChunker(stride=stride)`. This breaks math inside the chunker (`start += self.stride`) and triggers the exception fallback.
> 
> For fix I think we should only pass `stride` to the chunker if it is not `None`.
> 
> 2. In `semantica/split/methods.py`, the call to `relation_extractor.extract(text)` is missing the required second argument.
>    Here the fix should be changing it to `relation_extractor.extract(text, entities)`.
> 3. In `tests/split/test_chunkers.py`, update `test_split_sliding_window` and `test_split_graph_based` to explicitly assert that `c.metadata.get("method") != "recursive"` for the chunks. That way the test actually proves the target path ran successfully.

Hi, just implemented and pushed these changes.

ZohaibHassan16 (approved):
Merge ready.

PR #906, opened 2026-08-10
fix(ingest): SSRF protection for Web and API ingestors (#867)

### LINKED ISSUE #867: security(ingest): WebIngestor and APIIngestor have no SSRF protection — private IPs, loopback, and cloud metadata endpoints are reachable from user-supplied URLs
### Summary

`WebIngestor.ingest_url()`, `WebIngestor.crawl_sitemap()`, and `api_ingestor.py` contain no application-level protection against Server-Side Request Forgery (SSRF). A user-supplied URL is validated only by a `urlparse` netloc check before being passed to `requests.Session.get()`. Every RFC-1918 private IP range, every loopback address, and the cloud metadata endpoint at `169.254.169.254` bypass this check entirely. Additionally, `file://localhost/...` passes the URL validation code in semantica — it is currently stopped only incidentally by the `requests` library's own `InvalidSchema` exception, with no semantica-side control.

This is a high-severity finding for a project positioned as infrastructure for regulated, cloud-deployed enterprise systems. In an AWS, GCP, or Azure environment, a crafted URL of `http://169.254.169.254/latest/meta-data/iam/security-credentials/<role>` would return plaintext IAM credentials including `AccessKeyId`, `SecretAccessKey`, and `SessionToken`.

---

### Reproduction

**Step 1 — Confirm WebIngestor makes real TCP connections (no mocking):**

```python
import socket, threading, time
from semantica.ingest.web_ingestor import WebIngestor

PORT = 19827
received = threading.Event()

def server():
    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
    s.bind(("127.0.0.1", PORT))
    s.listen(1)
    s.settimeout(8)
    try:
        conn, _ = s.accept()
        received.set()
        conn.recv(4096)
        conn.send(b"HTTP/1.1 200 OK\r\nContent-Length: 4\r\nContent-Type: text/plain\r\n\r\ntest")
        conn.close()
    finally:
        s.close()

threading.Thread(target=server, daemon=True).start()
time.sleep(0.1)

ingestor = WebIngestor(respect_robots=False, delay=0)
ingestor.ingest_url(f"http://127.0.0.1:{PORT}/probe")
print("Server received connection:", received.is_set())
````

Output (confirmed):

```text
Server received connection: True
```

**Step 2 — Confirm private IPs pass all application-level validation and reach urllib3:**

```python
from unittest.mock import patch
import urllib3.connectionpool as pool
from semantica.ingest.web_ingestor import WebIngestor

ingestor = WebIngestor(respect_robots=False, delay=0)
attempts = []

def intercepting_urlopen(self, method, url, **kw):
    attempts.append((self.host, self.port, url))
    raise Exception("intercepted at urllib3.urlopen")

urls = [
    ("RFC-1918 10.x",   "http://10.0.0.1/"),
    ("RFC-1918 192.168","http://192.168.1.1/"),
    ("Loopback 127.x",  "http://127.0.0.1:9999/internal"),
]

with patch.object(pool.HTTPConnectionPool, "urlopen", intercepting_urlopen):
    for label, url in urls:
        attempts.clear()
        try:
            ingestor.ingest_url(url)
        except Exception:
            pass
        if attempts:
            h, p, path = attempts[0]
            print(f"CONFIRMED {label}: urllib3.urlopen called → host={h} port={p} path={path}")
        else:
            print(f"BLOCKED {label}: did not reach urllib3")
```

Output (confirmed):

```text
CONFIRMED RFC-1918 10.x: urllib3.urlopen called → host=10.0.0.1 port=80 path=/
CONFIRMED RFC-1918 192.168: urllib3.urlopen called → host=192.168.1.1 port=80 path=/
CONFIRMED Loopback 127.x: urllib3.urlopen called → host=127.0.0.1 port=9999 path=/internal
```

**Step 3 — Cloud metadata endpoint `169.254.169.254`:**

The same code path confirmed above for `10.0.0.1` applies without change to `169.254.169.254`. This was not independently observed in this test environment because Windows does not configure a link-local interface by default, so the OS routing table silently drops the packet before it reaches urllib3's connection pool — not because semantica filters it. On any AWS EC2 instance, GCP Compute Engine VM, Azure VM, or ECS/EKS container (where `169.254.169.254` is always routable), the metadata service would respond with IAM/service-account credentials.

**Step 4 — `file://localhost` netloc-check bypass:**

```python
from urllib.parse import urlparse

url = "file://localhost/etc/passwd"
p = urlparse(url)
print(f"scheme='{p.scheme}', netloc='{p.netloc}'")
print(f"Validation check 'not scheme or not netloc' = {not p.scheme or not p.netloc}")
```

Output:

```text
scheme='file', netloc='localhost'
Validation check 'not scheme or not netloc' = False
```

`file://localhost/...` has both a non-empty `scheme` and a non-empty `netloc`, so `ingest_url()`'s only validation passes without raising `ValidationError`. The `requests` library happens to raise `InvalidSchema` for `file://` because it has no registered file handler — but this is downstream library behavior, not a semantica security control. A `requests` version or session adapter that does support `file://` (e.g., via a registered `FileAdapter`) would transparently read local files.

---

### Root Cause

`semantica/ingest/web_ingestor.py`, `ingest_url()` method. The entire URL validation block is:

```python
# Validate URL format
try:
    parsed = urlparse(url)
    if not parsed.scheme or not parsed.netloc:
        raise ValidationError(
            f"Invalid URL format: {url}. "
            "URL must include scheme (http/https) and netloc (domain)."
        )
except Exception as e:
    raise ValidationError(f"Invalid URL: {url}") from e
```

This check only verifies that *a* scheme and *a* netloc are present. It does not:

* Validate that the scheme is `http` or `https` (blocking `file://`, `gopher://`, etc.)
* Resolve the hostname to an IP and reject private/loopback/link-local ranges
* Validate the netloc against any allowlist or denylist

The same absence of validation applies to `crawl_sitemap()` (which calls `ingest_url()` per URL extracted from a sitemap) and to `api_ingestor.py` (confirmed: no IP/scheme filtering present). The sitemap path adds a second vector: a malicious or compromised sitemap at a legitimate URL could embed `http://10.x.x.x/...` entries that the crawler dutifully fetches.

---

### Suggested Fix

Add an `_validate_url_for_request(url: str)` function that runs before any network call in `ingest_url()`, `crawl_sitemap()`, and the equivalent path in `api_ingestor.py`. The function should:

1. **Explicit scheme allowlist** — reject any scheme that is not `http` or `https` before doing anything else, regardless of netloc. This closes the `file://`, `gopher://`, and similar scheme vectors with a direct check rather than relying on downstream library behavior:

```python
if parsed.scheme not in ("http", "https"):
    raise ValidationError(f"URL scheme '{parsed.scheme}' is not permitted. Only http and https are allowed.")
```

2. **Hostname resolution + IP range check** — resolve the final IP of the hostname and reject connections to private address space:

```python
import ipaddress, socket as _socket

BLOCKED_NETWORKS = [
    ipaddress.ip_network("10.0.0.0/8"),
    ipaddress.ip_network("172.16.0.0/12"),
    ipaddress.ip_network("192.168.0.0/16"),
    ipaddress.ip_network("127.0.0.0/8"),
    ipaddress.ip_network("169.254.0.0/16"),  # link-local / cloud metadata
    ipaddress.ip_network("::1/128"),
    ipaddress.ip_network("fc00::/7"),
]

def _is_ssrf_target(hostname: str) -> bool:
    try:
        resolved = _socket.getaddrinfo(hostname, None)
        for _, _, _, _, sockaddr in resolved:
            addr = ipaddress.ip_address(sockaddr[0])
            if any(addr in net for net in BLOCKED_NETWORKS):
                return True
    except _socket.gaierror:
        pass  # DNS failure — let requests handle with its own error
    return False
```

Apply this check after scheme validation, raising `ValidationError` if it returns `True`, unless the caller explicitly opts in via a constructor parameter (e.g., `allow_private_ips=True` for internal/trusted deployments).

Note that hostname resolution introduces a TOCTOU window (the IP can change between the DNS check and the actual connection). A defense-in-depth approach is to combine the above check with an outbound firewall rule at the infrastructure level. The validation function narrows the window to a normal DNS TTL and closes the obvious cases.

pravit-amp:
## Summary
- Add shared `validate_url_for_request` SSRF guard
- Wire into WebIngestor, SitemapCrawler, and RESTIngestor
- Opt-in `allow_private_ips` for trusted internal deployments
- Closes #867

## Test plan
- [X] `pytest tests/ingest/test_ssrf_protection.py -v`
- [x] CI green

### DIFF
--- a/semantica/ingest/api_ingestor.py
+++ b/semantica/ingest/api_ingestor.py
@@ -38,6 +38,7 @@ except (ImportError, OSError):
 from ..utils.exceptions import ProcessingError, ValidationError
 from ..utils.logging import get_logger
 from ..utils.progress_tracker import get_progress_tracker
+from .ssrf import validate_url_for_request
 
 
 @dataclass
@@ -78,11 +79,14 @@ class RESTIngestor:
 
         Args:
             config: Optional REST API ingestion configuration dictionary
-            **kwargs: Additional configuration parameters (merged into config)
+            **kwargs: Additional configuration parameters (merged into config).
+                Recognized keys include ``allow_private_ips`` (default False) to
+                opt into fetching private/loopback/link-local endpoints.
         """
         self.logger = get_logger("api_ingestor")
         self.config = config or {}
         self.config.update(kwargs)
+        self.allow_private_ips = bool(self.config.get("allow_private_ips", False))
 
         # Initialize session with retry strategy
         self.session = requests.Session()
@@ -103,7 +107,10 @@ class RESTIngestor:
         # Initialize progress tracker
         self.progress_tracker = get_progress_tracker()
 
-        self.logger.debug("REST API ingestor initialized")
+        self.logger.debug(
+            "REST API ingestor initialized (allow_private_ips=%s)",
+            self.allow_private_ips,
+        )
 
     def ingest_endpoint(
         self,
@@ -137,7 +144,7 @@ class RESTIngestor:
                 - metadata: Additional metadata
 
         Raises:
-            ValidationError: If endpoint is invalid
+            ValidationError: If endpoint is invalid or fails SSRF checks
             ProcessingError: If request fails
         """
         tracking_id = self.progress_tracker.start_tracking(
@@ -148,6 +155,10 @@ class RESTIngestor:
         )
 
         try:
+            validate_url_for_request(
+                endpoint, allow_private_ips=self.allow_private_ips
+            )
+
             # Prepare request
             request_headers = self.session.headers.copy()
             if headers:
@@ -196,6 +207,11 @@ class RESTIngestor:
                 },
             )
 
+        except ValidationError:
+            self.progress_tracker.stop_tracking(
+                tracking_id, status="failed", message=f"Invalid endpoint URL: {endpoint}"
+            )
+            raise
         except requests.exceptions.RequestException as e:
             self.progress_tracker.stop_tracking(
                 tracking_id, status="failed", message=str(e)
--- a/semantica/ingest/api_ingestor.py
+++ b/semantica/ingest/api_ingestor.py
@@ -38,7 +38,7 @@ except (ImportError, OSError):
 from ..utils.exceptions import ProcessingError, ValidationError
 from ..utils.logging import get_logger
 from ..utils.progress_tracker import get_progress_tracker
-from .ssrf import validate_url_for_request
+from .ssrf import request_with_ssrf_guard
 
 
 @dataclass
@@ -155,19 +155,17 @@ class RESTIngestor:
         )
 
         try:
-            validate_url_for_request(
-                endpoint, allow_private_ips=self.allow_private_ips
-            )
-
             # Prepare request
             request_headers = self.session.headers.copy()
             if headers:
                 request_headers.update(headers)
 
-            # Make request
-            response = self.session.request(
-                method=method,
-                url=endpoint,
+            # Make request (SSRF-safe; validates URL and each redirect hop)
+            response = request_with_ssrf_guard(
+                method,
+                endpoint,
+                session=self.session,
+                allow_private_ips=self.allow_private_ips,
                 headers=request_headers,
                 params=params,
                 data=data,
--- a/semantica/ingest/api_ingestor.py
+++ b/semantica/ingest/api_ingestor.py
@@ -38,7 +38,7 @@ except (ImportError, OSError):
 from ..utils.exceptions import ProcessingError, ValidationError
 from ..utils.logging import get_logger
 from ..utils.progress_tracker import get_progress_tracker
-from .ssrf import request_with_ssrf_guard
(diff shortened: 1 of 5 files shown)


ZohaibHassan16:
Qodo is flagging "duplicate URL validation", which I think is a false flag. Can you change the code comment from:

 ```
# SSRF guard: scheme allowlist + private/loopback/link-local checks
# (request_with_ssrf_guard re-validates the URL and each redirect hop)
```

to:

```
# Validate before robots check: RobotsChecker.read() makes an unguarded
# HTTP request to <host>/robots.txt and must not reach blocked addresses.
```

ZohaibHassan16 (approved):
Merge ready.

PR #907, opened 2026-08-11
test(deduplication): cover ClusterBuilder, MergeStrategyManager, and batch paths

### LINKED ISSUE #866: test(deduplication): ClusterBuilder, MergeStrategyManager, and batch deduplication paths have no test coverage
## Summary

`semantica/deduplication/` exports 5 main classes across 10 source files totaling 3,933 lines. The module's single test file (`tests/deduplication/test_deduplication.py`, 540 lines) provides reasonable coverage of `DuplicateDetector` and `SimilarityCalculator`, but the following are untested or lightly covered:

**Untested classes:**
- `ClusterBuilder` — union-find and hierarchical clustering for batch deduplication at scale
- `MergeStrategyManager` — the strategy pattern layer that manages which merge strategy is applied and how property conflicts are resolved
- `PropertyMergeRule` — custom per-property merge rules

**Untested paths in tested classes:**
- `EntityMerger.merge_duplicates()` with a real `DuplicateGroup` (not just `merge_entities()` convenience function)
- `SimilarityCalculator.calculate_similarity()` with `method="embedding"` (requires embeddings, but can be mocked)
- Incremental detection (`detect_new_duplicates()`) — the O(n×m) path that compares new entities against an existing set

## Why This Matters

`ClusterBuilder` is the code path used when deduplicating large-scale knowledge graphs (`build_clusters()` method). A bug in the union-find implementation would silently produce wrong cluster assignments — entities that should be merged aren't, or unrelated entities get grouped — with no test catching it.

## Suggested Contribution

Add tests for the three untested classes. Good starting points:

1. **`ClusterBuilder`** — build clusters from a list of known duplicate pairs and verify entities in the same cluster share a cluster ID. Verify two unrelated entities get different cluster IDs.
2. **`MergeStrategyManager`** — apply a strategy to merge two conflicting entities and verify the correct field value is chosen.
3. **`PropertyMergeRule`** — define a custom rule (e.g., "always take the longer string for the `description` field") and verify it overrides the default strategy.

pravit-amp:
## Summary
- Adds `tests/deduplication/test_cluster_merge_coverage.py` with 34 tests for the untested / lightly covered paths called out in #866
- Covers `ClusterBuilder` (shared vs distinct cluster IDs, union-find merges), `MergeStrategyManager` / `PropertyMergeRule` (including custom longer-string resolution), `EntityMerger.merge_duplicates()`, embedding similarity, and incremental detection
Closes #866
## Test plan
- [x] `pytest tests/deduplication/test_cluster_merge_coverage.py -v`
- [x] `pytest tests/deduplication/ -q` (88 passed)

### DIFF
--- a/tests/deduplication/test_cluster_merge_coverage.py
+++ b/tests/deduplication/test_cluster_merge_coverage.py
@@ -152,23 +152,34 @@ class TestClusterBuilderCoverage(unittest.TestCase):
 
     def test_update_clusters_adds_matching_entity(self):
         """Incremental update attaches a near-duplicate into an existing cluster."""
-        builder = ClusterBuilder(similarity_threshold=0.4, min_cluster_size=2)
-        existing = [
-            Cluster(
-                cluster_id="cluster_0",
-                entities=[
-                    {"id": "a1", "name": "Apple Inc.", "type": "Company",
-                     "properties": {"industry": "Technology"}},
-                ],
-            )
-        ]
-        new_entities = [
-            {"id": "a2", "name": "Apple", "type": "Company",
-             "properties": {"industry": "Tech"}},
-        ]
-        result = builder.update_clusters(existing, new_entities)
-        clustered_ids = {e["id"] for c in result.clusters for e in c.entities}
-        self.assertTrue({"a1", "a2"}.issubset(clustered_ids) or len(result.clusters) >= 1)
+        builder = ClusterBuilder(similarity_threshold=0.8, min_cluster_size=2)
+        a1 = {
+            "id": "a1",
+            "name": "Apple Inc.",
+            "type": "Company",
+            "properties": {"industry": "Technology"},
+        }
+        a2 = {
+            "id": "a2",
+            "name": "Apple",
+            "type": "Company",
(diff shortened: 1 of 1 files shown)


ZohaibHassan16:
Thanks for addressing the Qodo finding quickly. However there is still one issue, that is prov assertion is checking the wrong key.



`merged_entity["merged_from"]` is always written by `MergeStrategyManager.merge_entities()` regardless of `preserve_provenance` , so the test passes even if `_add_provenance()` is deleted entirely. What `preserve_provenance=True` actually does is call `EntityMerger._add_provenance()`, which writes to `merged_entity["metadata"]["provenance"]`. That path is currently untested.


Replace the provenance assertions with:
```python
provenance = op.merged_entity.get("metadata", {}).get("provenance", {})
self.assertIn("merged_from", provenance)
self.assertIn("merge_count", provenance)
self.assertEqual(provenance["merge_count"], len(op.source_entities))
```

And ideally add a counterpart test with EntityMerger(preserve_provenance=False) asserting that metadata.provenance is absent , otherwise a regression that always writes provenance would go undetected too.

PR #908, opened 2026-08-11
test(context): skip symlink test without Windows privilege

ssynb:
## Summary

Skip the markdown symlink-safety test on Windows only when the OS reports that the current process lacks the privilege required to create a symbolic link.

## Problem

`test_markdown_export_rejects_symlink_without_touching_target` fails before exercising `AgentMemory.export()` on Windows environments without Developer Mode or `SeCreateSymbolicLinkPrivilege`:

```
OSError: [WinError 1314] A required privilege is not held by the client
```

That is an environment capability failure, not a failure of the export safety behavior under test.

## Changes

- catch the Windows-specific `winerror == 1314` from `Path.symlink_to`
- skip only in that case
- re-raise every other `OSError`, preserving genuine failures

## Testing

```
uv run --no-project --python 3.12 --with pytest --with networkx --with numpy --with pydantic --with scipy --with scikit-learn --with pyyaml --with requests --with python-dateutil --with rdflib --with gensim pytest tests/context/test_agent_memory_markdown.py tests/context/test_context_retriever_hybrid.py tests/context/test_query_result_unwrap.py -q
```

Result: `73 passed, 1 skipped in 2.15s`.

Also ran `git diff --check`.


ZohaibHassan16:
Please address the Qodo finding , define a named constant for 1314 (e.g. _ERROR_PRIVILEGE_NOT_HELD = 1314) instead of the bare number.

Also sys is missing from the imports at the top of the file , add import sys alongside the existing import errno line otherwise the sys.platform check will raise a NameError on every platform.

ssynb:
Addressed in 7192ad6: added the named _ERROR_PRIVILEGE_NOT_HELD constant and updated the comparison. I also confirmed that import sys is present alongside import errno in the current PR head (it was already included in the original commit). Local verification on Python 3.12: 45 passed, 1 skipped; Black and isort checks pass.

ZohaibHassan16:
Thank you. LGTM

ZohaibHassan16 (approved):
Approved.

PR #909, opened 2026-08-11
security: require API-key auth on all Explorer API routes (GHSA-j4mq)

KaifAhmad1:
## Summary

Fixes GHSA-j4mq-hprp-987v: every Explorer API route (bulk import/export, delete, LLM-backed ontology generation, SPARQL, etc.) was mounted with no authentication, and both server entrypoints bind `0.0.0.0` by default — anyone reaching the port got full read/write/delete on the graph, with no per-caller logging to trace it afterward.

- New `require_auth` dependency (`explorer/dependencies.py`): checks `X-API-Key` against `SEMANTICA_API_KEY`. **Fails closed** — refuses with 503 if no key is configured, rather than silently serving anonymously. `SEMANTICA_ALLOW_ANONYMOUS=true` opts out explicitly for local dev.
- Wired into all 11 API routers in both `explorer/app.py` and `server.py`. `/health`, `/api/info`, static assets, and the SPA catch-all stay public (the frontend needs to load before it can attach a key to its own API calls).
- `/ws/graph-updates` handshake checks the same key via header or `?api_key=` query param (browsers can't set custom headers on a WS handshake) before accepting the connection.
- Default bind changed from `0.0.0.0` → `127.0.0.1` in `server.py`'s `main()` and `cli.py`'s `server start`. `cli.py` warns if a non-loopback host is passed explicitly without a key configured.
- Startup logging reports the resolved auth mode (enabled / anonymous / unconfigured) in both app factories.
- Deploy recipes that expose a public endpoint by default now document/auto-generate `SEMANTICA_API_KEY`: `docker-compose.yml`, Railway, Fly, Render. `docker-compose.dev.yml` (local-only) defaults to anonymous for dev convenience.

## Test plan

- [x] New `tests/explorer/test_explorer_auth.py` (11 tests): fail-closed default (503), missing/wrong/correct key (401/200), `SEMANTICA_ALLOW_ANONYMOUS=true` opt-in, public routes stay public, WS handshake via header and via query param.
- [x] New `tests/explorer/conftest.py`: defaults the ~200 pre-existing explorer tests to `SEMANTICA_ALLOW_ANONYMOUS=true` so they keep exercising route logic without needing a key — env vars are read fresh per-request in the dependency, not cached at app-construction time, so this works even with module-scoped `TestClient` fixtures.
- [x] Full `tests/explorer` + `tests/vector_store` suite: 428 passed, 36 skipped. The only failures (6, in `test_pinecone_store.py`) are pre-existing and unrelated to this change (Pinecone client mocking issue, confirmed unmodified by this PR).
- [x] Manual end-to-end check: imported `server.py`, confirmed `/health` stays open and `/api/graph/nodes` requires auth.


ZohaibHassan16:
Pushed the fix for Qodo's CORS finding, awaiting scan checks to merge it @KaifAhmad1

ZohaibHassan16 (approved):
Merge ready.

PR #910, opened 2026-08-11
security: sanitize Cypher labels/relationship types/property keys (GHSA-482h)

KaifAhmad1:
## Summary

Fixes GHSA-482h-hw99-h62p: node labels and property keys passed to `create_node`/`create_relationship` were interpolated directly into Cypher strings in the Neptune, Neo4j, and FalkorDB graph stores. Property *values* are parameterized, but labels and keys can't be bound as query parameters, and nothing validated them — so a document-derived entity type or property name (the normal ingest path) could close the current Cypher token early and append arbitrary statements (e.g. `DETACH DELETE`), running with the application's database credentials.

- New shared `semantica/graph_store/query_sanitize.py`: `sanitize_identifier()` generalizes `age_store.py`'s existing `_sanitize_label`/`_sanitize_rel_type` (the only backend that already validated this) into a helper the other backends can import without creating an import cycle with `graph_store.py`/`methods.py`.
- Applied at every label/relationship-type/property-key interpolation site in `amazon_neptune.py`, `neo4j_store.py`, `falkordb_store.py`, `graph_store.py` (`degree_centrality`'s own query builder), and `methods.py` (`update_relationship`'s own query builder) — covers `create_node`, `create_nodes`, `create_relationship`, `get_nodes`, `get_relationships`, `get_neighbors`, `shortest_path`, `update_node`, `create_index`, and all relationship-type filters across the three backends.
- `depth`/`max_depth` path-length parameters are also cast to `int` before interpolation as defense-in-depth (already typed `int`, but Python doesn't enforce that at runtime).

## Test plan

- [x] New `tests/graph_store/test_cypher_injection.py` (12 tests): unit tests on `sanitize_identifier` directly, plus the advisory's own injection payload (`N}) MATCH (victim) DETACH DELETE victim //`) run against Neptune/Neo4j/FalkorDB `create_node`/`create_relationship`, asserting the malicious query is never built or sent (query-capture stubs) and that legitimate labels/keys still work.
- [x] Regression tests added to `tests/test_graph_store.py` (`degree_centrality`) and `tests/test_graph_store_methods.py` (`update_relationship`) for the two query builders that live outside the three backend files.
- [x] Full graph_store-related suite: 224 passed, 3 skipped, no regressions.
- [x] Ran the broader `tests/explorer` + `tests/ingest` suites for indirect regressions — clean aside from 11 pre-existing failures in `tests/ingest/test_web_ingestor.py` and related files, confirmed unrelated (this PR touches only `graph_store/*` and its tests; those failures predate this branch, from the SSRF-hardening merge already on `main`).


KaifAhmad1:
Ready for the review. @Sameer6305 @ZohaibHassan16

Sameer6305 (dismissed):
Completed a security-focused review of this PR, including the Cypher identifier sanitization and the affected Neo4j/Neptune/FalkorDB query paths.

I also identified and fixed the remaining Neo4j depth/max_depth injection paths and the GraphStore hops forwarding path, with regression coverage added. The relevant test suite passes, and the follow-up test-quality issues have been cleaned up.

I found no remaining blocking issues. Approving.

PR #911, opened 2026-08-11
security: validate triplet IRIs before SPARQL interpolation (GHSA-8vgg)

KaifAhmad1:
## Summary

Fixes GHSA-8vgg-8mr4-r236: `Triplet.subject`/`.predicate` (and, in some builders, `.object`) were interpolated directly into SPARQL update/query strings in the Blazegraph and RDF4J stores, and into a SELECT filter in the Jena store. A subject containing `>` closes the `<...>` IRI token early, so the rest of the value is parsed as more SPARQL. Entity names are document text in the normal ingest pipeline, so anyone whose content gets processed could append operations like `CLEAR ALL`, running with the application's store credentials.

Applied the existing `sparql_escaping.validate_uri` (already used by `anzo_store.py`, the one backend that was already hardened — this PR generalizes its approach rather than inventing a new one) at every subject/predicate/object interpolation site:

- **`blazegraph_store.py`**: `_build_insert_data`, `_triplets_to_rdf` (currently unreachable dead code — its output isn't used by any caller — but fixed for consistency/future-proofing), `bulk_load`'s `graph` option, `get_triplets`'s filter, `delete_triplet`.
- **`rdf4j_store.py`**: `_triplets_to_ntriples`, `get_triplets`'s filter, `delete_triplet`. (`add_triplets`'s `graph` option was already validated.)
- **`jena_store.py`**: `get_triplets`'s filter — the only vulnerable site. `add_triplets`/`delete_triplet` already use rdflib's native Python API (`Graph.add`/`.remove` with `URIRef`) rather than building query strings, so they were never exploitable this way.

## Test plan

- [x] New `tests/triplet_store/test_sparql_injection.py` (12 tests): reproduces the advisory's own injection payload (`http://example.com/a> ... ; CLEAR ALL ; INSERT DATA { ...`) against all three backends' write and read paths, asserting the malicious query is never built or sent, plus confirms legitimate input still works.
- [x] Full `tests/triplet_store` suite: 330 passed, no regressions.

**Side note, not part of this fix:** while adding read-path coverage I found that `jena_store.py`'s `get_triplets()` builds syntactically invalid SPARQL for its WHERE-clause filters (missing a `FILTER()`/separator before the equality conditions — `{ ?s ?p ?o ?s = <...> }` isn't valid SPARQL). This is a pre-existing correctness bug, unrelated to the injection fix (it predates this branch and isn't security-relevant), so I left it alone here — worth a separate follow-up issue.


KaifAhmad1:
@Sameer6305 @ZohaibHassan16 Ready for the review

ZohaibHassan16:
Applied the fixes Qodo was asking for @KaifAhmad1

Sameer6305:
Found one regression during the security review: RDF4JStore.delete_triplet() currently rejects legitimate literal-valued objects because the new validation is applied unconditionally to the object. I’m implementing the fix while preserving the URI/object injection protection and adding regression coverage. I’ll push the fix to the PR shortly.

KaifAhmad1:
> Found one regression during the security review: RDF4JStore.delete_triplet() currently rejects legitimate literal-valued objects because the new validation is applied unconditionally to the object. I’m implementing the fix while preserving the URI/object injection protection and adding regression coverage. I’ll push the fix to the PR shortly.

Sure

Sameer6305 (commented):
Final review from my side is complete. The RDF4J literal-object regression has been fixed and the security coverage remains intact. The PR is ready to merge once the CHANGELOG.md conflict is resolved. After that, no further changes are needed from my side.

Sameer6305:
@KaifAhmad1 are you fixing the conflict or should i do it??

PR #870, opened 2026-08-09
fix(mcp): report package version

oiahoon:
## Description

The MCP server reported a stale `0.4.0` version even though the package is at `0.6.0`. Both public MCP version surfaces now reuse the package-level `semantica.__version__` value so future package version updates are reflected automatically.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #863

## Changes Made

- Use the package version in the MCP initialize response.
- Use the same version in the `semantica://schema/info` resource.
- Add focused regression coverage for both protocol paths.

## Testing

- [x] Tested locally
- [x] Added tests for new functionality
- [x] Package builds successfully (`python -m build`)

### Test Commands

```bash
PYTHONPATH=. python3 -m unittest discover -s tests -p 'test_mcp_server_version.py' -v
python3 -m py_compile semantica/mcp_server/__init__.py tests/test_mcp_server_version.py
black --check tests/test_mcp_server_version.py
isort --check-only tests/test_mcp_server_version.py
flake8 --max-line-length=88 --extend-ignore=E203,W503 tests/test_mcp_server_version.py
python -m build --no-isolation
git diff --check
```

## Documentation

- [ ] Updated relevant documentation
- [ ] Added code examples if applicable
- [ ] Updated API reference if adding new APIs
- [ ] Updated cookbook if adding new examples
- [x] No documentation changes needed

## Breaking Changes

**Breaking Changes**: No

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] Package builds successfully

PR #913, opened 2026-08-11
fix(mcp): report the package version instead of a stale hardcoded 0.4.0

### LINKED ISSUE #863: bug(mcp): MCP server reports version 0.4.0 to clients — should be 0.6.0 (pyproject.toml)
## Summary

The MCP server hardcodes `"version": "0.4.0"` in two places, while the package is at `0.6.0`. This version is returned directly to every connecting MCP client — Claude Desktop, Windsurf, Cline, Continue, VS Code Copilot, and any other MCP-aware tool will display `semantica 0.4.0` in their server info panel even if the user installed `0.6.0`.

## Reproduction

Install semantica and connect any MCP client to the server. The `initialize` handshake response and the `semantica://schema/info` resource both return the wrong version:

```python
# From semantica/mcp_server/__init__.py, lines confirmed present:
SERVER_INFO = {
    "name": "semantica",
    "version": "0.4.0",   # ← stale
}
```

```python
# _read_resource for semantica://schema/info:
return {
    "name": "Semantica",
    "version": "0.4.0",   # ← stale
    "tools": [t["name"] for t in TOOLS],
    "resources": [r["uri"] for r in RESOURCES],
}
```

Both paths confirmed via direct inspection and `pyproject.toml`:

pyproject.toml version: 0.6.0
All version strings in mcp_server/init.py: ['0.4.0', '0.4.0']

## Root Cause

The version string in `mcp_server/__init__.py` was not updated when the package version advanced from 0.4.0 to 0.6.0. There is no dynamic link between the two — the MCP server hardcodes its own version literal.

## Suggested Fix

Two options:

**Option A (simple):** Update both hardcoded `"0.4.0"` strings to `"0.6.0"` in `semantica/mcp_server/__init__.py`.

**Option B (durable):** Read the version dynamically from the package metadata so it never drifts again:

```python
from importlib.metadata import version as _pkg_version
_SEMANTICA_VERSION = _pkg_version("semantica")

SERVER_INFO = {
    "name": "semantica",
    "version": _SEMANTICA_VERSION,
}
```

And in `_read_resource`:
```python
return {
    "name": "Semantica",
    "version": _SEMANTICA_VERSION,
    ...
}
```

Option B prevents this from recurring on every future release. This is a two-line fix.

erensh27:
Fixes #863

**Problem:** the MCP server hardcoded `"version": "0.4.0"` in `SERVER_INFO` and in the `semantica://schema/info` resource while the package is at `0.6.0` (`pyproject.toml` / `semantica/__init__.py`). Every connecting MCP client displays the stale version in its server info panel.

**Fix:** import `__version__` from the package and use it in both places, so the reported version can never drift from the installed package again.

**Test:** `tests/mcp/test_mcp_server_version.py` asserts both `SERVER_INFO["version"]` and the `semantica://schema/info` resource version equal `semantica.__version__`.

Verified: `pytest tests/mcp/test_mcp_server_version.py` (2 passed).

### DIFF
--- a/semantica/mcp_server/__init__.py
+++ b/semantica/mcp_server/__init__.py
@@ -47,6 +47,8 @@ import os
 import sys
 from typing import Any
 
+from semantica import __version__
+
 # ── logging ────────────────────────────────────────────────────────────────
 _log_level = getattr(logging, os.environ.get("SEMANTICA_LOG_LEVEL", "WARNING").upper(), logging.WARNING)
 logging.basicConfig(stream=sys.stderr, level=_log_level,
@@ -477,7 +479,7 @@ def _read_resource(uri: str) -> dict:
     if uri == "semantica://schema/info":
         return {
             "name": "Semantica",
-            "version": "0.4.0",
+            "version": __version__,
             "tools": [t["name"] for t in TOOLS],
             "resources": [r["uri"] for r in RESOURCES],
         }
@@ -490,7 +492,7 @@ def _read_resource(uri: str) -> dict:
 
 SERVER_INFO = {
     "name": "semantica",
-    "version": "0.4.0",
+    "version": __version__,
 }
 
 CAPABILITIES = {
--- a/tests/mcp/test_mcp_server_version.py
+++ b/tests/mcp/test_mcp_server_version.py
@@ -1,0 +1,18 @@
+"""MCP server must report the package version, not a hardcoded string.
+
+Regression test for #863: SERVER_INFO and the semantica://schema/info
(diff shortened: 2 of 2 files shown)


KaifAhmad1:
Thanks for the fix, @erensh27! This turned out to be a duplicate — #870 landed the same fix (import `semantica.__version__` into `SERVER_INFO` and the `semantica://schema/info` resource) and merged on 2026-08-12, closing #863. `main` already has this behavior, which is why this branch now shows a conflict. Closing this one, but appreciate you catching and fixing it independently.

PR #915, opened 2026-08-11
fix: resolve ReDoS in _PREFIX_DECL regex (CodeQL py/polynomial-redos #1897)

Sameer6305:
## Summary

Fixes the polynomial regular expression ReDoS vulnerability reported by CodeQL
alert `py/polynomial-redos` (issue #1897) in `semantica/explorer/routes/sparql.py`.

## Root Cause

The `_PREFIX_DECL` compiled pattern had an ambiguous trailing quantifier:

```python
# BEFORE (vulnerable)
_PREFIX_DECL = re.compile(r"^\s*(?:PREFIX\s+\S+|BASE)\s*<[^>]*>\s*", re.IGNORECASE | re.MULTILINE)
```

The trailing `\s*` after `<[^>]*>` overlaps with the `[^>]*` character class on
inputs that contain no closing `>` (e.g. `base<` followed by many `!<` repetitions).
The regex engine explores every possible split between the two quantifiers, causing
**O(n²) polynomial backtracking**.

The data-flow path confirmed by CodeQL:
`req.query` (user input) → `_is_read_only_query()` → `_PREFIX_DECL.sub("", cleaned)`

## Fix

```python
# AFTER (safe)
_PREFIX_DECL = re.compile(
    r"^[ \t]*(?:PREFIX[ \t]+\S+|BASE)[ \t]*<[^>]*>[ \t]*(?:\n|$)",
    re.IGNORECASE | re.MULTILINE,
)
```

Two targeted changes eliminate the ambiguity:

1. **All `\s` quantifiers replaced with `[ \t]`** — matches only horizontal
   whitespace, which is disjoint from the `[^>]*` IRI body match space.
2. **Trailing `\s*` replaced with `[ \t]*(?:\n|$)`** — matches horizontal
   whitespace then a hard line boundary. `[^>]*` and `[ \t]*` have no character
   overlap, so the engine has exactly one way to match and cannot backtrack.

Additionally, a `_SPARQL_MAX_QUERY_LEN = 10_000` character guard is added at the
entry of `_is_read_only_query` as defence-in-depth — oversized input is rejected
before any regex work begins, bounding worst-case cost even against future pattern
changes.

## Verification

- ReDoS payload `base<` + `!<` × 5,000 chars completes in **< 1 ms** (previously unbounded).
- Normal `PREFIX`/`BASE` stripping and all read-only query detection logic confirmed correct.
- 203 existing tests pass; 1 pre-existing unrelated failure (Windows symlink privilege
  error in markdown export test, not related to this change).

## References

- CWE-1333 (Improper Control of a Resource Through its Lifetime)
- CWE-730 / CWE-400 (ReDoS / Uncontrolled Resource Consumption)
- OWASP: Regular expression Denial of Service
- CodeQL rule: `py/polynomial-redos`

### DIFF
--- a/semantica/explorer/routes/sparql.py
+++ b/semantica/explorer/routes/sparql.py
@@ -52,21 +52,25 @@ _FORBIDDEN_KEYWORDS = re.compile(
 # the IRI (`BASE <...>`, vs. `PREFIX ex: <...>`), so the prefix-name token
 # is optional.
 #
-# ReDoS note: the original pattern ended with `<[^>]*>\s*` where the
-# trailing `\s*` could overlap with the `[^>]*` character class on inputs
-# that contain no closing `>`, causing polynomial backtracking (CodeQL
-# py/polynomial-redos, issue #1897).  The fix replaces the ambiguous `\s*`
-# suffix with `[ \t]*(?:\n|$)` which matches only horizontal whitespace
-# followed by a hard line boundary, so there is no character-class overlap
-# and the engine cannot split the match in multiple ways.
+# ReDoS fix (CodeQL py/polynomial-redos, issue #1897):
+#
+# The original pattern `<[^>]*>\s*` was vulnerable because `\s*` (which
+# matches newlines) could overlap with `[^>]*` on inputs that contain no
+# closing `>` (e.g. `base<!!<!<...`), forcing the engine to explore every
+# possible split between the two quantifiers — O(n²) backtracking.
+#
+# The fix uses `<[^>\r\n]*>` for the IRI body: excluding CR and LF from
+# the character class means the IRI match can never span a line boundary,
+# and the disjoint trailing `[ \t]*` (horizontal whitespace only) has zero
+# character-class overlap with `[^>\r\n]*`, so the engine has exactly one
+# way to match.  No end-of-line anchor is needed or used, which correctly
+# handles both inline prologues (`PREFIX ex: <...> SELECT ...` on one line)
+# and CRLF line endings (`\r\n`) without any special casing.
 _COMMENT_LINE = re.compile(r"(?:^|(?<=\s))#[^\n]*", re.MULTILINE)
 _PREFIX_DECL = re.compile(
-    r"^[ \t]*(?:PREFIX[ \t]+\S+|BASE)[ \t]*<[^>]*>[ \t]*(?:\n|$)",
+    r"^[ \t]*(?:PREFIX[ \t]+\S+|BASE)[ \t]*<[^>\r\n]*>[ \t]*",
     re.IGNORECASE | re.MULTILINE,
 )
-
-
-_SPARQL_MAX_QUERY_LEN = 10_000  # chars; guards regex cost on uncontrolled input
 
 
 def _is_read_only_query(query: str) -> bool:
@@ -76,12 +80,11 @@ def _is_read_only_query(query: str) -> bool:
     checking the first keyword. Also rejects queries containing SPARQL Update
     keywords anywhere in the body, preventing injection via embedded strings
     or multi-statement tricks.
-    """
-    # 0. Reject excessively long inputs before any regex work (defense-in-depth
-    #    against ReDoS even if a future regex change reintroduces ambiguity).
-    if len(query) > _SPARQL_MAX_QUERY_LEN:
-        return False
 
+    Note: callers are responsible for enforcing any input-length limit *before*
+    calling this function so that an oversized-query rejection can be surfaced
+    as a distinct, actionable error rather than the generic read-only message.
+    """
     # 1. Remove single-line comments that could hide the real query type
     cleaned = _COMMENT_LINE.sub("", query)
     # 2. Remove PREFIX/BASE declarations
@@ -175,6 +178,11 @@ _SPARQL_MAX_ROWS = 5_000     # hard cap on returned rows
 _SPARQL_TIMEOUT_S = 30       # seconds before abandoning the await
 _SPARQL_MAX_CONCURRENT = 4   # semaphore: max simultaneous executions
 _SPARQL_MAX_GRAPH_NODES = 50_000  # cap on graph nodes/edges to prevent OOM
+# Defense-in-depth against ReDoS: reject inputs longer than this before any
+# regex work so that even a future regex regression is bounded.  Checked in
+# execute_sparql() (not inside _is_read_only_query) so the route can return
+# a distinct, actionable error message rather than the generic read-only one.
+_SPARQL_MAX_QUERY_LEN = 10_000  # chars
 
 # Semaphore caps how many graph.query calls run concurrently so that
 # timed-out threads (which keep running in the pool) cannot crowd out
@@ -203,6 +211,24 @@ async def execute_sparql(
     req: SparqlRequest,
     session: GraphSession = Depends(get_session),
 ):
+    # Resource-limit check: reject oversized queries before any regex work.
+    # This is intentionally a separate, earlier check from _is_read_only_query
+    # so clients receive a specific, actionable message rather than the generic
+    # read-only rejection, and operators can tune _SPARQL_MAX_QUERY_LEN without
+    # touching query-semantics code.
+    if len(req.query) > _SPARQL_MAX_QUERY_LEN:
+        return SparqlResponse(
+            columns=[],
+            rows=[],
+            total=0,
+            error=(
+                f"Query exceeds the maximum allowed length of "
+                f"{_SPARQL_MAX_QUERY_LEN:,} characters "
+                f"({len(req.query):,} received). "
+                f"Please shorten your query."
+            ),
+        )
+
     if not _is_read_only_query(req.query):
         return SparqlResponse(
             columns=[],
--- a/semantica/explorer/routes/sparql.py
+++ b/semantica/explorer/routes/sparql.py
@@ -51,8 +51,22 @@ _FORBIDDEN_KEYWORDS = re.compile(
 # the query. BASE declarations have no prefix name between the keyword and
 # the IRI (`BASE <...>`, vs. `PREFIX ex: <...>`), so the prefix-name token
 # is optional.
+#
+# ReDoS note: the original pattern ended with `<[^>]*>\s*` where the
+# trailing `\s*` could overlap with the `[^>]*` character class on inputs
+# that contain no closing `>`, causing polynomial backtracking (CodeQL
+# py/polynomial-redos, issue #1897).  The fix replaces the ambiguous `\s*`
+# suffix with `[ \t]*(?:\n|$)` which matches only horizontal whitespace
+# followed by a hard line boundary, so there is no character-class overlap
+# and the engine cannot split the match in multiple ways.
 _COMMENT_LINE = re.compile(r"(?:^|(?<=\s))#[^\n]*", re.MULTILINE)
-_PREFIX_DECL = re.compile(r"^\s*(?:PREFIX\s+\S+|BASE)\s*<[^>]*>\s*", re.IGNORECASE | re.MULTILINE)
+_PREFIX_DECL = re.compile(
+    r"^[ \t]*(?:PREFIX[ \t]+\S+|BASE)[ \t]*<[^>]*>[ \t]*(?:\n|$)",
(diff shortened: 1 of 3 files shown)


ZohaibHassan16:
Reviewing it.

ZohaibHassan16:
There is one minor functional regression with the new regex. Because the new pattern expects (?:\n|$) immediately after the trailing spaces, it now fails to match valid SPARQL prefixes that contain an inline comment.

For example, this valid prefix will now be completely ignored: PREFIX ex: <http://example.org/> # This is my comment\n

We can easily restore inline comment support without reintroducing any ReDoS vulnerabilities by inserting a non-backtracking comment consumer (?:#[^\n]*)? just before the newline boundary.

Could you update the _PREFIX_DECL regex to this?

```
_PREFIX_DECL = re.compile(
    r"^[ \t]*(?:PREFIX[ \t]+\S+|BASE)[ \t]*<[^>]*>[ \t]*(?:#[^\n]*)?(?:\n|$)",
    re.IGNORECASE | re.MULTILINE,
)

```

ZohaibHassan16:
Also @Sameer6305 , I have identified two critical issues with the current implementation that need to be addressed before we can merge:

1. I think 10,000 characters limit is too restrictive or narrow. It is far too small for our enterprise graph workloads.

Automated Text-to-SPARQL generators and SHACL shape validations frequently generate queries between 15,000 and 35,000 characters. Plus, batch filtering using VALUES blocks with 100+ URIs will easily exceed 10,000 characters.
We shouldn't equate length with complixy, a 120 character query can cause a CPU DoS via a cartesian join, whereas a 30,000-character VALUES lookup executes in 5ms. Since we already have robust execution safeguards (_SPARQL_TIMEOUT_S, _SPARQL_MAX_ROWS and concurrency limits), I recommend increasing _SPARQL_MAX_QUERY_LEN to 100_000 (or making it configurable via an environment variable). This will successfully block megabyte-scale spam attacks without breaking our legitimate automated pipelines.

2.  The current approach uses a global regex replacement (_PREFIX_DECL.sub("", cleaned)) to strip prefixes from the query. Because this is not context-aware, it will corrupt any legitimate query that happens to use the word PREFIX inside a string literal (e.g., SELECT ?s WHERE { ?s rdfs:comment "PREFIX example" }).

To fix this, we should completely remove the .sub() stripping logic. Instead, we can update the existing _ALLOWED_QUERY_TYPES with a verbose, anchored regex that safely jumps over the preamble (including inline comments) without modifying the query payload at all:


```
_ALLOWED_QUERY_TYPES = re.compile(
    r"""
    \A                     
    (?:                      
        \s+             
      | \#[^\n]*             
      | PREFIX\s+[\w\-]*:\s*<[^>]*>  
      | BASE\s+<[^>]*>      
    )*
    \s*                     
    (SELECT|ASK|CONSTRUCT|DESCRIBE) 
    \b                    
    """,
    re.IGNORECASE | re.VERBOSE,
)
```

Sameer6305:
> There is one minor functional regression with the new regex. Because the new pattern expects (?:\n|$) immediately after the trailing spaces, it now fails to match valid SPARQL prefixes that contain an inline comment.
> 
> For example, this valid prefix will now be completely ignored: PREFIX ex: http://example.org/ # This is my comment\n
> 
> We can easily restore inline comment support without reintroducing any ReDoS vulnerabilities by inserting a non-backtracking comment consumer (?:#[^\n]*)? just before the newline boundary.
> 
> Could you update the _PREFIX_DECL regex to this?
> 
> ```
> _PREFIX_DECL = re.compile(
>     r"^[ \t]*(?:PREFIX[ \t]+\S+|BASE)[ \t]*<[^>]*>[ \t]*(?:#[^\n]*)?(?:\n|$)",
>     re.IGNORECASE | re.MULTILINE,
> )
> ```
Hey @ZohaibHassan16, thanks for the suggestion, we actually took your proposed pattern and ran it through the full test suite to make sure we weren't missing anything. Here's what we found:

**The regression doesn't exist in the current code.**

The key is that `_is_read_only_query` runs two substitutions in sequence, `_COMMENT_LINE.sub()` fires *first*, then `_PREFIX_DECL.sub()` runs on the already-cleaned string. So your example:

```
PREFIX ex: <http://example.org/> # This is my comment\n
```

becomes this after step 1:

```
PREFIX ex: <http://example.org/> \n
```

...which is just a plain space-terminated PREFIX line that the current pattern handles perfectly. We verified this for LF, CRLF, and multiple PREFIX lines with comments, all pass cleanly through the pipeline.

**There are also two issues with the proposed replacement we wanted to flag:**

1. It restores `<[^>]*>` instead of the fixed `<[^>\r\n]*>` — that's the original ReDoS-vulnerable IRI body we patched in this PR.

2. The `(?:\n|$)` end anchor re-breaks the **inline prologue** case we fixed in the previous review round. `PREFIX ex: <...> SELECT ...` on a single line would fail to strip again, causing valid read-only queries to get rejected.

I will add comment to the code so its less surprising for future reader

Sameer6305:
> Also @Sameer6305 , I have identified two critical issues with the current implementation that need to be addressed before we can merge:
> 
> 1. I think 10,000 characters limit is too restrictive or narrow. It is far too small for our enterprise graph workloads.
> 
> Automated Text-to-SPARQL generators and SHACL shape validations frequently generate queries between 15,000 and 35,000 characters. Plus, batch filtering using VALUES blocks with 100+ URIs will easily exceed 10,000 characters. We shouldn't equate length with complixy, a 120 character query can cause a CPU DoS via a cartesian join, whereas a 30,000-character VALUES lookup executes in 5ms. Since we already have robust execution safeguards (_SPARQL_TIMEOUT_S, _SPARQL_MAX_ROWS and concurrency limits), I recommend increasing _SPARQL_MAX_QUERY_LEN to 100_000 (or making it configurable via an environment variable). This will successfully block megabyte-scale spam attacks without breaking our legitimate automated pipelines.
> 
> 2. The current approach uses a global regex replacement (_PREFIX_DECL.sub("", cleaned)) to strip prefixes from the query. Because this is not context-aware, it will corrupt any legitimate query that happens to use the word PREFIX inside a string literal (e.g., SELECT ?s WHERE { ?s rdfs:comment "PREFIX example" }).
> 
> To fix this, we should completely remove the .sub() stripping logic. Instead, we can update the existing _ALLOWED_QUERY_TYPES with a verbose, anchored regex that safely jumps over the preamble (including inline comments) without modifying the query payload at all:
> 
> ```
> _ALLOWED_QUERY_TYPES = re.compile(
>     r"""
>     \A                     
>     (?:                      
>         \s+             
>       | \#[^\n]*             
>       | PREFIX\s+[\w\-]*:\s*<[^>]*>  
>       | BASE\s+<[^>]*>      
>     )*
>     \s*                     
>     (SELECT|ASK|CONSTRUCT|DESCRIBE) 
>     \b                    
>     """,
>     re.IGNORECASE | re.VERBOSE,
> )
> ```

Hey @ZohaibHassan16, thanks for the detailed feedback. These are both important points, so I went through them again with focused local tests to make sure the conclusions were solid. Here’s where I landed:

**Point 1 :-  Query length limit (10,000 chars)**

I agree that 10,000 characters may be too conservative for real enterprise workloads. The limit was added mainly as defense-in-depth against ReDoS, not as a query complexity limit, and your point that length ≠ complexity is valid. For example, a 120-character cartesian product can be much more expensive than a 30,000-character `VALUES` block.

I’d suggest increasing the limit to **50,000 characters** as a reasonable middle ground:

* Gives more room for Text-to-SPARQL, SHACL validation, and large `VALUES` blocks
* Still prevents unnecessarily large, megabyte-scale inputs from reaching the regex pipeline
* The existing `_SPARQL_TIMEOUT_S`, `_SPARQL_MAX_ROWS`, and `_SPARQL_MAX_CONCURRENT` safeguards independently control actual execution cost

For the environment variable suggestion, I agree it would be useful for operator flexibility. I’d lean toward keeping that as a follow-up, though, since it adds config and documentation overhead to a security-focused PR.

@KaifAhmad1, would be good to get your thoughts on the exact limit as well (50k vs 100k) and whether you’d prefer the env-var configuration in this PR or as a follow-up.

**Point 2 :-  PREFIX inside string literals**

I also re-verified this locally with several focused test cases. The concern doesn't apply here because of how `_PREFIX_DECL` is anchored.

It uses `re.MULTILINE` with a `^` anchor, so it can only match at the **start of a line**. A string literal such as `"PREFIX ex: <http://bad.com/>"` normally appears inside the query body, so the `^` anchor won't match it.

I tested the realistic cases again, including:

* Mid-line string literals
* `FILTER` expressions
* Multi-line/indented values
* The edge case of a literal starting at column 0

All 5 cases remained uncorrupted. The column-0 case isn't valid SPARQL syntax anyway.

Regarding the proposed `_ALLOWED_QUERY_TYPES` replacement, the preamble-skip approach itself is reasonable, and I noticed we already use a similar pattern in `sparql_escaping.py` with `CONSTRUCT_QUERY_RE`.

However, the specific pattern proposed has a serious ReDoS issue:

```python
(?:
    \s+
  | ...
)*
\s*
(SELECT|...)
```

The outer `(?:\s+)*` and the trailing `\s*` both consume the same whitespace. For input containing leading whitespace that doesn't end in an allowed keyword, the regex engine can try many different ways of partitioning those spaces.

I reproduced this locally: **20 spaces already took ~800ms, and 30+ spaces caused it to hang.** So this is exponential backtracking and would introduce a worse vulnerability than the original CodeQL finding.

Making that pattern safe would require atomic grouping or possessive quantifiers, which Python's standard `re` module doesn't support. Because of that, I think the current two-step approach — strip comments, strip `PREFIX` declarations, then match the first keyword — is still the safer option. Each step is a simple anchored replacement without the overlapping quantifiers.

@ZohaibHassan16, I’d be interested in your thoughts on these findings, especially if you see another way to achieve the preamble-skip approach without introducing the backtracking issue.

And @KaifAhmad1 , happy to adjust `_SPARQL_MAX_QUERY_LEN` based on your preference for 50k vs 100k and whether you want the env-var option in this PR.

Overall, I’m happy to make the changes once we agree on the direction.

KaifAhmad1 (approved):
## Final review

Independently verified everything in this thread rather than trusting the prose back-and-forth:

**ReDoS fix — confirmed real.** Timed the exact pathological pattern from the report (`base<` + `!<` repeated) at 5,000 and 20,000 reps: 0.238ms → 0.841ms, ~3.5x for 4x input. That's linear scaling, not the ~16x a surviving quadratic blowup would show. Fixed.

**Sameer's rebuttal to @ZohaibHassan16's inline-comment regression claim — confirmed correct.** Ran `PREFIX ex: <http://example.org/> # This is my comment\n\nSELECT ...` through the actual pipeline step by step: `_COMMENT_LINE.sub()` strips the comment first (leaving `PREFIX ex: <http://example.org/> \n\n...`), *then* `_PREFIX_DECL` runs on that already-cleaned text — which it handles fine since there's no comment left to trip over. `_is_read_only_query` correctly returns `True`. The regression @ZohaibHassan16 described doesn't reproduce; no code change needed there.

**@ZohaibHassan16's second comment (query-length limit, context-unaware `.sub()`) — real points, but non-blocking for this PR:**
- The 10,000-char limit is a legitimate product-tuning question, but it's *defense-in-depth* layered on top of a regex that's now provably linear-time regardless of input length — the actual ReDoS fix doesn't depend on this number. Raising it is a reasonable follow-up, not a correctness fix.
- The context-unaware `.sub()` stripping a line that happens to start with literal "PREFIX "/"BASE " inside a multi-line string value is real but narrow (only triggers if such a line is the very first thing on its own line). More importantly, I checked: `graph.query(req.query)` (line 253) always executes the **original, unmodified** query — the `cleaned` string is only used for the allowlist yes/no decision. So even in the worst case here, this can only cause a legitimate query to be wrongly *rejected*; it can never let something malicious *through*, since rdflib's parser still gates whatever actually executes. Worth a follow-up issue, not a blocker.

**Testing:** 225 `tests/explorer` + 82 SPARQL-specific tests, all passing on this branch.

Approving — this closes the CodeQL alert correctly and doesn't introduce the regressions raised in review. The two open points from @ZohaibHassan16 are worth a fast follow-up PR but shouldn't hold up shipping the ReDoS fix itself.

PR #916, opened 2026-08-11
security: DNS check-then-use pinning for SSRF fetcher, close object-IRI gap

KaifAhmad1:
## Summary

Two follow-up hardening items that were flagged as secondary/deferred while fixing GHSA-8c7v-62gr-hj6g and GHSA-8vgg-8mr4-r236:

### 1. DNS check-then-use (TOCTOU) window in the ontology URL fetcher

GHSA-8c7v-62gr-hj6g's own "Fix" section called this out as a secondary gap: `_validate_fetch_url()` resolved and validated a hostname once, but `_fetch_url_sync()` then let `requests` resolve the same hostname again independently at connect time — a low-TTL or rebinding DNS answer could differ between the two lookups, reopening the SSRF window the validation exists to close.

`_validate_fetch_url()` now returns the validated IP, and a new `_make_pinned_session()` builds a per-hop `requests.Session` whose connection pool is pinned directly to that IP (bypassing DNS resolution for the connection entirely), while explicitly restoring the real hostname as the outgoing HTTP `Host` header and, for HTTPS, the TLS SNI `server_hostname`/`assert_hostname` — so the connection reaches the validated IP but still presents (and is verified against) the real hostname's identity, keeping virtual hosting and certificate validation correct.

**Implementation note**: an earlier draft of this fix set urllib3's `_dns_host` post-construction, assuming it was decoupled from `host` (true in some urllib3 releases). In the version this project installs (2.7.0), `host` is a `@property` that reads/writes `_dns_host` directly, so that approach would have silently changed the Host header too, without erroring — it was caught by an end-to-end test against a real local server before landing. The actual fix instead pins the connection pool's `host` directly and restores the real identity via an explicit `Host` header + SNI params.

### 2. Pre-wrapped object IRIs skipped full validation (Blazegraph, RDF4J)

Flagged by automated review (Codex) during GHSA-8vgg-8mr4-r236's fix: a triplet object already wrapped in `<...>` only had its inner content checked for a literal space or `>` in `_format_object_for_sparql`/`_format_object_for_ntriples`, not run through `sparql_escaping.validate_uri()` like the unwrapped-object branch. Both branches now validate identically.

## Test plan

- [x] `tests/explorer/test_ontology_dns_pinning.py` (6 tests):
  - 2 real, non-mocked end-to-end tests against local HTTP servers proving the connection reaches the pinned IP for a hostname that's never actually resolved (and, separately, that a *different* real DNS answer is ignored in favor of the pin)
  - 2 real-TLS tests using a generated self-signed certificate: one proving SNI/cert-hostname verification checks the **real hostname** (not the pinned IP) and succeeds; one negative control proving a hostname/cert mismatch is still correctly **rejected**, not silently bypassed. Both gracefully skip if the optional `cryptography` package isn't installed.
  - 2 unit tests on `_validate_fetch_url`'s new return value and unchanged rejection behavior
- [x] `tests/explorer/test_ontology_ssrf.py` updated for the new per-hop session construction (mocks `_make_pinned_session` directly rather than the module-level `requests.get`, since `_fetch_url_sync` no longer calls it)
- [x] 4 new tests in `tests/triplet_store/test_sparql_injection.py` for the object-IRI fix (malicious pre-wrapped IRI rejected, legitimate pre-wrapped IRI still accepted, both stores)
- [x] Full `tests/explorer` + `tests/triplet_store` suite: 566 passed, no regressions


Sameer6305:
@KaifAhmad1 on it.

Sameer6305 (approved):
Reviewed the full PR diff against the latest `semantica-agi/main`, including the DNS-pinning implementation, redirect/SSRF handling, object-IRI validation, connected code paths, and regression tests.

**Key checks:**

* Verified DNS pinning with real local HTTP servers, including an unresolvable hostname and a hostname resolving to a different IP.
* Confirmed redirects are manually followed and each redirect target is re-validated with fresh IP pins.
* Confirmed proxy environment variables are disabled for the SSRF-sensitive fetcher.
* Reviewed the pre-wrapped IRI fix in both Blazegraph and RDF4J.
* `tests/explorer`: **234 passed**
* `tests/triplet_store`: **331 passed, 1 skipped**
* `test_ontology_dns_pinning.py`: **12 passed**
* `test_ontology_ssrf.py`: **4 passed**
* `test_sparql_injection.py`: **20 passed**
* `git diff --check`: clean.

**Suggestion:** Consider documenting an explicit supported `urllib3` compatibility range since the pinning implementation relies on urllib3 internals. This is **non-blocking** and does not affect the approval.

**Final verdict: Approved , no actionable security or correctness issues found.**

PR #917, opened 2026-08-11
security: validate WebSocket Origin against the CORS allowlist (GHSA-4643)

KaifAhmad1:
## Summary

Fixes GHSA-4643-wpgq-w329: `CORSMiddleware` doesn't cover WebSocket handshakes at all (Starlette's CORS support only wraps HTTP), so under `SEMANTICA_ALLOW_ANONYMOUS=true` — the mode `docker-compose.dev.yml` ships — `is_valid_api_key`'s anonymous-mode bypass accepted a `/ws/graph-updates` connection from **any** origin. Loopback binding isn't a boundary against a browser: any page the operator has open can still reach `ws://localhost:8000/ws/graph-updates` directly, and `ConnectionManager.broadcast` sends every `graph_mutation` to every connected socket with no per-connection scoping.

Combined with `/api/import` accepting `multipart/form-data` (a CORS-safelisted content type, so cross-origin POSTs skip preflight entirely), a hostile page could write to the graph over REST and read the result back over the unauthenticated WebSocket — demonstrated end-to-end in the report with a real uvicorn server and a real websockets client, full node payload delivered off-origin.

**Not affected**: any deployment with `SEMANTICA_API_KEY` configured — the handshake already rejects without a valid key in that mode. This is an anonymous-mode-only, development-configuration exposure, not a production one.

## Fix

Check the handshake's `Origin` header against `app.state.explorer_settings['allowed_origins']` — the same list `CORSMiddleware` already enforces for HTTP — before the key check. A missing `Origin` (native/CLI clients, which never set the header — only browsers do) is still allowed through, since the browser is the only threat this closes.

## Test plan

- [x] `test_websocket_rejects_hostile_origin_under_anonymous_mode` — pins the core regression
- [x] `test_websocket_rejects_hostile_origin_even_with_correct_key` — defense in depth: Origin is checked *before* the API key, so a leaked/guessed key alone can't hijack the socket
- [x] `test_websocket_accepts_allowlisted_origin_under_anonymous_mode` — legitimate dev-mode usage still works
- [x] `test_websocket_accepts_missing_origin_under_anonymous_mode` — native/CLI clients still work
- [x] Full `tests/explorer` suite: 226 passed, no regressions


KaifAhmad1:
@Sameer6305 Review this one also

PR #919, opened 2026-08-11
fix(export): log exceptions in DistanceExporter metric helpers instead of silently swallowing

### LINKED ISSUE #874: bug(export): DistanceExporter silently swallows metric computation errors, producing None values in exported data that are indistinguishable from legitimate "no path" results

Karunasagar12:
Fixes #874

## Problem

All four private metric helpers (`_betweenness`, `_hop_distance`, `_weighted_distance`, `_semantic_similarity`) catch exceptions with bare `except Exception` and return sentinel values (`None` or `{}`) with no logging. This makes it impossible to distinguish computation failures from legitimate "no path" results in exported data.

## Fix

Add `logger.debug()` calls in each exception handler with the exception message and (where applicable) the source/target node IDs. The sentinel return values are unchanged — no API or output format break.

Operators with debug logging enabled can now observe failed computations; downstream consumers still see the same export schema.

## Tests

Added `tests/test_distance_exporter_logging.py` verifying:
- Each helper emits the expected debug message on exception
- Successful computations produce no spurious failure logs

### DIFF
--- a/semantica/export/distance_exporter.py
+++ b/semantica/export/distance_exporter.py
@@ -71,7 +71,10 @@ class DistanceExporter:
         try:
             result = self._centrality.calculate_betweenness_centrality(graph_dict)
             return result.get("betweenness", {}) if isinstance(result, dict) else {}
-        except Exception:
+        except Exception as exc:
+            logger.debug(
+                "betweenness_centrality computation failed: %s — returning empty dict", exc
+            )
             return {}
 
     def _hop_distance(self, graph_dict: Dict[str, Any], src: str, tgt: str) -> Optional[int]:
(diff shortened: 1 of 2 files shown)


KaifAhmad1:
Thanks for the resubmission, @Karunasagar12 — closing this one too as a duplicate of #879, for the same reasons as #890 (which this PR largely re-implements from a new branch):

- #879 logs at `warning` with `exc_info=True`; this PR logs at `debug` with only `str(exc)`, no traceback — flagged by Qodo here too.
- #879 fixes the root cause of why these logs wouldn't be seen at all: `get_logger(__name__)` double-prefixes to `semantica.semantica.export.distance_exporter`, which `setup_logging()` never configures. This PR's own test file hardcodes that broken double-prefixed name as the *expected* logger name rather than fixing it, so the underlying visibility problem would remain even with debug logging enabled.

#879 is already reviewed, tested (71 passing in `tests/export/`), and approved. If you'd like to contribute the per-metric error/status export field discussed as a follow-up in #879's review (a schema change, intentionally out of scope there), that would be a solid next PR — happy to take a look.

Karunasagar12:
Makes sense — thanks for the detailed explanation. The double-prefixed logger issue is a good catch, I missed that entirely.

I'll pick up the per-metric error/status field as a follow-up. Will look at #879's review discussion and put together a schema proposal.

PR #926, opened 2026-08-12
fix(context): make ContextGraph.add_edge idempotent by deduping on edge_id

### LINKED ISSUE #922: bug(context): ContextGraph.add_edge has no dedupe — identical edges are stored repeatedly under one shared edge ID, and re-ingest doubles the edge set
## Bug Description

`ContextGraph._add_internal_edge` appends to `self.edges`, `edge_type_index`, and `_adjacency` unconditionally, with no check for an edge that is already present. Adding the same edge twice therefore stores it twice.

Edge identity is content-derived: `_resolve_edge_identity` builds `edge_id` from `(source_id, target_id, edge_type, weight, metadata, valid_from, valid_until)`. Two identical `add_edge` calls produce two edge objects that share **one** `edge_id`. The graph already considers them the same edge — it just keeps both copies.

Nodes do not behave this way. `self.nodes` is keyed by ID, so repeated `add_node` is idempotent. The asymmetry is the bug: edge ingest is not idempotent, node ingest is.

## Steps to Reproduce

1. Create a `ContextGraph`
2. Add the same edge twice
3. Inspect `len(graph.edges)`, the set of `edge_id`s, and `density()`

## Expected Behavior

Re-adding an edge that already exists is a no-op, as it is for nodes. `len(edges)` stays at 1, and `density()` stays within `[0.0, 1.0]`.

Genuinely parallel edges must still be supported — and they are unaffected by a dedupe, because differing metadata or type produces a different content-derived `edge_id` (verified below).

## Actual Behavior

Every call appends another copy. `edges` grows without bound, all copies sharing a single `edge_id`, and `density()` exceeds its mathematical maximum.

## Environment

- **OS**: macOS 26.0 (arm64)
- **Python Version**: 3.12.13

## Code Example

```python

g = ContextGraph(advanced_analytics=False)
g.add_node("a", "t")
g.add_node("b", "t")

for _ in range(3):
    g.add_edge("a", "b", "rel")

print("edges:          ", len(g.edges))
print("distinct ids:   ", len({e.edge_id for e in g.edges}))
print("density:        ", g.density())
print("stats:          ", g.stats()["edge_count"])
print("get_edge_data:  ", g.get_edge_data("a", "b"))
```

Output:

```
edges:           3
distinct ids:    1
density:         1.5
stats:           3
get_edge_data:   {'id': '4d432e4f-c11e-542e-aea3-a4acbdfbdb66', 'familyId': '4d432e4f-...', 'type': 'rel', 'weight': 1.0}
```

Deterministic — 10/10 identical runs. `density()` of `1.5` is impossible for a simple digraph; `max_edges` is computed as `n * (n - 1)` at `context_graph.py:1429`.

### Parallel edges are not affected by a dedupe

```python
g = ContextGraph(advanced_analytics=False)
g.add_node("a", "t"); g.add_node("b", "t")
g.add_edge("a", "b", "rel", confidence=0.9)
g.add_edge("a", "b", "rel", confidence=0.5)
g.add_edge("a", "b", "other")
print(len({e.edge_id for e in g.edges}))   # 3 — all distinct
```

Three genuinely different edges get three distinct IDs, so deduping by `edge_id` preserves multigraph semantics.

## Blast Radius

**Re-ingest doubles the edge set.** Via the public ingest path:

```python
g = ContextGraph(advanced_analytics=False)
ents = [{"id": "alice", "type": "person"}, {"id": "acme", "type": "org"}]
rels = [{"source_id": "alice", "target_id": "acme", "type": "works_at"}]
for run in range(3):
    g.build_from_entities_and_relationships(ents, rels)
    print(f"run {run + 1}: edges={len(g.edges)}")
```

```
run 1: edges=1
run 2: edges=2
run 3: edges=3
```

**Save/load cycles compound it.** Restoring a backup into a graph that already holds the same data doubles the edges each cycle — 9 → 18 → 36 → 72. A periodic refresh or restore job grows the edge set geometrically.

**Downstream effects:** `stats()["edge_count"]` and `density()` are inflated; `get_edge_data` returns only the first of the duplicates; memory grows without bound on repeated ingest.

## Interaction with #852 (worth resolving before that PR merges)

Open PR #852 (`feat(context): add ContextGraph Markdown round-trip`) adds an exporter that validates edge-ID uniqueness and raises when it fails:

```python
raise ValueError(f"Cannot export ContextGraph: duplicate edge ID(s): {duplicates}.")
```

That guards the symptom, but this bug produces duplicate IDs trivially, so the guard fires on ordinary graphs. Checked out against that PR's head:

```
single edge              edges=1 -> markdown export OK
same edge added twice    edges=2 -> ValueError: duplicate edge ID(s): '4d432e4f-...'
re-ingest x2             edges=2 -> ValueError: duplicate edge ID(s): '7a45d9b2-...'
```

The third case is a plain re-ingest — two calls to `build_from_entities_and_relationships` with the same relationship list, which is what any refresh job does. Such a graph can never be exported to Markdown.

Flagging so the reviewer can decide deliberately: fix the cause here and the guard becomes unreachable, or keep the guard and accept that ordinary graphs cannot round-trip.

## Additional Context

- `semantica/context/context_graph.py:1505` — `_add_internal_edge`, the unconditional append
- `semantica/context/context_graph.py:227` — `_resolve_edge_identity`, the content-derived ID
- `semantica/context/context_graph.py:1429` — `density()`, where `max_edges = n * (n - 1)`
- #732 (closed) fixed this same class of bug in `Reasoner.add_rule`, which suggests dedupe is the expected contract
- Found while stress-testing `ContextGraph` for idempotency and concurrency

## Possible Solution

Give `_add_internal_edge` an existence check keyed on `edge_id`, mirroring how `self.nodes` already dedupes by node ID — e.g. maintain an `edge_id -> ContextEdge` index and return `False` when the ID is already present, so `edges`, `edge_type_index`, and `_adjacency` stay in step. Since identity is content-derived, this leaves parallel edges with distinct attributes untouched.

Two decisions for a maintainer:

1. Should a repeat `add_edge` be a silent no-op returning `False`, or should it update the existing edge in place? The node path currently overwrites, so the two are not consistent today either.
2. Whether to fix this before #852 merges, given the interaction above.

Happy to submit a PR once there's a decision on (1).

pravit-amp:
Fixes #922

`ContextGraph._add_internal_edge` appended every edge unconditionally, so adding the same edge twice stored two copies sharing one content-derived `edge_id`. This inflated `stats()["edge_count"]`, pushed `density()` past 1.0, and made re-ingest double the edge set on every run.

## Changes

- Added an `edge_id -> ContextEdge` index (`_edge_index`), mirroring how `self.nodes` dedupes by node ID.
- `_add_internal_edge` now returns `False` when the `edge_id` already exists, before touching `edges`, `edge_type_index`, or `_adjacency`, and before firing the mutation callback (no phantom ADD_EDGE audit events).
- The two state-reset paths (`load` and `clear()`) also clear the new index.
- Four regression tests: repeat `add_edge` is a no-op, parallel edges with distinct attributes are preserved, re-ingest via `build_from_entities_and_relationships` stays at one edge, and `clear()` resets the dedupe index.

## Notes

- This implements the **silent no-op** option from the issue discussion (option 1). Happy to switch to update-in-place if maintainers prefer.
- This also resolves the interaction with #852: ordinary re-ingested graphs no longer trip that PR's duplicate-edge-ID export guard.

## Testing

- Full `tests/context/` suite passes (484 tests).
- All repro scenarios from #922 now produce expected output (1 edge, density 0.5, re-ingest stays at 1).


KaifAhmad1 (approved):
Verified this end-to-end, not just read the diff — pulled the branch into a worktree and ran it:

- `_edge_index` dedupe check sits inside the existing `self._lock` (RLock, so no deadlock with the outer lock in `add_edge`), runs before node auto-creation and before the mutation callback fires — a repeat `add_edge` is a true no-op, no phantom `ADD_EDGE` audit event.
- Grepped every write path to `self.edges` — `_add_internal_edge` is the only append site, and all callers (bulk `add_edges`, builders, `load_from_file`, `merge`, decision-graph helpers) route through it, so the index can't go stale. Both `load_from_file()` and `clear()` also clear `_edge_index`.
- Ran all three repro scripts from #922 directly against this branch: `edges=1`/`density=0.5` for repeated `add_edge` (was 3/1.5), parallel edges with distinct attributes still produce 3 distinct IDs, and triple re-ingest via `build_from_entities_and_relationships` stays at `edges=1` across all 3 runs (was 1→2→3).
- Full `tests/context/test_context.py`: 31/31 passed, including the 4 new regression tests.
- Checked the interaction with `add_edge`'s pre-existing SKOS hierarchy cycle check — re-adding an identical `skos:broader`/`narrower` edge doesn't trip a false cycle error.

Clean fix, mirrors the existing node-dedupe pattern, good test coverage. Also resolves the interaction with #852's Markdown export guard as described.

Thanks for the thorough writeup and reproducible repro cases in the issue, @pravit-amp — made this easy to verify. Approving.

Issue #927, opened 2026-08-12
SSRF: `FeedIngestor` (RSS/Atom feed ingestion) has no protection against internal/private network targets


ZohaibHassan16:
## Summary
`semantica/ingest/feed_ingestor.py` fetches feed URLs with plain `requests.get/head()` and no SSRF protection , no scheme allowlist or priv /link local IP validation, inlcuding default redirect-following. This is despite a working SSRF guard (`semantica/ingest/ssrf.py`) already existing in the same package and being correctly used by `web_ingestor.py` and `api_ingestor.py`.

## Severity
**High**

## Affected code
- `FeedIngestor.ingest_feed()` - `semantica/ingest/feed_ingestor.py:641`
- `FeedIngestor.discover_feeds()` - `semantica/ingest/feed_ingestor.py:696, 726, 744`
- `FeedMonitor.check_updates()` - `semantica/ingest/feed_ingestor.py:488`

None of these call sites import or use `ingest/ssrf.py`'s `validate_url_for_request()` / `request_with_ssrf_guard()`. URL validation in `ingest_feed()` (lines 629-634) only checks that `urlparse(url).scheme` and `.netloc` are non-empty , it doesn't check the scheme is `http`/`https` and doesn't resolve or check the target IP at all.

## Reachability
`ingest_feed()` is a documented public entry point: `semantica/ingest/methods.py:685` (`ingest_feed(source, method="rss"/"atom"/"discover")`), directly exposed to any caller that lets a user or agent supply a feed URL , e.g. `ingest("https://example.com/feed.xml")` auto-detects and routes here (`methods.py:1390`).

## Proof of concept
Verified live against this codebase:

1. Started a local HTTP server on `127.0.0.1:8765` serving a fake RSS feed (simulating an internal-only service, e.g. stand-in for a cloud metadata endpoint or internal API).
2. Ran:
   ```python
   from semantica.ingest.feed_ingestor import FeedIngestor
   FeedIngestor().ingest_feed("http://127.0.0.1:8765/feed.xml")

PR #928, opened 2026-08-12
security: apply SSRF guard to feed ingestion requests

ZohaibHassan16:
## Summary
See #927 

## Fix
Route all outbound requests in `feed_ingestor.py` (`ingest_feed`, `discover_feeds`, `FeedMonitor.check_updates`) through `request_with_ssrf_guard()`, matching the other ingestors. Added `allow_private_ips` config option for consistency.

## Testing
- `pytest tests/ingest/test_feed_ingestor.py`: 12/12 pass.
- Fixed 2 tests that mocked the wrong request layer.


ZohaibHassan16:
Addressed Qodo findings in the latest commit.

Sameer6305 (dismissed):
Thanks @ZohaibHassan16  for the fix! I reviewed the changes against the existing SSRF implementation and verified the affected feed ingestion paths. Also ran the relevant test suite, 47/47 tests pass. The SSRF protection looks correct and I didn't find any remaining security or compatibility issues.

@KaifAhmad1 , approved from my side and ready for your final review and merge.

KaifAhmad1 (approved):
Approving — thanks both.

@ZohaibHassan16, clean fix for #927: all 5 call sites (ingest_feed, the three requests inside discover_feeds, and FeedMonitor.check_updates) now go through the existing request_with_ssrf_guard(), matching web_ingestor.py/api_ingestor.py. Also appreciated the self-caught test-mock fix — a leftover requests.get patch would've kept test_discover_feeds_empty green without exercising the real code path.

@Sameer6305, thanks for the review pass and confirming the 47/47 run.

Independently verified this one rather than just reading the diff: ran the full test_feed_ingestor.py suite (12/12) and test_ssrf_protection.py (35/35) locally, grepped for any requests.get/head/post call sites the fix might have missed (none), and reproduced the issue's own PoC against a live local server — FeedIngestor().ingest_feed("http://127.0.0.1:8765/feed.xml") now raises ValidationError instead of succeeding. Also confirmed the shared ssrf.py guard re-validates every redirect hop, so this closes the redirect-chain bypass too, not just the direct-IP case.

Added a changelog entry (43ced6f) — including one carried-over, non-blocking note: discover_feeds()'s existing except Exception: continue loops now also swallow ValidationError from a blocked candidate the same way they already swallowed network failures. Not an SSRF bypass (the request is blocked before it reaches the network), just a missed opportunity to log it distinctly. No blockers — ready to merge.

PR #929, opened 2026-08-12
fix(context): take the lock in ContextGraph.to_dict()

### LINKED ISSUE #923: bug(context): ContextGraph.to_dict() is the only reader without the lock — RuntimeError under concurrent writes
## Bug Description

`ContextGraph.to_dict()` iterates `self.nodes.values()` and `self.edges` without acquiring `self._lock`. It is the only reader on the class that does not take the lock. Calling it while any other thread mutates the graph raises `RuntimeError: dictionary changed size during iteration`.

`ContextGraph` is explicitly implemented as thread-safe — `self._lock = threading.RLock()` at `context_graph.py:452`, used at 28 other call sites in the same file. `to_dict` was simply missed.

## Steps to Reproduce

1. Build a `ContextGraph`
2. Mutate it from one thread (`add_node` in a loop)
3. Call `to_dict()` from another thread

## Expected Behavior

`to_dict()` returns a consistent snapshot, as every other reader does. `stats()`, `density()`, `find_nodes()`, `find_edges()`, `get_neighbors()`, `get_nodes_by_label()`, `state_at()`, and `save_to_file()` all hold the lock and are unaffected by concurrent writes.

## Actual Behavior

`RuntimeError: dictionary changed size during iteration`, raised from `context_graph.py:1773`.

Reproduced in **20/20 trials**, median 1.04s to failure with a 1-second race window. It is a race, so not logically deterministic, but it fires on the first attempt every time under any real write load.

## Environment

- **OS**: macOS 26.0 (arm64)
- **Python Version**: 3.12.13

## Code Example

```python
import threading, time

g = ContextGraph(advanced_analytics=False)
for i in range(200):
    g.add_node(f"s{i}", "t")

stop, err = threading.Event(), []

def writer():
    i = 0
    while not stop.is_set():
        g.add_node(f"w{i}", "t")
        i += 1

def reader():
    while not stop.is_set():
        try:
            g.to_dict()
        except Exception as e:
            err.append(e)
            stop.set()
            return

tw, tr = threading.Thread(target=writer), threading.Thread(target=reader)
tw.start(); tr.start()
time.sleep(2); stop.set()
tw.join(); tr.join()

print("error:", err[0] if err else "none")
```

## Error Messages / Logs

```
RuntimeError: dictionary changed size during iteration
  File "semantica/context/context_graph.py", line 1773, in to_dict
    for n in self.nodes.values():
```

## Additional Context

**Every sibling reader is safe.** I probed nine read methods under identical concurrent write load. Only `to_dict` fails:

| method | result |
| --- | --- |
| `stats()` | safe |
| `density()` | safe |
| `find_nodes()` | safe |
| `find_edges()` | safe |
| `get_neighbors(hops=2)` | safe |
| `get_nodes_by_label()` | safe |
| `state_at()` | safe |
| `save_to_file()` | safe |
| **`to_dict()`** | **RuntimeError** |

**How it was missed.** Commit `1d1ae398` ("fix: add thread safety and pagination to ContextGraph", closing #378) introduced the `RLock` and added 26 `with self._lock:` blocks, removing none. `to_dict` already existed at that point and was not among them.

`save_to_file` is safe only incidentally — it holds the lock and builds its own payload inline (`context_graph.py:989`) rather than delegating to `to_dict`, so it never reaches the unguarded loops.

**Affected callers.** `to_dict()` is the ordinary way to snapshot or export a graph, and snapshotting while another thread writes is precisely the scenario the lock exists for:

- `semantica/context/agent_context.py:1943` — `_capture_checkpoint_state` fallback path
- `semantica/triplet_store/triplet_store.py:184` — knowledge-graph conversion
- Directly by user code and the bundled examples

## Possible Solution

Wrap the body in the existing lock, matching every sibling reader:

```python
def to_dict(self) -> Dict[str, Any]:
    """Export graph to dictionary format."""
    with self._lock:
        ...
```

`self._lock` is an `RLock`, so this composes safely with any caller that already holds it. Happy to submit a PR with a regression test that runs the race above.

pravit-amp:
## Description

`ContextGraph.to_dict()` iterated `self.nodes.values()` and `self.edges` without holding `self._lock`, so a concurrent writer raised `RuntimeError: dictionary changed size during iteration`. It was the only reader on the class that did not take the lock — `stats()`, `density()`, `find_nodes()`, `find_edges()`, `get_neighbors()`, `get_nodes_by_label()`, `state_at()` and `save_to_file()` all hold it.

Commit `1d1ae398` introduced the `RLock` and added 26 `with self._lock:` blocks; `to_dict` already existed and was not among them. `save_to_file` is safe only incidentally — it holds the lock and builds its payload inline rather than delegating to `to_dict`, so it never reaches the unguarded loops.

Beyond the `RuntimeError`, the unguarded body could also return a torn snapshot: the `statistics` k reads `len(self.nodes)`/`len(self.edges)` *after* building the node and edge lists, so a write landing in between yields counts that contradict the payload they describe.

`self._lock` is an `RLock`, so this composes with callers that already hold it (`build_from_conversation` and `build_from_documents` both `return self.to_dict()` from inside a locked block). Neither external caller — `agent_context.py:1943` nor `triplet_store.py:184` — defines a lock of its own, so there is no ordering inversion.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #923

## Changes Made

- `semantica/context/context_graph.py` — wrapped the `to_dict()` body in `with self._lock:`, matching every sibling reader. The body iswise unchanged; the diff is large only because of the re-indent.
- `tests/context/test_context_graph_thread_safety.py` — new regression tests.

## Testing

- [x] Tested locally

Reproduced the issue's script verbatim on `main` — `RuntimeError` on the first attempt; clean across 5 runs after the fix.

Five new tests:

| test | catches |
| --- | --- |
| `test_to_dict_waits_for_the_lock` | Deterministic — holds `_lock` in the main thread, asserts a worker's `to_dict()` blocks, then completes on release. No race window needed. |
| `test_to_dict_is_reentrant_for_a_caller_holding_the_lock` | The `RLock` composition guarantee (passes both ways — a deadlock guard, not a bug detector). |
| `test_to_dict_does_not_raise_during_concurrent_writes` | The reported `RuntimeError`. |
| `test_to_dict_snapshot_is_internally_consistent` | `statistics` counts contradicting the emitted lists. |
| `test_snapshot_node_ids_are_uniqtorn read emitting the same node twice. |

Four of the five fail against the unfixed method; all five pass with it.

Suite results: `tests/context/` 485 passed. `tests/triplet_store` + `tests/kg` 787 passed, 1 failed — `test_jena_store_empty_graph.py::test_empty_graph_with_concurrent_operations`, which fails identically on unmodified `main`. Pre-existing and unrelated.

## Documentation

## Breaking Changes

**Breaking Changes**: No

## Checklist

## Additional Notes

The lock is held for the whole snapshot, so `to_dict()` on a large graph now blocks writers for its duration. That is the same contract `save_to_file()` and `state_at()` already have, and it's the point — a snapshot that can't be torn.


ZohaibHassan16 (approved):
LGTM. All the race conditions are properly covered. Thanks for fixing flaky test issues as well. Approved.

PR #932, opened 2026-08-12
security(context): harden Markdown import against TOCTOU symlink races

### LINKED ISSUE #856: security(context): harden Markdown import against TOCTOU symlink races on platforms without O_NOFOLLOW
## Description

`AgentMemory.import_data(..., format="markdown")` now rejects symbolic-link import paths and, on platforms that provide `O_NOFOLLOW`, protects the final file open against a symlink being introduced between validation and reading.

However, on platforms where `O_NOFOLLOW` is unavailable, the final open still relies on the earlier `Path.is_symlink()` validation. This leaves a potential time-of-check/time-of-use (TOCTOU) window between validating the path and opening it.

An attacker or another process with the ability to modify the imported filesystem path could potentially replace a validated regular Markdown file with a symbolic link after the validation step but before the final `open()`/read operation. The import could then follow the replacement link and read content from a different filesystem location than the path originally validated.

## Current flow

The relevant flow is approximately:

1. Check whether the supplied path is a symbolic link.
2. Validate that it exists and is a file/directory.
3. Enumerate Markdown files when importing a directory.
4. Re-check individual paths where applicable.
5. Open and read the file.

On platforms supporting `O_NOFOLLOW`, the final file descriptor open provides an additional protection against a symlink being introduced after validation.

On platforms without `O_NOFOLLOW`, the final open does not have an equivalent no-follow guarantee and therefore depends on the earlier path checks.

## Security impact

This is primarily a filesystem race-condition hardening issue.

The normal import path is unaffected when the filesystem is not concurrently modified. Exploitation requires the attacker/process to have the ability to modify or replace files within the directory being imported during the validation-to-open window.

If exploited, Markdown import could read a file through a symlink that was not present when the path was initially validated.

The impact is therefore dependent on the permissions and trust boundary of the filesystem containing the Markdown import path.

## Scope

This issue is specifically about the remaining TOCTOU window on platforms without `O_NOFOLLOW`.

The following protections introduced by #851 should remain unchanged:

- Direct symbolic-link files are rejected.
- Broken symbolic links are rejected.
- Symbolic-link directories are rejected.
- Symbolic-link Markdown entries discovered inside import directories are rejected.
- On platforms supporting `O_NOFOLLOW`, the final file open rejects a symlink introduced between validation and opening.
- The opened descriptor is verified to reference a regular file.

This issue does **not** propose reverting or weakening those protections.

## Proposed direction

Investigate a platform-appropriate way to obtain equivalent no-follow semantics on platforms where `O_NOFOLLOW` is unavailable.

The solution should preferably:

- Avoid relying solely on a path-based `is_symlink()` check before opening.
- Minimize or eliminate the validation-to-open TOCTOU window.
- Preserve the existing behavior for ordinary Markdown files and directories.
- Continue rejecting symbolic links deterministically with an actionable error.
- Avoid introducing platform-specific behavior that silently weakens the security guarantees.
- Include regression coverage for the supported platforms and explicitly document any unavoidable platform limitations.

For Windows in particular, investigate whether an equivalent handle-based or reparse-point-aware open/validation strategy can provide the required guarantee without following a substituted symbolic link/reparse point.

## Relationship to #851

This is a follow-up security-hardening concern identified while reviewing #851.

#851 addresses the direct symlink-import problem and uses `O_NOFOLLOW` where the platform provides it. This issue covers the remaining race window on platforms without that primitive.

It should be treated as an independent follow-up rather than expanding #851 beyond its current scope.

## Acceptance criteria

- [ ] Identify a supported cross-platform mechanism for preventing or detecting symlink/reparse-point substitution between validation and file opening.
- [ ] Ensure Markdown import cannot silently follow a substituted symlink on supported platforms.
- [ ] Preserve existing behavior for regular Markdown files and directories.
- [ ] Add regression tests covering the relevant race/substitution scenario where practical.
- [ ] Document any platform-specific limitations if a complete no-follow guarantee cannot be implemented everywhere.

## Priority

Security hardening / follow-up to #851.

lakshanmuruganandam:
Closes #856

### Summary of Changes
- Refactored `AgentMemory._read_markdown_path` file reading into a dedicated helper `_read_markdown_file_content`.
- Enforced low-level file descriptor opening via `os.open` with `os.O_NOFOLLOW` (on platforms supporting it) and verified file descriptor regular file mode with `os.fstat` (`stat.S_ISREG`).
- Rejects symlink substitutions between path validation and file reading.
- Added regression unit test `test_markdown_import_file_open_security_rejects_symlink` in `tests/context/test_agent_memory_markdown.py`.


Sameer6305:
@lakshanmuruganandam can you fix the qodo reviews before we review it?
and also comment on the issue you are fixing so i can assign it to you.

Sameer6305 (dismissed):
Thanks @lakshanmuruganandam for the security hardening work on this PR. I went through the implementation in detail, including the full Markdown import flow, the affected code paths, platform-specific behavior, and the regression coverage.

### Review summary

The core implementation in this PR is strong:

* `O_NOFOLLOW` is used where supported, providing kernel-level protection against symlink substitution during `os.open()`.
* `fstat()` + `S_ISREG` verifies that the opened descriptor refers to a regular file.
* Existing direct-file and directory symlink protections are preserved.
* The file-descriptor ownership/cleanup path was reviewed.
* POSIX behavior and the Windows fallback were specifically examined.

### Findings and fixes

During the review, we identified a few gaps that were worth addressing:

1. The Windows fallback does not have `O_NOFOLLOW`, leaving a platform-specific TOCTOU limitation. We verified that eliminating this completely would require platform-specific Windows APIs outside the project's current scope. We therefore documented the limitation explicitly rather than implying a stronger guarantee than the implementation provides.

2. The original security regression test exercised the private helper and could fail on Windows without the required symlink privilege handling. This was fixed with the appropriate Windows privilege skip.

3. Added public-API coverage to verify that `import_data(..., format="markdown")` actually rejects symlink paths.

4. Added directory-import coverage to verify that symlinked Markdown entries are not imported.

5. Added coverage for the `fstat()` / `S_ISREG` defense against non-regular files.

These changes were made in commit **`7d2431d`**.

The targeted Markdown/context test suite was then verified with **46 passed, 4 skipped, 0 failures**; the skips are the expected Windows symlink-privilege cases.

I also reviewed the final diff and cleaned up the temporary review/probe artifacts. No unrelated regressions or additional security issues were found.

### Final verdict

**Approved from my side.** ✅

@KaifAhmad1, the PR is ready for merge from my side after your final review.

KaifAhmad1 (approved):
Reviewed the diff directly (not just the bot summaries) and ran the test suite locally against this branch: `46 passed, 4 skipped` (skips are the expected Windows symlink-privilege cases).

The core fix is solid:
- `os.open()` with `O_NOFOLLOW` closes the TOCTOU window atomically on POSIX (ELOOP → clear `ValueError`).
- `os.fstat()` + `stat.S_ISREG()` rejects non-regular files even after a successful open.
- Directory imports now exclude symlinked entries, consistent with the single-file path.

Two non-blocking items, already acknowledged in-thread:
- Windows has no `O_NOFOLLOW`, so the `is_symlink()` pre-check is the only defense there — documented inline rather than overclaiming, matches the scope of #856.
- Directory import silently skips symlinked entries rather than raising like the single-file path does (Qodo finding #3) — a UX inconsistency, not a security gap, since the content is never read either way.

Thanks @lakshanmuruganandam for the hardening work, and @Sameer6305 for the thorough follow-up review and test coverage. Approving.

PR #933, opened 2026-08-12
perf(semantic_extract): cache spaCy pipelines instead of reloading per call

### LINKED ISSUE #889: perf(semantic_extract): extract_entities_ml and two relation extractors call spacy.load() on every invocation — ~120 ms per call, 60x the actual work

dex0shubham:
Closes #889

### DIFF
--- a/semantica/semantic_extract/methods.py
+++ b/semantica/semantic_extract/methods.py
@@ -109,6 +109,7 @@ License: MIT
 import re
 import difflib
 from concurrent.futures import ThreadPoolExecutor, as_completed
+from functools import lru_cache
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 from ..utils.exceptions import ProcessingError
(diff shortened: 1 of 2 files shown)


dex0shubham:
Closing as a duplicate of #886 — I missed that #889 was already assigned to @joseedson18jc with the fix in flight. Sorry for the noise.

PR #935, opened 2026-08-12
test(semantic_extract): guard openai-dependent tests and assert on the logger, not stdout

### LINKED ISSUE #934: test(semantic_extract): 4 failing tests in test_pr482_deepseek_openai — unguarded optional openai import, and stdout assertions against a logging implementation
## Summary

`tests/semantic_extract/test_pr482_deepseek_openai.py` has 4 failing tests caused by two independent, unrelated defects. Both are test-side bugs — `semantica/semantic_extract/providers.py` behaves correctly.

A fifth test passes vacuously and can never catch a regression.

## Defect 1 — three tests import the optional `openai` SDK with no skip guard

`openai` ships only in the `llm-openai` / `llm-deepseek` extras (`pyproject.toml:103`, `:108`). It is **not** in the `dev` extra (`pyproject.toml:218-229`), so a contributor following CONTRIBUTING (`pip install -e ".[dev]"`) does not have it.

Three unguarded imports:

- `tests/semantic_extract/test_pr482_deepseek_openai.py:177` — in `TestDeepSeekInstructorPath._make_provider`
- `:187` — `test_generate_typed_instructor_openai_isinstance_check`
- `:319` — `test_generate_typed_deepseek_uses_openai_client_for_instructor`

The failure mode changes with run order, which makes this confusing to diagnose:

- **File run alone, `openai` absent:** `ModuleNotFoundError: No module named 'openai'`
- **Full-directory run:** a sibling test injects a `MagicMock` for `openai` into `sys.modules`, so `MagicMock(spec=OpenAI)` receives a mock as its spec and raises `unittest.mock.InvalidSpecError: Cannot spec a Mock object.`

Note that `_make_provider` at `:177` has no callers — it is dead code.

The two affected tests are sound; they pass once `openai` is installed. They just need a guard so the other 17 tests in the module still run without the extra. The repo already has the idiom — `pytest.importorskip("groq")` in `tests/semantic_extract/test_llm_smoke_groq.py:12`, and `unittest.skipUnless` in `tests/integration/test_relations_groq.py:9`.

## Defect 2 — two tests assert on stdout against a logging implementation

`BaseProvider.generate_typed` emits its verbose confirmation through the logger, not `print` (`semantica/semantic_extract/providers.py:441-444`):

```python
verbose_mode = kwargs.get("verbose", False) or self.config.get("verbose", False)
if verbose_mode:
    self.logger.debug(
        "[BaseProvider.generate_typed] Typed response received via instructor (%s).",
        provider_name,
    )
```

But the tests capture `sys.stdout` and assert the string appears there:

- `:231` `test_generate_typed_verbose_true_prints` → `:249` patches `sys.stdout`, asserts `assertIn("generate_typed", ...)` — always fails
- `:285` `test_generate_typed_verbose_from_config` → `:304` same pattern — always fails

Both fail deterministically, in isolation, regardless of installed extras:

```
AssertionError: 'generate_typed' not found in ''
```

These read as tests written against an earlier `print()` that was later converted to `logger.debug` without updating them.

### Related: a test that passes for the wrong reason

`:259` `test_generate_typed_verbose_false_no_print` asserts stdout is empty. Since the implementation never writes to stdout, this assertion holds no matter what `verbose_mode` does — including if the guard broke entirely. It should assert on the logger too.

## Reproduction

```bash
pip install -e ".[dev]"
pytest tests/semantic_extract/test_pr482_deepseek_openai.py -q
```

## Expected

All tests in the module either pass or skip explicitly, and the verbose tests assert against the behaviour the implementation actually has.

## Suggested fix

1. Module-level soft import of `OpenAI` plus a `unittest.skipIf` decorator on the two tests that genuinely need the real class, so the other 17 still run. Delete the unused `_make_provider`.
2. Point the three verbose assertions at the logger (`assertLogs`, or patching `provider.logger`) instead of `sys.stdout`. Keep `logger.debug` — a library should not print to stdout.

Note `assertNoLogs` is 3.10+, and the project declares `requires-python = ">=3.8"`, so the negative case needs a different approach.

I have a branch with this implemented and the full `tests/semantic_extract/` suite green, and would like to take this on if the approach looks right.

dex0shubham:
Closes #934


dex0shubham:
Addressed the assertLogs note in a follow-up commit. Both forms are valid API, but the consistency point is fair — all 11 assertLogs call sites in the suite pass a name string, and tests/reasoning/test_reasoner.py uses this exact .logger.name form. Also rebased onto main to clear the behind-status.

ZohaibHassan16:
Looks good to me

dex0shubham:
@ZohaibHassan16 thanks for the look — could you submit that as a formal approval when you get a chance? The gate still shows REVIEW_REQUIRED. Branch is a few commits behind main now but conflict-free; happy to rebase once it's approved so the update doesn't dismiss the review.

ZohaibHassan16 (approved):
LGTM. Approving.

KaifAhmad1 (approved):
Verified locally: reproduces all 4 pre-fix failures from #934 and confirms 17 passed / 2 skipped (openai not installed) on this branch — approving for merge.

PR #937, opened 2026-08-12
fix(explorer): /api/decisions returns 422 — coerce decision timestamp to str

logan-jl-cc:
## Problem

`GET /api/decisions` returns **HTTP 422 (Unprocessable Entity)** for any graph that contains real decisions, which makes the **Decisions workspace** in the Knowledge Explorer fail to load entirely — no decision can ever be listed.

## Root cause

`ContextGraph.record_decision()` stores the decision timestamp as a POSIX **float** (e.g. `1786513069.69`):

```python
# semantica/context/context_graph.py — record_decision
metadata["timestamp"] = datetime.now().timestamp()   # float
```

But the explorer response model types it as `Optional[str]`:

```python
# semantica/explorer/schemas.py
    ...
```

`_node_to_decision()` in `semantica/explorer/routes/decisions.py` passes the raw float through:

```python
timestamp=properties.get("timestamp"),   # float, not str
```

Pydantic (strict mode) rejects the `float → str` coercion and raises a validation error, so the endpoint returns 422:

```
ValueError: 1 validation error for DecisionResponse
timestamp
  Input should be a valid string [type=string_type, input_value=1786513069.69, input_type=float]
```

This affects every endpoint that builds a `DecisionResponse` (`/api/decisions`, `/api/decisions/{id}`, `/api/decisions/{id}/precedents`).

## Fix

Coerce the timestamp to `str` at the boundary (preserving `None`), so the value matches the declared schema regardless of how the source store serializes it:

```python
timestamp=None if properties.get("timestamp") is None else str(properties.get("timestamp")),
```

Minimal, one-line, no schema change.

## Verification

Before — `GET /api/decisions` on a graph with 3 recorded decisions:
```
HTTP/1.1 422 Unprocessable Entity
{"detail":"Invalid input"}
```

After — same request:
```
HTTP/1.1 200 OK
[ {"decision_id":"…","category":"loan_underwriting","outcome":"approved","confidence":0.94,"timestamp":"1786513069.69", ...}, ... ]
```

The Decisions workspace then renders all 3 decisions, and clicking one opens its causal chain (`Causal Chain · 7 steps`) as expected.

## Notes

- Did not change `DecisionResponse.timestamp` typing (kept `Optional[str]`); the explorer frontend already consumes it as a string. Coercing at the adapter is the lowest-risk fix.
- An alternative would be to widen the field to `Optional[Union[str, float]]`, but that pushes serialization ambiguity onto every consumer; coercing at the single boundary is cleaner.

## Checklist

- [x] Fix is minimal (1 line) and scoped to the explorer route adapter
- [x] No schema or API-shape change
- [x] Verified manually against a real `ContextGraph` with recorded decisions
- [ ] Unit test for `_node_to_decision` with a float timestamp (happy to add if maintainers want)


Sameer6305:
@logan-jl-cc can handle the qodo reviews before we review it.

Sameer6305 (approved):
**Approved — great work on identifying and addressing the decision timestamp regression.** 👍

The original issue was valid: `record_decision()` stores timestamps as POSIX floats, while `DecisionResponse.timestamp` is exposed as a string. The important part of the review was tracing that value through the graph and Explorer layers rather than only checking whether the 422 disappeared.

### What was solid

* The original failure mode was correctly identified.
* The affected decision endpoints were correctly covered.
* The fix remained scoped to the Explorer boundary rather than changing graph storage or the public schema.
* The regression coverage now exercises the real `record_decision()` → graph → Explorer → HTTP path.
* The tests cover float/int timestamps, existing strings, `None`/missing values, and invalid numeric values.
* The final implementation preserves the existing ISO-8601 API contract.
* The changes remain focused with no unrelated production-code modifications.

### What was missing / what we corrected

The main issue was that the PR branch had become stale relative to `main`.

`main` already contained `DecisionResponse._normalize_timestamp`, which is the correct normalization boundary. The original route-level `str()` cast bypassed that validator and changed float timestamps into numeric strings such as `"1786513069.69"` rather than the established ISO-8601 representation.

We therefore removed that route-level cast and strengthened the regression coverage.

The correction was pushed in:

`b4073539` — `fix(explorer): preserve decision timestamp normalization`

The final verification passed:

* **31/31 decision-related tests**
* **272/272 Explorer tests**
* `git diff --check` clean

The final PR preserves the existing API behavior while adding comprehensive regression coverage.

**Thanks for the contribution — the underlying bug was real, and the investigation helped surface an important interaction with the newer timestamp normalization already on `main`. With the correction and regression coverage now in place, this is ready to merge.**

PR #939, opened 2026-08-12
fix(normalize): accept unit aliases during conversion

Mr-Neutr0n:
### Summary
- normalize unit aliases before category validation and conversion-factor lookup
- add canonical aliases for feet, yards, miles, and gallons
- cover abbreviated and plural units with regression tests

### Verification
- `uv run --no-project --with pytest --with chardet --with loguru --with tqdm --with numpy --with pandas pytest tests/normalize/test_number_normalizer.py -q`
- 7 passed

Fixes a documented API path where `kg`, `g`, and other aliases were rejected before conversion.


ZohaibHassan16 (commented):
The fix looks right.

There's also another bug this fixes as a side effect that wasn't mentioned. The category check is `from_category == to_category`, and both default to `None`. So if you pass two aliases from different categories that aren't in the canonical lists, `None == None` ends up being True and the conversion goes through when it shouldn't.

On main right now:

```python
convert_units(1, "kg", "ft")   # -> 3.28... should raise
convert_units(1, "gal", "lb")  # -> 8.34... should raise
```

With this PR applied, both correctly raise `ValidationError`. Since the units get normalized first, `"kg"` becomes `"kilogram"` and `"ft"` becomes `"foot"`, so they resolve to their actual categories and the mismatch gets caught.

Could you add one more regression test for this case?

```python
def test_convert_rejects_mismatched_categories_even_for_aliases(self):
    with self.assertRaises(ValidationError):
        self.converter.convert_units(1, "kg", "ft")
```

Right now I don't think we have a test for rejecting different categories, even with canonical unit names. Since this fix makes that check work properly again, it'd be good to have a test so it doesn't regress later.

Also checked the `foot` / `yard` / `mile` / `gallon` additions to `normalize_unit()`. They are needed. Those units already exist in `conversion_factors`, but they were missing from the normalization map, so without adding them this fix would still be incomplete for those four.

All 7 tests passed.


Once the extra regression test is added, looks good to merge.

Mr-Neutr0n:
Good catch on the None == None path. Added the regression test as you wrote it, plus the gal/lb pair from your example so both mismatch directions are locked. Both raise ValidationError now that units normalize to canonical names before the category check.

Full file: 10 passed locally.

ZohaibHassan16 (approved):
LGTM

PR #940, opened 2026-08-12
fix(normalize): validate symbol currencies

Mr-Neutr0n:
### Summary
- include every currency code exposed by the symbol parser in the supported-code list
- keep symbol parsing and `validate_currency_code()` consistent
- add a round-trip regression test for all configured currency symbols

### Verification
- `uv run --no-project --with pytest --with chardet --with loguru --with tqdm --with numpy --with pandas pytest tests/normalize/test_number_normalizer.py -q`
- 7 passed

This keeps parsed RUB, KRW, ILS, NGN, and PKR values from being rejected by the same normalizer API that produced them.

### DIFF
--- a/semantica/normalize/number_normalizer.py
+++ b/semantica/normalize/number_normalizer.py
@@ -611,13 +611,15 @@ class CurrencyNormalizer:
         # Check for currency code
         if not currency_code:
             for code in self.currency_codes:
-                if code in currency_input.upper():
+                match = re.search(
+                    rf"(?<![A-Z]){re.escape(code)}(?![A-Z])",
+                    currency_input.upper(),
+                )
+                if match:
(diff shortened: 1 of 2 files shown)


Mr-Neutr0n:
Thanks for flagging this. I confirmed the substring matcher could classify words such as ruby and wilson as currency codes after the supported-code list was expanded.

The latest commit uses alphabetic boundaries, while still accepting digit-adjacent forms such as 100RUB. Regression coverage now includes valid RUB inputs and the false-positive word cases.

Verification: the focused number-normalizer command passes 8 tests, and both touched Python files compile cleanly.

ZohaibHassan16 (approved):
Looks good. Approved.

PR #941, opened 2026-08-12
fix(kg): align GraphBuilder raw-text extraction defaults with the documented contract

### LINKED ISSUE #930: GraphBuilder extraction defaults differ from build() docstring
Hi! I noticed a mismatch between the documented defaults and the actual runtime defaults in `GraphBuilder`.

## Current Behavior

In `GraphBuilder._extract_from_text()`, raw-text extraction currently defaults to LLM-based extraction:

```python
ner_method = options.get("ner_method", "llm")
relation_method = options.get("relation_method", "llm")
triplet_method = options.get("triplet_method", "llm")
```

Source:  
https://github.com/semantica-agi/semantica/blob/main/semantica/kg/graph_builder.py#L231-L243

But the `GraphBuilder.build()` docstring says:

```text
extract_relations: Whether to extract relations from text (default: False)
ner_method: NER method to use (default: "ml")
triplet_method: Triplet extraction method (default: "pattern")
```

Source:  
https://github.com/semantica-agi/semantica/blob/main/semantica/kg/graph_builder.py#L287-L305

## Why This Is Surprising

This is surprising relative to the standalone extractor defaults:

- `NamedEntityRecognizer()` defaults to spaCy
- `NERExtractor()` defaults to `method="ml"`
- `RelationExtractor()` defaults to `method="pattern"`
- `TripletExtractor()` defaults to `method="pattern"`

So it looks like the public docs and the `GraphBuilder` runtime behavior are not aligned.

## Expected Behavior

Either:

1. `GraphBuilder` should use the documented local/pattern-based defaults, or
2. the docstring and examples should explicitly state that raw-text graph construction defaults to LLM extraction.

## Why This Matters

The current behavior may surprise users because it changes which extraction backend is used and may require extra provider configuration, API keys, and network access.

## Suggested Fix

Align `_extract_from_text()` defaults with the `build()` docstring, or update the docs/examples to make the LLM defaults explicit.

dex0shubham:
Closes #930 #944


ZohaibHassan16:
Please address the issues found by Qodo bot

dex0shubham:
Thanks @ZohaibHassan16 — went through all three Qodo findings.

**#2 spaCy load on each text** — real, and introduced by this PR. `NERExtractor.__init__` loads its spaCy model eagerly when the method includes `"ml"`, which the previous `"llm"` default never reached, so a multi-document build would have reloaded the model once per source. Extractors are now cached per `(kind, method)` on the builder via `GraphBuilder._get_extractor()`. Qodo has since marked this one Resolved. Verified the new tests fail without the cache.

**#3 Mocks hide extractor interactions** — fair, and addressed in the same commit. Added `TestGraphBuilderDefaultsRunOffline`, which runs the real default path with no extractor mocks and asserts `create_provider` is never called. That's the test that would have caught #2.

**#1 Triplets ignore relation_method** — the behaviour is real (`TripletExtractor` re-extracts relations when `relations is None`), but it's pre-existing rather than introduced here: `main` has identical code and this PR doesn't touch that data flow. Fixing it means changing how relations propagate between extraction stages, which is wider than aligning defaults. Happy to open a separate issue and take it there — say the word and I will, or I can fold it into this PR if you'd rather have it in one go.

`kg` suite: 466 passed. flake8 on `graph_builder.py` unchanged from main (106 findings, all pre-existing).

ZohaibHassan16:
Hi @dex0shubham , I found one issue with the extractor cache. 

`NERExtractor`, `RelationExtractor` and `TripletExtractor` support a list of methods for fallback, for example:

```
ner_method=["pattern", "ml"]
```

Before this PR, GraphBuilder passed the list directly to the extractor and it worked but now `_get_extractor()` uses `method` inside the cache key:

```
key = (kind, method)

```

Since a list cannot be use as a dict key, this throws type error.


You can reproduce it with:

```
GraphBuilder(
    merge_entities=False,
    resolve_conflicts=False,
)._extract_from_text(
    "x",
    [],
    [],
    ner_method=["pattern", "ml"],
    extract_triplets=False,
)
```

Can you convert list methods to a tuple only for the cache key , also please add a test for fallback method lists.

dex0shubham:
Good catch @ZohaibHassan16 — reproduced your snippet exactly (`TypeError: unhashable type: 'list'`), and it's a clean regression from the extractor cache I added in the previous round. All three extractors take `method` as a list for fallback ordering, and the cache used it as a dict key directly.

Fixed as you suggested: the key now uses `tuple(method) if isinstance(method, list) else method`, and the extractor still receives the original list.

Tests added for fallback lists:
- your exact repro (`ner_method=["pattern", "ml"]`)
- list methods on all three extractors at once
- equal lists reuse a single cached extractor
- different orderings (`["pattern", "ml"]` vs `["ml", "pattern"]`) get separate extractors

**One extra worth calling out**, since it showed up while running your repro: it also logged `Entity extraction failed: 'GraphBuilder' object has no attribute '_extraction_stats'`. That one is pre-existing, not from this PR — `_extraction_stats` is only initialised inside `build()`, so calling `_extract_from_text()` directly raises an `AttributeError` that the extraction path's broad `except` swallows and reports as a failed extraction. I seeded it in `__init__` too, so your repro runs clean and a genuine `AttributeError` in that block can't be silently mislabelled. `build()` still resets it per run. Happy to pull that back out if you'd rather keep this PR strictly to the defaults change.

`kg` suite: 470 passed. flake8 on `graph_builder.py` still matches main (106 findings, all pre-existing).

ZohaibHassan16:
Yes, thank you. We can open separate issue for  remaining Qodo relation point

ZohaibHassan16 (dismissed):
Looks good to me. All relevant findings addressed. Approved.

KaifAhmad1:
@dex0shubham @ZohaibHassan16 @Sameer6305 let's fold this into #941 instead of tracking it separately as #944 — this PR already has two unresolved review threads blocking merge, so no reason to split the fix across two PRs.

@dex0shubham could you add the fix here: in `_extract_from_text()`, initialize `relations = None` before the `extract_relations` block, keep the extracted list when `extract_relations=True` succeeds, and pass `relations=relations` into `trip_extractor.extract_triplets(...)` — right now only `entities=` is forwarded, so `TripletExtractor` silently re-derives relations via `triplet_method` instead of using the ones already extracted with `relation_method`.

Once that's in, please resolve both open Qodo threads on this PR. I'll close #944 as folded into this PR once it lands.

dex0shubham:
Folded in as requested @KaifAhmad1 — relations is now initialized to None, holds the extracted list when extract_relations=True succeeds, and is forwarded as relations= into extract_triplets(). Disabled or failed extraction still passes None, so TripletExtractor keeps its current self-derivation on the default path. Three tests cover those states and all fail without the fix. kg suite: 473 passed. Resolving both Qodo threads now — #944 is safe to close as folded here.

KaifAhmad1:
@ZohaibHassan16 @Sameer6305 Ready for the review Guys

Sameer6305 (dismissed):
Thanks @dex0shubham for addressing the review findings and tightening up the implementation.

I re-reviewed the complete PR, including the related extraction flow, caching behavior, fallback methods, relation forwarding, backward compatibility, and the affected test coverage. The fixes for #930 and #944 are working as expected, and the kg suite passes **473/473**.

While reviewing, I also found a few unrelated/pre-existing issues, including some weak test assertions and deprecation warnings in provenance-related code. These are not introduced by this PR, so I’ll raise separate issues for them rather than expanding the scope here.

Overall, the PR looks good to me and is **approved from my side**.

@KaifAhmad1, please take a final look and proceed with the merge if everything looks good.

dex0shubham:
Heads up @Sameer6305 @KaifAhmad1 — #878 merged and conflicted with this PR, so I rebased and the force-push dismissed the approval. Sorry for the churn. The conflict was real rather than mechanical: #878 rewrote the build() docstring and documented the LLM defaults as the contract, so the two PRs resolved the same mismatch in opposite directions. I kept #878's formatting and corrected the values per the decision on #930, and added a changelog line recording that. kg suite: 473 passed. Ready for another look.

KaifAhmad1 (approved):
Verified the conflict resolution directly against `4b44ad5` (pulled the branch, ran the suite locally rather than trusting the UI state):

- The `#878`/`#930` clash in `build()`'s docstring is resolved correctly, not just textually — `ner_method`/`relation_method`/`triplet_method`/`extract_relations` now read `"ml"`/`"pattern"`/`"pattern"`/`False` in the Google-style block `#878` introduced, so the docs no longer contradict the code this PR changes. The added changelog note explaining the `#878`/`#930` interaction is accurate.
- `CHANGELOG.md` keeps both entries (`#878` and this one), no content lost.
- `python -m pytest tests/kg/` on the exact head commit: **473 passed**, matching what's claimed in the PR description.
- Extractor cache, list-method-to-tuple key conversion, relation forwarding into `extract_triplets`, and `_extraction_stats` seeding all check out against the actual extractor source, not just the diff.

Not waiting on a re-review from `@Sameer6305` for this — his approval covered the same code (dismissed only because the conflict-resolution push post-dated it), and I've independently re-verified the resulting state myself.

Approving. The two in-progress checks (MSDO, security-scan) are generic repo-wide scans unrelated to this diff — merge once they're green.

KaifAhmad1:
@dex0shubham thanks for flagging the rebase up front instead of letting it land quietly — no need to apologize for the churn, that's exactly the right way to handle a real conflict like this one.

Wanted to close the loop for everyone here, since this PR went through a lot of rounds:

- **@dex0shubham** — appreciated the persistence through three separate review passes (the Qodo findings, the fallback-list cache regression, the #944 relations-forwarding fold-in, and now the #878 conflict), and the transparency at each step about what was pre-existing vs. introduced.
- **@ZohaibHassan16** — the fallback-list cache catch (`ner_method=["pattern", "ml"]` → `TypeError: unhashable type: 'list'`) was a real regression with a clean repro. Good catch.
- **@Sameer6305** — thanks for the thorough independent re-review of the full extraction flow, caching, and backward compatibility, and for routing the unrelated provenance findings to a separate issue instead of scope-creeping this one.

I independently verified the #878 conflict resolution myself against the exact head commit (`4b44ad5`) rather than trusting the UI: the `build()` docstring now correctly reads `"ml"`/`"pattern"`/`"pattern"`/`False`, `CHANGELOG.md` keeps both entries, and `pytest tests/kg/` gives 473 passed on that commit. Approval stands — not waiting on a re-review from Sameer since his covered the same code and was only dismissed by the conflict-resolution push, and I've re-verified the result myself.

Merging once the two in-progress checks (MSDO, security-scan — generic repo-wide scans, unrelated to this diff) go green.

PR #942, opened 2026-08-12
fix seed SSRF

ZohaibHassan16:
## Summary

Fixed SSRF issue in `SeedDataManager.load_from_api()`.

Previously it was directly using `requests.get()` with the user provided API URL. Now it uses the existing `request_with_ssrf_guard()` helper.

## Changes

- Replaced direct `requests.get()` call with `request_with_ssrf_guard()`
- Blocks localhost, private IPs, link-local IPs and cloud metadata endpoints
- Redirects are checked too, so a public URL cannot redirect to an internal URL
- Updated API loading test to mock the SSRF helper
- Added tests for:
  - Direct loopback URL blocking
  - Redirect to cloud metadata endpoint blocking

## Testing

```bash
pytest tests/test_seed_manager.py tests/ingest/test_ssrf_protection.py -q


ZohaibHassan16:
Qodo bot's finding is irrelevant to this PR and that is intentional for this PR.

`allow_private_ips` needs its own API/config design and documentation. I think it is better handled in a separate follow-up issue instead of expanding this security fix.

Sameer6305 (dismissed):
Verified the fix end-to-end against `upstream/main`. The original SSRF vulnerability is confirmed, and the change to `request_with_ssrf_guard()` correctly closes the affected path.

I also verified the redirect protection, DNS/IP validation, blocked internal addresses, and relevant error handling. The focused tests and existing SSRF coverage pass successfully, with no blocking regressions found.

I found a few unrelated pre-existing issues during the review as well:

* `Authorization` headers can be forwarded across cross-origin redirects.
* The caller-provided `headers` dictionary can be mutated when adding `Authorization`.
* There is a now-unnecessary `except (ImportError, OSError)` around the old lazy `requests` import.

These are outside the scope of this PR, so I’ll file separate issues for them.

The PR is correct and **approved from my side**.

KaifAhmad1:
@Sameer6305 @ZohaibHassan16 Fix conflicts.

ZohaibHassan16:
Cleaned the PR @KaifAhmad1

Issue #943, opened 2026-08-12
Support explicit private network API sources for trusted seed deployments.


ZohaibHassan16:
## Summary

`SeedDataManager.load_from_api()` now uses SSRF protection and blocks private , loopback, link local and cloud metadata addresses by default.

Some trusted internal deployments may need to load seed data from an internal API. Add documented opt in for this use case without weakening the secure default.

## Proposed change

Add `allow_private_ips` support to `SeedDataManager`.

- Default to False.
- Read the option from `SeedDataManager` config
- Parse the value using the shared `parse_bool()` helper.
- Pass it to `request_with_ssrf_guard(...., allow_private_ips=...)`
- Document that it should be enabled only for the trusted internal deployments.

## Acceptance criteria

- Private targets remain blocked by default.
- `allow_private_ips=True` allows trusted private/loopback/link-local API targets.
- Invalid string values are rejected.
- Redirect handling remains enabled.
- Tests cover the default block and explicit opt-in behavior.
- Seed API documentation includes a security warning.

Note that it should be taken on after #942 is merged.

yunaremaia:
Implemented in the PR above — delegates to the shared SSRF guard with the documented opt-in.

ZohaibHassan16:
Please ask for the issue to be assigned before opening the PR for it next time @yunaremaia . Someone else (or I) can be working on it

yunaremaia:
You are right, my apologies — I should have asked for the assignment before opening the PR. The intent behind #959 was to implement this exact proposal (the `allow_private_ips` opt-in for trusted internal sources, delegated to the shared SSRF guard), and I would like to keep working on it if that is still available.

Could you assign #943 to me? If someone else is already working on it or you would prefer to handle it yourself, just let me know and I will close #959 to keep the queue clean.

For future contributions I will follow the assignment-first workflow.

ZohaibHassan16:
Thank you. I will

PR #950, opened 2026-08-12
fix(seed): guard SeedDataManager.load_from_api against SSRF

### LINKED ISSUE #936: Guard SeedDataManager.load_from_api() against SSRF
## Summary

`SeedDataManager.load_from_api()` fetches  a caller provided URL using `requests.get()` without validating the URL, res add or redirects. This allows callers that pass untrusted API source config to cause outbound requests to loopback or link local endpoints.

## Relevant Code

- semantica/seed/seed_manager.py
- SeedDataManager.load_from_api()
- Request sink: approx 495

```
response = requests.get(full_url, headers=request_headers, timeout=30)
```

## Effect

If a user can control API URL, they could make the server sen requests to services that are normally only available inside server's network.

For example, they may be able to request:

- localhost
- Private IP addresses such as 10.xxxx or 192.168.x.x
- Link local addresses
- Cloud metadata endpoints
- Inteneral URLs reached through a redirect from a public website

At the moment, this method is not exposted through a public API route in repo. So not directly exploitable rn.

Still this should be fixed because any future feature that lets users configure seed API sources could expose it. 

## Fix

We already have SSRF protection in `semantica.ingest.ssrf` so we should use that here instead of calling requests directly.

```

from ..ingest.ssrf import request_with_ssrf_guard

repsonse = request_with_ssrf_guard(

        "Get",
         full_url,
         headers=request_headers,
         timeout=30
)
```

This will make sure that the URL is valid before sending the request. It also checks redirects so a normal public URL cannot redirect the request to an internal IP.

If someone needs to load data from internal API, that should be an explicit option and clearly documented.

yunaremaia:
Closes #936

## Summary

Guards `SeedDataManager.load_from_api()` against SSRF: callers passing untrusted API URLs could previously trigger outbound requests to loopback or link-local endpoints (e.g. cloud metadata services).

**`is_safe_api_url()`** (new, in `semantica/seed/seed_manager.py`):
- Resolves the hostname via `socket.getaddrinfo` and checks every returned address against `ipaddress` flags: rejects loopback, private, link-local, reserved, multicast, and unspecified (IPv4 and IPv6, including `[::1]`)
- Rejects non-`http(s)` schemes (e.g. `file://`)
- Returns `False` for unresolvable hostnames instead of raising

`load_from_api()` now raises `ProcessingError` with a clear message before any request is made when the target is unsafe.

**Tests** — 11 new cases in `tests/seed/test_seed_manager.py`:
- Blocked: `127.0.0.1`, `localhost`, `192.168.x`, `169.254.169.254` (metadata), `10.x`, `[::1]`, `file://`
- Allowed: public IPs (`8.8.8.8`)
- Integration: `load_from_api("http://127.0.0.1...")` raises `ProcessingError`

Also updated the existing `test_load_from_api` (used the non-resolving `api.example.com`, which the guard now rejects) to a public IP to preserve the test's original intent.

## Note

- Pre-existing failure: `TestSeedDataManager::test_load_from_csv` fails on `main` (0 records parsed from the mocked CSV) — unrelated to this change, observed before this branch.
- DNS rebinding (TOCTOU between validation and connect) is noted as a limitation in the guard's docstring; full mitigation requires connecting through a validating proxy or re-checking the resolved IP at connect time.

### DIFF
--- a/semantica/seed/seed_manager.py
+++ b/semantica/seed/seed_manager.py
@@ -32,17 +32,59 @@ License: MIT
 """
 
 import csv
+import ipaddress
 import json
+import socket
 from dataclasses import dataclass, field
 from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Union
+from urllib.parse import urlparse
 
 from ..utils.exceptions import ProcessingError, ValidationError
 from ..utils.helpers import read_json_file, write_json_file
 from ..utils.logging import get_logger
 from ..utils.progress_tracker import get_progress_tracker
 from ..utils.types import EntityDict, RelationshipDict
+
+
+def is_safe_api_url(api_url: str) -> bool:
+    """SSRF guard: reject URLs that resolve to non-public addresses.
+
+    Blocks loopback, private, link-local, and IPv6 loopback ranges (issue #936).
+    Raises no exception — returns False for unsafe or unparseable URLs.
+    """
+    try:
+        parsed = urlparse(api_url)
+        if parsed.scheme not in ("http", "https"):
+            return False
+        hostname = parsed.hostname
+        if not hostname:
+            return False
+        # Resolve hostname (DNS rebinding is outside this check's scope; the
+        # connection itself is not re-validated after resolution).
+        try:
+            infos = socket.getaddrinfo(hostname, None)
+            addresses = {info[4][0] for info in infos}
+        except socket.gaierror:
+            return False
+        for addr in addresses:
+            try:
+                ip = ipaddress.ip_address(str(addr).split("%")[0])
+            except ValueError:
+                continue
+            if (
+                ip.is_loopback
+                or ip.is_private
+                or ip.is_link_local
+                or ip.is_reserved
+                or ip.is_multicast
+                or ip.is_unspecified
+            ):
+                return False
+        return True
+    except (ValueError, OSError):
+        return False
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


ZohaibHassan16:
Thanks for the fix. I found a couple of security issues that need to be fixed before merging.

### Redirects are still not protected

`load_from_api()` only checks the first URL, then it does:

```python
requests.get(full_url, headers=request_headers, timeout=30)
```

`requests` follows redirects by default. This means a safe public URL can pass `is_safe_api_url()` first, then redirect to `127.0.0.1`, a private IP, or a cloud metadata endpoint.

We already have `request_with_ssrf_guard()` in `semantica.ingest.ssrf`. It checks every redirect too. We should use that existing helper here instead of adding a separate guard.

### Non-global IP ranges are still allowed

The new guard blocks some internal ranges but it does not require the IP to be globally routable.

For example, this currently returns `True`:

```
is_safe_api_url("http://100.64.0.1/")
```

`100.64.0.0/10` is shared address space and should not be allowed for a public-only SSRF policy.

Using the existing shared SSRF helper should fix this too and keep the same SSRF rules across the project.

Please also add tests for a public URL that redirects to a loopback or metadata URL, and make sure the redirected request is never made.

Also, the error message includes the full rejected URL. It can contain tokens or query parameters, so it would be better not to include the full URL in the exception message.

yunaremaia:
Closing this PR in favor of #959, which resolves #936 with the repo's shared SSRF guard.

Since opening this, `semantica/ingest/ssrf.py` landed upstream (in #906, fixing #867) — a complete shared guard with redirect-chain validation and bounded DNS resolution. PR #959 plugs `SeedDataManager.load_from_api` into that shared guard (`request_with_ssrf_guard`), which blocks loopback/private/link-local by default exactly as this PR did, and adds the `allow_private_ips` opt-in requested in #943.

Consolidating on the shared guard avoids two parallel SSRF implementations in the codebase. The tests from this PR's intent are covered in #959 (block-by-default + opt-in). Thank you for the issue — it drove the right fix.

PR #951, opened 2026-08-12
fix(triplet_store): stop CONSTRUCT detection matching inside a leading comment

### LINKED ISSUE #931: bug(triplet_store): CONSTRUCT_QUERY_RE misclassifies SELECT/ASK queries whose leading comment contains the word CONSTRUCT — affects all four SPARQL backends
## Bug Description

`CONSTRUCT_QUERY_RE` (`semantica/triplet_store/sparql_escaping.py:53`) is meant to identify a query whose *form* is CONSTRUCT, and its docstring states the intent explicitly:

> This prevents false-positives from SELECT/ASK queries that merely contain the word "CONSTRUCT" inside a string literal or comment (e.g. a literal value of `"please CONSTRUCT this"` or a comment line).

The string-literal case works. The **comment** case does not. Any SELECT or ASK query whose *leading* comment contains the word `CONSTRUCT` is misidentified as a CONSTRUCT query.

The cause is regex backtracking. The comment-skipping alternative is:

```
| \#[^\n]*             # comments (until newline)
```

`[^\n]*` is greedy but backtracks. Given `# CONSTRUCT in a comment\nSELECT ...`, the engine consumes `#`, then gives back everything after it, then `\s*` matches the space, and `CONSTRUCT` — the one *inside the comment* — satisfies the query-form keyword. The newline is never reached. The matched text is literally `'# CONSTRUCT'`.

## Steps to Reproduce

1. Take any SELECT or ASK query
2. Put a comment on the first line containing the word `CONSTRUCT`
3. Pass it to any store's `_is_construct_query`, or to `execute_sparql` without an explicit `result_format`

## Expected Behavior

Detection depends only on the SPARQL query form. A comment mentioning `CONSTRUCT` is inert, exactly as a string literal containing `CONSTRUCT` already is.

## Actual Behavior

The query is treated as CONSTRUCT. In `execute_sparql`, that selects the CONSTRUCT branch, which sends `Accept: text/turtle` and parses the response body as Turtle. A SELECT response is not Turtle, so the call fails with a message that points at the wrong thing:

```
ProcessingError: Failed to parse CONSTRUCT response as Turtle: <parse error>
```

Raised at `blazegraph_store.py:213`, `rdf4j_store.py:271`, `anzo_store.py:226`.

## Environment

- **OS**: macOS 26.0 (arm64)
- **Python Version**: 3.12.13

## Code Example

No server needed to reproduce the misclassification:

```python
from semantica.triplet_store.sparql_escaping import CONSTRUCT_QUERY_RE

cases = [
    ("# CONSTRUCT in a comment\nSELECT * WHERE { }",   False),
    ("#CONSTRUCT\nSELECT * WHERE { }",                 False),
    ("# we will CONSTRUCT later\nSELECT * WHERE { }",  False),
    ("#\n# CONSTRUCT\nSELECT * WHERE { }",             False),
    ("# TODO: CONSTRUCT\nASK { ?s ?p ?o }",            False),
]
for query, expected in cases:
    got = CONSTRUCT_QUERY_RE.search(query) is not None
    print(f"{'BUG ' if got != expected else 'ok  '} expected={expected} got={got}  {query!r}")
```

All five report `expected=False got=True`.

It reaches the public detection method of every backend (each delegates to this regex — `blazegraph_store.py:134`, `rdf4j_store.py:116`, `anzo_store.py:152`, `jena_store.py:81`):

```python
q = "# CONSTRUCT in a comment\nSELECT * WHERE { }"

from semantica.triplet_store.blazegraph_store import BlazegraphStore
from semantica.triplet_store.rdf4j_store import RDF4JStore
from semantica.triplet_store.anzo_store import AnzoStore
from semantica.triplet_store.jena_store import JenaStore

for cls in (BlazegraphStore, RDF4JStore, AnzoStore, JenaStore):
    print(cls.__name__, cls._is_construct_query(None, q))
```

```
BlazegraphStore True
RDF4JStore True
AnzoStore True
JenaStore True
```

## Error Messages / Logs

The regex match itself shows the mechanism — note the match stops inside the comment:

```
>>> CONSTRUCT_QUERY_RE.search("# CONSTRUCT in a comment\nSELECT * WHERE { }").group(0)
'# CONSTRUCT'
```

Downstream, `execute_sparql` takes the CONSTRUCT branch and fails at the Turtle parse with `ProcessingError: Failed to parse CONSTRUCT response as Turtle: ...`. That path needs a live endpoint, so it is reasoned from the code rather than captured here; the misclassification above is directly reproducible offline.

## Additional Context

Correctly handled today, for contrast — these are not affected:

| query | detected as CONSTRUCT | correct |
| --- | --- | --- |
| `CONSTRUCT { } WHERE { }` | yes | yes |
| `PREFIX e: <http://e/>\nCONSTRUCT { } WHERE { }` | yes | yes |
| `# a comment\nCONSTRUCT { } WHERE { }` | yes | yes |
| `SELECT * WHERE { ?s ?p "please CONSTRUCT this" }` | no | yes |
| `SELECT * WHERE { } # CONSTRUCT` | no | yes |
| `# CONSTRUCTION notes\nSELECT * WHERE { }` | no | yes |

The trailing-comment case is safe only because `\A` anchors the pattern — it never reaches a comment that isn't at the start. Only *leading* comments are affected.

Leading comments are common in hand-written and generated SPARQL (headers, licences, TODOs), and a comment mentioning CONSTRUCT is unremarkable — e.g. `# CONSTRUCT variant handled elsewhere`.

- `semantica/triplet_store/sparql_escaping.py:53` — the regex
- `semantica/triplet_store/sparql_escaping.py:58` — the backtracking comment alternative
- #754, #322 — prior work on CONSTRUCT template support, where this regex was centralised
- Found while fuzzing the escaping module with `rdflib`'s SPARQL and Turtle parsers as oracles. `escape_literal` itself passed every payload I threw at it (28 injection attempts, exact round-trip through both parsers) — this is the one defect I found in the module.

## Possible Solution

Stop the comment alternative from backtracking, so a comment must be consumed to end of line. Either make it atomic:

```
| (?>\#[^\n]*)          # comments (until newline) — atomic, no backtracking
```

(Python 3.11+ supports atomic groups; the project supports 3.8+, so this needs a guard.)

Or require the newline, which is portable:

```
| \#[^\n]*(?:\n|\Z)     # comments (until newline or end of input)
```

The `(?:\n|\Z)` form keeps a trailing comment at end-of-input working. Either fix leaves all six currently-correct cases above unchanged.

A regression test over the table in Additional Context would pin both directions. Happy to submit a PR.

pravit-amp:
## Description

`CONSTRUCT_QUERY_RE` skipped comments with a bare `\#[^\n]*`. The trailing `*` backtracks, so for `# CONSTRUCT ...\nSELECT ...` the engine gave back everything after the `#` and the CONSTRUCT *inside the comment* satisfied the query-form keyword — the match ends at `'# CONSTRUCT'`, never reaching the newline.

All four SPARQL backends delegate to this regex, so an affected SELECT/ASK took the CONSTRUCT branch of `execute_sparql`, which sends `Accept: text/turtle` and parses the body as Turtle, failing with a misleading `Failed to parse CONSTRUCT response as Turtle`.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #931

## Changes Made

- `semantica/triplet_store/sparql_escaping.py` — require a comment to reach a line terminator: `\#[^\n\r]*(?:[\n\r]|\Z)`. This makes backtracking structurally impossible: if the character class gives a character back, the next character is by definition not a terminator, so the group cannot match.
- `tests/triplet_store/test_construct_query_detection.py` — 8 tests / 139 subtests over two case tables (12 must-detect, 15 must-not), run against the regex directly **and** against `_is_construct_query` on all four backends.

## Testing

- [x] Tested locally

Verified the regression guard actually fails: reverting the regex to its previous form produces 42 failures across all four backends; restoring the fix gives 8 passed / 139 subtests.

### Test Commands

```bash
pytest tests/triplet_store/test_construct_query_detection.py -q
```

```bash
pytest tests/triplet_store/ -q
```

Results: `8 passed, 139 subtests passed` and `338 passed, 1 skipped` plus one pre-existing failure (see Additional Notes).

## Documentation

The regex is annotated inline explaining why the terminator is required, so the constraint is not silently re-simplified later.

## Breaking Changes

**Breaking Changes**: No

Queries previously detected as CONSTRUCT still are — the 12 must-detect cases cover bare, lowercase, mixed-case, `PREFIX`/`BASE` preambles, and comment-then-CONSTRUCT forms. Only the misdetected leading-comment cases change behaviour.

## Checklist

## Additional Notes

**Why not LF-only.** The obvious `\#[^\n]*(?:\n|\Z)` fixes the false positives but introduces a false *negative*: `# c\rCONSTRUCT {...}` is a valid CONSTRUCT query, and with LF-only terminators the comment swallows the rest of the input so the query form is never found. Both LF and CR are accepted because the SPARQL grammar ends a comment at either. `test_carriage_return_terminates_a_comment` pins this in both directions.

**Why not an atomic group.** `(?>\#[^\n]*)` and the possessive `\#[^\n]*+` also fix the false positives, but both are Python 3.11+ and raise `re.error` at import on 3.8–3.10 — and since `CONSTRUCT_QUERY_RE` is compiled at module scope, that would be an import-time crash. CI pins 3.11/3.12 while `pyproject.toml` declares `>=3.8`, so the regression would go green here. The portable form sidesteps the question; happy to switch if you'd rather raise the floor.

**Pre-existing test failure, unrelated to this PR.** `tests/triplet_store/test_jena_store_empty_graph.py::TestJenaStoreEmptyGraph::test_empty_graph_with_concurrent_operations` fails on this branch, but it also fails deterministically (3/3 runs) on clean `main` with these changes stashed — `assert len(store.graph) == 5` gets `2`, i.e. concurrent writes to `JenaStore` losing triples. Happy to open a separate issue for it.


ZohaibHassan16 (approved):
Approving this. I also checked the bug and the fix myself instead of only going by the PR description.

I was able to reproduce the false positive on `main` with:

`"# CONSTRUCT is not what this does\nSELECT ?s WHERE { ?s ?p ?o }"`

That incorrectly matches `CONSTRUCT_QUERY_RE` on `main`. From what I can see, `[^\n]*` is able to backtrack and eventually match `CONSTRUCT` from inside the comment. With the regex change in this PR, it no longer matches, which is the expected behavior.


I couldn't reproduce the mentioned existing failure in `test_empty_graph_with_concurrent_operations`. It passed 3/3 for me on clean `main`. Since that test is doing concurrent writes from 5 threads against an in-memory `JenaStore`, it could just be timing or environment related. Either way, it doesn't look related to this PR.

LGTM.

Issue #953, opened 2026-08-12
[BUG] export_yaml silently writes an empty export when the input dict lacks its expected keys — no warning, logs report success


pravit-amp:
## Bug Description

`SemanticNetworkYAMLExporter.export_semantic_network` reads only three keys off the input mapping — `entities`, `relationships`, `triplets` (`yaml_exporter.py:126-131`) — each via `.get(key, [])`. Any dict that doesn't use those exact key names produces a structurally valid YAML file with every collection empty.

Nothing signals the loss: no exception, no warning, `None` returned, and the progress log reports `[COMPLETED] Exported semantic network to YAML` followed by `Exported YAML to: <path>`. The caller's only clue is opening the file.

`method="schema"` has the same shape of defect via a different key set, emitting empty `classes`/`properties`/`namespaces`.

## Steps to Reproduce

1. Build a dict of records using any key names other than `entities`/`relationships`/`triplets`
2. Call `export_yaml(data, path)`
3. The call succeeds; the written file contains no records

## Expected Behavior

Either the records are exported, or the caller is told they were not — a `ValidationError` naming the expected keys, or at minimum a logged warning that no recognized key was found.

Silently writing an empty export while logging success is the one outcome that gives the caller no way to detect the problem.

## Actual Behavior

A valid YAML file is written containing only metadata and empty collections. Exit is clean and the logs report success.

## Environment

- **OS**: macOS 26.0 (arm64)
- **Python Version**: 3.12.13
- **Semantica Version**: 0.6.5 (`main` @ `1a3dd503`)
- **Installation Method**: source (editable)

## Code Example

```python
import os
import tempfile

from semantica.export.methods import export_json, export_yaml

rows = [{"id": "1", "name": "Acme"}, {"id": "2", "name": "Beta"}]

d = tempfile.mkdtemp()
json_path, yaml_path = os.path.join(d, "o.json"), os.path.join(d, "o.yaml")

# What export_json produces:
export_json(rows, json_path)
print(open(json_path).read())
# {"data": [...2 records...], "count": 2, "metadata": {...}}

# Feed that same shape to export_yaml:
export_yaml({"data": rows}, yaml_path)
print(open(yaml_path).read())
```

```yaml
metadata:
  exported_at: '2026-08-12T11:31:55.828342'
  version: '1.0'
entities: []
relationships: []
triplets: []
```

Both records are gone. The same happens for every unrecognized shape I tried:

| input | result |
| --- | --- |
[...]

Issue #859, opened 2026-08-09
test(visualization): test suite fails non-deterministically in full-suite order due to sys.modules pollution across test files


Sameer6305:
## Summary

Eight tests in `tests/visualization/` fail when the full visualization suite is collected and run together by pytest, but every one of them passes when run in isolation (its own file). This is a test infrastructure problem — `AnalyticsVisualizer`, `EmbeddingVisualizer`, `KGVisualizer`, and the other visualizers are not broken. The failures are caused by module-level `sys.modules` mutation that persists across test files and corrupts the mock state that later test files depend on.

## Failing Tests (only when run as `pytest tests/visualization/`)

tests/visualization/test_optional_dependencies.py::TestOptionalDependencies::test_analytics_visualizer_without_plotly
tests/visualization/test_optional_dependencies.py::TestOptionalDependencies::test_semantic_network_visualizer_without_plotly
tests/visualization/test_optional_dependencies.py::TestOptionalDependencies::test_temporal_visualizer_without_plotly
tests/visualization/test_visualization.py::TestVisualization::test_kg_visualizer_layouts_init
tests/visualization/test_visualization_advanced.py::TestVisualizationAdvanced::test_visualize_2d_projection
tests/visualization/test_visualization_advanced.py::TestVisualizationAdvanced::test_visualize_centrality_rankings
tests/visualization/test_visualization_advanced.py::TestVisualizationAdvanced::test_visualize_similarity_heatmap
tests/visualization/test_visualization_comprehensive.py::TestVisualizationComprehensive::test_ontology_visualizer

All 8 pass individually:

```bash
pytest tests/visualization/test_visualization_advanced.py        # 6 passed
pytest tests/visualization/test_optional_dependencies.py         # 5 passed
pytest tests/visualization/test_visualization_comprehensive.py   # all passed
pytest tests/visualization/test_visualization.py                 # all passed
```

## Root Cause

Multiple test files inject mocks at **module level** (top-level `sys.modules` assignments outside any test class or fixture), rather than inside a scoped pytest fixture or `unittest.TestCase.setUp`. For example, `test_visualization_advanced.py` and `test_visualization_comprehensive.py` both begin with:

```python
sys.modules['plotly'] = MagicMock()
sys.modules['plotly.graph_objects'] = MagicMock()
# etc.
```

These assignments run at **import time** during pytest collection. When pytest collects the full directory, it imports all test files before running any of them. The order of collection (alphabetical) means:

[...]

PR #954, opened 2026-08-12
fix(export): reject non-mapping input to the YAML exporters

### LINKED ISSUE #952: [BUG] export_yaml raises AttributeError on List[Dict], a type its own signature declares as supported
## Bug Description

`export_yaml`'s signature declares `data: Union[Dict[str, Any], List[Dict[str, Any]]]`, but passing a list raises `AttributeError`. `YAMLExporter.export_semantic_network` calls `.get()` on the argument, which only works for a mapping.

Both dispatch methods are affected — `method="semantic_network"` (the default) and `method="schema"`.

The sibling exporters accept the identical list and round-trip it correctly, so a list is clearly meant to be valid input across the export API; `export_yaml` is the outlier.

## Steps to Reproduce

1. Import `export_yaml` from `semantica.export.methods`
2. Call it with a list of dicts — the second type in its declared signature
3. `AttributeError` is raised

## Expected Behavior

A list of dicts exports to YAML, consistent with `export_json` and `export_csv`, which accept the same value.

Failing that, a clear `ValidationError` naming the unsupported shape — not an `AttributeError` leaking from an internal `.get()` call.

## Actual Behavior

`AttributeError: 'list' object has no attribute 'get'`, raised from inside the exporter. No file is written.

## Environment

- **OS**: macOS 26.0 (arm64)
- **Python Version**: 3.12.13
- **Semantica Version**: 0.6.5 (`main` @ `1a3dd503`)

## Code Example

```python
import os
import tempfile

from semantica.export.methods import export_csv, export_json, export_yaml

rows = [{"id": "1", "name": "Acme"}]

for fn, ext in ((export_json, "json"), (export_csv, "csv"), (export_yaml, "yaml")):
    d = tempfile.mkdtemp()
    try:
        fn(rows, os.path.join(d, f"out.{ext}"))
        print(f"{fn.__name__:14} OK   files={os.listdir(d)}")
    except Exception as e:
        print(f"{fn.__name__:14} FAIL {type(e).__name__}: {e}")
```

```
export_json    OK   files=['out.json']
export_csv     OK   files=['out.csv']
export_yaml    FAIL AttributeError: 'list' object has no attribute 'get'
```

Both YAML methods fail on the same input:

```python
export_yaml([{"id": "1"}], "out.yaml", method="semantic_network")  # AttributeError
export_yaml([{"id": "1"}], "out.yaml", method="schema")            # AttributeError
```

## Error Messages / Logs

```
  File "<string>", line 5, in <module>
  File "semantica/export/methods.py", line 537, in export_yaml
    exporter.export(data, file_path, **kwargs)
  File "semantica/export/yaml_exporter.py", line 167, in export
    yaml_content = self.export_semantic_network(data, **options)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "semantica/export/yaml_exporter.py", line 126, in export_semantic_network
    **semantic_network.get("metadata", {}),
      ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'list' object has no attribute 'get'
```

The wrapper also logs `Failed to export YAML: 'list' object has no attribute 'get'` before re-raising.

## Additional Context

- `semantica/export/methods.py:496` — the declared signature `Union[Dict[str, Any], List[Dict[str, Any]]]`
- `semantica/export/yaml_exporter.py:126` — the `.get()` call that fails
- `semantica/export/yaml_exporter.py:167` — dispatch into `export_semantic_network`
- Deterministic; no timing or ordering involved
- Found while checking export round-trip fidelity across formats. `export_json` and `export_csv` were also exercised with adversarial values (embedded newlines, CRLF, quotes, tabs, non-ASCII) and preserved all of them correctly — this is the only fidelity failure I hit.

## Possible Solution

Two directions, and the choice is a maintainer's call since they imply different contracts:

1. **Normalize a list at the exporter boundary**, e.g. treat `[...]` as `{"entities": [...]}`, mirroring how `export_json` and `export_csv` handle the same input. Keeps the export API consistent, and makes the declared signature true.

2. **Narrow the type hint to `Dict[str, Any]`** and raise `ValidationError` with a message naming the expected shape. Smaller change, but leaves `export_yaml` accepting less than its siblings for no obvious reason.

I'd lean toward (1) for consistency, but happy to implement whichever you prefer.

pravit-amp:
## Description

`export_yaml` declared `Union[Dict[str, Any], List[Dict[str, Any]]]`, but both YAML exporters read their payload by key. A list reached `semantic_network.get(...)` and surfaced as a bare `AttributeError: 'list' object has no attribute 'get'` from inside the exporter — an error naming neither the offending argument nor the shape expected.

Reject rather than wrap. These formats distinguish entities from relationships from triplets, so inferring which collection a bare list represents would silently mislabel the records. Wrapping it under an unrecognised key is worse: the exporter reads only `entities`/`relationships`/`triplets`, so any other key writes a structurally valid file with every collection empty, trading a loud failure for silent data loss.

Validation lives in the exporters rather than the convenience wrapper, matching the existing precedent in `Neo4jCSVExporter._normalize_graph`, so direct users of the classes get the same contract as callers of `export_yaml`.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #952

## Changes Made

- `semantica/export/yaml_exporter.py` — added `_require_mapping(data, expected_keys)`, called from both entry points (`export_semantic_network` and `export_ontology_schema`), each naming its own key set. Raises `ProcessingError` with the same message shape as the existing `Neo4jCSVExporter` check.
- `semantica/export/methods.py` — narrowed `export_yaml`'s `data` hint to `Dict[str, Any]`, documented why a list is rejected rather than guessed at, and added a `Raises:` section.
- `tests/export/test_yaml_exporter_input_validation.py` — 9 tests / 46 subtests covering both directions.

Membership is tested with `isinstance(data, Mapping)` rather than `dict`, so `OrderedDict`, `defaultdict` and other mappings that worked before keep working.

Resulting messages:

```
Cannot export object of type 'list': expected a dict with 'entities'/'relationships'/'triplets'.
Cannot export object of type 'list': expected a dict with 'classes'/'properties'.
```

## Testing

- [x] Tested locally

Both YAML methods and both exporter classes are covered, directly and through the wrapper. Rejection cases span list-of-dicts (the shape from #952), empty list, tuple, list of scalars, `str`, `bytes`, `int`, `None` and `set`. Acceptance cases pin that valid mappings still export, records survive the round trip, `{}` is still accepted, and mapping subclasses still work.

Verified the guard actually fails: removing both `_require_mapping` calls produces **40 failures**; restoring them gives 9 passed / 46 subtests.

Verified nothing else broke by diffing the full-suite failure *set* against `main` rather than comparing counts — identical, 200 either way, with no failure introduced. (Raw counts are unreliable on this suite; repeat baseline runs gave 240 and 200, which is the order-dependent flakiness tracked in #859.)

Also checked the two non-test callers of the changed API: `examples/capability_gap_context_graphs_example.py` passes `context_graph.to_dict()` (a dict), and `tests/test_notebooks_simulation.py` exercises the YAML export path and passes.

### Test Commands

```bash
pytest tests/export/test_yaml_exporter_input_validation.py -q
```

```bash
pytest tests/export/ tests/test_export_module.py tests/test_export_methods_wrapper.py -q
```

```bash
python -m build
```

Results: `9 passed, 46 subtests`, `93 passed, 46 subtests`, and a clean `semantica-0.6.5` sdist + wheel.

## Documentation

The behaviour is documented in the affected docstrings — `export_yaml` now states the expected mapping shape and why a list is rejected, and both exporter methods carry a `Raises:` section. No prose docs referenced the list form.

## Breaking Changes

**Breaking Changes**: No

No working call changes behaviour: passing a list never succeeded, it raised `AttributeError`. The exception type for that input changes from `AttributeError` to `ProcessingError`, so the only code affected is anything catching `AttributeError` specifically to detect this case — which would have been catching an unintended leak from the exporter internals.

The type hint narrows from `Union[Dict[str, Any], List[Dict[str, Any]]]` to `Dict[str, Any]`, which makes the annotation describe what the function actually accepted all along.

## Checklist

`yaml_exporter.py` is black- and isort-clean, matching its state on `main`. `methods.py` was already black-unclean on `main` (at `export_neo4j_csv`, ~line 744); my hunk is conformant and I left the pre-existing formatting alone rather than widening the diff. flake8 on the two changed source files went from 4 findings to 3; the remainder are pre-existing `F401`/`E501`.

## Additional Notes

**Deliberately scoped to `export_yaml`.** Six other exporters raise the identical bare `AttributeError` on a list — `export_rdf`, `export_graph`, `export_lpg`, `export_owl`, `export_arango`, and `generate_report`. I kept this PR to the function named in #952 rather than widening it, but the same `_require_mapping` treatment applies to each and I'm happy to open a follow-up issue or extend this PR, whichever you prefer.

**Related.** #953 covers the neighbouring failure mode this fix deliberately avoids: a *mapping* whose keys aren't recognised currently exports an empty file with no warning. That is why a list is rejected here rather than normalised into a wrapper — normalising would route it straight into that silent path.

### DIFF
--- a/semantica/export/methods.py
+++ b/semantica/export/methods.py
@@ -494,7 +494,7 @@ def export_graph(
 
 
 def export_yaml(
-    data: Union[Dict[str, Any], List[Dict[str, Any]]],
+    data: Dict[str, Any],
     file_path: Union[str, Path],
     method: str = "semantic_network",
     **kwargs,
@@ -504,13 +504,26 @@ def export_yaml(
 
     This is a user-friendly wrapper that exports data to YAML format.
 
+    Unlike :func:`export_json` and :func:`export_csv`, which treat a list as
+    opaque records, both YAML methods are keyed formats: they distinguish
+    entities from relationships from triplets (and classes from properties
+    for ``method="schema"``). A bare list is therefore rejected rather than
+    guessed at, since inferring which collection it represents would silently
+    mislabel the records.
+
     Args:
-        data: Data to export (semantic network, entities, relationships)
+        data: Data to export, as a mapping. For ``method="semantic_network"``,
+            keyed by 'entities'/'relationships'/'triplets'; for
+            ``method="schema"``, by 'classes'/'properties'.
         file_path: Output YAML file path
         method: Export method (default: "semantic_network")
             - "semantic_network": Semantic network YAML export
             - "schema": Schema YAML export
         **kwargs: Additional options passed to YAML exporters
+
+    Raises:
+        ProcessingError: if ``data`` is not a mapping, or if ``method`` is not
+            a known YAML export method.
 
     Examples:
         >>> from semantica.export.methods import export_yaml
--- a/semantica/export/yaml_exporter.py
+++ b/semantica/export/yaml_exporter.py
@@ -21,6 +21,7 @@ Author: Semantica Contributors
 License: MIT
 """
 
+from collections.abc import Mapping, Sequence
 from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Union
@@ -29,6 +30,34 @@ from ..utils.exceptions import ProcessingError, ValidationError
 from ..utils.helpers import ensure_directory
 from ..utils.logging import get_logger
 from ..utils.progress_tracker import get_progress_tracker
+
+
+def _require_mapping(data: Any, expected_keys: Sequence[str]) -> None:
+    """Reject non-mapping export input with an actionable error.
+
+    Both YAML exporters read their payload by key. Handed a sequence (or any
+    other non-mapping), every lookup would fail with a bare
+    ``AttributeError: 'list' object has no attribute 'get'`` from deep inside
+    the exporter, which tells the caller nothing about the shape expected.
+
+    A list is rejected rather than wrapped: these formats distinguish
+    entities from relationships from triplets, so inferring which one a bare
+    list represents would silently mislabel the records.
+
+    Args:
+        data: Candidate export payload.
+        expected_keys: Key names this exporter reads, named in the error so
+            the caller learns the expected shape.
+
+    Raises:
+        ProcessingError: if ``data`` is not a mapping.
+    """
+    if not isinstance(data, Mapping):
+        keys = "/".join(f"'{key}'" for key in expected_keys)
+        raise ProcessingError(
+            f"Cannot export object of type '{type(data).__name__}': "
+            f"expected a dict with {keys}."
+        )
 
 
 class SemanticNetworkYAMLExporter:
@@ -96,6 +125,13 @@ class SemanticNetworkYAMLExporter:
                 - metadata: Metadata dictionary (optional)
             **options: Additional export options (unused)
 
+        Raises:
+            ProcessingError: if ``semantic_network`` is not a mapping. A bare
+                list of records cannot be exported here because this format
+                distinguishes entities, relationships, and triplets, and
+                guessing which one a list represents would silently mislabel
+                it.
+
         Returns:
             String containing YAML representation of semantic network
 
@@ -107,6 +143,8 @@ class SemanticNetworkYAMLExporter:
             ... }
             >>> yaml_str = exporter.export_semantic_network(network)
         """
+        _require_mapping(semantic_network, ("entities", "relationships", "triplets"))
+
         # Track YAML export
         tracking_id = self.progress_tracker.start_tracking(
             file=None,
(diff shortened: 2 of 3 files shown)


pravit-amp:
Heads up for reviewers: I've already implemented a fix for #953 (the neighboring "unrecognized keys silently export an empty file" issue this PR's description calls out as deliberately out of scope) on `fix/953-yaml-export-silent-empty` in my fork.

It's stacked on this branch (`fix/952-export-yaml-non-mapping-input`) since it builds on the `_require_mapping` guard added here. I'll open that PR once this one merges, so it targets `main` cleanly instead of showing this PR's commits too.

pravit-amp:
Superseded by the consolidated PR above, which carries this commit plus the fixes for #953 and #956. Closing in favour of reviewing them together — #953's fix is written against the resolver from #956, so they weren't separable.

PR #957, opened 2026-08-12
feat(context): add retraction and purge to ContextGraph

### LINKED ISSUE #955: [FEATURE] Retraction and purge for ContextGraph — no way to remove a node or edge without discarding the whole graph
## Problem Statement

`ContextGraph` has 56 public methods and none of them removes anything. The only way to take a node or edge out of a graph is `clear()`, which discards the entire graph.

```python
>>> from semantica.context import ContextGraph
>>> pub = [m for m in dir(ContextGraph) if not m.startswith("_")]
>>> len(pub)
56
>>> [m for m in pub if any(k in m.lower() for k in ("remove", "delete", "drop", "prune", "forget", "erase"))]
[]
```

The MCP server has the same gap: the five tool modules (`decisions`, `export`, `extraction`, `graph`, `reasoning`) expose no deletion tool, so an agent can write to shared context indefinitely and never retract.

This sits awkwardly against how the project is positioned. The docs state GDPR readiness in several places — for example `docs/getting-started.md:16` ("HIPAA, SOX, GDPR, FDA 21 CFR Part 11 infrastructure built in") and `docs/glossary.md:203`. Right-to-erasure is not optional in those settings, and today removing one data subject's entity from a context graph means exporting to a dict, filtering by hand, and rebuilding the graph — losing provenance in the process.

The capability exists elsewhere in the codebase, which makes the core graph the outlier:

- `AgentMemory.forget()` / `.delete_memory()` (`semantica/context/agent_memory.py:420,574`)
- `AgentContext.forget()` / `.delete_conversation()` (`semantica/context/agent_context.py:654,1458`)
- `GraphStore.delete_node()` / `.delete_relationship()` (`semantica/graph_store/graph_store.py:673,709`)

## Proposed Solution

Add retraction to `ContextGraph` as a **provenance-recorded event**, not a silent drop.

A hard delete would be the wrong primitive here. In a system whose value proposition is an auditable trail, a fact vanishing without a record is itself an audit failure: afterwards you cannot answer "was this ever in the graph, and who removed it?" Retraction should mean the fact stops being active going forward, while history remains queryable.

Concretely:

- `retract_node(node_id, *, reason=None, at=None)` and `retract_edge(edge_id, ...)` — close the validity window, record who/why/when
- `purge_node(node_id, ...)` — genuine hard removal for erasure requests, where the obligation is that the data is actually gone. Records a tombstone (that a purge occurred, when, and why) without retaining the purged content
- Index consistency maintained across `nodes`, `edges`, `_adjacency`, `node_type_index`, `edge_type_index` — the same invariant `clear()` already has to uphold
- Corresponding MCP tools so agents can retract, not only append

The distinction matters: retraction is for "this fact is no longer true," purge is for "this data must not exist." Compliance needs both, and conflating them produces either an unusable audit trail or an unmet erasure obligation.

## Alternatives Considered

**Rebuild the graph without the node.** Works today via `to_dict()` → filter → `from_dict()`. It is O(graph) per deletion, drops provenance, and every caller reimplements it. It also cannot express "removed at time T for reason R."

**Hard delete only (`remove_node`).** Simplest, and matches `GraphStore.delete_node()`. But it produces exactly the silent-loss failure mode this project treats as a bug elsewhere (#953, #922), and it makes point-in-time queries lie about the past.

**Retraction only, no purge.** Cleaner conceptually, but it cannot satisfy a GDPR Article 17 erasure request, since the data is still present.

**Handle it at the storage backend.** `GraphStore` backends already have `delete_node`. But `ContextGraph` is in-memory and independent of those backends, so this leaves the default path — the one the quickstart and MCP server use — with no answer.

## Use Cases

1. **Right-to-erasure request.** A data subject asks for removal. An operator purges every node and edge referencing them and can later evidence that the purge happened, when, and under what authority — without the graph retaining the erased content.

2. **Correcting a bad extraction.** An extractor produced a wrong entity or a hallucinated relationship. It is retracted with a reason; decisions recorded before the retraction still explain themselves against the graph as it stood, so the audit trail of past decisions stays intact.

3. **Long-running agent memory hygiene.** An agent accumulates context indefinitely. Stale or superseded facts are retracted so `find_active_nodes()` and retrieval reflect current state, while `state_at()` can still reconstruct what the agent believed at any earlier point.

4. **Multi-tenant / project teardown.** Removing one tenant's or project's subgraph without discarding the shared graph, which `clear()` would force.

## Impact Assessment

- **Who would benefit?** Anyone running Semantica in a regulated setting (the stated target: finance, healthcare, legal, government), plus anyone running a long-lived agent whose context graph grows without bound. Also MCP users, who currently have an append-only interface.
- **Priority**: High — it is the gap between a documented compliance claim and the API surface.
- **Breaking Changes**: No. Purely additive; existing behaviour is untouched.
- **Dependencies**: None new. It builds on machinery already shipped — the temporal validity window and `state_at()` — and would pair naturally with the `provenance` module for the retraction record.

## Implementation Ideas

The temporal machinery needed for retraction already works. `ContextNode`/`ContextEdge` carry `valid_from`/`valid_until` with an `is_active()` check, and `state_at()` honours them:

```python
>>> g = ContextGraph(advanced_analytics=False)
>>> g.add_node("alice", "person")
>>> g.add_node("bob", "person", valid_until="2020-01-01T00:00:00Z")
>>> [n["id"] for n in g.find_active_nodes()]
['alice']
>>> [n["id"] for n in g.state_at("2019-06-01T00:00:00Z")["nodes"]]
['alice', 'bob']
>>> [n["id"] for n in g.state_at("2026-01-01T00:00:00Z")["nodes"]]
['alice']
```

So retraction is largely "close the window and record why", rather than a new subsystem:

```python

graph = ContextGraph()
graph.add_node("alice", "person", email="alice@example.com")
graph.add_edge("alice", "acme", "works_at")

# Retract: no longer true going forward, still visible in history.
graph.retract_node("alice", reason="employment ended", at="2026-01-01T00:00:00Z")

graph.has_node("alice")                                   # False - not active
graph.state_at("2025-06-01T00:00:00Z")                    # still present
graph.get_retraction("alice")
# {"retracted_at": "2026-01-01T00:00:00Z", "reason": "employment ended", ...}

# Purge: genuinely gone, tombstone retained.
graph.purge_node("alice", reason="GDPR Art. 17 request #4471")
graph.state_at("2025-06-01T00:00:00Z")                    # no longer present
graph.get_tombstone("alice")
# {"purged_at": "...", "reason": "GDPR Art. 17 request #4471"}  # no purged content
```

Open questions worth settling before implementation:

1. Should retracting a node cascade to its edges, or leave dangling edges for the caller to handle? Cascading is likely right, but it changes edge counts in ways callers may not expect.
2. Should `retract_*` be idempotent (retracting twice is a no-op) or raise on an already-retracted id?
3. Does `purge_*` need to reach into `AgentMemory` and any bound vector store to be a complete erasure, or is graph-scope purge the right boundary for a first cut?

I would rather agree these before writing code than discover the intended contract in review.

## Additional Context

- `README.md:97` and `README.md:992` describe a single shared context graph across agents; an append-only shared surface is harder to operate as the number of writers grows.
- Related in spirit: #922 (edge ingest not idempotent, so the graph grows on re-ingest) and #953 (silent empty export). Both are cases where the graph's write path lacked a control that callers assumed existed.
- Prior art: RDF reification and named-graph retraction; bitemporal databases (valid time vs transaction time), which is effectively what `valid_from`/`valid_until` already model here.

## Contribution

- [x] I can help with testing

pravit-amp:
## Description

`ContextGraph` had 56 public methods and none that removed anything — the only option was `clear()`, which discards the whole graph. This adds two operations with deliberately different contracts.

**Retraction** (`retract_node`, `retract_edge`) closes the entity's validity window. It stops being active going forward, but `state_at()` before the retraction still returns it, so decisions recorded against it remain explainable. This reuses the `valid_from`/`valid_until` machinery already present rather than introducing a new subsystem.

**Purge** (`purge_node`, `purge_edge`) removes the entity outright — from history as well as the active view — leaving a tombstone that records *that* a purge happened and why, but never the purged content.

The distinction matters: a hard delete alone makes point-in-time queries lie about the past, while retraction alone cannot satisfy an erasure obligation.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #955

## Changes Made

- `semantica/context/context_graph.py` — `retract_node`, `retract_edge`, `purge_node`, `purge_edge`, plus `get_retraction`, `get_tombstone`, `list_retractions`, `list_tombstones`
- Internal helpers `_emit_mutation`, `_incident_edges`, `_drop_edge_from_indexes`; `clear()` now also resets the retraction and tombstone records
- `tests/context/test_context_graph_retraction.py` — 23 tests

**No changes to `change_management` were needed.** `MutationRecord` already documents `REMOVE_NODE`/`REMOVE_EDGE` in its operation vocabulary (`change_log.py:121`) — the audit trail anticipated removal that nothing emitted. Retraction emits `UPDATE_NODE`, purge emits `REMOVE_NODE`, matching the documented contract.

Two details worth review attention:

- `_incident_edges` scans `self.edges` rather than `_adjacency`. `_adjacency` is keyed by source only, so using it would silently leave inbound edges pointing at a removed node.
- Purge updates `edges`, `edge_type_index` and `_adjacency` together under the lock, so the indexes cannot drift — the same invariant `clear()` upholds.

**Three design decisions.** The issue left these open and they weren't answered, so they're defaults rather than settled contracts — happy to change any:

1. **Cascade defaults to `True`.** Leaving edges active around an inactive node means `find_active_nodes()` drops the node while its relationships still read as current. Both operations accept `cascade=False`.
2. **Idempotent.** Retracting an already-retracted entity returns `False` and preserves the original reason, rather than raising.
3. **Purge is graph-scope only.** `AgentMemory` and a bound vector store are not reached — documented as one step of an erasure workflow, not the whole of it.

Retraction is visible to the activity-aware views (`find_active_nodes`, `state_at`, `is_active`) and not to membership checks (`has_node`, `stats`), which still count the retained record. That match


KaifAhmad1 (approved):
Approving after a full pass: initial review, a bug found and fixed, and the resulting merge-conflict cleanup.

**Design**: clean split between the two contracts this needed — `retract_node`/`retract_edge` close the validity window (reusing `valid_from`/`valid_until`/`state_at()` rather than a new subsystem), `purge_node`/`purge_edge` are the destructive, tombstone-only counterpart for erasure obligations retraction can't satisfy. Index consistency (`nodes`, `edges`, `edge_type_index`, `_adjacency`, `node_type_index`) is maintained the same way `clear()` already has to. Mutation-callback payloads are snapshotted inside the lock and emitted after it's released — verified this holds even when the callback itself mutates the graph (`TestMutationEmissionIsSelfContained` clears the graph mid-emission and asserts every record still fires).

**Bug found in initial review, now fixed**: `retract_edge`/`purge_edge` resolved "the edge" for a given `edge_id` via the first matching object only, and `retract_node`'s cascade deduped against the live `_retractions` dict mid-loop. Since `edge_id` was content-derived and not unique before #926 landed, a graph holding two identical edges could have one silently left live/active while the call reported `True` and produced a tombstone/retraction claiming it was fully handled — undermining exactly the erasure guarantee this feature exists to provide. Fixed in `e711b557`: both direct operations now act on every edge matching the id under one record, and the cascade's dedup check is snapshotted before the loop starts. 5 new regression tests (`TestDuplicateEdgeId`) reproduce all three paths and were verified to fail pre-fix.

**Changelog** (`5328a977`) documents the feature and the review-caught fix, matching the repo's convention.

**Merge conflict** from #926 landing in the meantime is resolved (`00e801ec`) — merged `main` in, combined the two `### Added` entries, updated the changelog's "fixed during review" wording now that #926 is merged (was "pending"), and rewrote `TestDuplicateEdgeId`'s fixture to construct a duplicate edge directly rather than through `add_edge()`, since #926's dedupe means two `add_edge()` calls can no longer produce one — the test now correctly targets the realistic remaining case (a graph that already holds a duplicate, e.g. from a pre-#926 save).

**Verified**: full `tests/context/` suite — 565 passed, 3 skipped, none related to this change — plus all 49 tests in `test_context_graph_retraction.py`, and the cross-graph-link teardown logic traced against the real `link_graph`/`navigate_to`/`resolve_links` implementations to confirm the marker-node and link-deregistration assumptions hold.

One non-blocking note carried over from the initial review: purge only deregisters cross-graph links this node exits through; a link another `ContextGraph` instance holds *into* a node purged here isn't reached. Same boundary as the already-documented `AgentMemory`/vector-store exclusion — fine to leave out of scope, worth a one-line doc mention alongside that note if there's a follow-up.

Thanks for the thorough design writeup and for flagging the open design decisions up front, @pravit-amp. Approving.

PR #958, opened 2026-08-12
fix(export): harden YAML export input handling

### LINKED ISSUE #952: [BUG] export_yaml raises AttributeError on List[Dict], a type its own signature declares as supported

### LINKED ISSUE #953: [BUG] export_yaml silently writes an empty export when the input dict lacks its expected keys — no warning, logs report success

### LINKED ISSUE #956: [FEATURE] refactor(export): centralize graph-payload key normalization — four competing idioms across nine exporters produce divergent results
## Problem Statement

Graph payloads circulate through this codebase under two vocabularies — `entities`/`relationships` and `nodes`/`edges` — and there is no shared contract for reconciling them. Each consumer decides for itself, and they don't agree.

Auditing `origin/main`, 48 files read these keys via `.get()`:

| | files |
| --- | --- |
| accept **both** vocabularies | 19 |
| `entities`/`relationships` only | 27 |
| `nodes`/`edges` only | 2 |

Within `semantica/export/` alone, six of nine exporters handle both — but via at least three different idioms:

```python
neo4j_csv_exporter.py:498   nodes = graph.get("nodes") or graph.get("entities") or []
arango_aql_exporter.py:210  nodes = knowledge_graph.get("nodes", entities)
lpg_exporter.py:164         nodes = knowledge_graph.get("nodes", entities)
csv_exporter.py:412,436     entities = kg.get("entities", []); nodes = kg.get("nodes", [])
```

`get("nodes", entities)` and `get("nodes") or get("entities")` diverge whenever `nodes` is present but empty: the first yields `[]`, the second falls through to `entities`.

That divergence is observable. One payload, six exporters:

```python
import os, tempfile
from semantica.export import methods as m

payload = {"entities": [{"id": "e1", "name": "Acme"}], "nodes": [],
           "relationships": [], "edges": []}

for name in ("export_json", "export_arango", "export_neo4j_csv",
             "export_graph", "export_lpg", "export_csv"):
    d = tempfile.mkdtemp()
    try:
        getattr(m, name)(payload, os.path.join(d, "out"))
        blob = "".join(open(os.path.join(r, f), errors="ignore").read()
                       for r, _, fs in os.walk(d) for f in fs)
        print(f"{name:18} {'CONTAINS Acme' if 'Acme' in blob else 'DROPPED the entity'}")
    except Exception as e:
        print(f"{name:18} {type(e).__name__}: {e}")
```

```
export_json        CONTAINS Acme
export_arango      CONTAINS Acme
export_neo4j_csv   CONTAINS Acme
export_graph       CONTAINS Acme
export_lpg         DROPPED the entity
export_csv         ValidationError: No data to write. Data list is empty.
```

Three outcomes from identical input: exported, silently dropped, rejected. `lpg` and `arango` use the *same* idiom and still diverge, because the aliased value feeds different downstream paths.

This is not hypothetical. `ContextGraph.to_dict()` emits `nodes`/`edges`, and the YAML exporter reads `entities`/`relationships` — which is why the shipped example at `examples/capability_gap_context_graphs_example.py:566` writes an empty file (#953). That is one instance of a class.

## Proposed Solution

A single normalization helper in `semantica/utils/`, adopted by every graph-payload consumer, defining one contract:

- which keys are recognized, including aliases
- which vocabulary wins when both are present, and what "present but empty" means
- what happens when nothing is recognized — currently the source of the silent-empty exports

Fixing exporters one at a time (as #952 and #953 do) is correct for the bugs already filed, but each fix re-decides the same questions locally. A shared helper makes the contract reviewable in one place and testable once.

## Alternatives Considered

**Fix each exporter as bugs surface.** What's happening now. It works, but the contract stays implicit and the next exporter added will pick its own idiom. Three of nine exporters are still entities-only.

**Standardize on one vocabulary and migrate callers.** Cleanest end state, but a breaking change for anyone passing the other form, and `ContextGraph.to_dict()` would have to change shape or every caller of it would.

**Normalize at the `ContextGraph.to_dict()` boundary instead.** Narrower, but only helps payloads that originate from a `ContextGraph`. Dicts assembled by hand, loaded from JSON, or produced by `kg` builders still reach exporters unnormalized.

**Do nothing.** Defensible if the divergence is considered acceptable, but the silent-drop case (`export_lpg` above) is the same failure mode already treated as a bug in #953.

## Use Cases

1. **Exporting a context graph.** `export_yaml(graph.to_dict(), path)` is the obvious call and currently produces an empty file; `export_lpg` with an empty `nodes` key silently drops entities. A user has no way to know which exporters understand which shape.

2. **Switching export formats.** A pipeline that works with `export_json` and is later pointed at `export_lpg` or `export_csv` changes behaviour with no code change on the caller's side, and no error to explain it.

3. **Adding a new exporter or consumer.** A contributor has no canonical example to follow — six existing implementations disagree — so a new one adds a seventh idiom.

## Impact Assessment

- **Who would benefit?** Anyone exporting graph data, and anyone building on the `export` module. Contributors most of all: the current answer to "how should my consumer read a graph payload?" is "copy whichever neighbour you happened to open."
- **Priority**: Medium. No single instance is critical, but it generates recurring bugs (#953 is one) and each is fixed in isolation.
- **Breaking Changes**: Yes, narrowly. Exporters that currently drop records under one shape would start emitting them, and consumers relying on the current inconsistent behaviour would see output change. Every affected case is one that silently loses data today, so the change is toward correctness — but it isn't a no-op and shouldn't be described as one.
- **Dependencies**: Overlaps #952 and #953, which add input validation to the YAML exporters. Whatever contract those settle should be the one this generalizes; worth landing them first so this consolidates a decision rather than pre-empting it.

## Implementation Ideas

```python
from semantica.utils import normalize_graph_payload

# Returns a canonical payload, or raises when nothing is recognized.
payload = normalize_graph_payload(
    {"nodes": [{"id": "n1"}], "edges": [{"id": "e1"}]}
)
# {"entities": [{"id": "n1"}], "relationships": [{"id": "e1"}], "triplets": []}

# Already canonical: unchanged.
normalize_graph_payload({"entities": [{"id": "n1"}]})
# {"entities": [{"id": "n1"}], "relationships": [], "triplets": []}

# Unrecognized, non-empty: raises rather than returning empty collections.
normalize_graph_payload({"data": [{"id": "n1"}]})
# ProcessingError: no recognized keys; reads 'entities'/'nodes'/...
```

Each exporter then opens with one line instead of its own `or` chain:

```python
def export(self, data, file_path, **options):
    data = normalize_graph_payload(data)
    ...
```

Questions worth settling before implementation, since they are exactly what the current implementations answer differently:

1. When both `nodes` and `entities` are present and non-empty, which wins? `neo4j_csv` picks `nodes`; a caller supplying both probably has a reason, so raising may be safer than a silent precedence rule.
2. Is `{"nodes": []}` "no nodes" or "unspecified, fall back to entities"? This is the exact divergence between the two idioms above.
3. Should the helper raise or return empty collections when nothing is recognized? #952/#953 lean toward raising; this should match whatever they settle.

## Additional Context

- #953 — YAML exporter silently writes an empty file for an unrecognized mapping. One instance of this class.
- #952 — `export_yaml` raising `AttributeError` on a documented input type. Same area; establishes the input-contract direction.
- `semantica/export/neo4j_csv_exporter.py:496-512` — `_normalize_graph`, the closest thing to a reference implementation; a shared helper would generalize it.
- `semantica/context/context_graph.py` — `to_dict()` emits `nodes`/`edges`, which is what makes this reachable from the library's own graph type.
- Verified on `origin/main` @ `18f1d55`, Python 3.12.13, in a clean worktree so in-flight branches did not affect the counts.

## Contribution

- [x] I can help with testing

pravit-amp:
## Description

Consolidates the YAML-export input-handling work into one PR. Supersedes #954, which carried the first commit alone.

The three changes are one story and are hard to review apart: #953's fix is written against the resolver added for #956, and #952 and #953 reject adjacent shapes of the same bad input through the same code path. All three issues are assigned to me.

Today, three different malformed inputs to `export_yaml` fail three different ways — one loud but uninformative, one silent, one correct only by accident:

```python
export_yaml(records, "out.yaml")             # AttributeError from deep inside the exporter
export_yaml({"data": records}, "out.yaml")   # succeeds, writes a file with every record gone
export_yaml(graph.to_dict(), "out.yaml")     # succeeds, writes an empty file (nodes/edges)
```

After this PR the first two raise with an actionable message, and the third exports its records.

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring

## Related Issues

Closes #956
Closes #952
Closes #953

## Changes Made

- **`normalize_graph_payload()` in `utils.helpers` (#956)** — graph payloads circulate under two vocabularies, `entities`/`relationships` and `nodes`/`edges`, and consumers each reconciled them locally with competing idioms, so the same payload could be exported, silently dropped, or rejected depending on which consumer read it. This becomes the single place that decision is made, adopted by `LPGExporter`, `ArangoAQLExporter` and `Neo4jCSVExporter`. `LPGExporter` in particular dropped every entity when `nodes` was present but empty — the shape `JSONExporter` emits.
- **Non-mapping input rejected by the YAML exporters (#952)** — both read their payload by key, so a list surfaced as a bare `AttributeError` naming neither the argument nor the expected shape. Rejected rather than wrapped: these formats distinguish entities from relationships from triplets, so guessing which one a list holds would mislabel the records.
- **Unrecognized mappings rejected (#953)** — a mapping keyed by anything the exporters do not read serialized to a structurally valid file with every collection empty, with no exception, no warning, and the progress log reporting a completed export. `export_semantic_network`, `export_for_pipeline` and `export_ontology_schema` now validate first.
- **`ContextGraph.to_dict()` exports correctly** — the `nodes`/`edges` aliases come with the shared resolver, so the most direct path from this library's own graph type to YAML (used in `examples/capability_gap_context_graphs_example.py`) no longer produces an empty file.
- **`export()` serializes before creating the output directory**, so a rejected export leaves nothing behind.

### Design notes for reviewers

- **Two exception types, deliberately.** A payload of the wrong *type* cannot be exported at all and raises `ProcessingError`, matching `Neo4jCSVExporter._normalize_graph`. A mapping whose *contents* are unusable raises `ValidationError`, matching `normalize_graph_payload`. `_require_mapping` runs first at every entry point, so a non-mapping never reaches the resolver. Happy to flatten to one type if you'd prefer.
- **`metadata` alone does not make a payload recognized.** An `export_json` envelope carries one, so accepting it would readmit the exact silent-empty export this is most likely needed for.
- **An empty mapping still exports.** An empty graph is legitimate and has no records to lose.
- **Two non-empty spellings of one collection are refused** rather than guessed at — there is no basis for preferring either, and picking one would silently discard the other.
- **`CSVExporter` and `JSONExporter` are deliberately excluded** from the resolver: they write `entities`, `relationships`, `nodes` and `edges` as separate outputs by design rather than reconciling two spellings of one collection, so normalizing there would rename output files.

## Testing

- [x] Tested locally

New coverage:

- `tests/utils/test_normalize_graph_payload.py` — resolver behaviour, plus end-to-end assertions run through the real exporters rather than mocks, since the behaviour under test is that the exporters now agree.
- `tests/export/test_yaml_exporter_input_validation.py` — non-mapping rejection, both directions, including that no file is written.
- `tests/export/test_yaml_exporter_key_recognition.py` — every input shape from #953's table, a `ContextGraph.to_dict()` round-trip built from a real graph so it breaks if `to_dict()` changes vocabulary, and an assertion that no success is logged for a rejected export.

### Test Commands

```bash
pytest tests/export tests/utils tests/context tests/test_export_module.py tests/test_export_methods_wrapper.py tests/test_notebooks_simulation.py
# 641 passed, 62 subtests passed

python -m build
# Successfully built semantica-0.6.5.tar.gz and semantica-0.6.5-py3-none-any.whl
```

`black`, `isort` and `flake8` are clean on every line this PR touches. (`black` still reports pre-existing trailing whitespace in unrelated docstrings of `utils/helpers.py` and `export/methods.py`, left alone to keep the diff scoped.)

## Documentation

`semantica/export/export_usage.md` gains an "Accepted Input" section with the recognized key sets per method; `docs/reference/export.md` notes the contract inline. Every public method touched has its `Raises:` section updated. No cookbook changes — no new examples were added.

## Breaking Changes

**Breaking Changes**: Yes

Three call patterns that previously returned cleanly now raise:

| Call | Before | After |
| :--- | :--- | :--- |
| `export_yaml({"data": [...]}, path)` | wrote a file with every record dropped | `ValidationError` |
| `export_yaml({"nodes": [...]}, path, method="schema")` | wrote empty `classes`/`properties` | `ValidationError` |
| `export_yaml([...], path)` | `AttributeError` from inside the exporter | `ProcessingError` |

Migration: pass the records under a recognized key — `{"entities": [...]}` / `{"nodes": [...]}` for `semantic_network`, `{"classes": [...]}` for `schema`.

This was raised on #953 before implementation, and raising was chosen over warning: the current logs affirmatively report success, so a caller has no way to detect the loss. Anyone whose call now raises was writing an empty file and losing data. `{}` still exports, so genuinely-empty graphs are unaffected. A `nodes`/`edges` payload that previously wrote an empty file now writes its records — a behaviour change, but in the direction the caller intended.

## Checklist

## Additional Notes

Assignment: #953 was assigned by @KaifAhmad1 on the issue; #952 and #956 are also assigned to me.

Reviewable commit-by-commit — the four commits are ordered refactor → #952 → #952 review fixes → #953, each self-contained.

One follow-up deliberately left out of scope: `CSVExporter` and `JSONExporter` still emit both vocabularies as separate outputs. That is by design today, but worth a look if the project wants one canonical on-disk shape.


Sameer6305:
@pravit-amp can you handle the qodo findings before we review it.

pravit-amp:
> @pravit-amp can you handle the qodo findings before we review it.

Findings 1 and 3 are fixed in ae190e4. On finding 2 — narrowing export_yaml to reject List[Dict] is the intended resolution of [#952](https://github.com/semantica-agi/semantica/issues/952) rather than a regression against it: the old signature declared list support the code never had (it raised AttributeError), and wrapping a bare list under a guessed key would mislabel records that this format keeps distinct. The declared signature, docstring, and the Breaking Changes section of the PR description are updated to match.

Sameer6305 (changes requested):
Thanks @pravit-amp for putting this together. I appreciate the effort here, especially centralizing the `entities`/`nodes` and `relationships`/`edges` reconciliation instead of continuing to fix the same ambiguity independently in each exporter. The resolver design around populated-vs-empty aliases and refusing conflicting non-empty aliases is a solid direction, and the added end-to-end coverage is useful.

I reviewed the implementation against `semantica-agi/main`, including the connected exporters and public export paths. I found two issues that I think should be addressed before merge:

### 1. `normalize_graph_payload()` still accepts invalid collection values

`semantica/utils/helpers.py` — `_resolve_collection()`

The resolver currently treats any truthy value as a collection and does:

```python
if value:
    return list(value)
```

This means malformed payloads such as:

```python
{"entities": "abc"}
```

are normalized to:

```python
{"entities": ["a", "b", "c"], ...}
```

while:

```python
{"entities": 42}
```

raises a raw:

```text
TypeError: 'int' object is not iterable
```

The problem is that the new helper is now the central validation boundary, but invalid collection values can still either be silently transformed or leak low-level exceptions from `list()` into the exporters.

This is particularly risky because downstream exporters expect graph records, not arbitrary iterables. The resulting error then points at exporter internals rather than the malformed `entities`/`nodes` field.

**Suggested fix:** validate collection values before converting them. At minimum reject strings, bytes, mappings, and non-iterable scalar values with an actionable `ValidationError`. Ideally also validate that the resulting records have the expected mapping/record shape.

Please add regression tests for at least:

* `{"entities": "abc"}`
* `{"entities": 42}`
* `{"nodes": {"id": "n1"}}`
* equivalent relationship/edge cases

The resolver should either return a valid graph collection or fail at the normalization boundary.

### 2. Neo4j CSV still bypasses the new "unrecognized payload" rejection

`semantica/export/neo4j_csv_exporter.py` — `_normalize_graph()`

The new helper is called as:

```python
normalize_graph_payload(graph, require_recognized=False)
```

That means a non-empty mapping containing no recognized graph keys still normalizes to empty `entities`/`relationships` instead of raising.

For example, a payload such as:

```python
{"data": [{"id": "e1"}]}
```

can still reach the Neo4j exporter as an empty graph and produce empty CSV output.

This is directly at odds with the central contract introduced by this PR: the helper's default behavior is specifically designed to prevent an unrecognized mapping from becoming a valid-looking empty export.

I understand why the object/attribute path in Neo4j is kept separate, but for an actual mapping there does not appear to be a reason to opt out of the recognized-key validation.

**Suggested fix:** use the normal `normalize_graph_payload(graph)` behavior for mapping inputs, while retaining the existing attribute-based path for graph objects. Add a regression test asserting that an unrecognized non-empty mapping raises `ValidationError` rather than producing empty CSV files.

### Why I think these need to be fixed before merge

Both issues are at the boundary of the new shared normalization contract. The first allows malformed collection values to escape the new validation layer, and the second leaves one of the adopters capable of the exact silent-empty behavior this refactor is intended to eliminate.

Once these are addressed and the corresponding regression tests are added, the overall direction of the PR looks good from my side.

**Current verdict: Request changes.**

The changes are relatively contained, so I don't think the PR needs a redesign — tightening the normalization boundary and making the mapping path consistent in Neo4j should be sufficient.

pravit-amp:
> Thanks @pravit-amp for putting this together. I appreciate the effort here, especially centralizing the `entities`/`nodes` and `relationships`/`edges` reconciliation instead of continuing to fix the same ambiguity independently in each exporter. The resolver design around populated-vs-empty aliases and refusing conflicting non-empty aliases is a solid direction, and the added end-to-end coverage is useful.
> 
> I reviewed the implementation against `semantica-agi/main`, including the connected exporters and public export paths. I found two issues that I think should be addressed before merge:
> 
> ### 1. `normalize_graph_payload()` still accepts invalid collection values
> `semantica/utils/helpers.py` — `_resolve_collection()`
> 
> The resolver currently treats any truthy value as a collection and does:
> 
> ```python
> if value:
>     return list(value)
> ```
> 
> This means malformed payloads such as:
> 
> ```python
> {"entities": "abc"}
> ```
> 
> are normalized to:
> 
> ```python
> {"entities": ["a", "b", "c"], ...}
> ```
> 
> while:
> 
> ```python
> {"entities": 42}
> ```
> 
> raises a raw:
> 
> ```
> TypeError: 'int' object is not iterable
> ```
> 
> The problem is that the new helper is now the central validation boundary, but invalid collection values can still either be silently transformed or leak low-level exceptions from `list()` into the exporters.
> 
> This is particularly risky because downstream exporters expect graph records, not arbitrary iterables. The resulting error then points at exporter internals rather than the malformed `entities`/`nodes` field.
> 
> **Suggested fix:** validate collection values before converting them. At minimum reject strings, bytes, mappings, and non-iterable scalar values with an actionable `ValidationError`. Ideally also validate that the resulting records have the expected mapping/record shape.
> 
> Please add regression tests for at least:
> 
> * `{"entities": "abc"}`
> * `{"entities": 42}`
> * `{"nodes": {"id": "n1"}}`
> * equivalent relationship/edge cases
> 
> The resolver should either return a valid graph collection or fail at the normalization boundary.
> 
> ### 2. Neo4j CSV still bypasses the new "unrecognized payload" rejection
> `semantica/export/neo4j_csv_exporter.py` — `_normalize_graph()`
> 
> The new helper is called as:
> 
> ```python
> normalize_graph_payload(graph, require_recognized=False)
> ```
> 
> That means a non-empty mapping containing no recognized graph keys still normalizes to empty `entities`/`relationships` instead of raising.
> 
> For example, a payload such as:
> 
> ```python
> {"data": [{"id": "e1"}]}
> ```
> 
> can still reach the Neo4j exporter as an empty graph and produce empty CSV output.
> 
> This is directly at odds with the central contract introduced by this PR: the helper's default behavior is specifically designed to prevent an unrecognized mapping from becoming a valid-looking empty export.
> 
> I understand why the object/attribute path in Neo4j is kept separate, but for an actual mapping there does not appear to be a reason to opt out of the recognized-key validation.
> 
> **Suggested fix:** use the normal `normalize_graph_payload(graph)` behavior for mapping inputs, while retaining the existing attribute-based path for graph objects. Add a regression test asserting that an unrecognized non-empty mapping raises `ValidationError` rather than producing empty CSV files.
> 
> ### Why I think these need to be fixed before merge
> Both issues are at the boundary of the new shared normalization contract. The first allows malformed collection values to escape the new validation layer, and the second leaves one of the adopters capable of the exact silent-empty behavior this refactor is intended to eliminate.
> 
> Once these are addressed and the corresponding regression tests are added, the overall direction of the PR looks good from my side.
> 
> **Current verdict: Request changes.**
> 
> The changes are relatively contained, so I don't think the PR needs a redesign — tightening the normalization boundary and making the mapping path consistent in Neo4j should be sufficient.

Thanks for the careful read — both findings were real and are now fixed in a
follow-up commit on this branch.

**1. `_resolve_collection()` now validates before converting.**

A new `_coerce_records(key, value)` runs ahead of any truthiness test or
`list()` call, so a recognized key can no longer hold a non-collection. It
raises `ValidationError` naming the offending key for strings/bytes
(`{"entities": "abc"}` is no longer three character-records), mappings
(`{"nodes": {"id": "n1"}}`, with a hint to wrap it in a list or supply its
values), and non-iterable scalars (`{"entities": 42}` is a `ValidationError`,
not a `TypeError` leaked from inside an exporter). Elements are validated too,
with the index called out.

On your "ideally also validate record shape": records are accepted as either
mappings **or** attribute-carrying objects, rather than mappings only —
`Neo4jCSVExporter._record_to_dict()` legitimately handles dataclasses and
objects with `__dict__`, so a mapping-only rule would have broken that path.

Two judgment calls worth surfacing, happy to change either:

- `None` is read as an absent collection, the spelling a JSON round-trip
  produces for `[]`. It can't hide data loss: `{"entities": None, "data": [...]}`
  is still rejected by the `_require_nothing_dropped()` check.
- Every spelling present is validated, not just the one that wins, so a
  malformed `nodes` isn't excused by a well-formed `entities`.

**2. Neo4j CSV now uses the default contract for mappings.**

`_normalize_graph()` calls `normalize_graph_payload(graph)` for mappings; the
attribute path for graph objects is untouched. Since that was the only opt-out
in the tree, I also removed the `require_recognized` flag rather than leave a
documented route back into the silent-empty export — it was introduced in this
PR and had no other caller. Easy to restore as a public escape hatch if you'd
rather keep it.

**Tests** — 15 new cases covering the four shapes you listed across all five
collection keys, plus object/dataclass/tuple acceptance, the null and
malformed-alias cases, and an end-to-end pass asserting no malformed value
reaches `export_arango` / `export_neo4j_csv` / `export_lpg` / `export_yaml`.
The Neo4j regression asserts an unrecognized non-empty mapping raises
`ValidationError` *and* that `nodes.csv` / `relationships.csv` are never
created — the payload is normalized before any file is opened.

Docs updated in `export_usage.md` and `docs/reference/export.md`, including an
"Accepted Input" section for the Neo4j CSV path.

Full suite matches the pre-change baseline on my machine (the failures there
are all missing optional deps in `semantic_extract`/`visualization`, unrelated
to this change); `black` and `flake8` are clean on the touched code.

Sameer6305 (dismissed):
Thanks @pravit-amp for putting this together. The centralization of graph-payload normalization is a solid improvement, and I appreciate the effort to keep the behavior consistent across the exporters instead of continuing to handle `entities`/`nodes` and `relationships`/`edges` independently.

I also appreciate the follow-up on the review findings. The requested fixes were addressed cleanly in `49fe0174`, including validation of invalid collection values and making the Neo4j mapping path follow the shared recognized-key validation contract.

During the follow-up review, we also identified two additional edge cases in the new validation boundary:

- `_require_usable_schema()` could treat scalar schema metadata such as `version`, `uri`, `title`, or `description` as evidence that records had been exported, allowing records under an unread key to be silently dropped.
- `_is_record()` accepted modules and class/type objects through the generic `__dict__` path, allowing invalid records to reach exporter internals instead of failing at the normalization boundary.

These were addressed in `def27781` (`fix(export): close YAML schema and record validation gaps`), along with regression coverage. The changes are intentionally small and scoped to the validation boundary.

I verified the relevant coverage after the fixes:
- 101 targeted tests passing across the normalization, YAML validation, YAML key-recognition, and Neo4j exporter suites.
- `git diff --check` clean.
- Regression coverage added for the newly identified cases.
- The existing support for mappings, dataclasses, and attribute-bearing record objects is preserved.

Overall, the implementation now has a consistent validation boundary, rejects malformed collection values with actionable errors, prevents unrecognized mappings from becoming silent empty exports, and keeps the different graph-payload aliases consistent across the relevant exporters.

Thanks again for the careful implementation and for addressing the review feedback promptly.

Also One separate follow-up point from the broader review: we also noticed that some validation error paths can reflect caller-controlled key names without a length bound, which can result in unnecessarily large error/log messages for extremely long input keys. This is pre-existing and not introduced by this PR, so it does not block approval.

I'll raise this separately shortly so it can be tracked independently. @pravit-amp, if you're okay taking that follow-up, feel free to pick it up separately.

@KaifAhmad1, approved from our side. The PR is ready for your final review and merge.

KaifAhmad1 (approved):
Thanks @pravit-amp for the centralization work here, and @Sameer6305 for two thorough review rounds — this got the shared boundary to a solid place. Did a final pass before merge and found four more gaps that the earlier rounds didn't reach, all fixed in `ae1d5958`:

- `LPGExporter`/`ArangoAQLExporter` called `normalize_graph_payload()` with no type guard, so non-mapping input raised `ValidationError` from inside the resolver while YAML/`Neo4jCSVExporter` raised `ProcessingError` for the identical mistake — inconsistent with the exception-type contract this PR itself documents. The `_require_mapping()` guard is now shared from `utils/helpers.py` and used by all three.
- `Neo4jCSVExporter._normalize_graph` checked `isinstance(graph, dict)`, so a non-dict `Mapping` (`MappingProxyType`, `ChainMap`) was rejected even though the identical payload exported fine via the other three exporters. Now checks `isinstance(graph, Mapping)`.
- `normalize_graph_payload()` accepts dataclass/attribute-bearing object records, but `LPGExporter`/`ArangoAQLExporter` read them with `.get(...)` — an object-shaped record passed validation only to crash with a raw `AttributeError` once used. Records are now converted to plain dicts at the boundary (`_coerce_records` → new `_record_to_dict`), so every consumer gets a uniform shape.
- Two non-empty spellings of the same collection holding identical records in a different order were rejected as conflicting (plain list equality). Comparison is now an order-independent multiset of each record's canonical JSON form.

Added regression coverage for all four in `tests/utils/test_normalize_graph_payload.py`, updated 4 existing tests that asserted the old (unconverted) object-passthrough behavior, and merged `main` in to resolve the `CHANGELOG.md` conflict (`d056a8df`) — that was the only conflicting file, purely an adjacent-insertion clash with #957's entry.

Full suite post-merge: `pytest tests/export tests/utils tests/context tests/test_export_module.py tests/test_export_methods_wrapper.py tests/test_notebooks_simulation.py` — 771 passed, 4 skipped. `black`/`isort`/`flake8 --max-line-length=88` clean on every touched line. `python -m build` succeeds. Also added a CHANGELOG entry (`838639b3`) covering the full arc of this PR: @pravit-amp's original centralization work, both of @Sameer6305's review rounds, and this follow-up.

Nice collaborative fix, @pravit-amp and @Sameer6305 — approving to merge.

PR #959, opened 2026-08-12
feat(seed): allow_private_ips opt-in for trusted internal API sources

### LINKED ISSUE #943: Support explicit private network API sources for trusted seed deployments.

yunaremaia:
Closes #943

## Summary

Adds the documented opt-in for trusted internal deployments: `allow_private_ips` on `SeedDataManager`, without weakening the secure default.

**Changes** (`semantica/seed/seed_manager.py`):
- `load_from_api()` now delegates to the **shared SSRF guard** (`semantica/ingest/ssrf.py`, introduced in #906) via `request_with_ssrf_guard()` instead of raw `requests.get()` — this also gains redirect-chain validation and bounded DNS resolution for free
- New config option `allow_private_ips`, parsed through the shared `parse_bool()` helper (avoids the `bool("false") == True` pitfall), defaulting to **False**
- Docstring documents that it should be enabled only for trusted internal deployments

**Tests** (`tests/test_seed_manager.py`):
- `test_load_from_api_blocks_private_by_default` — `127.0.0.1` rejected, no request made
- `test_load_from_api_allows_private_when_configured` — with `config={"allow_private_ips": True}` the request proceeds and the flag reaches the guard
- Existing `test_load_from_api` updated to mock `request_with_ssrf_guard` (the call path changed)
- **19/19 green** in `test_seed_manager.py`, **25/25** across both seed suites

The follow-up in #936 (SSRF guard for `load_from_api`) is complementary: this PR plugs the same function into the repo's shared guard with the internal-deployment escape hatch requested here.

### DIFF
--- a/semantica/ingest/ssrf.py
+++ b/semantica/ingest/ssrf.py
@@ -33,6 +33,46 @@ _DEFAULT_MAX_REDIRECTS = 10
 _REDIRECT_STATUS_CODES = frozenset({301, 302, 303, 307, 308})
 _STRIP_BODY_ON_REDIRECT = frozenset({301, 302, 303})
 
+# Standard port per scheme (mirrors requests' DEFAULT_PORTS).
+_DEFAULT_PORTS = {"http": 80, "https": 443}
+
+
+def _should_strip_auth(old_url: str, new_url: str) -> bool:
+    """Decide whether credentials must not follow a redirect.
+
+    Mirrors ``requests.utils.should_strip_auth``: credentials are stripped
+    when the hostname changes, when the port changes (outside default
+    ports), or on an https -> http downgrade on the same host. The single
+    exception is an http -> https upgrade on default ports, which requests
+    treats as safe to keep the credential for.
+    """
+    old_parsed = urlparse(old_url)
+    new_parsed = urlparse(new_url)
+
+    if old_parsed.hostname != new_parsed.hostname:
+        return True
+
+    # Special case: allow http -> https redirect on standard ports.
+    if (
+        old_parsed.scheme == "http"
+        and old_parsed.port in (80, None)
+        and new_parsed.scheme == "https"
+        and new_parsed.port in (443, None)
+    ):
+        return False
+
+    changed_port = old_parsed.port != new_parsed.port
+    changed_scheme = old_parsed.scheme != new_parsed.scheme
+    default_port = (_DEFAULT_PORTS.get(old_parsed.scheme), None)
+    if (
+        not changed_scheme
+        and old_parsed.port in default_port
+        and new_parsed.port in default_port
+    ):
+        return False
+
+    return changed_port or changed_scheme
+
 _dns_executor: Optional[concurrent.futures.ThreadPoolExecutor] = None
 _dns_executor_lock = threading.Lock()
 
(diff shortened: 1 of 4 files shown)


yunaremaia:
Both findings addressed in `9288d847` — and thank you, the first one was a real credential-disclosure bug in `request_with_ssrf_guard`:

1. **Auth leaks on redirect** ✅ — `request_with_ssrf_guard` now strips `Authorization` and `Proxy-Authorization` from the headers when a redirect changes origin (`netloc`), copying the headers dict so the caller's dict is never mutated. Same-host redirects keep the credential (matching `requests` semantics). 2 new tests: cross-host redirect drops the bearer token, same-host keeps it. **37/37 green** in `test_ssrf_protection.py`.

2. **Docs omit redirect/metadata SSRF** ✅ — `load_from_api` docstring now documents cloud-metadata blocking (169.254.169.254) and that every redirect hop is re-validated before being followed.

Full suite: **62/62 green** across the seed + ssrf test modules.

ZohaibHassan16:
One thing of note: auth stripping fix from Qodo round only covers half the case it needs to. It compares `urlparse(url).netloc` before/after a redfirect and strips `Authorization`/`proxy-Authorization` if the host changes. But `netloc` doesn't include the scheme so a same host `https -> http` downgrade redirect doesn't trip it. The bearer token gets replayed in cleartext.

Repro:

```
request_with_ssrf_guard(
          "GET", "https://example.com/start",
           session=session,
            headers={"Authorization": "Bearer secret-token"},
}

```

Headers sent on downgraded hop: {'Authorization': 'Bearer secret-token'}

`reuqests`' own `should_strip_auth()` handles the very case (strips on hostname change or https -> http downgrade) which is presumably what "mirroring reuqests' cross host credential stripping" was going for , it just landed the host half and missed the scheme half. Neither of the two new tests (`test_strips_autorization_on_cross_host_redirect`, `test_keeps_authorization_on_same_host_redirect`_) exercises a downgrade on the same host, so it slipped through.

Given this PR's whole point is sending `Authorization: Bearer <api_key>` to external APIs, I had want this closed before merging.


Fix should be a one liner: compare scheme too, or just check `old_scheme == "https" and new_scheme == "http"` alongisde netloc check.

Everything else is good once this is in.

yunaremaia:
Excellent catch @ZohaibHassan16 — you are right, the netloc-only comparison missed the https -> http downgrade on the same host, and the Authorization header would have been replayed in cleartext. That is exactly the kind of leak this PR must not ship with.

**Fixed in `0abea559`:**

1. New `_should_strip_auth()` helper in `semantica/ingest/ssrf.py` that mirrors `requests.utils.should_strip_auth` semantics:
   - strips on **hostname change** (as before),
   - strips on **port change**,
   - strips on **https -> http scheme downgrade** (the gap you found),
   - keeps the credential only for the safe **http -> https upgrade on default ports** (requests own behavior, so upgrades keep working).
2. Two new regression tests in `tests/ingest/test_ssrf_protection.py`:
   - `test_strips_authorization_on_scheme_downgrade` — same host, https -> http redirect: second hop has no `Authorization` (this test fails on the previous code),
   - `test_keeps_authorization_on_scheme_upgrade` — http -> https on default ports: credential survives, matching requests.

**Verified locally:** 39/39 tests in `test_ssrf_protection.py` pass, plus 19/19 in `test_seed_manager.py`. No behavior change for cross-host stripping.

Thanks again for the sharp review — exactly the kind of gap that is hard to see from the inside.

ZohaibHassan16 (approved):
All tests passing. Approved.

PR #960, opened 2026-08-13
feat(export): add opt-in metric_errors column to DistanceExporter

Karunasagar12:
## Summary

Follow-up to #879 (as discussed in its review thread). Adds a `metric_errors` column to `DistanceExporter.compute_pairs()` that lets downstream consumers programmatically distinguish legitimate `None` results (no path between nodes) from computation failures.

## Usage

```python
# Opt-in: include 'metric_errors' in the column list
rows = exporter.compute_pairs(include=["hop_count", "weighted_distance", "metric_errors"])

# row["metric_errors"] == ""                         → all metrics succeeded
# row["metric_errors"] == "hop_count"                → hop computation raised
# row["metric_errors"] == "hop_count,weighted_distance" → both raised
```

## Design decisions

- **Opt-in only** — column only appears when `"metric_errors"` is passed in `include=`. Default export schema is completely unchanged (backward compatible).
- **Metric names, not exception messages** — stable for programmatic filtering (`df[df.metric_errors == '']`) without exposing internal error details. Exception details remain in the warning logs added by #879.
- **Comma-separated** — simple to parse, sortable, filterable. Easy to split in pandas: `df['metric_errors'].str.split(',')`.
- **Internal refactor** — helpers now return `(value, error_name | None)` tuples. This is an internal-only change; the public API (`compute_pairs`, `to_csv`, `to_dataframe`, etc.) is unchanged unless `metric_errors` is explicitly requested.

## Tests

Added `tests/export/test_distance_exporter_metric_errors.py` — 6 tests covering:
- Empty string on success
- Single metric failure recorded
- Multiple failures comma-separated
- Column absent when not requested
- Core distinction: no-path None vs error None
- Default schema unchanged

Also updated the existing `tests/export/test_distance_exporter.py` (from #879) to work with the new tuple return type.

```
============================== 77 passed in 0.88s ==============================
```


Karunasagar12:
Addressed both Qodo findings:

1. **Betweenness errors now tracked** — `_betweenness()` returns a `(dict, error)` tuple like the other helpers, so centrality failures appear in `metric_errors`. Previously they were silently swallowed even with the new column enabled.

2. **Removed unused `_ERROR_COLUMNS` constant** — dead code, the logic is handled inline in `compute_pairs()`.

All 77 tests in `tests/export/` pass.

ZohaibHassan16:
Solid work. Thanks for addressing Qodo findings.


It is merge ready, however there are few minor issues which I would just leave here:


- `include=["metric_errors"]` by itself does nothing useful , no metric names means nothing gets computed, so you just get `{"metric_errors": ""}` on every row no matter what. Tried it by forcing bfs to blow up and passing only `metric_errors`, still comes back empty. Docs show it paired with real metrics which is correct but someone's gonna try it alone first and be confused why it's always empty. Maybe a line in the docstring, or just default to full metrics when only metric_errors is passed.

- Leftover comment near the top about `_ERROR_COLUMNS` is now floating over nothing since you removed the constant.

-  distance_band still just says "distant" whenever hop_count is None, doesn't matter if that's a real no-path or a crash. Since this whole PR is about telling those apart, might be worth a note that people should check metric_errors for hop_count too if they care about distance_band being accurate.

KaifAhmad1 (approved):
Reviewed the diff and the full compute_pairs() loop logic — the (value, error) tuple refactor is correct, the no-path-vs-error distinction genuinely holds (verified via _hop_distance's empty-path vs exception branches), betweenness errors are now tracked per the earlier Qodo finding, and the default schema is unaffected when metric_errors isn't requested. 77 tests pass, including good coverage of the core distinction this PR exists for.

Pushed a changelog entry (4bbc4363) documenting the feature, including the include=["metric_errors"]-alone footgun Zohaib flagged as a known limitation, since that wasn't addressed in code. The leftover comment near the removed _ERROR_COLUMNS constant is still there but purely cosmetic — fine to clean up in a follow-up if anyone cares.

Approving.

PR #967, opened 2026-08-13
fix: decision persistence/query bugs, CJK similarity, and MCP graph query/update tools

### LINKED ISSUE #1211: [BUG] query_decisions(category=...) always returns [] in the same session as the write — find_nodes() exposes category under metadata, not at the top level
## Summary

`query_decisions(category=...)` returns `{"decisions": []}` for a decision recorded moments earlier **in the same server process**. The decision is in the graph — `get_graph_summary` counts it — it is simply unreachable through its own category filter.

Verified on `main` @ `7da3519ca78721a883617151dd08c74c01a9acd1`, semantica 0.6.6, Python 3.12.13, macOS arm64, clean venv, `pip install -e .`.

This is distinct from #1134 (nothing is persisted) and #1140 (decision embeddings / similarity search) — no restart and no serialization are involved here.

## Reproduction — one `semantica-mcp` stdio session, three calls in a row

```
record_decision(category="architecture", scenario=..., reasoning=..., outcome=..., confidence=0.9)
  -> {"decision_id": "...", "status": "recorded"}

get_graph_summary()
  -> {"node_count": 3, "decision_count": 1, "graph_ready": true}

query_decisions(category="architecture")
  -> {"decisions": []}
```

## Root cause

`semantica/mcp_server/__init__.py:161-162`:

```python
nodes = graph.find_nodes(node_type="decision")
results = [n for n in nodes if n.get("category") == category][:limit]
```

`ContextGraph.find_nodes()` (`semantica/context/context_graph.py:2114-2140`) projects every node to exactly four keys — `id`, `type`, `content`, `metadata` — with no top-level `category`. `record_decision()` stores the category in the decision node's properties, which `find_nodes` folds into `metadata`. So `n.get("category")` is always `None` and the comparison is never true, for any category.

Minimal in-process repro of that same expression, no MCP layer:

```python

g.record_decision(category="architecture",
                  scenario="Choose a persistence layer for the agent decision log",
                  reasoning="Postgres gives durability and transactional guarantees the JSON file cannot",
                  outcome="postgres", confidence=0.9)

nodes = g.find_nodes(node_type="decision")
print(sorted(nodes[0]))
print([n for n in nodes if n.get("category") == "architecture"])
print(len([n for n in nodes if (n.get("metadata") or {}).get("category") == "architecture"]))
```

Output:

```
['content', 'id', 'metadata', 'type']
[]
1
```

The node itself is intact — `metadata` holds `category`, `scenario`, `reasoning`, `outcome`, `confidence`, `timestamp`, `decision_maker`.

## Scope

Only the `category` branch is affected:

- `query_decisions()` with no arguments — fine, it uses the unfiltered `find_nodes(...)[:limit]`, so the `semantica://decisions/list` resource still works.
- `query_decisions(query=...)` — a different path (`find_similar_decisions`); that one is the subject of #1140.
- `query_decisions(category=...)` — always empty.

The failure is silent. No exception is raised, so the wrapping `except` never fires and the caller cannot tell "no decisions in that category" from "the filter cannot match anything".

## Suggested fix

Read the category from where `record_decision` puts it:

```python
results = [n for n in nodes
           if (n.get("metadata") or {}).get("category") == category][:limit]
```

`record_decision` also already maintains `self._decision_index[category]` (`context_graph.py:4216`), so a `ContextGraph`-level accessor over that index would be a more durable fix than filtering projected node dicts inside the server — and it would let the category filter compose with the `query` branch instead of being an either/or.

A regression test that records one decision and asserts `query_decisions(category=<that category>)` returns it would lock this down.

Thanks for the project — the decision-intelligence layer is the part we are evaluating, and everything else in that round-trip behaved as documented.

toratto:
A set of correctness fixes for decision persistence/querying and CJK text handling, plus three new MCP graph tools. Discovered while using Semantica as a local knowledge graph backend for a non-English (CJK) domain.

## Bug fixes

- **`_get_graph` (mcp_server)**: call `load_from_file` instead of the non-existent `graph.load`. Previously `SEMANTICA_KG_PATH` was silently ignored and the server started with an empty in-memory graph.
- **`query_decisions` (mcp_server)**: read `category` from `metadata.category`. The top-level category was always empty, so filtering by category returned nothing.
- **`find_precedents` / `query_decisions(query=)` (mcp_server)**: lower the default similarity threshold to `0.05` so short queries (esp. CJK, where token overlap is small) can still match.
- **`_calculate_decision_content_similarity` (context)**: add a character-bigram overlap-coefficient fallback so CJK scenarios (no whitespace tokenization) can match recorded decisions. The previous whitespace Jaccard was always `0` for CJK, so `find_precedents` never returned anything for Chinese input.
- **`load_from_file` (context)**: rebuild `_decisions` / `_decision_index` / `_entity_index` / `_temporal_index` from persisted decision nodes after loading. `save_to_file` does not serialize these internal indexes, so `find_precedents_by_scenario` and decision counts broke after every reload.
- **`extract_entities` / `extract_relations` (mcp_server)**: return the entity `text` field. Previously the spaCy type label was returned as `label` and the actual entity text was dropped. Also expose `model` / `language` / `method` params so non-English NER (e.g. `zh_core_web_sm`) can be used.
- **Explorer `GET /api/decisions`**: coerce the decision `timestamp` to `str`. `DecisionResponse.timestamp` is typed `str` but decision nodes store a float epoch, so the endpoint returned `422 Unprocessable Content`.

## New MCP tools

- **`query_graph`**: node detail, bidirectional neighbours (out-edges up to 5 hops + 1-hop in-edges), and keyword search over node id+content. Fills the gap that the MCP toolset had no graph-traversal/query capability.
- **`update_node`**: update node properties (e.g. action `status`: todo/doing/done) and persist back to `SEMANTICA_KG_PATH`.
- **`delete_node`**: soft-archive a node (`status=archived`, kept for history) and persist.

`update_node` / `delete_node` write back to `SEMANTICA_KG_PATH`, so mutations survive server restarts.

## Docker

- Pin the runtime image to `python:3.13-slim`. `gensim` (a core dependency) has no prebuilt `cp314` wheel and the slim base lacks `gcc`, so `pip install .[exporer]` failed on `python:3.14-slim`. Still satisfies `requires-python >=3.8`.


KaifAhmad1:
@toratto Can you fix qodo findings before we start reviewing it. @Sameer6305 @ZohaibHassan16

ZohaibHassan16:
Besides Qodo's list: this PR touches 4 files, ~300 lines, 7 bug fixes and 3 new MCP tools, and there's zero test coverage added anywhere.. Given findings 1 and 2 of Qodo bot above are the kind of thing a unit test on the reload path would've caught, I'd want at least minimal tests before merging, save/reload round-trip preserving metadata, and a basic smoke test on the 3 new MCP tools. @toratto

Sameer6305 (dismissed):
Thanks @toratto for the solid work on this PR. The decision persistence fixes, CJK similarity support, MCP graph loading, and the new graph tools address meaningful gaps, and the overall direction is good.

During review, I found a few correctness gaps that needed tightening:

* decision indexes could remain stale across reloads/mutations
* decision metadata/entities were not fully reconstructed after persistence
* `query_graph` had `limit=0` and inbound-result handling issues
* the initial bigram similarity approach could produce false 1.0 matches for very short queries
* Markdown/`from_dict`/`clear()` paths also needed to maintain the same decision-index invariant
* the PR had no regression coverage for the persistence and new MCP-tool paths

These were addressed in **`f454c48` — `fix: harden decision persistence and MCP graph tools`**, with the related regression coverage added as part of the same fix.

I’ve reviewed the final changes and the relevant test suites. **Approved from my side and ready for your final look, @KaifAhmad1 .**

KaifAhmad1:
## Final review

The `extract_relations` MCP tool crash (called `RelationExtractor.extract_relations()` without the required `entities` arg — TypeError on every invocation) is fixed in 84ccc7c0, with a regression test that fails against the prior code and passes now.

Re-reviewed the branch as it stands. Two correctness issues left that I'd consider blocking, plus two smaller ones:

**Blocking**

1. **CJK bigram-similarity fallback over-triggers on ordinary English text** (`context_graph.py` `_calculate_decision_content_similarity`, ~L5052). Verified by execution: two unrelated English decisions ("employee vacation request approval process" vs. "server infrastructure migration to cloud provider") score `0.163` similarity purely from incidental character-bigram overlap. Combined with this PR's new `min_similarity=0.05` default for `find_precedents`/`query_decisions(query=)`, unrelated decisions will now surface as "similar" for normal English queries — the CJK fix leaks into the general (non-CJK) path. Suggest gating the bigram fallback on a token/whitespace-density heuristic so it only kicks in for CJK-like input, or requiring both word_sim and bigram_sim to agree above a floor.

2. **`recorded_at` is lost on every reload** (`context_graph.py` `_rebuild_decision_indexes`/`_sync_decision_from_node`, ~L4899/L4984). Confirmed by execution: `record_decision()` sets an ISO timestamp, but `_add_decision_to_graph` never persists it as a node property, so it reverts to `""` after `save_to_file`→`load_from_file`, and also after any `update_node` call. This undercuts the PR's own "persistence round-trip fidelity" claim; not covered by `test_core_fields_preserved_after_reload` (checks category/scenario/outcome/confidence/entities/decision_maker, not `recorded_at`).

**Non-blocking**

3. `query_graph` neighbors mode: outbound honors `depth` (1–5 hops), but inbound is hardcoded to exactly 1 hop regardless of `depth` (`mcp_server/__init__.py` ~L423), contradicting the tool's documented "traverse up to depth hops" behavior.
4. `from_dict()`'s new call to `_rebuild_decision_indexes()` runs outside `self._lock` (`context_graph.py` ~L3507), unlike the two sibling call sites (`load_from_file`, `_load_markdown_directory`) added in this same diff, which do hold the lock. Race risk under concurrent mutation.

Minor cleanup (not blocking): `_tool_query_decisions`'s `n.get("category") == category` top-level check is dead code since `find_nodes()` never puts `category` at the top level (`mcp_server/__init__.py` ~L190); `query_graph`'s inbound scan calls `find_edges()` and materializes every edge in the graph for a single-node lookup, O(E) instead of O(degree); and `_rebuild_decision_indexes`/`_sync_decision_from_node` still duplicate ~40 lines of near-identical decision-reconstruction logic, which is how #2 above happened — a shared helper would prevent the next divergence.

Happy to push fixes for 1 and 2 if that's the plan for this PR.

KaifAhmad1 (approved):
Both blocking issues from my final review are now fixed in 88d73189, verified by direct execution and full test suite (679 passed, 14 skipped, no failures):

1. **Bigram similarity false-positives**: gated the character-bigram fallback in `_calculate_decision_content_similarity` to only activate for CJK-like scripts or single-token queries, matching its documented intent. Confirmed the reported unrelated-English-sentence case now scores 0.0 instead of 0.163.
2. **`recorded_at` lost on reload**: `_add_decision_to_graph` now persists `recorded_at` as a node property, so the existing read-back logic in `_rebuild_decision_indexes`/`_sync_decision_from_node` recovers it correctly. Verified with a save→load round trip.

Combined with the earlier `extract_relations` crash fix (84ccc7c0), the remaining non-blocking items (query_graph inbound depth, from_dict lock scope, dead code in the category filter, O(E) inbound scan, index-rebuild duplication) are reasonable follow-up work rather than merge blockers. Approving.

PR #972, opened 2026-08-13
refactor(seed): remove obsolete requests import handling in load_from_api

### LINKED ISSUE #949: refactor: remove obsolete requests import exception handling in SeedDataManager
## Description

During the review of PR #942, an obsolete exception-handling path was identified in `SeedDataManager.load_from_api()`.

Previously, the method performed a lazy import of `requests` and therefore had exception handling around that import.

PR #942 replaces the direct `requests.get()` call with the module-level `request_with_ssrf_guard()` helper, so the original lazy import is no longer present.

However, the surrounding `except (ImportError, OSError)` handling remains.

## Why this matters

The remaining exception handling no longer corresponds to an operation that can raise those import-related exceptions at that location.

This makes the control flow misleading and suggests that the method still performs a local `requests` import when it does not.

Dead or misleading exception handling makes future maintenance harder and can obscure the actual failure modes of the method.

## Expected behavior

Review the existing `except (ImportError, OSError)` block and remove it if it is no longer required.

If there is an intentional compatibility reason for retaining it, document that reason explicitly.

The cleanup should not alter the existing `ProcessingError` behavior for actual request, validation, parsing, or network failures.

## Scope

This is a **pre-existing cleanup issue** identified during the review of PR #942. It is not a blocker for the SSRF fix.

## Acceptance Criteria

* [ ] Verify whether the `ImportError` / `OSError` handling is still reachable or necessary.
* [ ] Remove the obsolete handling if it has no valid purpose.
* [ ] Preserve existing error handling for genuine API/request failures.
* [ ] Add or update tests if required.
* [ ] Run the relevant seed manager test suite successfully.

pravit-amp:
## Description

`SeedDataManager.load_from_api()` ended with `except (ImportError, OSError)` raising `ProcessingError("requests library not available. Install with: pip install requests")`. Both arms are wrong:

- **`ImportError` can't realistically fire.** `requests` is a *core* dependency (`dependencies` in `pyproject.toml`), not an optional extra, so the lazy import it was guarding served no purpose.
- **`OSError` was actively harmful.** `requests.exceptions.RequestException` subclasses `OSError`, so `ConnectionError`, `Timeout`, and the `HTTPError` from `raise_for_status()` were *all* caught by that arm and reported as "requests library not available" — telling users to reinstall a library that was already installed, while hiding the real failure and dropping the exception chain.

This removes the block and hoists the import to module scope, matching how the rest of the ingest stack (`api_ingestor`, `web_ingestor`, `ssrf`, …) imports `requests`. Genuine failures now fall through to the pre-existing `ProcessingError(f"Failed to load from API: {e}") from e`, which was already there and is unchanged.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring

Filed as a refactor, but the `OSError` overlap made it a real behavior bug, so both are ticked.

## Related Issues

Closes #949

## Changes Made

- Removed the obsolete `except (ImportError, OSError)` block from `SeedDataManager.load_from_api()`; request, parsing, and network failures now surface through the existing `Failed to load from API: {e}` path with `from e` chaining intact.
- Hoisted `import requests` out of the `try` body to module scope, with a comment recording *why* there is no import guard, so the lazy-import-plus-fallback pattern doesn't get reintroduced.
- Corrected the `load_from_api` docstring, which advertised `ProcessingError` "if ... requests library is not available".
- Added regression tests for `ConnectionError`, `Timeout`, `HTTPError`, a failing `raise_for_status()`, and a malformed JSON body — each asserting the real cause reaches the caller and the "requests library not available" text does not.

## Testing

- [x] Tested locally

Build not run — the change touches no packaging metadata, only a module body and a test file.

The new tests were verified to fail against the pre-change code (4 of the 5 fail; the malformed-JSON case passes before and after, since `ValueError` was never caught by the removed block — it's there to pin that behavior).

### Test Commands

```bash
pytest tests/test_seed_manager.py tests/seed/ -q
```

`28 passed`. `black`/`flake8` report diffs on both touched files, but those are pre-existing on `main` (unrelated lines); the added lines are clean.

## Documentation

The only stale doc was the method's own `Raises:` section, corrected in this PR.

## Breaking Changes

**Breaking Changes**: No

Every failure mode still raises `ProcessingError` from `load_from_api()`; only the message text changes, and only for cases that were previously mislabeled. Anything matching on the literal string `"requests library not available"` would need updating — a grep found no such callers or tests in the repo.

## Additional Notes

Two things for reviewers:

1. **Overlap with #942.** That PR replaces `requests.get()` here with `request_with_ssrf_guard()` and touches the same region, so whichever lands second needs a small rebase. #949 was filed as a pre-existing cleanup, explicitly not a blocker for the SSRF fix, so this is based on `main` rather than on that branch.
2. **Sibling left out of scope.** `load_from_database()` (same class, ~50 lines up) has the identical `except (ImportError, OSError)` shape, mislabeling real DB and socket errors as "Database ingestion module not available". Deliberately not touched here to keep the PR scoped to the linked issue — happy to fix it in a follow-up, or in this PR if you'd prefer them together.


ZohaibHassan16 (commented):
The fix in this PR looks stale against current main , which is probably why GitHub is showing a conflict. After #942, `seed_manager.py` doesn't call `requests.get()` or use `requests` directly anymore. It calls `request_with_ssrf_guard` from `..ingest.ssrf`.

So I don't think we need to hoist `import requests` anymore since there isn't a direct requests call to fix.

I tried the smaller change on current main: just remove the `except (ImportError, OSError)` block and don't change the imports. That fixes the issue. The same `ConnectionError` repro now correctly reports `"Failed to load from API: ..."` and keeps the real exception as the cause.

I also ran:

```text
tests/test_seed_manager.py tests/seed/
```

and everything passed.

The new tests need a small update too. They currently patch:

```python
@patch("requests.get")
```

but current main doesn't call that anymore in this path. They should patch:

```python
semantica.seed.seed_manager.request_with_ssrf_guard
```

instead. That's also the same pattern used by the tests that came in with #995/#942.

The docstring change is still correct and needed. 

Could you rebase this onto current main and update it to:

* remove only the `except (ImportError, OSError)` block
* don't add/hoist `import requests`
* update the test mocks to patch `request_with_ssrf_guard`

The regression coverage itself looks good .

pravit-amp:
> The fix in this PR looks stale against current main , which is probably why GitHub is showing a conflict. After #942, `seed_manager.py` doesn't call `requests.get()` or use `requests` directly anymore. It calls `request_with_ssrf_guard` from `..ingest.ssrf`.
> 
> So I don't think we need to hoist `import requests` anymore since there isn't a direct requests call to fix.
> 
> I tried the smaller change on current main: just remove the `except (ImportError, OSError)` block and don't change the imports. That fixes the issue. The same `ConnectionError` repro now correctly reports `"Failed to load from API: ..."` and keeps the real exception as the cause.
> 
> I also ran:
> 
> ```
> tests/test_seed_manager.py tests/seed/
> ```
> 
> and everything passed.
> 
> The new tests need a small update too. They currently patch:
> 
> ```python
> @patch("requests.get")
> ```
> 
> but current main doesn't call that anymore in this path. They should patch:
> 
> ```python
> semantica.seed.seed_manager.request_with_ssrf_guard
> ```
> 
> instead. That's also the same pattern used by the tests that came in with #995/#942.
> 
> The docstring change is still correct and needed.
> 
> Could you rebase this onto current main and update it to:
> 
> * remove only the `except (ImportError, OSError)` block
> * don't add/hoist `import requests`
> * update the test mocks to patch `request_with_ssrf_guard`
> 
> The regression coverage itself looks good .

Thanks — you're right, this was stale against main. Rebased onto current main and cut it down to the minimal change you described:

Removed only the except (ImportError, OSError) block.
No import requests in seed_manager.py — confirmed the module reaches the network solely through request_with_ssrf_guard, so there was nothing left to hoist.
Kept the docstring Raises: correction.
The three regression tests now patch semantica.seed.seed_manager.request_with_ssrf_guard, matching the pattern from #942/#995. The test file still imports requests to construct ConnectionError / Timeout / HTTPError — those subclassing OSError is exactly what the tests pin.
tests/test_seed_manager.py tests/seed/ — 33 passed.

I also checked the coverage actually regresses: restoring the old handler on top of current main makes 4 of the 5 new cases fail with "requests library not available" instead of the real cause. The malformed-JSON case passes either way (ValueError isn't an OSError), so it pins existing parse-failure behavior rather than guarding the regression.

Unrelated, spotted while rebasing: seed_manager.py on main imports request_with_ssrf_guard twice (lines 41 and 47, the second alongside parse_bool) — looks like a merge artifact from #942. Left it alone to keep this PR scoped; happy to open a separate issue.

ZohaibHassan16 (approved):
Re-reviewed after the update.

The obsolete error handler is removed, API failures now keep their real cause and the tests cover the request helper correctly.

Looks good to me. Approved.

PR #974, opened 2026-08-13
fix(seed): pass connection string to DBIngestor and stop mislabeling OSError

### LINKED ISSUE #973: [BUG] load_from_database never reaches the database — DBIngestor called without the connection string, and OSError mislabeled as a missing module
## Bug Description

`SeedDataManager.load_from_database()` has two defects in the same method:

1. **It never reaches the database.** Both branches call `DBIngestor` with the wrong arity — `db_ingestor.execute_query(query)` and `db_ingestor.export_table(table_name)` — but those methods are `execute_query(self, connection_string, query, **params)` and `export_table(self, connection_string, table_name, ...)`. The user's `connection_string` is passed to the `DBIngestor` constructor as config and then never used, so every call raises `TypeError` before any connection is attempted.

2. **`except (ImportError, OSError)` mislabels real failures.** The handler wraps the entire ~30-line body, so any `OSError` from the ingestor (socket errors, file-permission errors on SQLite paths, driver-level `IOError`s) is reported as *"Database ingestion module not available. Install required dependencies."* — advice that is unrelated to what actually failed. It also re-raises without `from e`, dropping the exception chain, so the original traceback is lost.

Defect 2 is latent behind defect 1 today: the method dies on `TypeError` before any code that could raise `OSError` runs. Fixing 1 without 2 would make the mislabeling live.

This is the sibling of #949, which fixed the same `except (ImportError, OSError)` pattern in `load_from_api()` (there, `requests.exceptions.RequestException` subclasses `OSError`, so every connection error was reported as a missing library). `load_from_database` was left out of that PR to keep it scoped to the linked issue.

## Steps to Reproduce

1. `pip install sqlalchemy` (see Additional Context — it is not declared in `pyproject.toml`)
2. Run the script in [Code Example](#code-example)
3. Observe that no connection is ever attempted, and that a raw `OSError` from the ingestor is reported as a missing module

## Expected Behavior

1. `load_from_database("sqlite:///:memory:", query="SELECT 1")` connects using the supplied connection string and returns the query results as a list of dicts, per the method's docstring and `Example:` block.
2. A genuine connection or query failure raises `ProcessingError("Failed to load from database: <real cause>")` with the original exception chained via `from e`. Only an actual failure to import `DBIngestor` produces the "module not available" message.

## Actual Behavior

1. Every call raises `ProcessingError` wrapping a `TypeError` about a missing positional argument, regardless of the connection string. The method is non-functional for both the `query` and `table_name` paths.
2. Any `OSError` raised inside the block is reported as *"Database ingestion module not available. Install required dependencies."*, with `__cause__` set to `None`.

## Environment

- **OS**: macOS 26.5.2 (arm64)
- **Python Version**: 3.12.13
- **Semantica Version**: 0.6.5

## Code Example

```python
from unittest.mock import patch

from semantica.seed.seed_manager import SeedDataManager

manager = SeedDataManager()

# 1. Neither branch ever reaches the database
for kwargs in ({"query": "SELECT 1"}, {"table_name": "users"}):
    try:
        manager.load_from_database("sqlite:///:memory:", **kwargs)
    except Exception as e:
        print(f"{list(kwargs)[0]:>11} -> {type(e).__name__}: {e}")

# 2. A raw OSError from the ingestor is reported as a missing module.
#    (Patched here because defect 1 prevents reaching this code path at all.)
import semantica.ingest.db_ingestor as dbi

with patch.object(dbi.DBIngestor, "execute_query",
                  side_effect=OSError(111, "Connection refused")):
    try:
        manager.load_from_database("postgresql://u:p@10.0.0.9/db", query="SELECT 1")
    except Exception as e:
        print(f"  raw OSError -> {type(e).__name__}: {e}  (cause={e.__cause__!r})")
```

## Error Messages / Logs

```
      query -> ProcessingError: Failed to load from database: DBIngestor.execute_query() missing 1 required positional argument: 'query'
 table_name -> ProcessingError: Failed to load from database: DBIngestor.export_table() missing 1 required positional argument: 'table_name'
  raw OSError -> ProcessingError: Database ingestion module not available. Install required dependencies.  (cause=None)
```

## Additional Context

- **Why this was never caught:** `tests/test_seed_manager.py::test_load_from_database` mocks `DBIngestor` entirely, so the mocked `execute_query(query)` accepts any signature and the arity mismatch is invisible. `test_load_from_database_import_error` patches `sys.modules` and only exercises the `ImportError` arm. Neither test touches the real call, which is exactly the failure mode `CLAUDE.md` warns about: *"A test that mocks away the exact interaction you're trying to regression-test doesn't catch the bug it claims to."*
- **The `ImportError` arm should be kept, not removed** (unlike in #949): `from ..ingest.db_ingestor import DBIngestor` is a genuine import that can fail, and the message is accurate for it. It just needs to be narrowed to the import statement itself.
- **`sqlalchemy` is not declared anywhere in `pyproject.toml`** — not as a core dependency and not as an optional extra — despite `DBIngestor` requiring it. `DatabaseConnector.connect()` already handles this precisely (`"sqlalchemy is required for database ingestion. Install with: pip install sqlalchemy"`), which is a further argument that the broad catch-all in `seed_manager` adds nothing but noise. Whether to declare it as an extra is out of scope here and probably deserves its own issue.
- Related: #949 (`load_from_api` cleanup), #942 (SSRF fix touching the same file).

## Possible Solution

In `semantica/seed/seed_manager.py`:

1. Pass the connection string through to the ingestor: `db_ingestor.execute_query(connection_string, query)` and `db_ingestor.export_table(connection_string, table_name)`.
2. Move `from ..ingest.db_ingestor import DBIngestor` into its own narrow `try` / `except ImportError` that raises the "module not available" `ProcessingError`, and drop `OSError` from the handler entirely.
3. Let everything else fall through to the existing `except Exception as e: raise ProcessingError(f"Failed to load from database: {e}") from e`, preserving the cause.

In `tests/test_seed_manager.py`:

4. Assert the ingestor is called with the connection string (`mock.execute_query.assert_called_once_with(conn_str, query)`) so the arity is actually pinned, and add a regression test that a raw `OSError` surfaces the real cause rather than the install message.

pravit-amp:
## Description

`SeedDataManager.load_from_database()` never reached a database. Both branches called `DBIngestor` with the wrong arity:

```python
db_ingestor = DBIngestor(config={"connection_string": connection_string})
result = db_ingestor.execute_query(query)          # execute_query(self, connection_string, query, **params)
table_data = db_ingestor.export_table(table_name)  # export_table(self, connection_string, table_name, ...)
```

The user's connection string went into the constructor config and was never read, so every call died on `TypeError` before a connection was attempted:

```
ProcessingError: Failed to load from database: DBIngestor.execute_query() missing 1 required positional argument: 'query'
```

The method's `except (ImportError, OSError)` block compounded it. It wrapped the whole ~30-line body, so any `OSError` from the ingestor — socket errors, permission errors on SQLite paths, driver-level `IOError`s — was reported as *"Database ingestion module not available. Install required dependencies."*, and re-raised without `from e`, discarding the cause. That arm was latent behind the `TypeError`; fixing the arity alone would have made it live.

This is the sibling of #949, which removed the same handler from `load_from_api()`. It was deliberately left out of that PR to keep it scoped, then investigated separately — which is how the arity bug surfaced.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #973

## Changes Made

- Pass the connection string through to the ingestor: `execute_query(connection_string, query)` and `export_table(connection_string, table_name)`.
- Narrowed the import guard to the `DBIngestor` import statement itself, keeping the "module not available" message where it is accurate and chaining the cause with `from e`.
- Dropped `OSError` from that handler so genuine failures fall through to the pre-existing `ProcessingError(f"Failed to load from database: {e}") from e`.
- Switched the `DBIngestor` patch in the tests to `autospec=True` and pinned both calls with `assert_called_once_with(...)`.
- Added a parametrized regression test for `OSError` / `PermissionError`, and extended the existing import-error test to assert the cause is chained.

## Testing

- [x] Tested locally

Build not run — no packaging metadata is touched, only a module body and a test file.

All four `load_from_database` tests were verified to fail against the pre-change source and pass after, including the pre-existing `test_load_from_database`, which asserted the wrong arity.

Beyond the unit tests, a live call now reaches `DatabaseConnector.connect()` with the real connection string on both branches instead of raising `TypeError`. End-to-end against an actual database stops at `"sqlalchemy is required for database ingestion"` — that dependency is not declared anywhere in `pyproject.toml`, which is noted in #973 and left alone here.

### Test Commands

```bash
pytest tests/test_seed_manager.py tests/seed/ -q
```

`25 passed`. `black` / `flake8` report diffs on both files, but they are pre-existing on `main` and unrelated to these lines (the test module has never been black-formatted; added code follows the surrounding style).

## Documentation

The docstring's `Raises:` section — "If database connection fails, query fails, or DBIngestor module is not available" — is accurate as written and now actually describes what the code does.

## Breaking Changes

**Breaking Changes**: No

The method previously failed unconditionally, so there is no working behavior to break. Every failure mode still raises `ProcessingError`; only the message changes, and only for cases that were mislabeled. Nothing in the repo matches on the `"Database ingestion module not available"` string apart from the test that asserts it, which still passes.

Worth flagging for reviewers: this PR makes `load_from_database()` genuinely attempt a connection for the first time, so callers who were silently swallowing its `ProcessingError` will now see real driver and connection errors instead of a `TypeError`. That is the intended fix, but it is a visible change in what reaches those callers.

## Additional Notes

- **Why CI never caught this:** `test_load_from_database` patched `DBIngestor` with an unspecced `MagicMock`, which accepts any call signature, so the missing argument was invisible. `test_load_from_database_import_error` only exercised the `ImportError` arm. `autospec=True` is what turns that test into a real guard — per `CLAUDE.md`, *"A test

### DIFF
--- a/semantica/seed/seed_manager.py
+++ b/semantica/seed/seed_manager.py
@@ -398,11 +398,13 @@ class SeedDataManager:
             ...     entity_type="Person"
             ... )
         """
-        # Guard only the import itself: a broader guard would report unrelated
-        # failures below as a missing module.
+        # OSError is caught here because module loading can fail on the
+        # filesystem, which does mean the module is unavailable. The guard stays
+        # wrapped around the import alone: over the body below it would report
+        # unrelated connection and driver failures as a missing module.
         try:
             from ..ingest.db_ingestor import DBIngestor
-        except ImportError as e:
+        except (ImportError, OSError) as e:
             raise ProcessingError(
                 "Database ingestion module not available. Install required dependencies."
             ) from e
--- a/semantica/seed/seed_manager.py
+++ b/semantica/seed/seed_manager.py
@@ -398,20 +398,27 @@ class SeedDataManager:
             ...     entity_type="Person"
             ... )
         """
+        # Guard only the import itself: a broader guard would report unrelated
+        # failures below as a missing module.
         try:
             from ..ingest.db_ingestor import DBIngestor
+        except ImportError as e:
+            raise ProcessingError(
+                "Database ingestion module not available. Install required dependencies."
+            ) from e
 
+        try:
             # Initialize DB ingestor
             db_ingestor = DBIngestor(config={"connection_string": connection_string})
 
             # Execute query or export table
             if query:
                 # Execute custom query
-                result = db_ingestor.execute_query(query)
+                result = db_ingestor.execute_query(connection_string, query)
                 records = result if isinstance(result, list) else [result]
             elif table_name:
                 # Export table
-                table_data = db_ingestor.export_table(table_name)
+                table_data = db_ingestor.export_table(connection_string, table_name)
                 records = table_data.rows if hasattr(table_data, "rows") else []
             else:
                 raise ProcessingError("Either 'query' or 'table_name' must be provided")
@@ -428,10 +435,6 @@ class SeedDataManager:
             self.logger.info(f"Loaded {len(records)} records from database")
             return records
 
-        except (ImportError, OSError):
-            raise ProcessingError(
-                "Database ingestion module not available. Install required dependencies."
-            )
         except Exception as e:
             raise ProcessingError(f"Failed to load from database: {e}") from e
 
--- a/tests/test_seed_manager.py
+++ b/tests/test_seed_manager.py
@@ -1,5 +1,6 @@
 
 import pytest
+import builtins
 import os
 import json
 import csv
@@ -177,6 +178,23 @@ def test_load_from_database_import_error(seed_manager):
             seed_manager.load_from_database("sqlite:///:memory:", query="SELECT 1")
         assert "Database ingestion module not available" in str(excinfo.value)
         assert isinstance(excinfo.value.__cause__, ImportError)
+
+# Module loading can fail on the filesystem, which escapes as OSError rather
+# than ImportError and would otherwise bypass the ProcessingError contract.
+def test_load_from_database_import_os_error(seed_manager):
+    real_import = builtins.__import__
+
+    def failing_import(name, *args, **kwargs):
+        if "db_ingestor" in name:
+            raise OSError(5, "Input/output error")
+        return real_import(name, *args, **kwargs)
+
+    with patch("builtins.__import__", failing_import):
+        with pytest.raises(ProcessingError) as excinfo:
+            seed_manager.load_from_database("sqlite:///:memory:", query="SELECT 1")
+
+    assert "Database ingestion module not available" in str(excinfo.value)
+    assert isinstance(excinfo.value.__cause__, OSError)
 
 @patch("requests.get")
 def test_load_from_api(mock_get, seed_manager):
--- a/tests/test_seed_manager.py
+++ b/tests/test_seed_manager.py
@@ -109,36 +109,66 @@ def test_load_from_json_not_found(seed_manager):
     with pytest.raises(ProcessingError):
         seed_manager.load_from_json("non_existent.json")
 
-@patch("semantica.ingest.db_ingestor.DBIngestor")
+# autospec so the mock enforces DBIngestor's real signatures: an unspecced
+# MagicMock accepts any arity and hid the missing connection_string argument.
+@patch("semantica.ingest.db_ingestor.DBIngestor", autospec=True)
 def test_load_from_database(mock_db_ingestor_cls, seed_manager):
-    mock_db_ingestor = MagicMock()
-    mock_db_ingestor_cls.return_value = mock_db_ingestor
-    
+    mock_db_ingestor = mock_db_ingestor_cls.return_value
+
     # Mock execute_query result
     mock_db_ingestor.execute_query.return_value = [{"id": 1, "name": "Alice"}]
-    
+
     records = seed_manager.load_from_database(
         connection_string="sqlite:///:memory:",
         query="SELECT * FROM users",
         entity_type="User"
     )
-    
+
     assert len(records) == 1
     assert records[0]["id"] == 1
     assert records[0]["entity_type"] == "User"
-    mock_db_ingestor.execute_query.assert_called_once_with("SELECT * FROM users")
+    mock_db_ingestor.execute_query.assert_called_once_with(
+        "sqlite:///:memory:", "SELECT * FROM users"
+    )
 
     # Mock export_table result
     mock_table_data = MagicMock()
     mock_table_data.rows = [{"id": 2, "name": "Bob"}]
     mock_db_ingestor.export_table.return_value = mock_table_data
-    
+
     records = seed_manager.load_from_database(
         connection_string="sqlite:///:memory:",
         table_name="users"
     )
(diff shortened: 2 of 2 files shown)


pravit-amp:
@Sameer6305 — could you help me understand the process here?

You assigned #973 to me on 2026-08-14 04:38, after I commented asking to take it and opened this
PR on 2026-08-13. #995 was opened for the same issue on 2026-08-15 05:41 and merged at 12:18,
which auto-closed #973. This PR has been open since the 13th and is now superseded.

No issue with @yzxcj797 over it — the fix is fine and the outcome for the codebase is the same.
My question is about the workflow rather than that PR.

CONTRIBUTING asks contributors to comment on an issue, so
that's what I did here. If an assignment doesn't reserve the issue in practice, I'd genuinely
rather know — I'd stop asking and just open PRs, which seems worse for everyone but at least
wouldn't leave work stranded.

Two concrete things:

1. Should I close this PR? It's superseded and just sitting in the queue.
2. Is there something I should do differently to make an in-flight PR visible? This one was
   cross-referenced on #973 from the day it opened, so I'm not sure what else would have
   surfaced it.

Asking partly because it's recurring: #1022 and #1023 were opened last night for #1014 and #1015,
which duplicate #1016 and #1017 that I'd already opened after commenting on both issues. Those two
issues are still unassigned, so nobody's done anything wrong — but a clear answer on how
assignment is meant to work would help me avoid spending time on work that gets superseded.

Sameer6305:
Hi @pravit-amp, I understand your concern, and you're right to raise it. Since you had already been assigned #973 and had an active PR, #995 being merged instead was understandably frustrating.

We’re a small team handling a high volume of issues and PRs, and this was a gap in our workflow. We’ll discuss this internally and make sure assigned issues and existing contributor PRs are properly considered before another PR is picked up.

For #973, since the issue is already resolved, please close this PR. Thanks for raising this constructively, and please continue with your other issues/PRs, we’ll handle the assignment flow more carefully going forward.

pravit-amp:
> Hi @pravit-amp, I understand your concern, and you're right to raise it. Since you had already been assigned #973 and had an active PR, #995 being merged instead was understandably frustrating.
> 
> We’re a small team handling a high volume of issues and PRs, and this was a gap in our workflow. We’ll discuss this internally and make sure assigned issues and existing contributor PRs are properly considered before another PR is picked up.
> 
> For #973, since the issue is already resolved, please close this PR. Thanks for raising this constructively, and please continue with your other issues/PRs, we’ll handle the assignment flow more carefully going forward.

Thanks @Sameer6305, I appreciate the response and the plan to tighten up the assignment flow. Closing this one out since [#973](https://github.com/semantica-agi/semantica/issues/973) is already resolved by [#995](https://github.com/semantica-agi/semantica/issues/995) — no concerns from my side. I'll carry on with my other open issues/PRs.

PR #978, opened 2026-08-14
fix(kg): remap relationship endpoints after entity resolution

T1mn:
## Summary

- Remap relationship endpoints from merged entity IDs to the canonical ID.
- Ignore unhashable entity IDs instead of failing graph construction.
- Skip remapping when entity resolution produced no merged IDs.
- Add regression tests for both cases.

## Tests

- `pytest -q tests/kg/test_graph_builder_external.py tests/kg/test_kg.py tests/kg/test_algorithms.py tests/kg/test_entity_pipeline.py`
- 58 passed

### DIFF
--- a/semantica/kg/graph_builder.py
+++ b/semantica/kg/graph_builder.py
@@ -283,18 +283,30 @@ class GraphBuilder:
             if not isinstance(entity, dict):
                 continue
 
-            canonical_id = entity.get("id") or entity.get("entity_id")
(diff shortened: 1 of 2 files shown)


ZohaibHassan16 (approved):
Verified this against the EntityResolver and confirmed this branch fixes it. Also checked the unhashable-ID and no-op-when-no-merges cases, both are good to go. All 58 tests passed.

Approved.

PR #980, opened 2026-08-14
Fix/explorer backend failure states

### LINKED ISSUE #977: Explorer UI doesn't handle backend being down: graph load just hangs forever
## Summary
Was poking around the Explorer frontend and noticed it doesn't quite handle the API failing. If the backend is down or errors out, the graph workspace just spins forever on a progress bar and the landing page still says "SYSTEM ONLINE" either way.

Putting these in one issue since they're kind of the same problem (UI hides backend failures) and the fixes are in nearby code.

Tested both with the backend running (643 nodes / 1,080 edges from the primekg demo graph) and with it stopped.

---

## 1. Graph workspace hangs forever if the load fails (main one)

**Repro:**
1. Don't start the backend (or kill it)
2. Open the app, click "Open Semantica Explorer"
3. It sits on `Preparing graph session · STAGE 1/6` with the bar saying `LIVE`

I left it 35+ seconds, it never errors or times out. No retry either, you have to reload the page.

Dug into as to why, there's three things stacking up:

- `GraphWorkspace.tsx:1228` only pulls `{ data: summary, isLoading, isFetching }` out of `useLoadGraph()`. It never reads `isError` or `error`.
- `useLoadGraph.ts:463` has `retry: 0` so it fails immediately, but `loadingProgress` still has the last progress frame sitting in it. So this line:
 
```
  // GraphWorkspace.tsx:1904
  const showLoadingOverlay = !graphReady && (isLoading || isFetching || Boolean(loadingProgress));
```

stays `true` forever : `isLoading`/`isFetching` are both false by then but `loadingProgress` is truthy.

- `GraphLoadingOverlay.tsx:169` only takes `progress`, `visible`, `showGraphBehind`. There's no error prop, so it can't show an error even if we passed one.

One thing to flag: `GraphWorkspaceShell.tsx:354` actually *does* handle `isError, error` properly. But I grepped and nothing imports it, looks like dead code. So the working version exists, it's just not the one being used. Probably why this slipped through.

**Fix idea:**

- Add `error` + `onRetry` props to `GraphLoadingOverlay`, show an error card with a Retry button instead of the progress stuff
- Pull `isError`, `error`, `refetch` in `GraphWorkspace` and pass them down, plus add the error case to `showLoadingOverlay`
- Separately we should probably decide what to do with `GraphWorkspaceShell.tsx` , having two copies that drifted apart is how this happened

---

## 2. Landing page always says "SYSTEM ONLINE"

`App.tsx:1577-1582` : the green dot and text are just hardcoded:

```
<div className="landing-status-bar">
  <div className="landing-status-dot" />
  <span className="landing-status-text">System Online</span>

```

With the backend completely stopped it still shows green "SYSTEM ONLINE". The metrics do fall back to "Live" / "Ready" / "Standby" but nothing tells you anything's actually wrong.

The annoying part is the component already has `stats.ready` in state and uses it for the 4th metric card. It's just never hooked up to the status bar. Should be a pretty small fix.

---

## 3. Smaller stuff I noticed

- No way to dismiss search results, the results strip (`GraphWorkspace.tsx:2841`) has no close button and pushes the graph down until you search again. The selected-edge card right next to it has a Close button, so it feels inconsistent.
- Search scores show as `96.900` / `138.000` (`result.score.toFixed(3)`), the 3 decimals don't really mean anything to a user.
- No typeahead on search : nothing happens while typing, you have to submit to see anything, even though `/api/graph/search` is there.

---

## Setup I used

- Frontend: vite dev server (6.4.3)
- Backend: `python -m semantica.explorer --graph demo_out_smoke/primekg_demo_graph.json`
- Checked with backend up and with it stopped

ZohaibHassan16:
## Summary

Fixes #977.

This fixes a few Explorer UI issues around backend errors and search.

The main issue was that if the backend was down or returned an error, the graph page would just keep showing the loading/progress UI forever. The landing page also always showed "System Online" even when the backend could not be reached.

I also fixed a few smaller search UX issues I found while testing this.

### Graph load error state + retry

`useLoadGraph()` already returned `isError` and `error`, but they were not being used.

Because retries are disabled (`retry: 0`), a failed graph request would leave the loading overlay stuck on the last progress state. There was no error message or retry option, so the only way to try again was refreshing the whole page.

`GraphLoadingOverlay` now takes `error` and `onRetry`.

When loading fails, it shows:

* the actual fetch error message
* a Retry button
* retry without needing a full page reload

### Landing page connection status

The landing page status was hardcoded to always show the green dot and "System Online".

I replaced the old `ready` boolean with three states:

* `checking`
* `online`
* `offline`

The status dot, text and fourth metric card now all use the same state, so they should stay in sync.

When the backend cannot be reached, the page now shows "Backend Unreachable" instead of "System Online".

### Search result dismiss + score display

The search results strip did not have a close button, so it stayed open and kept taking space from the graph until another search happened.

I added a small header with:

* result count
* dismiss button

I also changed relevance scores to display as whole numbers instead of values like `96.900` or `138.000`.

### Search typeahead

The graph search input only did something after submitting the form.

I added a debounced typeahead using the existing search endpoint.

Current behavior:

* waits 250ms before searching
* shows matching suggestions in a combobox
* Up/Down arrows move through suggestions
* Enter selects the highlighted suggestion
* clicking a suggestion jumps to that node
* Escape closes the dropdown
* `aria-activedescendant` tracks the active result

Selecting a result also clears the current query.

## Testing

I tested these against the running app:

* Killed the backend -> graph workspace showed the error card with the real fetch error
* Landing page changed to red "Backend Unreachable"
* Restarted the backend and clicked Retry -> graph loaded again without refreshing the page
* Landing page changed back to green "System Online"
* Typeahead returned real backend results while typing
* Arrow keys moved the selected suggestion correctly
* `aria-activedescendant` updated with the selected suggestion
* Clicking a suggestion focused the node and cleared the query
* Escape closed the typeahead without clearing the query
* Search result dismiss button removed the results strip and gave the canvas its full height again
* Search scores displayed as whole numbers
* `npx tsc -b` passes
* `npm run test:graph-store`, `test:graph-workspace` and `test:plugin-registry`: 42/42 tests pass
* `npm run build` succeeds
* Explorer bundle structure is unchanged


ZohaibHassan16:
Fixed Qodo's finding

Sameer6305 (dismissed):
Hi @ZohaibHassan16, verified the fix mechanism directly against the code, it's solid.

useLoadGraph() genuinely returns real isError/error/refetch from React Query, showLoadingOverlay correctly resolves once isError is true even with stale loadingProgress sitting around, Retry calls a real refetch() (not just a UI reset), and the landing page's checking/online/offline state is driven by an actual fetch outcome with no stuck-state path and no drift between the status dot and the 4th metric card. Good work on all of that.

One thing needs addressing before merge, though it's the root cause the issue itself called out, not just a symptom:

**GraphWorkspaceShell.tsx already had a working isError/error implementation, and it's still dead code after this PR.** It's not imported anywhere (confirmed — no file imports it, App.tsx loads GraphWorkspace directly). This PR adds the fix as a third, separate implementation in GraphWorkspace.tsx instead. So after this merges, we'd have three places with overlapping graph-loading/error-handling logic — GraphWorkspaceShell.tsx (dead), the old GraphWorkspace.tsx logic being replaced, and the new one — rather than fewer. The issue named "two copies that drifted apart" as the actual reason this bug slipped through in the first place, so leaving a third copy around undermines the fix.

Could you either:

1. Delete GraphWorkspaceShell.tsx if it's genuinely unused/superseded, or
2. If it's meant to be used somewhere (a route or feature flag I'm missing?), wire it up and reuse the same error-handling logic instead of duplicating it.

Either is fine, just want to avoid us being back here again in a few months with a fourth copy.

Small, non-blocking note: graphLoadError is typed Error | null in GraphWorkspace.tsx, but React Query actually returns it as unknown, the instanceof check handles it safely at runtime, so this isn't a bug, just worth tightening the type annotation while you're in there.

Requesting changes on the dead-code question, everything else here looks ready to go once that's resolved.

ZohaibHassan16:
Deliberately leaving the GraphWorkspaceShell.tsx deletion out of this PR, see #981 . Makes it easier to track and test.

ZohaibHassan16:
Also graphLoadError type is already correct as-is.

`useLoadGraph.ts` calls `useQuery<GraphLoadSummary>(...)` with only the first generic provided. In TanStack Query v5, the second generic (`TError`) defaults to `Error` and I couldn't find any `Register` override in the app changing that.

So `error` is actually typed as `Error | null`, not `unknown`.

I also checked it with TypeScript instead of only relying on the docs:

```ts
const asError: Error | null = error;   // works
const asString: string = error;        // TS2322
```

The compiler error names the type directly, so that confirms it too.

I’d still keep the `instanceof Error` check, just for a different reason. TypeScript types aren’t enforced at runtime, so a query function could still reject with something that isn’t actually an `Error`.

But there’s no `unknown` → `Error | null` cleanup to make here. It’s already typed as `Error | null`. Thank you for the thorough review.

Sameer6305:
@KaifAhmad1 approved from my side.

KaifAhmad1 (dismissed):
Verified independently (worktree + \`npm ci\`), not just reading the description:

- Confirmed in \`useLoadGraph.ts\` that \`useQuery\` returns the full result (isError/error/refetch all real) with \`retry: 0\`, and the uncaught \`fetch()\` inside \`queryFn\` genuinely propagates a backend-down failure instead of being swallowed — matches the root cause in #977.
- \`npx tsc -b\`: clean.
- \`test:graph-store\` / \`test:graph-workspace\` / \`test:plugin-registry\`: **42/42 passed**.
- \`npm run build\`: succeeds.
- Traced the stale-suggestion abort fix (\`AbortController\` in the typeahead effect) — correctly cancels in-flight requests on re-debounce and on selection-clear.

Thanks @ZohaibHassan16 for the thorough writeup and testing notes, and @Sameer6305 for catching the dead-code question early and getting it properly scoped out to #981 instead of blocking here. Approving for merge.

KaifAhmad1 (dismissed):
Re-approving after the changelog-only follow-up commit — no code changed, review above still stands.

KaifAhmad1 (approved):
Re-approving after merging main to resolve the CHANGELOG.md conflict with #932 — no functional code touched, review above still stands.

PR #983, opened 2026-08-14
fix(context): honor explicit causal edges in decision tracing

### LINKED ISSUE #975: trace_decision_chain() ignores relationships recorded via add_causal_relationship() (returns [] despite stored CAUSED edge)
## Summary

`trace_decision_chain()` never uses the causal relationships recorded via `add_causal_relationship()`. The edge is stored in the graph (visible as a `ContextEdge` with `edge_type='CAUSED'`), but `trace_decision_causality()` infers "potential causes" purely from shared NER entities plus timestamps. When entity extraction finds nothing (which for us was every decision text we tried, both German and English), the returned chain is always `[]` — even though an explicit `CAUSED` edge exists between the two decisions.

`analyze_decision_impact()` has a related problem: "direct_influence" is computed from shared entities/category, so explicitly linked decisions only ever show up under "indirect_influence".

## Reproduction (semantica 0.6.5, Python 3.13, pip install)

```python

a = g.record_decision(category="k", scenario="Server Alpha fails",
                      reasoning="PSU defect on server Alpha",
                      outcome="flagged", confidence=0.9)
b = g.record_decision(category="k", scenario="Failover to server Beta",
                      reasoning="Failover triggered because of server Alpha outage",
                      outcome="approved", confidence=0.9)
g.add_causal_relationship(a, b, relationship_type="CAUSED")

print(g._decisions[a]["entities"])   # []
print(g._decisions[b]["entities"])   # []
print(g.trace_decision_chain(b))     # []  <- expected: chain containing a -> b

edges = list(g.edges.values()) if isinstance(g.edges, dict) else list(g.edges)
print([e for e in edges if getattr(e, "edge_type", "") == "CAUSED"])
# [ContextEdge(source_id=<a>, target_id=<b>, edge_type='CAUSED', ...)]  <- the edge IS there
```

Observed: `trace_decision_chain(b)` returns `[]` although the explicit `CAUSED` edge from `a` to `b` is stored in the graph.

Expected: the chain should contain the explicitly recorded relationship — that is what the README's decision-intelligence lifecycle promises (`record → link → trace`).

## Analysis

In `context/context_graph.py`, `trace_decision_causality()` builds `potential_causes` only from `self._entity_index` overlap plus timestamp ordering. The edges created by `add_causal_relationship()` are never consulted. So the flagship "full causal ancestry" path silently depends on NER having extracted overlapping entities, and explicit links have no effect on the trace.

Same file, `analyze_decision_influence()`: `direct_influence` = shared entities + same category; the explicit causal edges only enter through `_find_indirect_decision_influence()`, so an explicitly `CAUSED` decision is reported as "indirect".

## Secondary (README drift)

The README example uses `add_causal_relationship()` with types `triggers · enables · causes · precedes`; the API raises `ValueError: Relationship type must be one of: ['CAUSED', 'INFLUENCED', 'PRECEDENT_FOR']`.

## Suggested fix

Make `trace_decision_causality()` traverse the explicit `CAUSED`/`INFLUENCED`/`PRECEDENT_FOR` edges first (they are the ground truth the caller recorded), and use entity/timestamp inference only as an additive heuristic — or at least document that the trace is entity-based and unrelated to `add_causal_relationship()`. In `analyze_decision_influence()`, explicitly linked decisions should count as direct influence.

Happy to provide more detail. We evaluated Semantica for an audited decision-ledger use case in a regulated context; precedent search (`find_similar_decisions`) and the policy gate (`check_decision_rules`) worked well for us — this trace disconnect was the one blocker.

hsd2514:
Closes #975

## Problem

`trace_decision_causality()` never consulted the relationships recorded via `add_causal_relationship()`. It inferred "potential causes" purely from shared NER entities plus timestamp ordering, so when entity extraction found nothing, `trace_decision_chain()` returned `[]` even though an explicit `CAUSED` edge was stored in the graph.

`analyze_decision_influence()` had the related problem: `direct_influence` was computed from shared entities and category only, so explicitly linked decisions surfaced under `indirect_influence`.

## Change

**`trace_decision_causality()`** now traverses the explicit `CAUSED` / `INFLUENCED` / `PRECEDENT_FOR` edges first — they are the ground truth the caller recorded — and keeps the entity/timestamp inference as an additive fallback for pairs with no explicit link. Each hop reports the actual relationship type instead of a generic `"influences"`.

Edges whose source has no decision record are skipped. This matters because `from_dict()` / `load_from_file()` restore nodes and edges but do not repopulate `_decisions`; without the guard such an edge raises `KeyError`, which the surrounding `except` converts into `[{"error": ...}]`, discarding every valid chain.

**`analyze_decision_influence()`** counts explicitly linked decisions as direct influence, and subtracts them from the indirect set so the same decision is no longer listed under both.

## Verification

Reproduced the issue on unfixed `main`, then confirmed both symptoms flip:

| | before | after |
|---|---|---|
| `trace_decision_chain(b)` | `[]` | `'Server Alpha fails' --CAUSED--> 'Failover to server Beta'` |
| `analyze_decision_impact(a).direct_influence` | `[]` | `['Failover to server Beta']` |
| `analyze_decision_impact(a).indirect_influence` | `['Failover to server Beta']` | `[]` |

Entity extraction still returns `[]` for both decisions in the fixed run, which is the point — the trace no longer depends on NER finding overlapping entities.

One note on the issue's reproduction: it uses `category="k"` for both decisions, so the category-match shortcut already placed the effect in `direct_influence` and masked the second symptom. The regression test uses differing categories to exercise it properly.

## Tests

Added `tests/context/test_decision_causal_edge_regression.py` (7 tests):

- the issue's reproduction — explicit edge drives the trace with zero entities extracted
- all three relationship types round-trip into the hop's `type`
- multi-hop `a → b → c` chains
- an edge referencing an unrecorded decision does not abort the trace
- explicit link counts as direct influence, with differing categories
- no decision listed under both direct and indirect
- entity-based inference still fires when no explicit edge exists, guarding the fallback path

6 of the 7 fail on unfixed `main` and all pass with the change. The seventh covers pre-existing entity-inference behavior and passes both ways by design.

No wider breakage: `pytest tests/context -k "decision or causal"` goes from 189 to 196 passing (exactly the new tests), with pre-existing failures unchanged.

## Note on formatting

I did not run `black` on `semantica/context/context_graph.py`. That file already fails `black --check` on pristine `main` by ~1214 lines, and 23 of 29 files in `tests/context/` do too, so reformatting would bury a 48-line fix under unrelated churn. The new code follows the surrounding style instead. Happy to reformat if you would prefer it.

## Secondary item from the issue

The README relationship-type drift the issue mentions appears to be already resolved on `main` — it documents `CAUSED` / `INFLUENCED` / `PRECEDENT_FOR`, which match the API. No change needed, so this PR leaves it alone.


ZohaibHassan16:
Thank you for opening the PR.

The main fix does work. I was able to reproduce the same before/after behavior from the PR. On main, `trace_decision_causality()` returns `[]` for a decision that has an explicit stored `CAUSED` edge when NER doesn't find any shared entities. On this branch, it returns the hop correctly.

Test results are also good.

That said, I was also able to reproduce all three Qodo findings and both Copilot findings.

**Zero weight becomes 1.0**

I tested this:

`add_edge(a, b, edge_type="CAUSED", weight=0.0)`

When tracing it, the result reports:

`edge_weight: 1.0`
`confidence_decay: 1.0`

The issue is here:

`float(getattr(edge, "weight", 1.0) or 1.0)`

A valid `0.0` gets treated the same as a missing value.

`add_causal_relationship()` always uses `weight=1.0`, so it won't hit this case, but `add_edge()` is public and can create CAUSED/INFLUENCED/PRECEDENT_FOR edges with other weights.

**Parallel causal edges get dropped**

I added two edges between the same source and target:

* `CAUSED`, weight `0.8`
* `INFLUENCED`, weight `0.3`

Only one chain comes back. The other edge disappears without any error.

It looks like this happens because `explicit_causes = {}` is keyed only by `source_id`, so adding the second edge overwrites the first one.

**The global `visited` set drops valid chains**

I tested this with a diamond-shaped graph:

`R → S → A → D`
`R → S → B → D`

The results included:

`A → D`
`S → A → D`
`R → S → A → D`
`B → D`
`S → B → D`

But `R → S → B → D` was missing.

From what I can tell, this happens because `S` was already added to the global `visited` set while walking the first branch. When the second branch reaches `S`, recursion stops there.

This seems like a normal graph shape, so it can silently lose valid deeper causal chains.

**Full edge scans instead of using the index**

I also checked the performance comment. `edge_type_index` already exists, is maintained, and is used in other parts of this file.

The new code has two places that scan all edges instead of using that index. One of those scans happens inside the DFS, so it can run again for every visited node.

So overall, the main fix is working, but I don't think this should merge yet.

The goal of the PR is to trust explicit causal edges as the source of truth, but right now that explicit-edge path can still lose or change data in normal cases:

* `0.0` weights become `1.0`
* multiple causal edges between the same nodes overwrite each other
* valid chains can disappear in branching graphs

At minimum I'd fix the weight handling and parallel-edge issue before merging. I think the `visited` behavior should also be fixed because silently dropping causal chains goes against what this PR is trying to solve.

hsd2514:
Thanks for the thorough review — reproducing all five findings by hand made them easy to act on. All four are fixed in `3ca91891`, and I used your exact reproductions to verify.

**Zero weight becomes 1.0**

The truthiness check is replaced with an explicit `None` check:

```python
weight = getattr(edge, "weight", None)
edge_weight = 1.0 if weight is None else float(weight)
```

With your `add_edge(a, b, edge_type="CAUSED", weight=0.0)`, the trace now reports `edge_weight: 0.0` and `confidence_decay: 0.0`. Your point about `add_edge()` being the reachable path here was the useful one — `add_causal_relationship()` alone could never hit it.

**Parallel causal edges get dropped**

`explicit_causes` is now a list of edges rather than a dict keyed by `source_id`. The entity fallback previously relied on dict-key membership, so it now uses a separate `explicit_cause_ids` set. Your `CAUSED` (0.8) + `INFLUENCED` (0.3) pair returns both hops with both weights.

**The global `visited` set drops valid chains**

Cycle detection is now per path — `path_ids` is threaded through the recursion instead of a traversal-wide `visited` set, with `max_depth` still bounding the walk. Your diamond returns all six paths, including the `R → S → B → D` that was missing.

You were right to push on this one. I had initially treated it as pre-existing and out of scope, but the PR's premise is that explicit edges are authoritative, and silently dropping chains contradicts that. I also added a test for a genuine cycle (`A → B → C → A`) to confirm removing the global set doesn't cost termination.

**Full edge scans instead of using the index**

Both call sites now use `edge_type_index`. For the trace specifically, a reverse index (`target_id -> [edges]`) is built once per call rather than inside the DFS, so the per-node rescan you identified is gone:

| | before | after |
|---|---|---|
| `trace_decision_causality` | O(depth × C) | O(C + paths) |
| `analyze_decision_influence` | O(E) | O(C) |

One caveat worth stating: per-path cycle detection makes the traversal worst-case exponential in `max_depth` on densely branching graphs, since every distinct path is enumerated. That is inherent to returning all chains rather than something the index addresses. Happy to cap or dedupe results if you would prefer a bound.

**Tests**

Four added, covering zero weights, parallel edges, your diamond, and cycle termination — 11 total in the file. Three of the four fail on the previous commit; the cycle test passes both ways by design, guarding termination rather than the fix.

I also compared the full `tests/context` suite against the previous commit by failing-test name: the sets are identical, with passes going 393 to 397 (exactly the new tests). The remaining failures and errors are pre-existing in my environment, from optional dependencies that are not installed.

One item I have deliberately left alone, in case you would like it changed: explicit edges are added to `direct_influence` in both directions, so `analyze_decision_impact(b)` lists `b`'s upstream cause. The pre-existing entity/category logic is already symmetric, so this is consistent with it, but it does sit awkwardly with the README describing that call as a downstream map. I am happy to narrow it to the source direction if you would rather it match the documented contract.

ZohaibHassan16:
All four bugs I had before are fixed now.

There is one new thing I think should be handled before merge though. Copilot's comment about combinatorial growth looks valid (and worse than it sounds).

I made a branching DAG and traced backwards from the end:

```text
14 layers, 575 nodes, branching=2   -> 12,286 reports, 0.04s
20 layers, 3,327 nodes, branching=2 -> 786,430 reports, 2.88s
```

So the number of reports can grow pretty fast. The default `max_depth=5` is fine , even with more branching I only got 363 reports and it returned immediately. The problem is when a caller passes a much larger `max_depth`.

I checked and this isn't exposed through a REST endpoint right now, but it is documented as an agent-callable SDK method in a few places. That makes me a bit worried about an agent asking to trace "as far back as possible" on a large graph and accidentally producing a huge response.

I get that the old implementation was bounded partly because it was incorrectly dropping valid paths, so this isn't really a new bug. It's more that fixing the correctness issue exposes the actual cost of returning every valid chain.

I think adding something simple like `max_chains` or `max_reports` would be enough. Once that limit is hit, stop collecting more results instead of letting one call grow without a bound.

Separately, I also checked the `direct_influence` direction issue you mentioned. The README describes `analyze_decision_impact()` as downstream-only but both the new explicit-edge logic and the older entity/category heuristic are currently symmetric. So that inconsistency was already there before this PR. I don't think this PR should try to fix only one side of it.

I'd like to see a cap on the number of generated chains before merging. The direction issue can be handled separately.

hsd2514:
Pushed `b89025de` to address the traversal bound I flagged in my previous comment, which the automated review also picked up.

Measured on a layered graph where every decision in a level causes every decision in the next (branching factor 5):

| `max_steps` | before | after |
|---|---|---|
| 5 (default) | 780 chains, 0.01 s | unchanged |
| 9 | 488,280 chains, 7.3 s | 10,001 chains, 0.22 s |
| 11 | did not finish in 120 s | 10,001 chains, 0.15 s |

`max_chains` defaults to 10000. Rather than capping silently, the traversal stops at the bound and appends a marker so the caller can always tell the result is partial:

```python
{"truncated": True, "max_chains": 10000, "message": "..."}
```

A warning is logged with the same detail, and `max_chains=None` restores unbounded behaviour. Silently dropping chains would have reintroduced exactly the failure this PR set out to fix, so it seemed worth the extra field. Graphs that fit inside the bound are unaffected — verified that a small graph returns no marker and the same chains as before.

Three tests added for this (truncation reported, small graphs unmarked, `None` opting out), bringing the file to 14.

I did not act on the other suggestion, maintaining a `decision_id`-keyed causal index inside `_add_internal_edge()`. It would need to stay consistent across `clear()` and `from_dict()`, and an index drifting out of sync with `_decisions` is the same class of problem that made the original trace abort on restored graphs. It reads as a worthwhile refactor on its own, just a larger change than this fix warrants. Happy to open a follow-up issue if you would like it tracked.

Full `tests/context` run compared against the pre-change commit by failing-test name: the sets are identical, passes went from 393 to 400 (exactly the new tests). Remaining failures and errors are pre-existing in my environment from optional dependencies that are not installed.

ZohaibHassan16 (approved):
LGTM. Ready to merge.

PR #984, opened 2026-08-14
refactor(explorer): remove dead graph workspace shell

### LINKED ISSUE #981: `GraphWorkspaceShell.tsx` is dead code but still getting bug fixes
## Summary

`explorer/src/workspaces/GraphWorkspace/GraphWorkspaceShell.tsx` is 862 lines, but it does not seem to be imported anywhere in the app or tests.

The component that is actually mounted and shipped is `GraphWorkspace.tsx`, which is currently 3,400+ lines.

It looks like these two files probably came from the same implementation at some point, but only `GraphWorkspace.tsx` stayed connected to the real app.

The bigger problem is that `GraphWorkspaceShell.tsx` is still getting real fixes even though it never runs.

Some recent commits that touched it:

* `6a0eecbe` fix(explorer): resolve #830 — Maximum update depth exceeded on Temporal panel open
* `343d2bc4` Fix #769: Resolve all react-hooks/set-state-in-effect lint errors project-wide
* `e3a3f601` fix(explorer): make distance intelligence API calls slash-safe
* `7c8dfbd3` feat(explorer): stabilize and refine grouped graph view

So there is still engineering work going into a component that no user actually reaches.

It also has no test coverage, which makes this harder to notice. A fix can be made there and everything still passes even if the production component has the same bug.

## Where I found this

I ran into this while fixing #977, where Explorer hangs when the backend is unavailable.

`GraphWorkspaceShell.tsx` already had proper `isError` / `error` handling for the graph load query, while the live `GraphWorkspace.tsx` did not.

I ended up using the dead component's implementation as a reference for the actual fix.

That seems like a pretty good example of the problem: useful fixes are landing in this file, but they are not reaching the code users actually run.

## Proposed fix

Remove `GraphWorkspaceShell.tsx` and anything that is only used by it.

`useGraphData.ts` is worth checking as part of this. It looks like it may be the data-loading hook for `GraphWorkspaceShell`, while the live component uses `useLoadGraph.ts`.

Before removing it, we should confirm `useGraphData.ts` is not used anywhere else.

Dead code with no tests is confusing, and in this case it is also causing people to spend time fixing code that never runs.

lakshayxi:
## Summary

- Remove the unused `GraphWorkspaceShell.tsx`.
- Remove `useGraphData.ts`, which had no consumers outside the dead shell.
- Remove `GraphRuntimeStage.tsx`, which was only consumed by the deleted shell.
- Keep the active `GraphWorkspace.tsx` / `useLoadGraph.ts` path and shared graph components unchanged.

Closes #981

## Validation

- `npx tsc -b`
- `npm run test:graph-store`
- `npm run test:graph-workspace`
- `npm run test:plugin-registry`
- `npm run build`
- `git diff --check`

`npm run lint` still reports existing unrelated repository lint findings; this change adds no new lint findings.


ZohaibHassan16:
Everything looks good overall. 

One thing I noticed though: `GraphRuntimeStage.tsx` is 479 lines and doesn't have any test coverage. It looks like it was only imported by `GraphWorkspaceShell.tsx`, which this PR deletes. I did another grep after the deletion and couldn't find anything else using it.

Since it seems to be dead code for the same reason as the files already being removed here, I think it would make sense to remove `GraphRuntimeStage.tsx` in this PR too instead of leaving it around.

Everything else looks good to me. If you can include that file too, I'll approve.

lakshayxi:
Good catch. Confirmed `GraphRuntimeStage.tsx` had no remaining consumers after removing the shell, so I removed it as well and reran the Explorer validation.

ZohaibHassan16 (approved):
Cleanup completed. Approved.

PR #991, opened 2026-08-15
docs(cookbook): add Change Management module notebook

LeonSGP43:
## What

Adds `cookbook/introduction/24_Change_Management.ipynb` — a practical guide to the `change_management` module (follows #990, the reasoning notebook).

## Why

The `change_management` module ships `change_management_usage.md` (1k+ lines) in the package but has no cookbook coverage. The notebook demonstrates the core workflow with verified examples:

- `ChangeLogEntry` with email-validated author (the `ValidationError` behavior is documented in the notebook markdown, not demonstrated in a cell)
- `InMemoryVersionStorage`: `save` / `get` / `list_all` / `exists` / `delete`
- Named release tags: `save_tag` / `get_tag`
- SHA-256 integrity: `compute_checksum` / `verify_checksum` with live tamper detection (intact `True` → tampered `False`)

## Verification

Every cell was executed against the module source (`change_log.py`, `version_storage.py`); outputs shown are real results. One subtlety documented: `verify_checksum` takes the whole snapshot dict and excludes its `checksum` field when recomputing.


LeonSGP43:
Addressed the install/claim mismatch: the verification note now states outputs were checked against the repository's `semantica/change_management/` source, and that the `pip install` cell may pull a newer release. Committed outputs are kept intentionally for the cookbook.

ZohaibHassan16 (commented):
Small thing, not blocking: the PR description makes it sound like the ValidationError case is demonstrated in the notebook, but looking at the cells it's only mentioned in the markdown text, there's no actual cell that triggers and catches it. Not wrong just slightly overstated, up to you if you want a cell added for that or don't bother.

On qodo's findings : the install/claim mismatch one looks properly fixed, intro cell now says outputs were verified against repo source and the pip install cell might pull a different version, that's exactly the right caveat.

The committed-outputs one qodo only flagged as informational here but same deal as the reasoning notebook : checked the raw json, execution_count is null on every cell and the print-based cell (cell 9) is saved as execute_result instead of stream, so this one wasn't actually run through jupyter either, values are just typed in (correctly, verified them all, but still)

same ask as the other one : restart kernel, run all, save, and this is good to merge from me

LeonSGP43:
Done in ffee6ac: ran the notebook through a fresh Jupyter kernel (restart + run all + save) — all 6 code cells now carry execution counts and the `print()` cell (cell 9) is saved as a proper `stream` output. Result values verified byte-identical to the previous committed outputs; only the output artifacts changed.

On the ValidationError point: agreed the description overstated it — I've updated the PR description to say the `ValidationError` behavior is documented in the notebook markdown rather than demonstrated in a cell (kept the scope as-is rather than adding a cell, per your "up to you").

ZohaibHassan16 (approved):
Kernel produced now. Output artifacts are good now. LGTM.

PR #992, opened 2026-08-15
docs(cookbook): add Seed Data module notebook

LeonSGP43:
## What

Adds `cookbook/introduction/25_Seed_Data.ipynb` — a practical guide to the `seed` module (follows #990 / #991).

## Why

The `seed` module ships `seed_usage.md` in the package but has no cookbook coverage. Seed data is the recommended way to bootstrap a KG with trusted entities before extraction runs. The notebook demonstrates with verified examples:

- `register_source(name, format, location, entity_type=...)` with a CSV source
- `load_source(name)` — records enriched with `entity_type` / `source` provenance
- `create_foundation_graph()` — entities (confidence 1.0) / relationships / metadata
- `validate_quality(graph)` gating before commit

## Verification

Every cell was executed against `semantica/seed/seed_manager.py`; outputs shown are real results from a live 2-record CSV round trip.


LeonSGP43:
Re: `docs/cookbook.md` index link — agreed it's needed. I'll add index entries for the whole notebook series (22–25) in a single follow-up PR once the sibling PRs (#989/#990/#991) merge, to avoid adjacent-line conflicts between the four branches and dangling links to not-yet-merged files. Re: fixed CSV filename — kept as-is for demo readability (single file, content printed in the same cell), but happy to switch to a temp path if maintainers prefer.

ZohaibHassan16 (commented):
The fixed companies.csv filename written with mode="w" into the current directory. If someone's got their own companies.csv sitting in whatever folder they run this from, this silently overwrites it, and the notebook never cleans up after itself either way so it leaves the file behind regardless. saw your comment about keeping it for demo readability, but silently clobbering a user's file is worth avoiding even for demo purposes : tempfile.TemporaryDirectory() or NamedTemporaryFile would fix it in like 3-4 lines without changing what's printed in the cell.

On the index link : your reasoning for batching that into a follow-up PR makes sense.

On outputs : same as the other two notebooks, checked the raw json and execution_count is null everywhere, so this wasn't run through jupyter for real either. Same ask: restart kernel, run all, save.

LeonSGP43:
Both addressed in 661269e:

**File clobbering:** the CSV now goes into a session-scoped `tempfile.mkdtemp(prefix="semantica-seed-")` directory instead of the working directory, so a user's own `companies.csv` can never be overwritten. I used `mkdtemp()` rather than a `with TemporaryDirectory()` block because cell 5 loads the source in a later cell, so the file has to persist for the kernel session (OS cleans up `/tmp` leftovers). Nothing printed in any cell changes — the registered path only appears inside the package's own progress line, never in the notebook's documented output.

**Real execution:** ran the notebook through a fresh Jupyter kernel (restart + run all + save) — all 6 code cells carry execution counts, `print()` cells are saved as `stream` outputs. Result values verified identical to the previous committed outputs (the only textual change is dict reprs now being pretty-printed multi-line by IPython's display, values themselves unchanged). One note: the package's seeding progress line in cell 5's `stream` output embeds the session's temp path — that's genuine run output and will differ on any re-run, which felt preferable to trimming it by hand.

Happy to switch to a `NamedTemporaryFile`-style cleanup cell at the end if you'd prefer the temp dir removed within the notebook itself.

ZohaibHassan16 (approved):
Data loss risks and fabricated outputs are resolved. LGTM.

PR #1000, opened 2026-08-15
fix(kg): make k-shortest path search side-effect free

T1mn:
## Summary

- Keep Yen's k-shortest path search from mutating the input graph.
- Exclude temporary nodes and edges during spur searches.
- Reject repeated-node paths and add regression tests.

## Tests

- `pytest -q tests/kg/test_path_finder.py tests/kg/test_kg.py tests/kg/test_algorithms.py tests/kg/test_entity_pipeline.py -k 'not path_response'`
- 112 passed, 2 deselected


ZohaibHassan16 (approved):
Approving this. 

This PR also fixes another issue in `main` that wasn't mentioned in the PR description. The outer loop keeps using `paths[-1]`, but `paths` isn't updated until after that loop finishes. So `paths[-1]` stays as `first_path` which means it keeps generating candidates from the same path and can miss deeper alternatives. This version keeps a candidate heap and generates new spur paths from the last accepted path each round, which is the expected behavior for Yen's algorithm.

LGTM.

Issue #982, opened 2026-08-14
Explorer keeps sending duplicate API requests when the backend is down even with `retry: 0`


ZohaibHassan16:
## Summary

When the Explorer backend is unavailable, opening the graph workspace fails repeatedly (instead of once).

It keeps sending requests to:

* `/api/graph/nodes`
* `/api/temporal/snapshot`

in a loop and does not stop.

This happens even though the query doing the initial graph fetch is configured with `retry: 0`.

I found this while working on #977, which fixes the graph workspace getting stuck on the loading UI when the backend is down.

The error handling added in #977 does not seem to cause this. The request loop was already happening underneath the old loading UI, it was just much harder to notice because the page stayed stuck on the fake progress state.

## Repro

1. Stop the Explorer backend, or leave it stopped.
2. Start the frontend.
3. Open Explorer and go into the graph workspace.
4. Watch either the Vite proxy logs or the browser Network tab.
5. The failed graph/temporal requests keep getting sent again instead of stopping after the first failure.

## Evidence

I reproduced this from a clean `npm run dev` session with the backend down the whole time.

The frontend then kept producing repeated requests for:

```text
/api/graph/nodes
/api/temporal/snapshot
```

even though the graph query itself has:

```ts
retry: 0
```

So this does not look like normal TanStack Query retry behavior.

## Expected behavior

If the backend is unreachable and the request is configured with `retry: 0`, the request should fail once and stay failed until something intentionally triggers another fetch, for example:

* the user clicks Retry
* the component remounts
* connectivity/focus refetch behavior intentionally runs

It should not continuously send the same failed requests by itself.

## Actual behavior

The graph workspace keeps causing new requests after the previous ones fail.

This creates an endless stream of failed proxy/network requests while the backend is offline.

## Notes

I confirmed this is separate from the #977 changes.

The loop reproduces with those commits removed/stashed, so it appears to be pre-existing behavior.

My guess is that something higher up is repeatedly causing a refetch or remount after the request enters an error state, rather than TanStack Query itself retrying the request.

`/api/temporal/snapshot` being part of the same loop may also help narrow down which workspace state/effect is triggering it.

## Suggested investigation

Things that seem worth checking:

* effects that call `refetch()` when graph state changes
* query keys being recreated or changed on every render
* workspace/component remounts after a failed load
* temporal snapshot effects tied to graph loading state
* focus/reconnect refetch settings
* state updates caused by the error path that trigger the loading path again

[...]

PR #1003, opened 2026-08-15
fix(explorer): gate temporal requests on graph load

lakshayxi:
Addresses #982.

## Summary

Prevents temporal API requests from running before the initial graph load has succeeded.

With the backend unavailable, `GraphWorkspace` previously still requested:

- `/api/temporal/bounds` on mount
- `/api/temporal/snapshot` once the timeline initialized and graph loading failed

Both requests are now gated on a successful graph summary. Empty graphs (`nodeCount: 0`) still count as a successful load.

## Investigation

I reproduced the backend-down path through the real Vite proxy on both current `main` and pre-#980.

In both cases I observed:

- `/api/graph/nodes`: 1 request
- `/api/graph/edges`: 0 requests
- `/api/temporal/bounds`: 1 premature request
- `/api/temporal/snapshot`: 1 premature request

I could not reproduce an autonomous repeated `/api/graph/nodes` loop during the idle period, so this patch does not change the existing TanStack Query configuration.

After the fix, with the backend unavailable:

- `/api/graph/nodes`: 1 request
- `/api/temporal/bounds`: 0 requests
- `/api/temporal/snapshot`: 0 requests

## Testing

- `npm --prefix explorer run test:graph-workspace`
- `npm --prefix explorer run test:graph-store`
- `npm --prefix explorer run test:plugin-registry`
- `npm --prefix explorer run build`
- `git diff --check upstream/main...HEAD`

Also verified the runtime behavior with:
- backend unavailable
- successful empty graph
- successful populated graph

### DIFF
--- a/explorer/package.json
+++ b/explorer/package.json
@@ -9,7 +9,7 @@
     "lint": "eslint .",
     "preview": "vite preview",
     "test:graph-store": "node --test tests/graphStore.multi-edge.test.mjs",
-    "test:graph-workspace": "node --import tsx --test tests/graphSceneState.display.test.ts",
+    "test:graph-workspace": "node --import tsx --test tests/graphSceneState.display.test.ts tests/temporalLifecycle.test.ts",
     "test:plugin-registry": "node --import tsx --test tests/pluginRegistry.temporal.test.mjs"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


ZohaibHassan16 (approved):
Thank you for raising the PR. Tested and confirmed. LGTM.

Issue #1015, opened 2026-08-15
[BUG] DBIngestor.execute_query and DataExporter.export_table_data always fail — sqlalchemy's text() is never imported in either method


pravit-amp:
**Title:** `[BUG] DBIngestor.execute_query and DataExporter.export_table_data always fail — sqlalchemy's text() is never imported in either method`

---

## Bug Description

`semantica/ingest/db_ingestor.py` imports `text` from `sqlalchemy` **function-locally**, inside
`DatabaseConnector.connect()` (line 129) and `DatabaseConnector.test_connection()` (line 195).
Those bindings are local to those two methods and do not leak to module scope.

Two other methods call `text(...)` without importing it:

| Method | Line | Call |
|---|---|---|
| `DataExporter.export_table_data` (def at 232) | 310 | `conn.execute(text(query))` |
| `DataExporter.export_table_data` | 331 | `conn.execute(text(count_query))` |
| `DBIngestor.execute_query` (def at 679) | 713 | `conn.execute(text(query), params)` |

`export_table_data` imports only `inspect` from sqlalchemy (line 272); `execute_query` imports
nothing. There is no module-level `text` binding anywhere in the file. Both methods therefore
raise `NameError` on every call.

Worse, both call sites sit inside `except Exception` blocks that re-wrap the failure, so the
`NameError` is reported as a `ProcessingError` with a message that reads like a database
problem rather than a missing import.

## Steps to Reproduce

1. Install Semantica with sqlalchemy available (reproduced with 2.0.52).
2. Run either code example below. Any database works — SQLite is enough, and the failure
   happens before a query is ever sent, so the contents are irrelevant.
3. See `ProcessingError` wrapping a `NameError`.

## Expected Behavior

`execute_query()` returns a list of row dictionaries, and `export_table_data()` returns a
`TableData` object, as documented in their docstrings and in
`docs/guides/ontology.md:428`.

## Actual Behavior

Both raise `ProcessingError` immediately on reaching the `text(...)` call, with the underlying
`NameError: name 'text' is not defined` chained via `from e`. No query is ever executed.

## Environment

- **OS**: macOS 26.5.2
- **Python Version**: 3.12.13
- **Semantica Version**: 0.6.5 (`origin/main` @ `55798512`)
- **Installation Method**: source (editable install)

## Code Example

```python
from semantica.ingest import DBIngestor

db = DBIngestor()
rows = db.execute_query(
[...]

Issue #530, opened 2026-05-04
[BUG] DuplicateDetector crashes because SimilarityCalculator does not import get_progress_tracker


ZohaibHassan16:
## Summary

`DuplicateDetector` crashes during initialization because `SimilarityCalculator` calls `get_progress_tracker()` without importing it.

## Environment

- OS: Windows
- Python: 3.12
- Semantica version reported by pip: 0.0.5

## Reproduction

```python
from semantica.deduplication import DuplicateDetector

detector = DuplicateDetector()
```

## Actual Result

```text
NameError: name 'get_progress_tracker' is not defined
```

The failing code is in:

```text
semantica/deduplication/similarity_calculator.py
```

## Expected Result

`DuplicateDetector()` should initialize successfully.

## Suggested Fix

Add:

```python
from ..utils.progress_tracker import get_progress_tracker
```

to `similarity_calculator.py`.

KaifAhmad1:
@ZohaibHassan16 
> ✅ **No fix needed — already resolved in `main`.**
> - **Import check passed:** `get_progress_tracker` is already imported in `similarity_calculator.py` — `DuplicateDetector()` initializes successfully with no `NameError`.
> - **Tests passed:** Full pytest run confirms **9/9 deduplication tests pass in 0.64s** with zero failures. The crash only affects the outdated pip-published version (`0.0.5`). Closing as resolved.

PR #1016, opened 2026-08-15
fix(parse): import get_progress_tracker in ExcelParser

### LINKED ISSUE #1014: [BUG] ExcelParser() raises NameError on every instantiation — get_progress_tracker is never imported
**Title:** `[BUG] ExcelParser() raises NameError on every instantiation — get_progress_tracker is never imported`

---

## Bug Description

`ExcelParser.__init__` calls `get_progress_tracker()` but the module never imports it, so
constructing the class raises `NameError`. Because the call is in `__init__` and not behind
any conditional, **every** instantiation fails — the class is unusable in all code paths.

`ExcelParser` is a documented public export (`semantica.parse.ExcelParser`, listed in
`semantica/parse/__all__`), so this breaks the entire Excel parsing surface of the `parse`
module.

## Steps to Reproduce

1. Install Semantica (any method — this is a source-level defect, not environment-dependent).
2. Run the code example below, or any code that reaches `ExcelParser()`.
3. See `NameError`.

## Expected Behavior

`ExcelParser()` initializes successfully and returns a usable parser instance, the same way
every other parser in `semantica.parse` does.

## Actual Behavior

`ExcelParser()` raises `NameError: name 'get_progress_tracker' is not defined` before the
constructor completes. No `ExcelParser` instance can be created, so `parse()`, `parse_sheet()`,
and every other method on the class are unreachable.

## Environment

- **OS**: macOS 26.5.2
- **Python Version**: 3.12.13
- **Semantica Version**: 0.6.5 (`origin/main` @ `55798512`)

## Code Example

```python
from semantica.parse import ExcelParser

parser = ExcelParser()  # raises NameError before returning
```

## Error Messages / Logs

```
  File "<stdin>", line 1, in <module>
  File "semantica/parse/excel_parser.py", line 74, in __init__
    self.progress_tracker = get_progress_tracker()
                            ^^^^^^^^^^^^^^^^^^^^
NameError: name 'get_progress_tracker' is not defined
```

## Additional Context

**This is a sibling of #530**, which was the identical bug — same missing import, same error
string — in `semantica/deduplication/similarity_calculator.py`, closed 2026-05-05. That fix
patched the single reported file without sweeping the rest of the codebase for the same
pattern, leaving this one behind.

I grepped every call site on `origin/main`:

```bash
for f in $(grep -rl "get_progress_tracker()" semantica/ --include="*.py"); do
  grep -q "import get_progress_tracker" "$f" || echo "MISSING: $f"
done
```

`semantica/parse/excel_parser.py` is the **only** remaining offender (the one other hit,
`semantica/utils/progress_tracker.py`, is the module that defines the function). So this
should be the last of them — but the check is worth re-running as part of the fix rather
than patching the one file blind.

**Why this was never caught:** `tests/parse/test_parse_comprehensive.py:13` imports
`ExcelParser` but never instantiates it, so the module-level import succeeds and the test
passes. No test in the repo constructs the class.

## Possible Solution

Add the missing import alongside the other `..utils` imports at
`semantica/parse/excel_parser.py:38-39`:

```python
from ..utils.exceptions import ProcessingError, ValidationError
from ..utils.logging import get_logger
from ..utils.progress_tracker import get_progress_tracker  # <-- add
```

This matches how the import is done in every other module that uses the tracker — e.g.
`semantica/ingest/db_ingestor.py:39`.

Please also add a regression test that actually **constructs** the parser, since an
import-only test is what let this ship:

```python
def test_excel_parser_constructs(self):
    self.assertIsNotNone(ExcelParser())
```

pravit-amp:
## Description

`ExcelParser.__init__` called `get_progress_tracker()` without importing it, so every
instantiation raised `NameError: name 'get_progress_tracker' is not defined`. The call sits in
the constructor with no conditional around it, so the class was unusable in all code paths —
`parse()`, `parse_sheet()`, and everything else on it were unreachable. `ExcelParser` is a
documented public export listed in `semantica.parse.__all__`, so this took out the module's
entire Excel surface.

This is the same defect as #530, which was `SimilarityCalculator` missing the identical import.
That fix patched the one reported file without sweeping the codebase for the pattern, which is
what left this one behind. I ran the sweep as part of this change: `semantica/parse/excel_parser.py`
was the last remaining call site without an import, and there are now none.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #1014

Context: #530 is the same missing-import bug in `semantica/deduplication/similarity_calculator.py`,
closed 2026-05-05 without a codebase-wide sweep.

## Changes Made

- Added `from ..utils.progress_tracker import get_progress_tracker` to
  `semantica/parse/excel_parser.py`, matching how every other module that uses the tracker
  imports it (e.g. `semantica/ingest/db_ingestor.py`).
- Swept every `get_progress_tracker()` call site in `semantica/` for the same missing import.
  `excel_parser.py` was the only remaining offender; the sweep is now clean.
- Added `tests/parse/test_parser_construction.py` — construction coverage for all 20 parser
  classes exported from `semantica.parse`, driven off `__all__` so parsers added later are
  covered without anyone updating the test file. Optional-dependency parsers (e.g.
  `DoclingParser` without `docling`) are allowed to raise `ImportError`, but the message must
  carry install guidance, so a genuine breakage can't hide behind that exemption.

### Why the tests are in a new file

They are deliberately **not** in `tests/parse/test_parse_comprehensive.py`. That file's `setUp`
patches `get_logger` and `get_progress_tracker` into every parse module that exposes them.
Before this fix `excel_parser` had no such attribute, so it was skipped — but with the import
added it would now be patched, and a regression test placed there would mock away the exact
call it exists to verify. The new file does no patching at all.

That file's existing `ExcelParser` coverage is also what let this ship: it imports the class
but never constructs it, so the module-level import succeeds and the test passes regardless of
whether `__init__` works.

## Testing

- [x] Tested locally

Each new test was confirmed to **fail without the fix**, rather than assumed to:

| Test | Without fix | With fix |
|---|---|---|
| `test_excel_parser_constructs` | `NameError` | pass |
| `test_excel_parser_wires_progress_tracker` | `NameError` | pass |
| `test_all_exported_parsers_construct` | SUBFAIL `ExcelParser`, 19 others pass | 20 subtests pass |

`tests/parse/` after the fix: **24 passed, 1 failed**. The failure is
`test_parse_comprehensive.py::test_pdf_parser`, which fails identically on unmodified `main`
(`AttributeError: module 'semantica.parse.pdf_parser' does not have the attribute 'pdfplumber'`
— `pdfplumber` not installed). Unrelated to this change.

Build: `Successfully built semantica-0.6.5.tar.gz and semantica-0.6.5-py3-none-any.whl`.

### Test Commands

```bash
# The regression tests
pytest tests/parse/test_parser_construction.py -v

# Full parse suite
pytest tests/parse/

# Confirm the tests actually catch the bug
git stash push semantica/parse/excel_parser.py
pytest tests/parse/test_parser_construction.py   # 3 failed
git stash pop

# The sweep, re-runnable
for f in $(grep -rl "get_progress_tracker()" semantica/ --include="*.py"); do
  grep -qE "import get_progress_tracker|^def get_progress_tracker" "$f" || echo "MISSING: $f"
done

# Build
python -m build
```

## Documentation

No public API, signature, or behavior changed — `ExcelParser` now does what its existing
documentation already says it does. Happy to add a `### Fixed` entry to `CHANGELOG.md` under
`[Unreleased]` if maintainers want one; I left it out because the entries in that file are
keyed by PR number, which doesn't exist until this is opened.

## Breaking Changes

**Breaking Changes**: No

The change is a single added import. Nothing that previously worked can break, because nothing
that touched this constructor previously worked.

## Checklist

`black --check` and `isort --check-only` both report no changes needed for the two files.

## Additional Notes

**On the sibling-bug pattern.** #530 and this issue are the same defect in different modules,
and the first fix didn't sweep. The `__all__`-driven test here is aimed at that specifically: it
covers the whole parser family rather than just the one class that was reported, so the next
parser added is covered automatically and this can't recur in this module by omission.

For reviewers who want to confirm the sweep independently, the one-liner is in the Test Commands
section above. The only other `undefined name` results from `pyflakes semantica/` are the three
`text` hits in `semantica/ingest/db_ingestor.py` (separate bug, filed as #1015) and six
`"Decision"` hits in `semantica/context/`, which are quoted annotations never evaluated at
runtime — a missing `TYPE_CHECKING` import, not a defect. Neither is in scope here.


ZohaibHassan16 (approved):
Approving.

`ExcelParser()` fails on main with `NameError: name 'get_progress_tracker' is not defined`. Adding the import fixes it.

Also ran `test_parser_construction.py` before the fix and got the 3 expected failures. After the fix, all 3 passed and the parser sweep passed all 20 subtests.

Ran the grep check too, and `excel_parser.py` is the only one with this missing import.

black/isort are clean and `python -m build` passes. 

And yeah, keeping this test out of `test_parse_comprehensive.py` is the right call since that file patches `get_progress_tracker` and would hide the bug.

LGTM.

PR #1017, opened 2026-08-15
fix(ingest): import sqlalchemy text where DBIngestor and DataExporter use it

### LINKED ISSUE #1015: [BUG] DBIngestor.execute_query and DataExporter.export_table_data always fail — sqlalchemy's text() is never imported in either method

pravit-amp:
## Description

`sqlalchemy.text` was imported **function-locally** inside `DatabaseConnector.connect()` and
`DatabaseConnector.test_connection()`. Those bindings don't leak to module scope, and two other
methods call `text(...)` without importing it:

| Method | Line | Call |
|---|---|---|
| `DataExporter.export_table_data` | 310 | `conn.execute(text(query))` |
| `DataExporter.export_table_data` | 331 | `conn.execute(text(count_query))` |
| `DBIngestor.execute_query` | 713 | `conn.execute(text(query), params)` |

Both methods raised `NameError` on every call, before any query reached the database. Because
the call sites sit inside `except Exception` handlers that re-wrap and re-raise, users saw
`ProcessingError: Failed to execute query: name 'text' is not defined` — a message that reads
like a database or connection fault rather than a missing import, which is a large part of why
this survived.

`docs/guides/ontology.md:428` documents `DBIngestor().execute_query(...)` as a supported entry
point, so the documented usage walked straight into it.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #1015

Not overlapping, but adjacent: #973 / #974 fix `load_from_database` calling `DBIngestor` without
a connection string and mislabeling `OSError`. That work is on the `seed/` side and doesn't touch
`db_ingestor.py`, so the two changes are independent.

## Changes Made

- `semantica/ingest/db_ingestor.py:272` — `from sqlalchemy import inspect` → `from sqlalchemy import inspect, text`, covering both call sites in `export_table_data`.
- `semantica/ingest/db_ingestor.py:711` — added `from sqlalchemy import text` in `execute_query`.
- Added `tests/ingest/test_db_ingestor_query.py` — 5 tests covering all three call sites against a temporary SQLite database, so they need no external service.

Two lines, no logic or signature changes.

I kept the file's existing **local-import** convention rather than hoisting `text` to module
scope. A module-level import would arguably prevent a recurrence, but changing the file's
dependency-handling style is a separate decision from fixing the bug, and I didn't want to make
it as a drive-by. Happy to switch if maintainers prefer it. Lines 415 and 601 have the same
`inspect` import but never call `text`, so they were left alone.

## Testing

- [x] Tested locally

All 5 new tests were confirmed to **fail without the fix**, then pass with it.

### The third call site needed proof, not inference

The pre-fix failure output only ever names lines 310 and 713 — line 331 never appears, because
310 raises first and short-circuits it. So "all tests fail pre-fix" does **not** by itself
demonstrate the `COUNT(*)` branch is covered. I verified that directly with a line tracer
instead of assuming:

```
export_table_data(no limit)   -> line 310 executed: True,  line 331 executed: True
export_table_data(limit=1)    -> line 331 executed: False
```

That branch only runs when no `limit` is passed, which is exactly why
`test_export_table_data_without_limit_counts_rows` exists as a separate case. Without it, one of
the three sites the bug touched would still be untested after this PR.

### This also repairs an existing failing test

Diffing the failure set before and after the change:

```
fixed by this change:
  tests/ingest/test_notebook_02.py::TestNotebook02DataIngestion::test_08_database_ingestion
newly broken:
  (none)
```

`tests/ingest/` goes from **12 failures to 11**. That test builds a real SQLite database and
drives `DatabaseConnector`, so the bug was already producing a red test in the repo — it just
wasn't attributable to a missing import without knowing the cause. Independent corroboration
that this is the right fix.

The remaining 11 failures (`test_web_ingestor`, `test_submodules`) are pre-existing and
unrelated — identical before and after this change.

### Test Commands

```bash
# The regression tests
pytest tests/ingest/test_db_ingestor_query.py -v

# Full ingest suite
pytest tests/ingest/

# Confirm the tests actually catch the bug
git stash push semantica/ingest/db_ingestor.py
pytest tests/ingest/test_db_ingestor_query.py   # 5 failed, NameError
git stash pop

# No undefined names left in the file
python -m pyflakes semantica/ingest/db_ingestor.py

# Build
python -m build
```

Build: `Successfully built semantica-0.6.5.tar.gz and semantica-0.6.5-py3-none-any.whl`.

## Documentation

No public API, signature, or behavior changed — both methods now do what their docstrings and
`docs/guides/ontology.md` already describe. Happy to add a `### Fixed` entry to `CHANGELOG.md`
under `[Unreleased]` if maintainers want one; I left it out because entries there are keyed by
PR number, which doesn't exist until this is opened.

## Breaking Changes

**Breaking Changes**: No

Nothing that previously worked can break, because neither method previously worked.

## Checklist

One note on style, flagged rather than silently handled: `black --check` reports that
`semantica/ingest/db_ingestor.py` would be reformatted — but it reports the same on unmodified
`main`, so the drift is pre-existing and not introduced here. I deliberately did **not** run
`black` on the file, because reformatting it would bury a two-line fix under a whole-file diff
and pull unrelated changes into a scoped PR. The new test file is clean under both `black` and
`isort`. If you'd like the formatting brought up to standard, I'm glad to do it as a separate
commit or a separate PR.

## Additional Notes

**Why this was invisible.** No test exercised either method. The only `execute_query` references
under `tests/` are `Mock()` stand-ins for the unrelated graph-store method of the same name, so
a grep for coverage looks reassuring and finds nothing real. Combined with the `except Exception`
re-wrap turning a `NameError` into a database-sounding `ProcessingError`, both the automated and
the human signal pointed away from the actual cause.

**On the pattern.** `pyflakes semantica/` catches this class of bug immediately — a name bound
only inside one function but used in another is invisible until that method actually runs. The
same sweep is what surfaced #1014. Worth considering as a CI step; the repo's current workflow
(`.github/workflows/ci.yml`) runs the Explorer frontend tests and a package build, but no Python
linting or test run, which is why both of these shipped.


pravit-amp:
## Merged `main` and resolved the conflict — and it cuts this PR down to one call site

The conflict itself was trivial, but resolving it revealed that `main` has since fixed most of what this PR reported. Flagging that up front, because the description above is now substantially out of date.

### The conflict

`main`'s 43b207c1 (`fix(security): address privately disclosed zip-slip, SQLi, SSRF, XSS, and SPARQLi findings`, @KaifAhmad1) inserts `_validate_sql_identifier` / `_validate_sql_fragment` calls immediately after the import line this branch edits, which is the whole reason git couldn't auto-merge. I kept `main`'s validation block; this branch's `text` on the import line had already merged cleanly.

### The part that matters: that commit also added `text` to the same import

So `main` already fixes two of the three `NameError` call sites reported here — both in `export_table_data`. What's left is `DBIngestor.execute_query`, which 43b207c1 never touched.

| | as opened | after merge |
|---|---|---|
| import fixes | 2 | **1** |
| call sites fixed | 3 | **1** (`execute_query`) |
| tests failing pre-fix | 5 | **2** |
| "repairs an existing failing test" (`test_08_database_ingestion`) | yes | **obsolete** — passes on `main` already |

The entire remaining code change is now:

```python
try:
+   from sqlalchemy import text
+
    with engine.connect() as conn:
        result = conn.execute(text(query), params)
```

### It is not obsolete, though

Verified rather than inferred — there's no module-scope `text` in `db_ingestor.py` on `main`, and running this PR's test file against an `origin/main` worktree:

- `test_execute_query_returns_rows`, `test_execute_query_binds_parameters` → **fail**: `ProcessingError: Failed to execute query: name 'text' is not defined`
- the three `export_table_data` tests → **pass** (`main`'s half)

`DBIngestor.execute_query` still raises `NameError` on `main` today, and `docs/guides/ontology.md:428` documents it as a supported entry point, so the documented usage still walks into it.

### Why the three now-passing tests are kept

They no longer demonstrate *this* fix, which would normally argue for dropping them. But `main` has **no other test coverage for `export_table_data`** — a method a security commit had just modified. Reverting `main`'s half of the import (`inspect, text` → `inspect`) fails all three, so they're live regression guards rather than restatements, and they also exercise the `where` / `order_by` validation path 43b207c1 added.

### Verification

- `pytest tests/ingest/test_db_ingestor_query.py` → **5 passed**
- `pytest tests/ingest/` → 7 failures, **diffed line-by-line against an `origin/main` worktree: identical set**. No new breakage from the merge.

I kept the file's local-import convention for the remaining fix, as originally described — 43b207c1 kept it too, so it stays consistent.

### Suggested follow-ups for maintainers

1. **Retitle / rewrite the description.** It leads with a three-call-site table and a "repairs an existing failing test" section that are both superseded. Suggest `fix(ingest): import sqlalchemy text in DBIngestor.execute_query (#1015)` — happy to rewrite the body.
2. **#1015 is now partially fixed on `main`.** Worth a note on the issue so the remaining scope is clear to anyone picking it up.
3. If maintainers would rather not carry a 189-commit merge, I can rebase or reopen this as a fresh three-line PR instead.

ZohaibHassan16 (approved):
LGTM.

PR #1021, opened 2026-08-16
fix(parse): warn when PDF parse yields no text layer (scanned PDFs)

### LINKED ISSUE #1020: Scanned PDFs parse 'successfully' with empty text and no warning
## Description

Scanned (image-only) PDFs parsed through the default pdfplumber route (`parse_pdf()` / `DocumentParser`) return `{"full_text": "", ...}` with progress status `"completed"` — no error, no warning.

Root cause: `PDFParser._parse_page` (`semantica/parse/pdf_parser.py`) does `page.extract_text() or ""`, which silently swallows the missing text layer.

## Impact

The failure only surfaces far downstream (e.g., zero entities extracted from the document), which makes it hard to trace back to "this PDF is a scan". OCR support already exists (`parse_pdf(..., method="docling", enable_ocr=True)`), but nothing tells the user to switch routes.

## Proposed fix

Log an actionable warning in `PDFParser.parse()` when every parsed page yields no text (and `extract_text` is enabled), pointing the user to `method="docling"` with `enable_ocr=True`. PR #1021 implements this.

## Related issues found while testing

1. `import semantica.parse` fails in a fresh Python process (verified on 3.11/3.12/3.13) unless something else imported `email.message` first: `semantica/parse/email_parser.py` uses `email.message.Message` at class-definition time but only does `import email`. Fix is one line (`import email.message`), included in the PR since the new tests can't import the package without it.
2. `tests/parse/test_parse_comprehensive.py::test_pdf_parser` patches `semantica.parse.pdf_parser.pdfplumber`, which never exists as a module attribute (pdfplumber is imported inside `PDFParser.parse`), so the test errors as soon as the file is collectable. Also included in the PR (test-only change).
3. `pdfplumber` is not declared in `pyproject.toml` dependencies — PDF parsing raises `ProcessingError` on a fresh install until the user installs it manually. Not addressed in the PR; flagging here for maintainers to decide where it belongs.

Happy to split items 1/2 into separate PRs if preferred.

shanyu910:
Closes #1020

> **Note:** the original PR body was accidentally truncated to `-` during submission; this is the restored description.

## Problem

Scanned (image-only) PDFs parsed through the default pdfplumber route return an empty `full_text` with progress status `"completed"` — no error, no warning — so the failure only surfaces far downstream (e.g., zero entities extracted). `PDFParser._parse_page` does `page.extract_text() or ""`, silently swallowing the missing text layer.

## Changes

Two commits, ordered so the second depends on the first:

1. **`fix(parse): import email.message and repair pdfplumber test mock`** — enabling fixes required for the parse test suite to run at all:
   - `email_parser.py` uses `email.message.Message` at class-definition time but only did `import email`, so `import semantica.parse` fails in a fresh Python process (verified on 3.11/3.12/3.13) unless something else imported `email.message` first. Added the missing `import email.message`.
   - `test_pdf_parser` patched `semantica.parse.pdf_parser.pdfplumber`, which never exists as a module attribute (pdfplumber is imported inside `PDFParser.parse`), so the test errored once the file became collectable. Injected a fake module via `patch.dict(sys.modules, ...)` instead.

2. **`fix(parse): warn when PDF parse yields no text layer (scanned PDFs)`** — the actual fix:
   - `PDFParser.parse()` now logs a warning when every parsed page yields no text while `extract_text` is enabled, pointing the user to `parse_pdf(..., method="docling", enable_ocr=True)`.
   - New `tests/parse/test_pdf_scanned_detection.py` covering: warning on text-empty PDFs, no warning when text is present, no warning when `extract_text=False` (table/image-only calls).

Deliberately **not** included (kept minimal): auto-fallback to Docling OCR behind a config flag — happy to do that as a follow-up if there's interest; it needs a design call on config naming and defaults.

## Testing

- `python -m pytest tests/parse/ -v` → **25 passed** (22 pre-existing + 1 repaired + 3 new); before this PR the directory could not even be collected in a fresh process due to the `email.message` bug.
- New tests verified red first (warning not emitted on main behavior), then green after the fix.
- `black --check` / `isort --check-only` / `flake8 --max-line-length=88 --extend-ignore=E203,W503` clean on all new/edited lines; remaining flake8 findings in `pdf_parser.py` (`io` unused, two long lines) are pre-existing and untouched.


Sameer6305:
Hi @shanyu910, thanks for contributing! I noticed this PR isn’t linked to an existing issue. If this is for an existing issue, could you please comment on it and share the link here?

If this is a new issue, please raise it first and share it here so I can assign it to you. This helps us properly track the work and avoid duplicate efforts.

shanyu910:
Hi @Sameer6305, good catch — the PR body (and the issue body, it turns out) were accidentally truncated to `-` during submission, which is why the closing reference never registered. Both are now restored, and this PR properly **closes #1020**.

#1020 describes the bug (scanned PDFs silently parse to empty `full_text` with status `"completed"`) plus two test-suite blockers this PR also carries: the missing `import email.message` in `email_parser.py` and the broken `pdfplumber` mock in `test_pdf_parser`.

Happy to have the issue assigned — I'd like to take this on. Also glad to split the two enabling fixes into separate PRs if you prefer.

Sameer6305 (approved):
## Final Review — PR #1021

Thanks for the solid work on this PR, and especially for keeping the core fix focused on the behavior reported in #1020. I re-reviewed the final branch after the updates and the `main` merges, and the implementation is now **merge-ready**.

### What was solid

* The root cause was correctly identified: the pdfplumber path silently converts missing text into `""`, making scanned PDFs appear to parse successfully.
* The warning is actionable and points users toward the existing Docling OCR route with `enable_ocr=True`.
* The warning is correctly gated by `extract_text`, so legitimate `extract_text=False` use cases aren't affected.
* The fix correctly distinguishes completely textless PDFs from mixed PDFs.
* The `email.message` import fix is valid and addresses the fresh-process import issue required for the parse tests.
* The `pdfplumber` test mocking was correctly repaired given that the dependency is imported inside `PDFParser.parse()`.
* The added tests provide meaningful coverage for the scanned-PDF behavior.
* The PR remains appropriately scoped despite the branch having `main` merged into it.

### What was initially missing / what we tightened

During review we found a few things worth tightening before merge:

1. **Progress-tracking ordering**

   * `stop_tracking()` was originally called before the parser determined whether the document actually contained text.
   * We moved the tracking call after the no-text detection so the tracker can accurately report `(no text layer detected)` for scanned PDFs.

2. **Large-string `strip()`**

   * The original `full_text.strip()` check could create a temporary copy of the entire concatenated PDF text.
   * We changed this to a lazy page-level check using `any(page.text.strip() for page in pages)`, preserving the whitespace semantics while avoiding the full-string copy.

3. **Edge-case test coverage**

   * Added coverage for whitespace-only extracted text.
   * Added mixed-PDF coverage to ensure a document containing both text and image-only pages does not produce a false-positive document-level warning.

4. **Test-file churn**

   * The review initially introduced broad formatting changes into `test_parse_comprehensive.py`.
   * Those unrelated formatting changes were removed, leaving the PR focused on the actual functional mock repair.

### Validation

The final branch was revalidated with:

* `python -m pytest tests/parse/ -v` → **30 passed**
* `import semantica.parse` → **OK**
* `git diff --check` → **clean**
* New/changed code is clean under the relevant lint checks.
* Remaining Black/isort/flake8 findings were verified as **pre-existing on the base branch**, not introduced by this PR.
* Scanned, normal, mixed, whitespace-only, `extract_text=False`, and zero-page scenarios were verified.

Overall, this is a good, focused fix that addresses the original usability problem without changing successful parse semantics or introducing an automatic OCR fallback.

**Approved from my side. Thanks for the work and for keeping the implementation minimal and testable.**

Issue #962, opened 2026-08-13
[FEATURE] Semantica CrewAI Integration


KaifAhmad1:
## Problem Statement

CrewAI is one of the most widely adopted multi-agent orchestration frameworks, but Semantica has no first-class integration for it. Semantica already has a working pattern for framework integrations under [integrations/](../../integrations/): existing packages wrap Semantica's core (`ContextGraph`, `AgentContext`, extractors) as framework-native tools, and degrade gracefully when the target framework isn't installed. CrewAI should follow the same pattern so Crews can use Semantica's context graph and decision intelligence as native `BaseTool`s.

## Proposed Solution

Add `integrations/crewai/`, following the structure and optional-dependency pattern already established for other integrations in this repo:

```
integrations/crewai/
├── __init__.py           # exports, __version__, CREWAI_AVAILABLE flag
├── kg_tool.py             # SemanticaKGTool(BaseTool): extract/add/query the ContextGraph
├── decision_tool.py       # SemanticaDecisionTool(BaseTool): record/query decisions, precedents, causal chains
├── knowledge_source.py    # SemanticaKnowledgeSource: CrewAI Knowledge source backed by ContextGraph
└── README.md
```

Each tool should follow CrewAI's `crewai.tools.BaseTool` contract (Pydantic `args_schema`, synchronous `_run(self, **kwargs)`), and must degrade gracefully (importable, but flagged unavailable) when `crewai` isn't installed.

### `SemanticaKGTool` (`kg_tool.py`)

Backed by `semantica.semantic_extract.NERExtractor`, `RelationExtractor`, and `semantica.context.ContextGraph`.

```python
class SemanticaKGToolInput(BaseModel):
    action: Literal["extract_entities", "extract_relations", "add_to_graph", "query_graph", "find_related"]
    text: Optional[str] = None
    query: Optional[str] = None
    entity: Optional[str] = None
    hops: int = 1

class SemanticaKGTool(BaseTool):
    name: str = "semantica_knowledge_graph"
    description: str = "Extract entities/relations from text and query a shared knowledge graph."
    args_schema: Type[BaseModel] = SemanticaKGToolInput

    def _run(self, action: str, **kwargs) -> str: ...
```

### `SemanticaDecisionTool` (`decision_tool.py`)

Backed by `semantica.context.AgentContext`, exposing `record_decision`, `find_precedents`, `trace_causal_chain`, `analyze_impact`, and `check_policy`.

[...]

PR #1027, opened 2026-08-16
feat(context): add ErasureCoordinator for cross-store entity erasure

### LINKED ISSUE #1018: [FEATURE] Erasure coordinator — purge_node() erases the graph but leaves the same content live in AgentMemory and the vector store
## Problem Statement

`ContextGraph.purge_node()` / `purge_edge()` (#957) deliberately scope themselves to a single
graph. The changelog entry says so explicitly:

> Purge is graph-scope only: `AgentMemory` and any bound vector store are not reached, so it is
> one step of an erasure workflow rather than the whole of it

I made the right call for #957 — `ContextGraph` holds no reference to `AgentMemory` or to a
vector store, and giving it one would invert the dependency. But it leaves the feature's stated
motivation unfinished. The changelog names GDPR Article 17 as the reason purge exists, and an
Article 17 erasure that removes the node while the same content survives verbatim as a memory
item and as an embedding is not an erasure. It's worse than not offering one, because
`purge_node()` returns `True` and writes a tombstone attesting the content is gone.

Today the caller is left to complete the workflow by hand, and the pieces they need are
non-obvious in a way that makes silent partial erasure the likely outcome:

```python
graph.purge_node("e1")                      # graph: done
memories = memory.find_by_entity("e1")      # BUG: limit defaults to 10
for m in memories:
    memory.delete_memory(m["id"])
```

`AgentMemory.find_by_entity(entity_id, limit=10)` (`agent_memory.py:1288`) stops at ten results
and returns `results[:limit]`. An entity referenced by more than ten memory items silently keeps
the rest — an erasure check computed from a page that was already truncated by the very `limit`
it was called with.
Every caller has to know to override the default, and nothing in the signature suggests they
must.

## Proposed Solution

A coordinator that owns references to the stores and drives erasure across all of them, returning
an auditable record of what was reached and what was not.

`ContextGraph` keeps its current graph-scope contract unchanged — the coordinator composes it
rather than modifying it.

The critical property is **honest partial reporting**. Some vector backends cannot delete at all
(see Impact Assessment), so the coordinator must never report success it didn't achieve. A
receipt that says "graph: erased, memory: 14 erased, vectors: unsupported on FAISS" is
actionable. A silent `True` is a compliance liability.

## Alternatives Considered

**1. Give `ContextGraph` the references and cascade from `purge_node()`.**
Rejected. It inverts the dependency (`context` → `vector_store`), contradicts the documented
graph-scope contract that #957 established deliberately, and would put a large diff into
`context_graph.py` — which currently has ~840 added lines pending across #852, #967 and #921.
A separate module keeps the change off the contended file entirely.

**2. Subscribe to the existing mutation callback.**
This is the strongest alternative and deserves a real look: `purge_node()` already emits
`REMOVE_NODE` through the audit-trail callback, so a subscriber could erase downstream without
any new coordination API. Rejected for this use case because the callback is fire-and-forget —
there's no return path, so a failed vector delete is invisible to the caller who needs to certify
the erasure, and there's no way to express "this backend cannot delete." The callback remains the
right mechanism for *observing* erasure (audit logging, replication); it's the wrong one for
*certifying* it. Both can coexist.

**3. Leave it to callers (status quo).**
Rejected. The `find_by_entity` limit trap above means the obvious hand-rolled version is wrong,
and each caller re-derives the same cascade with the same bug.

## Use Cases

1. **GDPR Article 17 request.** A data subject asks for erasure. The operator needs every copy
   of that entity gone — graph, memory, embeddings — and needs a record proving it, including
   any store that could not be reached so it can be handled out of band.
2. **Poisoned context.** A decision was recorded against source data later found to be wrong or
   fraudulent. Retraction preserves explainability for the decision, but the underlying content
   must not remain retrievable by similarity search, or it resurfaces in the next RAG lookup —
   the graph node is gone while its embedding still answers queries.
3. **Tenant offboarding.** A customer leaves and their data must be removed across every store
   the deployment binds, in one auditable operation rather than a checklist someone runs by hand.

## Impact Assessment

- **Who would benefit?** Any deployment holding personal or regulated data — the users purge was
  built for. Also anyone running Semantica as RAG memory, where a stale embedding is a
  correctness bug rather than only a compliance one.
- **Priority**: Medium-High. The capability is advertised (GDPR Art. 17 in the changelog) but
  only partly delivered, and the gap fails in the direction of a false assurance.
- **Breaking Changes**: No. New module, additive. No existing signature or behavior changes.
- **Dependencies**: None new. Builds on APIs that already exist:
  - `ContextGraph.purge_node()` / `purge_edge()` (#957)
  - `AgentMemory.find_by_entity()` (`agent_memory.py:1288`), `delete_memory()` (`:576`),
    `batch_delete()` (`:1563`)
  - `AgentMemory.delete_memory(skip_vector=False)` already cascades memory → vector store via its
    `_vector_ids` map, so that leg is solved. The missing link is graph → memory.

**Known limitation, and I'd rather state it up front than discover it in review:** three vector
backends have no delete method at all — `faiss_store.py`, `milvus_store.py`, `weaviate_store.py`.
FAISS in particular can't remove from a flat index without a rebuild. So erasure is genuinely not
completable on those backends today. That's precisely why the receipt reports `unsupported`
rather than swallowing it, and it may warrant its own issue: `delete_vectors()` is declared on the
`VectorStore` facade (`vector_store.py:786`) but not implemented across the backend set — the
same multi-backend divergence that makes a change tested only against the default pass locally
and break everywhere else.

## Implementation Ideas

```python
from semantica.context import ContextGraph, AgentMemory
from semantica.context.erasure import ErasureCoordinator

coordinator = ErasureCoordinator(graph=graph, memory=memory)

receipt = coordinator.erase_entity(
    "customer-4471",
    reason="GDPR Art. 17 request #882",
)

receipt.complete          # False -- at least one store could not be reached
receipt.to_dict()
# {
#   "entity_id": "customer-4471",
#   "reason": "GDPR Art. 17 request #882",
#   "erased_at": "2026-08-15T22:41:07+00:00",
#   "stores": {
#     "graph":   {"status": "erased",      "nodes": 1, "edges": 3},
#     "memory":  {"status": "erased",      "items": 14},
#     "vectors": {"status": "unsupported", "backend": "faiss",
#                 "detail": "backend exposes no delete; index rebuild required"},
#   },
# }
```

Sketch of the semantics:

- **Ordering.** Erase outward-in — vectors, then memory, then the graph last. The graph tombstone
  is the durable record that an erasure was attempted; writing it before the downstream work
  means a crash mid-cascade leaves a tombstone claiming more than happened.
- **Unbounded lookups.** The memory sweep must pass an explicit unbounded limit to
  `find_by_entity`, never the default of 10, and should assert it drained the set rather than
  trusting one page.
- **Partial failure is a result, not an exception.** A backend that can't delete is reported in
  the receipt; the coordinator continues with the remaining stores rather than aborting and
  leaving a half-erased state with no record of which half.
- **Optional stores.** `memory=None` / no bound vector store is normal; those legs report
  `not_configured`, distinct from `unsupported` and from `failed`.

## Additional Context

- Builds directly on #955 / #957 (retraction and purge for `ContextGraph`), whose changelog entry
  documents this exact gap as out of scope.
- #765 (wiki-style editable memory view) touches `AgentMemory` conceptually but is about viewing
  and editing rather than erasure — adjacent, not overlapping.
- The three delete-less vector backends noted above are arguably a separate issue; happy to split
  that out if maintainers prefer, since it blocks *completing* erasure on those backends but not
  the coordinator itself.
- Prior art: this is the shape most "right to be forgotten" implementations converge on — a
  fan-out erasure driver plus a receipt, rather than each store's delete being called ad hoc by
  application code.

## Contribution

- [x] I can help with testing

pravit-amp:
## Description

`ContextGraph.purge_node()` is graph-scope by design (#957), and its changelog entry says so explicitly. That leaves the feature's stated motivation unfinished: the changelog names GDPR Article 17 as the reason purge exists, and an Article 17 erasure that removes the node while the same content survives verbatim as an `AgentMemory` item and as an embedding is not an erasure. It's worse than not offering one, because `purge_node()` returns `True` and writes a tombstone attesting the content is gone.

This adds `ErasureCoordinator`, which owns references to the stores, drives erasure across all of them, and returns an auditable `ErasureReceipt` of what was reached and what was not.

`ContextGraph` keeps its graph-scope contract unchanged — the coordinator **composes** the existing public APIs rather than modifying them. Nothing in `context_graph.py` or `agent_memory.py` changes behaviorally (the only edit to `context_graph.py` is a docstring pointer), which also keeps this off a file with large diffs pending across #852, #967 and #921.

The property that matters is **honest partial reporting**. Three vector backends cannot delete at all, so the coordinator must never report a success it didn't achieve. A receipt reading `graph: erased, memory: 14 erased, vectors: unsupported on faiss` is actionable; a silent `True` is a compliance liability.

```python
from semantica.context import ErasureCoordinator

coordinator = ErasureCoordinator(graph=graph, memory=memory)
receipt = coordinator.erase_entity("customer-4471", reason="GDPR Art. 17 request #882")

if not receipt.complete:
    print(receipt.incomplete_stores)   # ['vectors'] -- handle out of band
```

### Scope note

Per @Sameer6305's guidance on #1018, this is the standalone follow-up covering the coordinator itself. #1024 by @yzxcj797 separately fixes the `find_by_entity()` truncation this issue also cited; the two do not overlap and this PR does not touch `agent_memory.py`. The memory sweep here is written to be correct **either way** — see "Deviations from the plan" below.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #1018

Builds on #955 / #957 (retraction and purge for `ContextGraph`), whose changelog entry documents this exact gap as out of scope. Adjacent to #1024, which fixes the `find_by_entity()` limit trap independently.

## Changes Made

- **New `semantica/context/erasure.py`**, exporting `ErasureCoordinator` and `ErasureReceipt` from `semantica.context`.
  - `erase_entity(entity_id, reason=..., at=..., vector_ids=...)` → `ErasureReceipt`
  - `erase_entities([...], reason=...)` → one receipt per entity, in order, so one entity's failure doesn't stop the rest
- **Five per-store statuses**, so a shortfall is never indistinguishable from a success: `erased`, `not_found`, `not_configured` (store never bound — normal), `unsupported` (store cannot delete at all; retrying won't help), `failed`. `receipt.complete` is `False` when any store reports `unsupported`/`failed`, and `receipt.incomplete_stores` names them. Note `erased` is a **weaker claim on the vectors leg** than on the memory leg — see Additional Notes.
- **Erasure runs outward-in: vectors → memory → graph.** The graph tombstone is the durable attestation that an erasure happened, so writing it first would let a crash mid-cascade leave a record claiming more than occurred. Erasing the graph last means a partial failure leaves the node *present* and the receipt incomplete — recoverable and honest; the reverse is neither.
- **Partial failure is a result, not an exception.** A store that raises is recorded as `failed` (with the exception type) and the remaining legs still run, rather than aborting into a half-erased state with no record of which half.
- **The memory sweep cannot be silently truncated.** It pages until dry — deleting as it goes, so the next page is the remainder — rather than passing one large `limit=` that's only correct until someone exceeds it, then **re-queries once afterwards** and reports `failed` with the residual count if anything survived. It also stops rather than spinning if `batch_delete` reports no progress on a non-empty page.
- **`unsupported` vector backends are detected by probing, not by calling and catching** — see below. Backends are reached under either supported name, `delete_vectors(ids)` (pinecone/qdrant) or `delete(ids)` (pgvector/sqlite-vec), and the receipt records which was used.
- **Every store is optional.** `vector_store` defaults to `memory.vector_store`, stays overridable for deployments binding a store the memory doesn't own, and accepts `False` to disable the leg. Vectors owned by memory items are removed by the memory leg's own `delete_memory()` cascade; the explicit vector leg covers entity-keyed embeddings written by something other than `AgentMemory`.
- **Docs**: `purge_node()`'s docstring now points at the coordinator, so callers reading the graph-scope caveat find the thing that completes the workflow; new sections in `docs/reference/context.md` and `semantica/context/context_usage.md`; CHANGELOG entry.

### Deviations from the plan I posted on #1018

Three things the code turned up that are worth a reviewer's attention:

1. **`find_by_entity()` returns items keyed `memory_id`, not `id`.** The cascade sketched in the issue body (`memory.delete_memory(m["id"])`) would `KeyError`. Handled, and a memory item carrying no identifier reports `failed` rather than looking erased.
2. **The `unsupported` probe has to look at the wrapped backend, not the facade.** `delete_vectors()` is declared on the `VectorStore` facade (`vector_store.py:786`) for *every* backend and only raises `NotImplementedError` once called, so probing the facade alone cannot tell a deletable backend from a delete-less one. Probing rather than calling-and-catching also keeps a missing method distinguishable from an `AttributeError` raised *inside* a working one — exactly where guessing wrong produces a false clean bill of health. `NotImplementedError` at call time is still caught and reported as `unsupported`; a store returning `False` is reported as `failed`.
3. **`erased_at` is normalized through `ContextGraph`'s own temporal normalizer.** Without it the receipt read `...T00:00:00Z` while the tombstone for the same erasure read `...T00:00:00` — an audit record contradicting the tombstone it attests to. Normalizing up front also rejects an unparseable `at` before any store is touched, instead of half way through the cascade.

I also **kept** the loop-until-dry memory sweep that my #1018 amendment said #1024 would make unnecessary. #1024 hasn't merged, so `limit=10` is still the default on `main`; the loop is correct under both the old and new defaults, so nothing here needs to change when #1024 lands.

### Answers to the open questions from #1018

- **Does the receipt persist?** Not in this PR — `to_dict()` is serialization-ready and safe to store, but I'd rather not guess at a storage contract. Easy follow-up if maintainers want it durable.
- **Does `erase_entity` accept edges?** No. Memory and vectors are keyed on entities, so an edge-scoped erasure has only one leg to run; callers should use `purge_edge()` directly.
- **Module placement:** `semantica/context/erasure.py`, as it reads as context-layer today. Happy to move it to a top-level `semantica/erasure/` if the intent is for it to eventually reach graph stores and triplet stores.

## Testing

- [x] Tested locally

New `tests/context/test_erasure_coordinator.py` — **30 tests**, run against **real** `ContextGraph` and `AgentMemory` instances rather than mocks. The bug this feature exists to prevent lives in the interaction between them, so mocking that interaction away would test nothing. Most vector stores are fakes, because the point of those tests is backend *shape* — but a fake is worth exactly as much as the assumption that a real store looks like it, so three tests run against a real `VectorStore(backend="inmemory")` as well.

- **The truncation regression**: 25 memory items on one entity, more than any plausible default limit. A coordinator that calls `find_by_entity()` once with the default fails this test.
- **All three vector-backend shapes** — `delete_vectors` / `delete` / neither → `erased` / `erased` / `unsupported` — plus the facade-over-a-delete-less-backend shape, and a store returning `False` → `failed`.
- **Partial failure**: a raising store is recorded as `failed`, the graph leg still runs, and nothing raises through to the caller.
- **Memory-leg failure modes**: residual items after the sweep, `batch_delete` making no progress, items with no identifier — all `failed`, never silently `erased`.
- **Idempotency**: a second `erase_entity()` reports nothing left to do rather than raising, matching how `purge_node()` already returns `False` on a repeat.
- **Receipt semantics**: `complete`/`incomplete_stores`, `to_dict()` copying rather than aliasing the store results, and `erased_at` agreeing with the tombstone's `purged_at`.
- **Against a real `VectorStore`**: the ids given are really removed (`count()` 2 → 0); the **full three-store cascade** with a real `ContextGraph` + `AgentMemory` + `VectorStore`, where the embedding is written by `AgentMemory.store()` and has to be gone afterwards — that exercises the memory leg's own `delete_memory()` vector cascade rather than the coordinator's model of it; and the `erased`-means-accepted caveat below.

### Test Commands

```bash
# New tests
pytest tests/context/test_erasure_coordinator.py -q
# 30 passed

# Full context suite (569 passed before this PR)
pytest tests/context/ -q
# 599 passed

# Build
python -m build
# Successfully built semantica-0.6.5.tar.gz and semantica-0.6.5-py3-none-any.whl

# Format / lint
black --line-length=88 semantica/context/erasure.py tests/context/test_erasure_coordinator.py
isort --profile black --line-length=88 semantica/context/erasure.py tests/context/test_erasure_coordinator.py
flake8 --max-line-length=88 --extend-ignore=E203,W503 semantica/context/erasure.py tests/context/test_erasure_coordinator.py
```

## Documentation

`docs/reference/context.md` gains an `ErasureCoordinator` section (constructor parameters, methods, the status table, and the receipt shape) plus a row in the Exported Classes table; `semantica/context/context_usage.md` gains a walkthrough; `purge_node()`'s docstring points at the coordinator. No cookbook notebook — happy to add one if that's wanted.

## Breaking Changes

**Breaking Changes**: No

New module, purely additive. No existing signature, return type, or default changes; the only edit to an existing Python file besides the package `__init__` is a docstring addition on `purge_node()`.

## Checklist

`black`, `isort --profile black`, and `flake8 --max-line-length=88 --extend-ignore=E203,W503` are all clean on the new files. The new test module emits zero warnings. Annotations use `typing` generics rather than the `collections.abc` subscripting that needs 3.9, per the declared 3.8 floor.

## Additional Notes

**`erased` is a weaker claim on the vectors leg than on the memory leg, and the receipt should be read with that in mind.** The memory leg re-queries after its sweep and only claims `erased` once nothing referencing the entity comes back. The vectors leg cannot do the same: backends delete by id and report success whether or not the id was there — `VectorStore(backend="inmemory").delete_vectors(["never-embedded"])` returns `True` with the count unchanged — and there is no portable "does this id exist" check to confirm against. So `vectors: {"status": "erased", "vector_ids": 1}` means *the store accepted the delete for one id*, not *one embedding was really removed*. Documented on `STATUS_ERASED` and `_erase_vectors()`, and pinned by a test so it can't quietly drift into a stronger-sounding claim. I'd rather state this in the open than have an operator read the receipt as more than it is — but if reviewers would prefer the leg attempt a `count()`-based confirmation where the backend exposes one, I'm happy to add it.

**Known limitation, unchanged by this PR and worth its own issue:** erasure still cannot be *completed* on FAISS, Milvus, or Weaviate — `faiss_store.py`, `milvus_store.py`, and `weaviate_store.py` expose no delete method at all, and FAISS in particular can't remove from a flat index without a rebuild. `delete_vectors()` is declared on the `VectorStore` facade but implemented across the backend set under at least three different names. That's the same multi-backend divergence that makes a change tested only against the default pass locally and break everywhere else. The coordinator ships reporting `unsupported` for those, and starts reporting `erased` once the backends gain a delete — no API change needed here. I'm happy to open that issue separately.

Reviewers may also want to weigh in on whether `complete` should treat `not_configured` as complete. It does today: a store that was never bound leaves no residue, and only `unsupported`/`failed` mean data survived. The alternative reading — that an unconfigured vector store means "you don't know whether embeddings exist elsewhere" — is defensible, but it would make every graph-only erasure report incomplete, which I think dulls the signal.


pravit-amp:
Pushed two follow-ups (`7a63ae5`, `9f99335`) and updated the description to match. Both came out of self-review rather than anything a reviewer flagged, so here's what changed and why.

### The vector-leg tests were only as good as my guess about the backends

The original tests asserted the three backend shapes the coordinator expects — `delete_vectors` / `delete` / neither — against fakes I wrote. That's circular: the fakes returned `True` because I made them, which said nothing about whether a real store's `True` means anything. Given this repo's recurring failure mode is a change verified against the default that reaches for internals and breaks on every other backend, "I tested the shapes I imagined" wasn't good enough.

`VectorStore(backend="inmemory")` runs without external services, so three tests now use it:

- the ids handed to it are really removed (`count()` 2 → 0)
- **the full three-store cascade** — real `ContextGraph` + `AgentMemory` + `VectorStore`, where the embedding is written by `AgentMemory.store()` and has to be gone afterwards. This is the one I most wanted: it exercises the memory leg's own `delete_memory()` vector cascade instead of my model of it, and it passes.
- the caveat below.

### The real backend surfaced a caveat the fakes structurally could not

`VectorStore(backend="inmemory").delete_vectors(["never-embedded"])` returns `True` and leaves the count untouched — it pops the ids and reports success whether or not they were there. No backend offers a portable "does this id exist" check, so the vectors leg cannot confirm the way the memory leg does (which re-queries after its sweep and only claims `erased` once nothing comes back).

So `vectors: {"status": "erased", "vector_ids": 1}` means **the store accepted the delete for one id**, not **one embedding was really removed**.

For a feature whose entire purpose is not overstating what was erased, that asymmetry shouldn't be buried. It's now documented on `STATUS_ERASED` and `_erase_vectors()`, called out in both status tables, and pinned by a test so it can't drift into a stronger-sounding claim later. If you'd prefer the leg attempt a `count()`-based confirmation where a backend exposes one, say the word — I left it out because it isn't portable and is racy under concurrent writes, but it's a small change.

### Docs

`docs/reference/context.md` had no `ErasureCoordinator` entry, so the API reference now carries one (constructor parameters, methods, status table, receipt shape) plus a row in the Exported Classes table.

Worth flagging separately: `purge_node()` / `retract_node()` appear nowhere in `docs/` — #957 shipped without touching the reference. Out of scope here, but happy to open a docs issue for it.

`tests/context/`: **599 passed** (569 before this PR). `black` / `isort` / `flake8` clean; new tests emit no warnings.

ZohaibHassan16:
This looks good overall.

There's a conflict in CHANGELOG.md against main. Could you rebase/merge and resolve it? Should just be your entry and a newer one both landing near the top of [Unreleased].

pravit-amp:
Thanks @ZohaibHassan16 — merged `main` in and pushed as 7d335571.

One wrinkle worth flagging, since it turned out not to be two entries landing in `[Unreleased]`: 0.6.6 was cut on 2026-08-20, which rolled this branch's other unreleased entries into the release and left `[Unreleased]` empty on main. Both sides still share an `### Added` heading, so git aligns on that and drops the `ErasureCoordinator` entry *inside* the released 0.6.6 section — filing an unreleased feature under a published release. Worth knowing if the same conflict shows up on other branches that predate the release.

Resolved so `[Unreleased]` holds only the `ErasureCoordinator` entry and 0.6.6 stays intact. Also dropped this branch's stale `- **First-class CrewAI integration** (#962)` stub in favour of main's finished `(#988, closes #962) by @Shindevrp`.

`tests/context/` is green on the merged tree — 630 passed, including the 39 erasure tests. GitHub now reports the PR as mergeable, so it's back to you whenever you have a moment.

ZohaibHassan16 (commented):
I found an issue with the receipt reporting.

The vector leg deletes `vector_ids` or `[entity_id]` before the memory leg runs. But `AgentMemory` has its own vector IDs for each memory item and `delete_memory()` deletes those internally.

The problem is if deleting one of those memory vector IDs fails, `AgentMemory.delete_memory()` catches the error and logs it, but still deletes the memory item.

Because of that, the coordinator thinks the memory delete was successful and can return:

```python
receipt.complete is True
```

even though the embedding is still in the vector store.

I reproduced this using `AgentMemory`.

Deleting:

```python
["customer-4471"]
```

returned `True` but deleting the actual vector ID used by the memory returned `False`.

The final receipt still reported that the vectors were erased and the receipt was complete.

I think we should either return the vector delete result from `AgentMemory`, or let the coordinator delete/verify the memory-owned vector IDs before marking the receipt as complete.

Also, we should add a regression test for when deleting a memory-owned vector fails.

The existing focused tests are passing (`88 passed`), but they only test the successful memory-vector cascade. They don't cover this failure case.

pravit-amp:
Thanks for merging `main` in @ZohaibHassan16 — I'd pushed my own merge for the CHANGELOG just before that, and yours has it as an ancestor, so nothing was lost. I rebased onto your head rather than merging again, so the branch stays linear.

Heads up that there's now a **behaviour fix** on top (cbff79c1), not just the conflict resolution — worth a look before you approve, since it changes what `receipt.complete` means in one case.

**The bug.** The vector leg deleted `vector_ids` or `[entity_id]`, and the memory leg relied on `AgentMemory.delete_memory()` to cascade to the vectors each memory item owns. That cascade is best-effort: `_delete_vector_ids()` raises when a backend returns `False`, `delete_memory()` catches it, logs a warning, and still returns `True`. So `batch_delete` counted the item, the residual re-query found no *items*, and the memory leg reported `erased` — while the embedding was still sitting in the vector store. Reproduced with a store that accepts the entity-keyed id and refuses the memory-owned one: `receipt.complete` came back `True` with the embedding live.

That's the exact failure this module exists to prevent — a receipt that overstates is worse than no receipt, because it attests the content is gone.

**The fix.** Memory-owned vector ids are now deleted through the coordinator's own vector leg, which reports honestly, instead of trusting the memory leg's cascade. They're collected before anything is deleted, while the items still exist to be enumerated, and unioned with any caller-supplied ids rather than replacing them.

**One thing I want to flag rather than bury**, since it contradicts what this PR's description promises. Making the receipt truthful required a small addition to `agent_memory.py`: a read-only `vector_ids_for(memory_id)` accessor. `_vector_ids` is private, and there's no portable way to ask a vector store what it still holds, so the coordinator had no way to make the claim honest without it. The PR previously said nothing in `agent_memory.py` changes — that no longer holds. No *existing* behaviour changes: `delete_memory()` still cascades best-effort and other callers are unaffected; the coordinator simply stops depending on that being reliable. The cost is that vectors get attempted twice, which is a no-op on a working store.

If you'd rather keep `agent_memory.py` untouched, the alternative is making `delete_memory()` propagate the vector failure — but that changes a public API's contract for every caller, and returning `False` when the item genuinely was deleted is its own kind of dishonest. I went with the additive accessor as the smaller blast radius, but I'm happy to switch if you disagree.

Four tests cover it: the refused-vector case, that memory-owned ids reach the store, that explicit `vector_ids` don't displace them, and the accessor's fallback. I checked the first three fail against the previous coordinator on the `receipt.complete` assertion specifically, rather than passing vacuously. 694 passed across `tests/context/` and the agno integration on your merged head.

ZohaibHassan16 (dismissed):
Thanks for the follow-up. The original issue looks fixed but I found 2 more cases where the receipt can still say `complete=True` even though a memory embedding is still left in the vector store.

First one is around `_all_vector_ids()`. It only reads the first 500 memory items, but `_erase_memory()` keeps paging after that and calls `batch_delete()`. That ends up using `delete_memory()`, where vector deletion is still best effort.

So for example, if we have 501 matching memory items and deleting the vector for item 501 fails, the first vector delete can still succeed, all the memory items get deleted and the final receipt can still say `complete=True` even though that last embedding is still there.

Can we collect/delete the memory-owned vector ids for each page before deleting that page's memory items? Also please add a test with 501 items where the vector delete for the last item returns `False`.

There is another case when `vector_store=False` or when the coordinator vector store is different from `memory.vector_store`.

The receipt is based on `self.vector_store` but `batch_delete()` still lets the memory store do its own best-effort vector delete. If that delete fails, it only gets logged, so again the receipt can look complete while the vector is still there.

I think we should either report both vector stores separately or skip the internal memory vector cascade once the coordinator already handled those vector ids.

Please add a regression test for separate coordinator and memory vector stores too.

ZohaibHassan16 (commented):
Thanks for the follow-up. The original issue looks fixed but I found 2 more cases where the receipt can still say `complete=True` even though a memory embedding is still left in the vector store.

First one is around `_all_vector_ids()`. It only reads the first 500 memory items, but `_erase_memory()` keeps paging after that and calls `batch_delete()`. That ends up using `delete_memory()`, where vector deletion is still best effort.

So for example, if we have 501 matching memory items and deleting the vector for item 501 fails, the first vector delete can still succeed, all the memory items get deleted and the final receipt can still say `complete=True` even though that last embedding is still there.

Can we collect/delete the memory-owned vector ids for each page before deleting that page's memory items? Also please add a test with 501 items where the vector delete for the last item returns `False`.

There is another case when `vector_store=False` or when the coordinator vector store is different from `memory.vector_store`.

The receipt is based on `self.vector_store` but `batch_delete()` still lets the memory store do its own best-effort vector delete. If that delete fails, it only gets logged, so again the receipt can look complete while the vector is still there.

I think we should either report both vector stores separately or skip the internal memory vector cascade once the coordinator already handled those vector ids.

Please add a regression test for separate coordinator and memory vector stores too.

Sameer6305:
Thanks @pravit-amp  for the thoughtful implementation and for addressing the review feedback thoroughly. The coordinator is well-scoped, the receipt semantics are clear, and the cross-store erasure and failure handling are well covered.

During review, we identified and fixed the important edge cases around memory-owned vector IDs:

* Fixed vector-ID collection across memory pages so items beyond the first batch cannot be missed.
* Added coverage for separate/disabled vector stores and failed memory-owned vector deletion.
* **The 501-item pagination test was initially getting stuck for 5+ minutes. We investigated and fixed the test setup, reducing it to ~17 seconds while still exercising the pagination boundary and verifying that the old buggy implementation fails.**
* Verified the full **47 erasure-coordinator tests pass**.
* Resolved the `CHANGELOG.md` conflict with current `main`.

The final implementation now avoids falsely reporting `receipt.complete=True` when a memory-owned embedding can remain.

**Approved from my side ✅ @KaifAhmad1, ready for your final review and merge.**

Sameer6305:
@pravit-amp can you fix the conflict.

Sameer6305 (approved):
## PR Review — `feature/erasure-coordinator` → `main` (closes #1018)

**Verdict: ✅ APPROVED — ready to merge**

---

### What this PR fixes

Before this PR, `ContextGraph.purge_node()` was the only erasure primitive in Semantica. It is graph-scope by design: the node is removed and a tombstone is written attesting the content is gone — but the exact same content could still be sitting live as an `AgentMemory` item and as a vector embedding. A GDPR Article 17 erasure that removes the graph node while the content remains retrievable by similarity search is not an erasure; it is worse than not offering one, because `purge_node()` returns `True` and writes a tombstone claiming the content is gone.

This PR completes the erasure workflow that `purge_node()` only starts.

---

### What was built

**New file: `semantica/context/erasure.py`**

`ErasureCoordinator` drives a cascade across every store the caller supplies — vector store, then memory, then graph — and returns an `ErasureReceipt` recording exactly what each store reported.

Example receipt:

| Store | Status | Details |
|---|---|---|
| **vectors** | `unsupported` | backend: `faiss` |
| **memory** | `erased` | 14 items |
| **graph** | `erased` | 1 node, 3 edges |

**Receipt:**
- `entity_id`: `customer-4471`
- `reason`: `GDPR Art. 17 request #882`
- `erased_at`: `2026-08-16T09:03:36.813220`
- `complete`: `False`

The important property is that the receipt explicitly reports the incomplete vector leg instead of claiming that the entire erasure succeeded.

**Key design properties:**

- **Composes existing APIs, changes nothing.** `ContextGraph` and `AgentMemory` keep their existing contracts. No existing signatures changed. The coordinator adds one new method to `AgentMemory` (`vector_ids_for`) and updates the `purge_node()` docstring to point callers here.
- **Honest partial reporting.** Each store reports one of five statuses: `erased`, `not_found`, `not_configured` (never bound — normal), `unsupported` (cannot delete; retrying won't help), `failed`. `receipt.complete` is `False` when any store reports `unsupported` or `failed`, with `receipt.incomplete_stores` naming them. FAISS, Milvus, and Weaviate expose no delete method — the coordinator says so rather than reporting a success it did not achieve.
- **Outward-in ordering: vectors → memory → graph.** The tombstone is the durable attestation that an erasure happened, so it is written last. A crash mid-cascade leaves the node present and the receipt incomplete — recoverable and honest. Writing the tombstone first would create a record claiming more than occurred.
- **Partial failure is a result, not an exception.** A store that raises is recorded as `failed` with the exception type, and the remaining stores still run. Aborting on first failure would leave a half-erased state with no record of which half.
- **Memory sweep cannot be silently truncated.** The coordinator sweeps in pages of 500 (deleting as it goes so the next page is the remainder), then re-queries once after the sweep and reports `failed` with the residual count if anything survived. Stops rather than spinning if `batch_delete` makes no progress.
- **Vector backend capability is detected by probing, not by calling and catching.** The `VectorStore` facade declares `delete_vectors()` for every backend and only raises `NotImplementedError` once called — probing the facade alone cannot tell a deletable backend from a delete-less one. The coordinator looks at `_backend_store` directly.

---

### What happened during this PR (workflow summary)

1. **Initial implementation** (`d1b37ff7`) — `ErasureCoordinator`, `ErasureReceipt`, five status constants, vector/memory/graph legs, helper functions, basic tests.

2. **Docs commit** (`f4b2d081`) — `docs/reference/context.md` and `context_usage.md` sections added with API table, status table, code examples, and the "always check the receipt" warning.

3. **Real VectorStore test** (`dd526d26`) — End-to-end test against `VectorStore(backend="inmemory")` rather than a fake, to catch shape assumptions that pass against fakes but break on real backends.

4. **Review fixes — round 1** (`d287c823`) — Addressed Qodo findings: constructor's "at least one store" guard was using `not vector_store` (rejecting valid falsey stores); `at` type annotations were narrower than the `ContextGraph` normalizer they delegate to; several receipt-shape edge cases.

5. **Memory-owned vector reporting** (`bac071f4`) — `AgentMemory.delete_memory()` deletes vectors best-effort and still returns `True` on failure. The coordinator cannot learn from the memory leg whether embeddings actually went away, so it collects all memory-owned vector IDs before any deletion starts and deletes them through the vector leg, which reports honestly. Added `AgentMemory.vector_ids_for()` to expose those IDs.

6. **Vector failure honesty** (`78deac78`) — No in-repo backend returns a bare `bool` — Qdrant returns `{"status": <UpdateStatus>}`, Pinecone `{"deleted": True}`. A bare `result is False` check called every dict a success. Added `_interpret_delete_result()` to handle bool, dict with named failure markers, `None` (void method), and everything else, with the backend's own payload preserved in `backend_result` stringified for JSON-serializability.

7. **Pagination fix** (`ebe1deb7`) — `_all_vector_ids()` originally called `find_by_entity(limit=500)` once, collecting only the first 500 items' vector IDs while `_erase_memory()` continued paging and deleted all 501+ items. The vector belonging to item 501 remained while the receipt reported `complete=True`. Fixed to use `limit=None` for the vector collection pass (unbounded, before any deletion), while `_erase_memory()` retains its defensive paged sweep.

8. **Batch timestamp fix** (`77dbffc6`) — `erase_entities()` was calling `erase_entity()` in a loop, each independently calling `_normalize_timestamp(None)` → `datetime.now()`. A batch GDPR request produced tombstones with diverging `purged_at` values per entity. Fixed: `erase_entities()` resolves the timestamp once before the loop and passes the resolved value to every `erase_entity()` call.

9. **Whitespace** (`9484dc2f`) — Stripped trailing whitespace from 11 blank lines (2 in `erasure.py`, 9 in the test file) flagged by `git diff --check`.

10. **CHANGELOG rebase** — Conflict between the PR branch and `upstream/main`'s newer `[0.6.7]` release was resolved correctly: ErasureCoordinator entry placed under `## [Unreleased]` / `### Added`, all upstream/main release history preserved intact, no duplicate entries.

11. **CHANGELOG count correction** (`b37f74d0`) — Stale test counts from the initial commit (`39 tests` / `608 passed`) updated to reflect the final state after all review rounds (`48 tests` / `738 passed`).

---

### Files changed

| File | Change |
|---|---|
| `semantica/context/erasure.py` | **New** — `ErasureCoordinator`, `ErasureReceipt`, all helpers |
| `semantica/context/agent_memory.py` | **+21 lines** — `vector_ids_for()` method |
| `semantica/context/context_graph.py` | **+5 lines** — `purge_node()` docstring update pointing to coordinator |
| `semantica/context/__init__.py` | **+4 lines** — exports `ErasureCoordinator`, `ErasureReceipt` |
| `semantica/context/context_usage.md` | **+76 lines** — usage guide section |
| `docs/reference/context.md` | **+95 lines** — API reference section |
| `CHANGELOG.md` | **+23 lines** — `[Unreleased]` entry |
| `tests/context/test_erasure_coordinator.py` | **New** — 48 tests |

**1,827 insertions, 1 deletion. 48 tests pass. Full `tests/context/` suite: 738 passed, 14 skipped. `git diff --check upstream/main...HEAD` → clean.**

---

### Known limitation (documented, not introduced by this PR)

Erasure cannot be completed on FAISS, Milvus, or Weaviate backends — `delete_vectors()` is declared on the `VectorStore` facade but not implemented in those backend stores. The coordinator ships reporting `unsupported` for them. When backend deletion lands for those three, the coordinator starts reporting `erased` with no API change needed here.

PR #1038, opened 2026-08-16
fix(config): honor boolean env overrides in Config.get()

### LINKED ISSUE #1035: Config.get() silently drops boolean environment overrides: the bool branch is unreachable because bool subclasses int
A boolean environment override is silently dropped by `Config.get()`, because the `bool` branch of its type dispatch can never run.

Every module config's `get()` has this chain:

```python
if isinstance(default, int):
    return int(value)
elif isinstance(default, float):
    return float(value)
elif isinstance(default, bool):
    return value.lower() in ("true", "1", "yes", "on")
return value
```

`bool` is a subclass of `int` in Python, so `isinstance(False, int)` is `True` and a boolean default takes the first branch. `int("true")` raises, the surrounding `except (ValueError, TypeError): pass` swallows it, and the default is returned as though no environment variable were set.

## Reproduction

```
CONFLICT_ZZTESTFLAG=true
```

```python
from semantica.conflicts.config import ConflictsConfig

c = ConflictsConfig()
c.get("zztestflag", False)   # -> False   the env var is ignored
c.get("zztestflag", "x")     # -> 'true'  the same env var, read fine
```

The string case is the control: the variable is found and returned, so this is the type dispatch, not the lookup. With `CONFLICT_ZZTESTFLAG=1` it returns `1`, an `int` rather than a `bool`, which is the other half of the same problem.

## Which modules are affected

The dead branch is in ten config modules:

```
semantica/conflicts/config.py:139      semantica/normalize/config.py:139
semantica/deduplication/config.py:140  semantica/ontology/config.py:140
semantica/embeddings/config.py:152     semantica/parse/config.py:139
semantica/export/config.py:151         semantica/split/config.py:126
semantica/ingest/config.py:172         semantica/kg/config.py:152
```

but it only bites in three of them, and the reason is worth knowing before choosing a fix. Seven modules also run a generic scanner in `_load_env_vars`:

```python
for key, value in os.environ.items():
    if key.startswith(env_prefix) and key not in env_mappings:
```

which puts any `PREFIX_*` variable into `_configs`, so `get()` returns from the config dict and never reaches the broken dispatch. `conflicts`, `deduplication` and `split` have no such scanner, so for any key outside their fixed `env_mappings` the broken path is the only route.

Measured in a separate process per module, setting `PREFIX_ZZTESTFLAG=true`:

```
conflicts       CONFLICT_      bool-default=False    str-default='true'    <- dropped
deduplication   DEDUP_         bool-default=False    str-default='true'    <- dropped
split           SPLIT_         bool-default=False    str-default='true'    <- dropped
embeddings      EMBEDDING_     bool-default=True     str-default=True
export          EXPORT_        bool-default=True     str-default=True
ingest          INGEST_        bool-default=True     str-default=True
kg              KG_            bool-default=True     str-default=True
normalize       NORMALIZE_     bool-default=True     str-default=True
ontology        ONTOLOGY_      bool-default=True     str-default=True
```

`parse` has the scanner too; I could not import it here because of an unrelated missing `docx` dependency, so I am not claiming a measurement for it.

## How reachable is it

Not from inside the repo today, as far as I can tell. The in-repo `.get(...)` calls with boolean defaults, for example `conflict_detector.py:123` and `:126`, are `dict.get` on a plain `self.config = config or {}`, not this class. The only in-repo uses of `conflicts_config.get()` are the docstring and `conflicts_usage.md`, both with a float default.

So this is a latent defect on a public surface rather than a live incident: `conflicts_config` is exported from `semantica/conflicts/__init__.py`, and the class docstring advertises the "config file -> environment -> defaults" fallback chain, so a caller reading that would reasonably expect `get("auto_resolve", False)` to honour `CONFLICT_AUTO_RESOLVE`.

Worth noting that the same file gets it right in `_load_env_vars`, where `type_func == bool` is compared by identity rather than by `isinstance`, so the two env paths in one class disagree.

## Fix

Moving the `bool` check above the `int` check is enough, and it is the same one-line reorder in all ten:

```python
if isinstance(default, bool):
    return value.strip().lower() in ("true", "1", "yes", "on")
elif isinstance(default, int):
    ...
```

The separate question is whether `conflicts`, `deduplication` and `split` should also get the generic `PREFIX_*` scanner the other seven have, so the two env paths stop disagreeing. That is a design call rather than a bug fix and I did not want to presume on it.

Happy to send a PR for the reorder plus a regression test, or for both, whichever you prefer.

## Environment

`semantica` 0.6.5, Python 3.12.10. Nothing here is platform specific.

Kyou12138:
Fixes #1035

## Summary

`Config.get()` silently dropped boolean environment overrides because the type dispatch checked `isinstance(default, int)` before `isinstance(default, bool)`. Since `bool` subclasses `int` in Python, the bool branch was unreachable:

- `CONFLICT_ZZTESTFLAG=true` + `get("zztestflag", False)` returned `False` (env var ignored)
- `CONFLICT_ZZTESTFLAG=1` returned `1` (int) instead of `True`

## Change

Reorder the dispatch so `bool` is checked first (before `int`), and strip whitespace before parsing truthy spellings. Applied to all ten config modules listed in the issue:

- `semantica/conflicts/config.py`
- `semantica/deduplication/config.py`
- `semantica/split/config.py`
- `semantica/embeddings/config.py`
- `semantica/export/config.py`
- `semantica/ingest/config.py`
- `semantica/kg/config.py`
- `semantica/parse/config.py`
- `semantica/ontology/config.py`
- `semantica/normalize/config.py`

## Tests

Added `tests/conflicts/test_config_bool_env.py` covering:
- truthy/falsy spellings (true/1/yes/on, false/0/no/off, case variants, whitespace)
- int/float/str parsing still works
- the three modules that reach the dispatch path (conflicts, deduplication, split)
- mapped boolean env vars in all ten modules parse as real bools

All tests pass: 12 new + 6 existing conflicts tests + 131 related module tests.

## Notes

Per the issue, the separate question of whether conflicts/deduplication/split should also gain the generic `PREFIX_*` scanner (so both env paths agree) is left as a design decision and not included here.


Sameer6305:
Thanks for picking up #1035. Could you please address the Qodo findings on the PR?

Also, please comment on the issue to confirm you’re taking it before opening a PR. This helps us track ownership and avoid duplicate work.

Once the fixes are pushed, I’ll take another look.

Kyou12138:
Thanks @Sameer6305! I've addressed the Qodo findings:

1. **Mapped bools ignore whitespace**: _load_env_vars() (mapped booleans) and the generic PREFIX_* scanner now also use alue.strip().lower(), matching the get() dispatch path. Added regression tests for mapped and scanned booleans with surrounding whitespace.

2. Also commented on the issue to confirm ownership as requested.

All 21 conflicts tests pass locally.

ZohaibHassan16 (changes requested):
Solid PR.  But this PR also includes the worker import cleanup from #1061 (`semantica/worker.py`). That is unrelated to this issue and should stay in its own PR.

ZohaibHassan16 (approved):
The boolean env parsing fix and test coverage look good.

Approved.

PR #1039, opened 2026-08-16
perf(semantic_extract): reuse cached spaCy model in NERExtractor

### LINKED ISSUE #997: perf(semantic_extract): cache spaCy model loading in NERExtractor
## Problem

`NERExtractor` can load the spaCy model directly with `spacy.load()` during initialization instead of using the model-loading cache already available in `semantica.semantic_extract.methods`.

This means repeated construction of `NERExtractor` instances can repeatedly load the same spaCy model from disk, adding significant startup/processing overhead.

## Current behavior

The `NERExtractor` initialization path uses a direct spaCy load when the `ml` extraction method is selected:

```python
spacy.load(self.model_name)
````

This bypasses the cached model-loading path introduced/used in `semantic_extract.methods`.

As a result, callers that repeatedly create `NERExtractor` instances can repeatedly incur the cost of loading the same model.

## Why this matters

spaCy model loading is substantially more expensive than processing an already-loaded document. Repeated model initialization can therefore:

* increase request latency;
* increase CPU and disk I/O;
* increase memory churn;
* become particularly expensive for batch processing;
* unnecessarily duplicate model instances in long-running applications.

This is especially relevant because `NERExtractor` is a higher-level extraction API and can be used independently of the Explorer endpoint.

## Suggested fix

Reuse the existing centralized spaCy model-loading/cache mechanism from `semantica.semantic_extract.methods` instead of calling `spacy.load()` directly.

The fix should preserve the existing model selection and error/fallback behavior.

Please also add a regression test that verifies repeated `NERExtractor` usage does not repeatedly call `spacy.load()` for the same model.

## Scope

This is intentionally separate from PR #886. PR #886 fixes the specific uncached extraction functions associated with #889, while this issue covers the remaining `NERExtractor` loading path.

Kyou12138:
Fixes #997

## Summary

`NERExtractor.__init__` loaded the spaCy model directly with `spacy.load()` when the `ml` method was selected, bypassing the process-level model cache (`load_spacy_model()`) already used by the extraction functions in `methods.py`. Repeated construction of `NERExtractor` instances therefore reloaded the same model from disk every time, adding significant startup/processing overhead in batch and long-running scenarios.

## Change

Route the init path through `methods.load_spacy_model()` (delayed import, consistent with the other extractors' import style), so the model is loaded once per process and shared between extraction functions and `NERExtractor`.

`OSError` semantics are preserved: a missing model still triggers the existing soft-fallback warning path, and other initialization failures still disable the ML runtime for that instance.

## Tests

Added `tests/semantic_extract/test_ner_extractor_cache.py`:
- model loaded once across multiple `NERExtractor` instances
- distinct models cached separately
- `NERExtractor` reuses a model already loaded by `methods`
- missing model still falls back (no regression)

All 4 new tests pass, plus the existing 18 extractor/cache tests.

### DIFF
--- a/semantica/semantic_extract/ner_extractor.py
+++ b/semantica/semantic_extract/ner_extractor.py
@@ -139,12 +139,16 @@ class NERExtractor:
         if not self.progress_tracker.enabled:
             self.progress_tracker.enabled = True
 
-        # Initialize spaCy model if ML method is used
+        # Initialize spaCy model if ML method is used. Route through the
+        # process-level model cache in methods.load_spacy_model() so repeated
+        # NERExtractor construction does not reload the same model from disk.
         self.nlp = None
         self._ml_runtime_usable = True
         if "ml" in self.method and SPACY_AVAILABLE:
             try:
-                self.nlp = spacy.load(self.model_name)
+                from .methods import load_spacy_model
+
+                self.nlp = load_spacy_model(self.model_name)
             except OSError:
                 self.logger.warning(
                     f"spaCy model {self.model_name} not found. ML method will fallback."
--- a/tests/semantic_extract/test_ner_extractor_cache.py
+++ b/tests/semantic_extract/test_ner_extractor_cache.py
@@ -1,0 +1,94 @@
+'''Regression tests for issue #997.
+
+NERExtractor.__init__ must reuse the process-level spaCy model cache
+(methods.load_spacy_model) instead of calling spacy.load() directly, so
+constructing many extractors does not reload the same model from disk.
+'''
+
+from types import SimpleNamespace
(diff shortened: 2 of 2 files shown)


CountClaw:
Closing this PR: the same fix (routing NERExtractor's spaCy load through methods.load_spacy_model()) landed in #1042 and is already merged into main. The NERExtractor init path now reuses the process-level model cache, so this change is redundant. Thanks!

Kyou12138:
Closing this PR: the same fix (routing NERExtractor's spaCy load through methods.load_spacy_model()) landed in #1042 and is already merged into main. The NERExtractor init path now reuses the process-level model cache, so this change is redundant. Thanks!

PR #1042, opened 2026-08-17
perf(split): avoid repeated spaCy model loading in split/chunking paths

### LINKED ISSUE #998: perf(split): avoid repeated spaCy model loading in split/chunking paths
## Problem

There are additional direct `spacy.load()` call sites in the split/chunking subsystem that bypass the existing spaCy model cache.

The affected code includes:

- `semantica/split/methods.py`
- `semantica/split/semantic_chunker.py`

These paths can independently load spaCy models instead of reusing an already-loaded model.

## Why this matters

spaCy model initialization is significantly more expensive than running inference against an already-loaded model.

When these split/chunking paths are invoked repeatedly, direct model loading can result in:

- unnecessary model initialization;
- increased latency;
- additional CPU and disk I/O;
- increased memory usage when multiple model instances exist;
- avoidable overhead in batch/document-processing workloads.

This also means the repository has multiple independent mechanisms for loading the same spaCy models, making model lifecycle and caching behavior harder to reason about.

## Suggested fix

Audit the spaCy loading paths in:

```text
semantica/split/methods.py
semantica/split/semantic_chunker.py
````

and determine whether they can safely reuse the existing centralized spaCy model-loading/cache mechanism.

The implementation should preserve any pipeline-specific configuration required by these components. In particular, do not blindly share a cached model if the caller requires different disabled pipeline components.

Add regression tests demonstrating that repeated operations reuse the appropriate model instance/configuration rather than repeatedly calling `spacy.load()`.

## Scope

This is a separate performance issue from PR #886/#889. It was discovered while reviewing the spaCy model-loading architecture, but these split/chunking call sites are outside the scope of that PR and should be handled independently.

Accute9:
## Description

Both `semantica/split/methods.py` and `semantica/split/semantic_chunker.py` independently loaded spaCy models instead of reusing an already-loaded model from the process-level model cache. This fix routes both files to use the model cache, with regression tests to verify that the cached model was reused as opposed to repeatedly calling `spacy.load()`.

## Type of Change

- [ ] Documentation update
- [X] Performance improvement
- [ ] Code refactoring

Closes #998 

## Changes Made

- Changed both `semantica/split/methods.py` and `semantica/split/semantic_chunker.py` to run `load_spacy_model()` getting the cached model instead of each calling `spacy.load()`
- OSError handling is still preserved (i.e missing model triggering warning)

## Testing

Added `tests/split/test_spacy_model_cache.py`

Tests:
- Splitter reusing cached model
- Semantic chunker reusing cached model
- Both using the cached model across instances, not reloading the model between them
- Distinct model names loading separately instead of using the cached model
- Missing the model name falls back instead of poisoning the cache

Passes all tests, in addition to 4 tests from `tests/split/test_splitter.py` and 38 tests from `tests/split/test_chunkers.py`.

- [X] Tested locally
- [X] Added tests for new functionality

## Documentation

- [X] No documentation changes needed

## Breaking Changes

**Breaking Changes**: No

## Checklist

- [X] My code follows the project's style guidelines
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings

## Additional Notes

Currently, `load_spacy_model()` does not accept `disable=[..]`. Thus, one of the added tests `test_no_disable_kwarg_requested` passes for now, but will fail if and when `load_spacy_model()` is changed, as the cache will return a model with the incorrect config. Did not want to speculatively build the disable-aware caching yet, as I wasn't sure if it was in the scope of this issue, but worth mentioning for later.


Sameer6305 (dismissed):
Thanks for the solid work on this PR, @Accute9 . The core approach was well thought out, especially routing the split/chunking paths through the existing centralized spaCy cache and adding regression coverage for model reuse, cache sharing, distinct models, and failed loads.

During my review, I found that `NERExtractor.__init__()` still had a direct `spacy.load()` path with the same underlying cache-bypass issue. I also found a few test patch targets that needed updating after the cache integration, and strengthened the cache-configuration assertion. I addressed these in **`c7415f2e` (`fix: complete spaCy model cache integration`)**, along with additional NER cache regression coverage.

I also verified that the remaining production `spacy.load()` calls are intentional: the centralized cache implementation itself and the separate `get_nlp_model()` path, which requires a different pipeline configuration.

The targeted tests pass, and the remaining 3 failures in the broader NER configuration suite were confirmed to be pre-existing on `main`.

I also found a few **unrelated issues** during the broader review. I’ll raise those separately soon rather than mixing them into this PR. If you’re interested, feel free to take those up once the issues are opened.

From my side, this PR is **approved and ready for final review and merge**.

@KaifAhmad1 ready for your final review and merge.

KaifAhmad1 (approved):
Final review, ready to merge.

**What this PR does:** routes `split/methods.py`'s `split_by_sentences()` and `split/semantic_chunker.py`'s `SemanticChunker.__init__` through the existing process-level spaCy cache (`load_spacy_model()` in `semantic_extract/methods.py`) instead of each calling `spacy.load()` independently, closing #998.

**Review history on this PR:**
- Sameer (@Sameer6305) caught that `NERExtractor.__init__()` had the same cache-bypass bug outside the two files named in #998, fixed it in the same PR, and added NER cache regression coverage.
- I did an independent pass and found one gap Sameer's fix didn't carry over: `SemanticChunker.__init__` only caught `OSError` around `load_spacy_model()`, while the sibling `NERExtractor` fix added a broader `except Exception` for a model that's installed but broken at runtime (e.g. incompatible config) — without it, `SemanticChunker()` would crash instead of degrading to fallback chunking like every other path in this PR. Fixed in a8194df, with a regression test mirroring the existing NERExtractor case for the same scenario.
- Added a CHANGELOG entry for #998/#1042.
- Resolved 3 outstanding review-thread comments (unused `spacy` binding, overly-strict test assertion) — confirmed both were already addressed in later commits (`0f252ab`/`c7415f2`), just left unresolved on GitHub.

**Verification:**
- `pytest tests/split/test_spacy_model_cache.py tests/split/test_splitter.py tests/split/test_chunkers.py`: all passing (12/12 in the cache test file, including the new broken-runtime test).
- Confirmed the 3 failures in `tests/test_ner_configurations.py` are pre-existing on `main`, unrelated to this PR.
- All CI checks green on the latest commit (build, security-scan, MSDO, Analyze Python, GitGuardian, etc.)

Nice work all around — approving.

PR #1056, opened 2026-08-17
test(export): guard Parquet tests on pyarrow itself, not the exporter import

### LINKED ISSUE #1054: test(export): 6 Parquet tests fail instead of skipping — the pyarrow guard checks the exporter import, not the dependency
## Summary

All 6 tests in `TestParquetExportRealData` (`tests/test_030_context_graph_realworld_extended.py`) fail on a clean `pip install -e ".[dev]"` install. They are *intended* to skip when `pyarrow` is missing — the guard is present and the docstring says so — but the guard checks the wrong thing, so it never fires.

`pyarrow` ships only in the `ingest-parquet` / `ingest-arrow` / `db-arrow` extras, not in `dev`, so any contributor following CONTRIBUTING hits this.

## The guard doesn't work

Each test opens with:

```python
try:
    from semantica.export import ParquetExporter
except ImportError:
    pytest.skip("pyarrow not installed")
```

`ParquetExporter` is a plain Python class, so that import **always succeeds** regardless of `pyarrow`. The dependency is checked lazily inside the export call instead — `semantica/export/parquet_exporter.py:142-146`:

```python
if not PARQUET_AVAILABLE:
    raise ImportError(
        "pyarrow is not installed. Please install it with: "
        "pip install pyarrow"
    )
```

So `pytest.skip()` is unreachable and the `ImportError` escapes from `exporter.export_entities(...)`, failing the test.

The intent is unambiguous — the class docstring at `tests/test_030_context_graph_realworld_extended.py:981` states:

> Requires: pyarrow (optional dep — tests skip if not installed).

## Affected tests

`tests/test_030_context_graph_realworld_extended.py::TestParquetExportRealData`

- `test_parquet_export_entities_to_file`
- `test_parquet_export_relationships_to_file`
- `test_parquet_export_knowledge_graph`
- `test_parquet_export_snappy_compression`
- `test_parquet_export_none_compression`
- `test_parquet_convenience_function`

## Reproduction

```bash
pip install -e ".[dev]"          # note: no pyarrow
pytest tests/test_030_context_graph_realworld_extended.py -q
```

```
ImportError: pyarrow is not installed. Please install it with: pip install pyarrow
semantica\export\parquet_exporter.py:143: ImportError
6 failed, 99 passed
```

## Expected

The 6 tests skip cleanly when `pyarrow` is absent, and run when it is present.

## Suggested fix

Guard on the dependency rather than on the exporter import, using the idiom already established elsewhere in this repo:

- `tests/ingest/test_parquet_ingestor.py:5` — `pa = pytest.importorskip("pyarrow")`
- `tests/ingest/test_arrow_ingestor.py:7` — `pa = pytest.importorskip("pyarrow")`

A single class-level guard replaces all six per-test `try`/`except` blocks.

I'd like to take this on if the approach looks right — I have it implemented and verified in both states (skipping without `pyarrow`, passing with it).

dex0shubham:
Closes #1054


ZohaibHassan16 (commented):
One thing before merging: I think Qodo's comment about `find_spec` is correct.

The comment here says this follows `test_parquet_ingestor.py` / `test_arrow_ingestor.py`, but I checked those files and they use `pytest.importorskip("pyarrow")`. That actually imports pyarrow, while `importlib.util.find_spec("pyarrow")` only checks if the module can be found.

So if pyarrow is installed but broken for some reason, like a missing compiled dependency or ABI issue, `find_spec` could still pass and the tests would fail later instead of skipping.

I think the class-level `pytestmark` approach is still the right one though. This file has a lot of unrelated test classes, so using `importorskip` for the whole module would skip tests that don't need pyarrow.

Could we just make the availability check do a real import instead?

```python
def _pyarrow_available() -> bool:
    try:
        import pyarrow  # noqa: F401
        return True
    except ImportError:
        return False


pytestmark = pytest.mark.skipif(
    not _pyarrow_available(),
    reason="pyarrow not installed",
)
```

That keeps the current test scoping, but also matches what the referenced tests are actually doing.

Once that's changed, looks good to me.

dex0shubham:
You're right on both counts @ZohaibHassan16 — `find_spec` only proves the module is discoverable, and my comment claiming it followed `test_parquet_ingestor.py` / `test_arrow_ingestor.py` was inaccurate, since those use `importorskip` and genuinely import. Sorry for the misleading comment.

Both are already fixed — I pushed `d0d38f0..` → the second commit (`test(export): guard on PARQUET_AVAILABLE so the skip matches the runtime check`) shortly before your review landed, so I think you were looking at the first commit. Current head has no `find_spec` and no reference to those two files.

I went with the exporter's own flag rather than a local helper:

```python
from semantica.export.parquet_exporter import PARQUET_AVAILABLE

pytestmark = pytest.mark.skipif(
    not PARQUET_AVAILABLE,
    reason="pyarrow not installed",
)
```

Two reasons over a local `_pyarrow_available()`:

1. It's the exact flag `ParquetExporter` gates on (`parquet_exporter.py:35-43`), so the skip condition can't drift from the runtime check — that drift is what caused this bug in the first place.
2. The exporter imports both `pyarrow` **and** `pyarrow.parquet`. A helper that imports only `pyarrow` would still miss a working `pyarrow` with a broken `pyarrow.parquet`.

Exception semantics are identical to your version — that `try` catches `ImportError`, which is also what a broken native install raises (`ImportError: DLL load failed while importing lib`).

Your point about scoping is exactly why this stays a class-level `pytestmark` rather than a module-level `importorskip` — the file has ~98 other tests that don't need pyarrow.

Verified in both states:
- pyarrow installed → 105 passed (all 7 Parquet tests run)
- pyarrow uninstalled → 98 passed, 7 skipped

That said, if you'd rather keep the check local and avoid the test importing a module-level flag from `semantica.export`, your `_pyarrow_available()` version is a fine call and I'll switch — just say the word.

ZohaibHassan16 (approved):
Approving.

Ran the tests, without `pyarrow` installed I got `98 passed, 7 skipped`, so that matches. I also checked that importing `PARQUET_AVAILABLE` at module level does not stop the other 98 tests in the file from being collected, so the class-level skip is working as expected.

Also, importing `PARQUET_AVAILABLE` directly from `parquet_exporter` is better than the helper I suggested before. I checked `parquet_exporter.py` again and it imports both `pyarrow` and `pyarrow.parquet` in the same `try` block before setting the flag.

My helper only checked `pyarrow`, so it could miss the case where `pyarrow` imports but `pyarrow.parquet` does not.

Using the same flag as the actual runtime code also means the test skip condition should stay in sync with the real check.

LGTM.

PR #1057, opened 2026-08-17
docs: fix broken star history chart in README

OctoBored:
The star history chart currently embedded in the README is broken and no longer renders, because it relies on the GitHub stargazer API which is now restricted. This PR switches the chart to a working alternative that uses a different data source requiring no API token, so the chart displays correctly again.


ZohaibHassan16 (commented):
One small thing before merging: the new markup uses raw `&` inside `href`, `srcset` and `src`, for example `...type=date&legend=top-left`, as caught by Qodo.

The README already uses `&amp;` in attribute URLs elsewhere, like the Trendshift badge near the top. Could you update these to match? It renders fine either way, but keeping it consistent with the existing markup would be better and it's a pretty small change.

Other than that, looks good to me. Thanks for finding a working alternative.

KaifAhmad1:
@OctoBored following up on @ZohaibHassan16's comment above — the `&` → `&amp;` fix in the four attribute values (href/srcset/src) is still the only thing standing between this and merge. Also flagged by Qodo, and it'd match the convention already used elsewhere in the README (e.g. the Trendshift badge). Once that's pushed and the branch is updated with main, this is good to go — thanks for finding a working replacement for the broken chart.

KaifAhmad1 (approved):
Pushed the outstanding fix (raw \`&\` → \`&amp;\` in the href/srcset/src attribute values) so this matches the README's existing convention (e.g. the Trendshift badge) — resolves the point Zohaib and Qodo raised. Verified the new \`star-history.dera.page\` domain: it's a legitimate, actively-maintained open-source fork (Mubelotix/SimRepo) built specifically as a token-free replacement after GitHub restricted the stargazer API the old chart relied on. Thanks for tracking down a working alternative, @OctoBored — LGTM now.

PR #1061, opened 2026-08-17
fix(worker): remove unused sys import and organize imports

aoright:
### Summary
- Remove unused `sys` import from `semantica/worker.py`.
- Organize and sort import declarations according to PEP 8.


ZohaibHassan16 (approved):
LGTM

PR #1071, opened 2026-08-17
test(ingest): track relationship provenance via ProvenanceManager

### LINKED ISSUE #1055: test(ingest): test_notebook_06 fails unconditionally — patches ProvenanceTracker.track_relationship, which was never implemented
## Summary

`tests/ingest/test_notebook_06.py::TestNotebook06MultiSourceIntegration::test_multi_source_integration_flow` fails unconditionally, because it patches a method that has never existed on `kg.ProvenanceTracker`.

This looks like a straggler from a cleanup the project has already done elsewhere.

## The failure

`tests/ingest/test_notebook_06.py:82`:

```python
with patch.object(provenance_tracker, 'track_relationship'):
    for rel in relationships:
        provenance_tracker.track_relationship(...)
```

```
AttributeError: <semantica.kg.provenance_tracker.ProvenanceTracker object at 0x...>
does not have the attribute 'track_relationship'
```

`patch.object` requires the attribute to exist, so this raises before the test body does anything.

`semantica/kg/provenance_tracker.py` defines `track_entity`, `get_all_sources`, `clear`, `query_recorded_between`, `revision_history`, and `export_audit_log` — no `track_relationship`.

## This was already known and cleaned up elsewhere

`CHANGELOG.md:563` records it explicitly:

> `kg.ProvenanceTracker` was a standalone in-memory implementation that never delegated to the unified `ProvenanceManager` backend; its own test suite asserted the existence of `get_lineage`, `track_relationship`, `track_entities_batch`, `get_provenance`, and `_use_unified`, none of which were ever implemented

And `tests/provenance/test_backward_compat.py:39-40` carries the matching note:

> NOTE: `test_track_relationship_unchanged` removed; it only exercised `tracker.track_relationship()`, which was never implemented on ...

So the sibling tests were removed at that time; this one was missed.

## Reproduction

```bash
pytest tests/ingest/test_notebook_06.py -q
```

Fails on current `main` (`4d379200`).

## Note on the current block

As written the block patches `track_relationship` and then immediately calls the patched mock, asserting nothing — so it provides no coverage even in principle.

## Two possible fixes

1. **Remove the block**, matching the precedent set in `tests/provenance/test_backward_compat.py`. Smallest diff, but drops the relationship step from this notebook-flow simulation.
2. **Use `ProvenanceManager` for the relationship step.** `ProvenanceManager.track_relationship` genuinely exists (`semantica/provenance/manager.py:417`), and `ProvenanceTracker` already emits a `DeprecationWarning` pointing there:

   > ProvenanceTracker is deprecated and will be removed in a future major version. Use `semantica.provenance.ProvenanceManager` instead.

   Slightly larger, but keeps the step meaningful and moves the test toward the API the project is migrating to.

I lean towards (2) and have it implemented, but happy to switch to (1) if you'd rather keep this strictly to deleting dead assertions. Let me know which you prefer and I'll open the PR.

dex0shubham:
Closes #1055


ZohaibHassan16 (approved):
Approved.

Ran the fixed test, passes.

LGTM.

PR #1077, opened 2026-08-18
fix(reasoning): implement RETE alpha/beta matching with Token model (#300)

### LINKED ISSUE #300: Implement RETE engine pattern matching (AlphaNode._matches, BetaNode._can_join)
## Summary

The RETE algorithm engine (`semantica/reasoning/rete_engine.py`) has correct network structure (alpha nodes, beta nodes, terminal nodes) but the two core matching methods are placeholder stubs that always return `True`.

## Primary Repo

`https://git-ext.saabusa.com/Saab-Inc/SKAPA-fusion-kg.git`

## Problem

```python
# rete_engine.py:79-82
def _matches(self, fact: Fact) -> bool:
    return True  # <- Always matches everything

# rete_engine.py:101-104
def _can_join(self, left_fact: Fact, right_fact: Fact) -> bool:
    return True  # <- Always joins everything
```

This means RETE fires ALL rules against ALL facts with no discrimination.

## Current State

- `reasoner.py` forward/backward chaining works correctly with regex-based pattern matching
- RETE is needed for scale optimization only (hundreds of rules, thousands of facts)
- Current CoT scenario scale (~50 entities, 8 rules) runs fine with naive `Reasoner`

## RETE vs Naive Reasoner

| Aspect | Reasoner (naive) | RETE |
|--------|-----------------|------|
| Logic | forward chaining | forward chaining |
| Results | identical | identical |
| Method | checks all rules x facts every cycle | caches partial matches, reprocesses only changes |
| Complexity | O(rules x facts^conditions) per cycle | amortized O(changes) |
| Scale | rules <50, facts <1,000 | rules 1000+, facts 10,000+ |

## Implementation Scope

Estimated ~40 lines of code across 2 methods:

1. **`AlphaNode._matches(fact)`**: Match fact predicate/arguments against the alpha node condition. Use unification logic similar to `Reasoner._match_pattern()`.

2. **`BetaNode._can_join(left, right)`**: Check that variable bindings from left and right facts are consistent (no binding conflicts).

## Files

- `semantica/reasoning/rete_engine.py` (lines 79-82, 101-104)
- Reference: `semantica/reasoning/reasoner.py` `_match_pattern()` (line 331)

## Priority

Low priority. Only needed when rule/fact scale exceeds naive Reasoner capacity.

cxzg007:
## Summary

Implements the two RETE matching methods requested in #300 and, on top of that, fixes a P1 chained-join defect surfaced while wiring them up.

Closes #300

## What changed

- **`unify_condition()`** (new module-level helper): regex-based unification reusing `Reasoner._match_pattern()`'s approach — `?var` becomes a named capture group, a variable repeated within one condition becomes a backreference, and it returns the bindings dict or `None`.
- **`AlphaNode._matches()`**: no longer a `return True` stub; unifies each fact against the node's condition and emits a single-fact **token**.
- **Token model**: propagation now carries a `Token(facts, bindings)` instead of bare facts. `BetaNode.join()` merges a left token with a right token, concatenating their facts in condition order and returning the merged token only when shared variables agree (conflict → `None`).
- **P1 fix (chained joins)**: rules with three or more conditions previously lost bindings and accumulated wrong facts at the third join, and a conflicting third condition could spuriously fire. Beta nodes now keep both `left_tokens` and `right_tokens` memories and join each new token against every token on the opposite side, so deep chains stay binding-consistent and third-level conflicts are suppressed.
- **Topology fix**: newly created beta nodes were never appended to their inputs' `children`, blocking propagation — now wired correctly.

## Behavior

Results are identical to the naive `Reasoner` (as #300 requires). Verified by an explicit parity test against `Reasoner._match_rule()`.

## Tests

`tests/reasoning/test_rete_engine.py`:
- `unify_condition` unit cases (single/multi variable, literal args, predicate mismatch, repeated-variable equality)
- alpha match/reject, beta consistent-join vs conflict-reject, single/multi-condition end-to-end
- `TestThreeConditionChain` (7 cases): valid three-condition match, third-level conflict suppression, insertion-order independence, complete in-order `Match.facts`, multiple left tokens joining one right fact, parity against `Reasoner._match_rule()`, and `reset()` clearing all token memory

`tests/reasoning/` — 64 passed, no regressions. `flake8 --max-line-length=88` clean.


KaifAhmad1 (approved):
## Final review

**The core fix is correct.** `AlphaNode._matches`/`BetaNode._can_join` go from `return True` stubs to real regex-based unification (`unify_condition`, mirroring `Reasoner._match_pattern`), and propagation is reworked around a `Token(facts, bindings)` model. I traced the join algorithm by hand: each beta node keeps `left_tokens`/`right_tokens` memories and joins every new token against the opposite side's full history — the standard incremental-RETE technique, which correctly handles chained (3+ condition) joins and is insertion-order independent. 68 tests cover the real edge cases (conflict suppression at the third join, permuted fact arrival, multiple left tokens joining one right fact, `reset()` semantics), plus an explicit parity test against `Reasoner._match_rule()`. The precompiled-regex perf fix and the `re.error`/unexpected-exception logging are both sound.

**Merge conflict — resolved and reverified.** #1096 merged after this branch's last rebase and added a rule-actions/provenance layer (`bind_reasoner()`, `execute_matches()` firing `Rule.actions`/legacy `handler`, `_executed_activations` dedup, `reset_action_history()`) on top of the *same* still-broken stubs, via an interim `_bindings_for_rule()` regex re-extraction. That's a structural conflict, not a cosmetic one. I pushed a reconciliation merge commit that keeps this PR's `Token`-based matching and layers #1096's action/provenance plumbing on top of it, sourcing `Match.bindings` straight from `Token.bindings` (making `_bindings_for_rule()`/`_extract_bindings()` redundant, so they're removed). Also fixed a casualty test broken by `Reasoner._match_rule()`'s return shape changing upstream (2-tuple → 3-tuple), and dropped an unrelated `.gitignore` byte-diff.

**Verified on the merged tree:** full `tests/reasoning/` suite — 106 tests, covering both this PR's RETE tests and #1096's rule-action tests — passes; `flake8 --max-line-length=88` clean. PR now shows `MERGEABLE`.

LGTM — approving.

PR #1079, opened 2026-08-18
fix(security): address privately disclosed zip-slip, SQLi, SSRF, XSS, and SPARQLi findings

KaifAhmad1:
## Summary

Fixes a set of runtime trust-boundary issues from a private security disclosure (local checkout @ `7c3372c0`), plus several lower-severity hardening items found during the same review, and two follow-up fixes found in a second review pass over this diff.

### Disclosed findings

- **Tarball restore path traversal (zip-slip)** — `backup restore`'s `tar.extract()` had no containment check. Added path-containment validation for every member, lexical + resolved symlink/hardlink target validation, and `filter="data"` on Python ≥3.12.
- **SQL injection in the DB exporter (latent)** — `DataExporter.export_table_data` was missing the `text` import (silent `NameError`), and `where`/`order_by`/`table_name`/`schema` were interpolated raw. Fixed the import and added identifier validation (allowlist) plus fragment validation (blocklist for statement separators, comments, UNION, DDL/DML keywords, time-based blind-injection primitives, schema-enumeration terms).
- **DNS-rebinding TOCTOU in the shared SSRF guard** — `request_with_ssrf_guard` validated a hostname, then let the HTTP client re-resolve it independently at connect time. Ported the IP-pinning pattern already used in `explorer/routes/ontology.py` into the shared ingest guard, so the connection is pinned to the exact IPs that were validated.
- **Stored XSS in HTML report generation** — report title/summary/metrics were interpolated into HTML with no escaping. Added `html.escape()` on every interpolated value.
- **SPARQL injection in AnzoStore's object position** — subject/predicate were validated via `validate_uri`, but the object position wasn't. Now validated the same way, matching the other triplet-store backends.

### Additional hardening from the same review

- Added the CGNAT range (`100.64.0.0/10`) to the SSRF blocklist.
- Allowlisted Apache AGE's `create_index` `index_type` parameter (was interpolated raw into a `USING` clause).
- Cast Neo4j's `limit` explicitly before interpolation into `LIMIT {limit}`, raising a clear `ValidationError` for invalid input instead of falling through to a generic error.
- Guarded the `ffprobe` filename argument against leading-dash option injection.
- MCP server no longer echoes raw exception text to JSON-RPC clients; logs full details server-side and returns a generic message plus the exception class name.

### Follow-up fixes from a second review pass

- Fixed a connection-pool leak introduced by the SSRF IP-pinning change: `Session.mount()` silently drops the adapter it replaces without closing it, so a multi-hop redirect chain on a reused session was leaking one pooled connection per hop.
- Added a per-session lock so two guarded requests sharing the same caller-supplied `Session` from different threads now serialize instead of racing on adapter/Host-header mutation.
- Added lexical symlink/hardlink target validation (reject absolute or `..`-traversing link targets) in the tar restore path, as defense-in-depth alongside the existing resolved-path check.

### Known limitation (documented, not silently claimed fixed)

The SQL fragment validation for `where`/`order_by` is a blocklist, not a grammar — it closes the disclosed UNION-exfiltration PoC and common injection primitives, but a boolean-blind subquery using no blocked keyword could still get through. This is inherent to `where`/`order_by` being a raw-SQL-fragment API by design; the docstrings now say explicitly that these parameters must be treated as trusted/operator input, not exposed to untrusted end users.

## Test plan

- [x] Re-ran each disclosed PoC (or an equivalent adversarial test) against the fix and confirmed it's blocked: zip-slip, SQLi (UNION + identifier breakout), DNS-rebinding TOCTOU, XSS, SPARQL injection.
- [x] Verified the SSRF proxy-bypass and session-concurrency fixes with dedicated tests (env-proxy fail-closed check under both owned and caller-supplied sessions; two-thread concurrent-session test proving serialization with no Host-header cross-contamination).
- [x] Full `tests/ingest/` suite: 266 passed, 2 skipped, 10 failures — identical failure set on this branch and on unmodified `main` (pre-existing, unrelated to this change).
- [x] Full regression sweep across every other touched module (`graph_store`, `export`, `triplet_store`, `parse`, backup/restore): 313 passed, 3 skipped, zero failures.


Sameer6305 (approved):
I completed a full security-focused review of this PR, including the complete diff, affected modules, related/caller code paths, existing security patterns, and the relevant test coverage. I also performed a second-pass validation of the security-sensitive areas rather than relying only on the tests or the PR description.

### Security findings reviewed

- **TAR/zip-slip:** Verified the restore-path containment checks, including lexical and resolved symlink/hardlink validation and the Python 3.12+ `filter="data"` protection. No remaining exploitable traversal path was found.
- **SQL injection:** Verified `DataExporter.export_table_data()` identifier validation and SQL-fragment handling, including identifier breakout, UNION/stacked-query attempts, comments, DDL/DML and time-based primitives. The documented limitation that `where`/`order_by` remain raw SQL fragments and can accept subqueries is real, but these parameters are operator/developer-controlled and are not exposed through HTTP or MCP in the current codebase, so this is not blocking.
- **SSRF/DNS rebinding:** Verified DNS validation, IP pinning, redirect re-validation, multi-IP handling, private/reserved/CGNAT blocking, proxy fail-closed behavior, session mutation/restoration, connection-pool cleanup, and concurrent caller-supplied session handling. The main DNS-rebinding TOCTOU is correctly addressed.
- **Stored XSS:** Audited all HTML interpolation in the report generator. Title, summary, metrics, keys and values are escaped correctly.
- **SPARQL injection:** Verified subject/predicate validation and, importantly, the previously unsafe object path including pre-bracketed URI objects. No remaining injection path was found.
- **Apache AGE:** Verified `index_type` is now strictly allowlisted and checked the other interpolated values.
- **Neo4j:** Verified `limit` is converted to an integer before interpolation and that other dynamic identifiers are sanitized/parameterized.
- **ffprobe:** Verified leading-dash option injection is prevented and the subprocess uses argument-list execution rather than a shell command.
- **MCP errors:** Verified raw exception messages are no longer returned to JSON-RPC clients while full details remain available server-side through logging.

### Additional findings

A few non-blocking hardening/test gaps were identified:

- `RobotsChecker` still uses `RobotFileParser.read()` outside the shared SSRF guard, leaving a narrow DNS-rebinding window for the robots.txt request.
- `trust_env` is not disabled on the first SSRF hop, although the adapter-level proxy check currently fails closed.
- TAR special-file handling could be made more explicit on Python versions before 3.12.
- Some adversarial security tests could be added for TAR traversal, XSS, pre-bracketed SPARQL objects, ffprobe option injection and MCP exception disclosure.
- The PR's claim about rerunning the disclosed PoCs is not fully substantiated by the repository's `poc_runner.py`; that runner covers different PoCs. The fixes themselves were independently verified during this review.

I consider these **follow-up hardening/test improvements rather than blockers**. I did not find a realistic exploitable path within the current Semantica trust boundaries that would require changes to this PR.

The five primary disclosed issues — **zip-slip, SQLi, SSRF DNS-rebinding, stored XSS, and SPARQL injection — are fixed correctly**, and the additional hardening changes were also verified.

@KaifAhmad1  **Approving from my side**

PR #1081, opened 2026-08-18
feat(context): add to_kg_dict() adapter for canonical KG shape

### LINKED ISSUE #1080: [FEATURE][DX] Add ContextGraph.to_kg_dict() to emit the canonical {entities, relationships} shape
### Problem

`ContextGraph.to_dict()` returns an internal shape — `{"nodes": [...], "edges": [...]}`
with `source` / `target` keys — but downstream consumers like `RDFExporter` and
`TemporalGraphQuery.query_time_range()` expect the canonical knowledge-graph shape:
`{"entities": [...], "relationships": [...]}` with `source_id` / `target_id`.

Today every user has to hand-map field names between the two APIs, e.g.:

```python
graph_dict = graph.to_dict()
kg = {
    "entities": [{"id": n["id"], "type": n["type"], "text": n["content"]}
                 for n in graph_dict["nodes"]],
    "relationships": [{"source_id": e["source"], "target_id": e["target"], "type": e["type"]}
                      for e in graph_dict["edges"]],
}
RDFExporter().export(kg, "out.ttl")

cxzg007:
Closes #1080

### What

Adds `ContextGraph.to_kg_dict(entities_only=False)`, the official adapter that converts the graph's internal `{"nodes", "edges"}` / `source`-`target` representation into the canonical `{"entities", "relationships"}` / `source_id`-`target_id` shape consumed by `RDFExporter` and `TemporalGraphQuery.query_time_range()`.

### Why

Users currently hand-map field names between `to_dict()` and the shape that downstream consumers expect. This is boilerplate and error-prone. `to_kg_dict()` removes that friction.

### Details

- Maps `nodes -> entities` (`content -> text`, keeps `type`, copies `properties` / `metadata`) and `edges -> relationships` (`source -> source_id`, `target -> target_id`, keeps `type` / `weight`, carries `id` / `familyId`)
- Preserves `valid_from` / `valid_until` on both entities and relationships when set
- `entities_only=True`: exports only `node_type == "entity"` nodes and drops relationships whose endpoints are not in the exported set (avoids dangling refs)
- Thread-safe (`with self._lock`); returned dicts are shallow copies isolated from internal state
- Returns `statistics` with `entity_count` / `relationship_count`

### Changes

- `semantica/context/context_graph.py`: new `to_kg_dict()` method (+82 lines, single hunk)
- `README.md`: two examples now use `graph.to_kg_dict()` instead of manual mapping
- `tests/context/test_to_kg_dict.py`: 7 new unit tests

### Testing

`pytest tests/context/test_to_kg_dict.py` -> 7 passed. Covered: basic shape, source_id/target_id mapping, temporal passthrough, statistics counts, entities_only node filtering, entities_only dangling-relationship drop, and returned-dict isolation.


ZohaibHassan16 (approved):
Approved.

Null properties: created a node with `properties=None` and `metadata=None` and ran `to_kg_dict()`. No crash, both come back as `{}`.

Id mismatch: created two nodes with int ids and an edge between them. Ran `to_kg_dict(entities_only=True)` and the edge is still there, with the ids converted to strings.

Endpoint compat: checked `GraphValidator` and `TemporalGraphQuery`. Both work with `source_id` / `target_id` now.

I did notice duplicate entities while testing the int id case. Looks like an existing `ContextGraph` issue though. `_add_internal_edge` creates a placeholder node using the string id even if the same id already exists as an int. Doesn't seem related to this PR.

New tests: 10 passed.

Good to merge.

PR #1084, opened 2026-08-18
fix(ontology): preserve #-terminated namespaces in SHACLGenerator base_uri (#1082)

### LINKED ISSUE #1082: SHACLGenerator mangles hash-namespace base_uri: '#' becomes '#/' so sh:targetClass never matches
## Summary

`SHACLGenerator.__init__` normalizes `base_uri` with `base_uri.rstrip("/") + "/"`. When the ontology namespace ends with `#` (the most common convention for RDF vocabularies), this produces `...#/`, so every generated URI (`sh:targetClass`, `sh:path`, shape URIs) lands in a different namespace than the instance data. Validation then silently passes because the shapes target nothing.

## Repro (semantica 0.6.5)

```python
from semantica.ontology.engine import OntologyEngine

MFG = "http://example.org/manufacturing#"
ontology = {
    "uri": "http://example.org/manufacturing",
    "name": "MfgMini", "version": "demo",
    "classes": [{"name": "Equipment", "uri": MFG + "Equipment", "label": "Equipment"}],
    "properties": [{"name": "hasSerialNumber", "uri": MFG + "hasSerialNumber",
                    "label": "hasSerialNumber", "type": "data",
                    "domain": "Equipment", "range": "string",
                    "required": True, "cardinality": {"min": 1, "max": 1}}],
}
print(OntologyEngine(base_uri=MFG).to_shacl(ontology, base_uri=MFG))
```

Output contains:

```turtle
@prefix ex: <http://example.org/manufacturing#/> .   #  <-- '#/' mangled namespace
...
    sh:targetClass ex:Equipment ;                    #  = ...#/Equipment, never matches ...#Equipment
```

Any data graph typed with `<http://example.org/manufacturing#Equipment>` validates as conforming even when it violates the shape, because the target class URI differs.

## Expected

`SHACLGenerator.generate()` already has the right guard (`if not base_uri.endswith("/") and not base_uri.endswith("#")`), but `__init__` runs `rstrip("/") + "/"` unconditionally beforehand. Suggest applying the same `#`-aware normalization in `__init__` (and anywhere else base_uri is normalized).

## Workaround

Post-process the generated Turtle: `shacl_str.replace(ns + "/", ns)` for hash namespaces.

Found while building executable corroboration demos for an ontology-engineering teaching repo (validated with pyshacl; after the workaround the auto-derived shapes correctly reject the violating instance).

changshenhan:
Closes #1082

## Root cause

`SHACLGenerator.__init__` normalized `base_uri` with `base_uri.rstrip("/") + "/"`. For a `#`-terminated RDF namespace — the most common convention for vocabularies — this produces `...#/`, so every generated URI (`sh:targetClass`, `sh:path`, shape URIs) lands in a different namespace than the instance data. SHACL validation then silently passes because the shapes target nothing.

`generate()` already had the correct `#`-aware handling (it only appends `/` when the URI ends in neither `/` nor `#`); `__init__` simply wasn't aligned with it.

## Fix

`__init__` now preserves a `base_uri` already ending in `/` or `#`:

```python
self.base_uri = (
    base_uri if base_uri.endswith(("/", "#")) else base_uri + "/"
)
```

`shapes_uri` inherits the fix (`...#` + `shapes` → `...#shapes`, correct RDF).

## Test

New `test_hash_namespace_base_uri_is_not_mangled` in `tests/ontology/test_ontology_advanced.py`:
- asserts `base_uri` stays `...manufacturing#` and `shapes_uri` becomes `...manufacturing#shapes`;
- runs the full generate + turtle path and asserts no `#/` namespace in the output.

The test fails on current `main` (verified against the old normalization) and passes with the fix. Full ontology suite: 76 passed.


ZohaibHassan16 (approved):
Good fix with proper edge cases. Tests passing. LGTM.

PR #1026, opened 2026-08-16
fix(kg): honor exact entity resolution

T1mn:
## Description

`EntityResolver(strategy="exact")` currently uses fuzzy duplicate detection, so similar names can be merged unexpectedly. It also drops non-duplicate entities without an `id` or `entity_id`.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Fixes #1031

## Changes Made

- Apply exact name matching with case and surrounding whitespace normalization.
- Preserve the existing fuzzy resolution path.
- Preserve entities without identifiers and avoid re-inserting merged id-less entities.
- Add regression tests.

## Testing

- [x] Tested locally
- [x] Added tests for new functionality
- [ ] Package builds successfully (`python -m build`); the local environment does not have the `build` module installed.

### Test Commands

```bash
pytest -q tests/kg/test_entity_resolver_exact.py tests/kg/test_core_components.py tests/kg/test_graph_builder_external.py
```

## Documentation

- [x] No documentation changes needed

## Breaking Changes

**Breaking Changes**: No

## Additional Notes

The change is limited to EntityResolver behavior. The full KG suite also depends on optional local packages already declared in the repository dependencies; CI will install the pinned environment.

PR #1086, opened 2026-08-18
fix(kg): preserve entity_id aliases during entity merging

### LINKED ISSUE #1085: [BUG] Preserve entity_id aliases during entity merging
## Bug Description

When duplicate entities use `entity_id` instead of `id`, entity resolution can merge the entities without preserving their identifiers or remapping relationship endpoints.

## Steps to Reproduce

1. Check out commit `a1194a15` from `main`.
2. Run:

```python

builder = GraphBuilder(
    merge_entities=True,
    resolve_conflicts=False,
)

graph = builder.build({
    "entities": [
    ],
    "relationships": [
        {
            "source_id": "alice:2",
            "target_id": "org:1",
            "type": "WORKS_FOR",
        }
    ],
})

print(graph["entities"])
print(graph["relationships"])
```

## Expected Behavior

The merged Alice entity should retain a canonical identifier, and the relationship should point to that identifier.

## Actual Behavior

The merged entity has `id=None`, `merged_from=[]`, and the relationship still points to `alice:2`, which is no longer present as an entity:

```text
entities: ... {'id': None, ..., 'merged_from': []} ...
relationships: ... {'source': 'alice:2', 'target': 'org:1'} ...
```

This leaves a dangling relationship endpoint in the resulting KG.

## Environment

- **OS**: macOS
- **Python Version**: 3.x
- **Semantica Version**: source checkout at `a1194a15`

## Code Example

The minimal reproduction above is self-contained.

## Error Messages / Logs

No exception is raised; the incorrect graph is returned.

## Additional Context

- Related to entity merging and relationship endpoint remapping in #978 and #1026.
- The issue is in the supported `entity_id` input path, so a regression test should cover entities that do not provide an `id` field.

## Possible Solution

Normalize `id` and `entity_id` to one canonical identifier during merge, provenance recording, and relationship endpoint remapping. Add a regression test that asserts the final relationship has no dangling endpoint.

T1mn:
## Description

Preserve `entity_id` aliases when duplicate entities are merged so relationship endpoints are remapped to the surviving entity.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #1085

## Changes Made

- Use one shared `get_entity_id()` helper across entity resolution and merging.
- Treat empty and other falsy identifiers as missing, matching existing KG behavior.
- Preserve source IDs in `merged_from` and merge provenance.
- Add regression coverage for entity ID aliases, falsy IDs, and relationship remapping.

## Testing

- [x] Tested locally

### Test Commands

```bash
pytest -q tests/deduplication tests/kg --ignore=tests/kg/test_node_embeddings.py --ignore=tests/kg/test_enhanced_algorithms_e2e.py --ignore=tests/kg/test_integration_comprehensive.py --ignore=tests/kg/test_provenance_integration.py
```

Result: `520 passed`.

The excluded Node2Vec tests require the existing optional `gensim` dependency. The local package build could not run because the `build` module is not installed.

## Documentation

## Breaking Changes

**Breaking Changes**: No

## Checklist

- [ ] My changes generate no new warnings

## Additional Notes

Focused follow-up to the entity merge and relationship remapping fixes in #978 and #1026.

### DIFF
--- a/semantica/deduplication/entity_merger.py
+++ b/semantica/deduplication/entity_merger.py
@@ -45,16 +45,12 @@ License: MIT
 from dataclasses import dataclass
 from typing import Any, Dict, List, Optional, Union
 
+from ..utils.entity_ids import get_entity_id
 from ..utils.exceptions import ProcessingError, ValidationError
 from ..utils.logging import get_logger
 from ..utils.progress_tracker import get_progress_tracker
 from .duplicate_detector import DuplicateDetector, DuplicateGroup
-from .merge_strategy import (
-    MergeResult,
-    MergeStrategy,
-    MergeStrategyManager,
-    _get_entity_id,
-)
+from .merge_strategy import MergeResult, MergeStrategy, MergeStrategyManager
 
 
 @dataclass
@@ -509,7 +505,7 @@ class EntityMerger:
         # Record source entities
         provenance["merged_from"] = [
             {
-                "id": _get_entity_id(e),
+                "id": get_entity_id(e),
                 "name": self._get_entity_value(e, "name"),
                 "source": self._get_entity_value(e, "metadata", {}).get("source") if hasattr(e, "metadata") or isinstance(e, dict) else None,
             }
--- a/semantica/deduplication/entity_merger.py
+++ b/semantica/deduplication/entity_merger.py
@@ -49,7 +49,12 @@ from ..utils.exceptions import ProcessingError, ValidationError
 from ..utils.logging import get_logger
 from ..utils.progress_tracker import get_progress_tracker
 from .duplicate_detector import DuplicateDetector, DuplicateGroup
-from .merge_strategy import MergeResult, MergeStrategy, MergeStrategyManager
+from .merge_strategy import (
+    MergeResult,
+    MergeStrategy,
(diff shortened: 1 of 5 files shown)


ZohaibHassan16 (approved):
Approved. Confirmed that the issue was: three entities keyed by entity_id, two near-dupe "Alice" names, one relationship pointing at alice:2. On main the merged entity comes back with id=None, merged_from=[] and the relationship's still pointing at alice:2 which doesn't exist anymore. Also noticed it swept "Acme" into the same merge group too, which the original issue didn't mention.

With this PR: merged entity gets id='alice:1', merged_from has both source ids and the relationship's source gets remapped to alice:1 instead of dangling. Acme stays separate like it should.

Checked the falsy-id handling too since that was the other Qodo comment. id=0 with entity_id set falls through to entity_id, id="" same thing and if both are falsy you get None instead of literal 0 masquerading as an id. Matches how the rest of the codebase already treats truthiness for these fields.

Good that this collapsed three separate id-extraction implementations into one helper :that drift between them is what caused the bug in the first place.

Ran the test command from the description, 520 passed.

LGTM.

Issue #318, opened 2026-02-15
[FEATURE] SHACL Shape Generation from OWL


KaifAhmad1:
## Strategic Context

> **"Semantica turns ontologies into executable data contracts."**

Most knowledge graph tools stop at two layers:

| Layer | Coverage Today |
|---|---|
| **Ontology** (structure / schema) | ✅ Semantica already builds this |
| **Graph** (data) | ✅ Semantica already builds this |
| **Constraints** (enforcement) | ❌ This RFC adds it |
| **Semantics** (embeddings) | ✅ Semantica already has this |

Adding the constraints layer completes a **hybrid reasoning system** — symbolic constraints (SHACL) alongside semantic retrieval (embeddings). Very few tools do both. This is a differentiator, not a checkbox.

The full pipeline this enables:

```
Text / Data
    ↓  Semantic Extract
Entities + Relations
    ↓  OntologyEngine.from_data()
OWL Ontology
    ↓  OntologyEngine.to_shacl()          ← this RFC: Phase 1
SHACL Shapes
    ↓  OntologyEngine.validate_graph()     ← this RFC: Phase 2
Validation Report (with plain-English explanations)
    ↓  (future) OntologyEngine.repair_graph()
Self-healed Graph ✅
```

---

## Table of Contents

1. [Summary](#1-summary)
2. [Motivation and Use Cases](#2-motivation-and-use-cases)
3. [Background: OWL → SHACL Mapping](#3-background-owl--shacl-mapping)
4. [Affected Files and Scope](#4-affected-files-and-scope)
5. [Detailed Design](#5-detailed-design)
   - 5.1 [Phase 1 — Generation: `OntologyEngine.to_shacl()`](#51-phase-1--generation-ontologyengineto_shacl)
   - 5.2 [Phase 2 — Runtime: `OntologyEngine.validate_graph()`](#52-phase-2--runtime-ontologyenginevalidate_graph)
   - 5.3 [Validation Result Model](#53-validation-result-model)
   - 5.4 [SHACL Generator: `SHACLGenerator`](#54-shacl-generator-shaclgenerator)
   - 5.5 [Serialization via `rdf_exporter.py`](#55-serialization-via-rdf_exporterpy)
   - 5.6 [Logging and Progress Tracking](#56-logging-and-progress-tracking)
   - 5.7 [Deferred Imports and Circular Import Safety](#57-deferred-imports-and-circular-import-safety)
[...]

PR #1094, opened 2026-08-19
fix(ontology): render real SHACL constraint values in explain_violations

### LINKED ISSUE #1093: [BUG] explain_violations renders hardcoded constraint values instead of real SHACL parameters
## Bug Description

`SHACLValidationReport.explain_violations()` renders **hardcoded / wrong constraint values** in its plain-English explanations:

- `min_count` and `max_count` are hardcoded to `1`, so a `sh:maxCount 3` violation is explained as "At most 1 value ...".
- `datatype` and `class` are filled with the raw violation **message** instead of the real `sh:datatype` / `sh:class` IRI.

The root cause is that `_run_pyshacl()` never reads the real constraint parameters from `sh:sourceShape` when building each `SHACLViolation`, so the formatter has no correct values to render.

## Steps to Reproduce

1. Define a SHACL shape with e.g. `sh:maxCount 3`, `sh:datatype xsd:integer`, `sh:class ex:Person`.
2. Validate data that violates those constraints via the pyshacl path.
3. Call `report.explain_violations()`and inspect each `violation.explanation`.

## Expected Behavior

Explanations reflect the **real** constraint parameters, e.g. "At most 3 values ...", the real datatype IRI, and the real required class IRI.

## Actual Behavior

- MaxCount violations always say "At most 1 value ...".
- Datatype / class explanations contain the violation message text rather than the actual `sh:datatype` / `sh:class`.

## Environment

- **OS**: macOS
- **Python Version**: 3.12
- **Semantica Version**: main (source checkout)

## Code Example

```python
from semantica.ontology.ontology_validator import _run_pyshacl

shacl = """
@prefix sh: <http://www.w3.org/ns/shacl#> .
@prefix ex: <http://example.com/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
ex:PersonShape a sh:NodeShape ;
    sh:targetClass ex:Person ;
    sh:property [ sh:path ex:age ; sh:datatype xsd:integer ; sh:maxCount 2 ] .
"""
data = """
@prefix ex: <http://example.com/> .
ex:john a ex:Person ; ex:age "x" ; ex:age 1 ; ex:age 2 ; ex:age 3 .
"""
report = _run_pyshacl(data, shacl)
report.explain_violations()
for v in report.violations:
    print(v.explanation)  # min/max/datatype/class are wrong
```

## Location

`semantica/ontology/ontology_validator.py` — `_run_pyshacl()` and `SHACLValidationReport.explain_violations()`.

## Possible Solution

Back-reference `sh:sourceShape` in `_run_pyshacl()` to read the real `sh:minCount` / `sh:maxCount` / `sh:datatype` / `sh:class`, store them on `SHACLViolation`, and let `explain_violations()` render those real values (falling back to a placeholder when unknown). Note: `sh:qualifiedMinCount` / `sh:qualifiedMaxCount` would remain unhandled and fall back to the placeholder.

cxzg007:
## Description

`SHACLValidationReport.explain_violations()` rendered hardcoded / wrong constraint values: `min_count`/`max_count` were hardcoded to `1`, and `datatype`/`class` were filled with the raw violation message instead of the real `sh:datatype` / `sh:class` IRI. The root cause is that `_run_pyshacl()` never read the real constraint parameters from `sh:sourceShape` when building each `SHACLViolation`, so the formatter had no correct values to render.

This PR back-references `sh:sourceShape` to populate the real constraint parameters on each violation and renders them in the plain-English explanations.

## Type of Change

## Related Issues

Closes #1093

## Changes Made

- `SHACLViolation`: add `min_count` / `max_count` / `datatype` / `class_` fields and include them in `to_dict()`
- `_run_pyshacl`: back-reference `sh:sourceShape` to extract the real `sh:minCount` / `sh:maxCount` / `sh:datatype` / `sh:class` values
- `explain_violations`: render the real values, falling back to `"?"` / descriptive text when unknown
- Add regression tests covering both the formatting path and the `sh:sourceShape` back-reference (the end-to-end test skips when `pyshacl`/`rdflib` are absent)

Note: `sh:qualifiedMinCount` / `sh:qualifiedMaxCount` are not handled and fall back to the `"?"` placeholder.

## Testing

- [x] Tested locally

### Test Commands

```bash
pytest tests/ontology/test_ontology_advanced.py -q   # 30 passed
```

## Breaking Changes

**Breaking Changes**: No

## Checklist

### DIFF
--- a/semantica/ontology/ontology_validator.py
+++ b/semantica/ontology/ontology_validator.py
@@ -33,6 +33,12 @@ class SHACLViolation:
     value: Optional[str] = None
     shape: Optional[str] = None
     explanation: Optional[str] = None
+    # Real constraint parameters extracted from the source shape (sh:sourceShape),
+    # used to render accurate plain-English explanations.
+    min_count: Optional[int] = None
+    max_count: Optional[int] = None
+    datatype: Optional[str] = None
+    class_: Optional[str] = None
 
     def to_dict(self) -> Dict[str, Any]:
         return {
@@ -44,6 +50,10 @@ class SHACLViolation:
             "value": self.value,
             "shape": self.shape,
             "explanation": self.explanation,
+            "min_count": self.min_count,
+            "max_count": self.max_count,
+            "datatype": self.datatype,
+            "class_": self.class_,
         }
 
 
@@ -118,10 +128,10 @@ class SHACLValidationReport:
                 focus_node=v.focus_node,
                 path=v.result_path or "",
                 value=v.value or "",
-                min_count=1,
-                max_count=1,
-                datatype=v.message or "",
-                class_=v.message or "",
+                min_count=v.min_count if v.min_count is not None else "?",
+                max_count=v.max_count if v.max_count is not None else "?",
+                datatype=v.datatype or "the expected datatype",
+                class_=v.class_ or "the required class",
             )
 
     def to_dict(self) -> Dict[str, Any]:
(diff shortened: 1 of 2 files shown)


KaifAhmad1 (approved):
Verified this fix end-to-end, not just by reading the diff — checked out the branch and ran it against real `pyshacl`/`rdflib`.

**Repro from #1093:** confirmed the bug was real on `main` (hardcoded `min_count=1`/`max_count=1`, message text used as datatype/class) and confirmed this PR fixes it — a `sh:maxCount 2` violation now correctly explains "At most 2 value(s)", datatype/class explanations show the real IRIs.

**Additional checks:**
- Tested a shape with 3 sibling property constraints (maxCount+datatype, minCount, class+minCount) to rule out cross-talk between property shapes sharing one node shape — each violation correctly picked up only its own constraint values via `sh:sourceShape`.
- Confirmed `sh:minCount 0` is handled correctly (`is not None` checks, not truthiness) — avoids a common footgun.
- Confirmed blank-node identity between the validation report's `sourceShape` and the original shapes graph resolves correctly (pyshacl uses the passed-in graph object directly here, no copy/skolemization mismatch).
- Ran the full `tests/ontology/test_ontology_advanced.py` suite on the branch: 30/30 passed.
- Grepped for other `SHACLViolation(...)` construction sites and `explain_violations()` callers — this is the only non-test code path for both, so no other route is left with the old bug.

**Non-blocking nit:** the new test `test_explain_violations_uses_real_constraint_values` docstring says "regression for PR #318" — that doesn't correspond to anything in this repo's history, looks like a copy-paste leftover. Worth a follow-up cleanup, doesn't need to block this merge.

Approving.

Issue #1099, opened 2026-08-19
[BUG] Entity ids and types are written into <> without IRI validation, so GraphBuilder's own defaults emit invalid Turtle


fabio-rovai:
### Summary

`serialize_to_turtle()` writes entity ids and types straight into `<...>` with no IRI validation:

```python
lines.append(f"<{entity_id}> a <{entity_type}> ;")
```

`GraphBuilder._process_item()` (`semantica/kg/graph_builder.py:141-146`) defaults `id` to the entity's surface text and `type` to the extractor's label. So the default pipeline path, with no unusual input, emits IRIs containing spaces.

### Version

semantica 0.6.5, Python 3.13.

### Reproduction

```python
from semantica.kg.graph_builder import GraphBuilder
from semantica.export.methods import export_rdf

source = {
    "entities": [
        {"id": "Acme Corp", "name": "Acme Corp", "type": "ORG", "confidence": 0.91},
        {"id": "Jane Doe", "name": "Jane Doe", "type": "PERSON", "confidence": 0.88},
    ],
    "relationships": [
        {"source": "Jane Doe", "target": "Acme Corp", "type": "works_for", "confidence": 0.8},
    ],
}
kg = GraphBuilder(resolve_conflicts=False).build(sources=[source])
export_rdf(kg, "out.ttl", format="turtle")
```

### Actual

```turtle
<Acme Corp> a <ORG> ;
    semantica:text "" ;
    semantica:confidence 0.91 .

<Jane Doe> <works_for> <Acme Corp> .
```

- Oxigraph (strict RDF 1.1): `Parser error at line 5 between columns 1 and 12: Invalid IRI code point ' '`
- rdflib 7.x: parses, but resolves the relative reference against the file location and warns `file:///.../out/Acme Corp does not look like a valid URI, trying to serialize this will break`

The second behaviour is the more dangerous one: the export looks like it worked, and the graph silently acquires `file://` identity that will not match anything downstream.

### Expected

Either reject non-IRI ids with a clear error, or mint a stable IRI in the `semantica:` namespace by percent-encoding or slugging the label. Bare relative references such as `<ORG>` and `<works_for>` should not be emitted at all, since there is no `@base` in the output.

### Note

`semantica/triplet_store/` already validates IRIs before SPARQL interpolation (#911). The export path has no equivalent check, so the two surfaces disagree about what a valid identifier is.

### Suggested fix

[...]

Issue #1098, opened 2026-08-19
[BUG] serialize_to_turtle interpolates literals unescaped: a quote or newline in entity text emits invalid Turtle


fabio-rovai:
### Summary

`RDFSerializer.serialize_to_turtle()` (`semantica/export/rdf_exporter.py:369`) builds the literal with an f-string and no escaping:

```python
lines.append(f'    semantica:text "{text}" ;')
```

Any `"`, newline, carriage return or backslash in the entity text produces syntactically invalid Turtle. Entity text comes from extraction over free text, so quotes and newlines are ordinary, not edge cases.

### Version

semantica 0.6.5, Python 3.13.

### Reproduction

```python
from semantica.export.methods import export_rdf

kg = {"entities": [{"id": "https://example.org/e3",
                    "text": 'He said "hi"\nthen left',
                    "type": "https://example.org/Person",
                    "confidence": 1.0}],
      "relationships": []}
export_rdf(kg, "out.ttl", format="turtle")
```

### Actual

```turtle
<https://example.org/e3> a <https://example.org/Person> ;
    semantica:text "He said "hi"
then left" ;
    semantica:confidence 1.0 .
```

Both parsers reject it:

- Oxigraph (strict RDF 1.1): `Parser error at line 6 between columns 30 and 32: hi is not a valid subject or graph name`
- rdflib 7.x: `BadSyntax ... Bad syntax (objectList expected)`

### Expected

Escaped per the Turtle grammar, either as an escaped short string (`\"`, `\n`, `\r`, `\\`) or a triple-quoted long string.

### Suggested fix

A single `_escape_literal()` helper applied everywhere a value is interpolated into quotes, or delegate serialization to rdflib, which is already a hard dependency (`rdflib>=6.2.0` in `pyproject.toml`). The same unescaped pattern appears in the OWL-Time branch of `_owl_time_triples_for_rel()` for `time:inXSDDateTimeStamp` values.

Related, previously fixed in a different module: #478 (OWLExporter generated invalid Turtle), #448 (BulkLoader), #450. The RDF exporter was not covered by those.

pkupt:
Same root as #1097 — I've claimed that one and plan to cover this escaping gap in the same PR, since both are in `serialize_to_turtle()` (rdf_exporter.py). Confirmed on current `main`: a `text` value containing `"` or a newline emits invalid Turtle. Happy to fix both together.

Sameer6305:
@pkupt you can take this one.
[...]

PR #1109, opened 2026-08-19
Declare the Semantica vocabulary, and mint entity IRIs deterministically (#1107, #1101)

### LINKED ISSUE #1101: [BUG] Missing-id fallback uses Python hash(): entity IRIs change between processes, and use a 'semantica:' URI scheme rather than the declared prefix
### Summary

When an entity has no `id`, `serialize_to_turtle()` synthesises one from Python's builtin `hash()`:

```python
entity_id = f"semantica:entity_{hash(entity_text)}"
```

`hash()` on `str` is randomised per process by PYTHONHASHSEED, so the same knowledge graph exported twice produces different IRIs. The same pattern is used for `rel_base_id` in `_owl_time_triples_for_rel()`.

### Version

semantica 0.6.5, Python 3.13.

### Reproduction

Run this twice, as two separate processes:

```python
kg = {"entities": [{"text": "Acme Corp", "type": "https://example.org/Org"}], "relationships": []}
print([l for l in open("out.ttl") if "entity_" in l])
```

### Actual

```
run 1: <semantica:entity_1966390833113473535> a <https://example.org/Org> ;
run 2: <semantica:entity_-1569031364158313747> a <https://example.org/Org> ;
```

### Why it matters

1. **Identity is not stable across runs.** Re-exporting the same graph yields a different graph. Diffing two exports, deduplicating against an earlier load, or resolving a provenance record written by an earlier run all break. This is hard to square with the project's provenance and audit-trail guarantees.
2. **The IRI is not what the prefix declares.** `<semantica:entity_...>` inside angle brackets is an absolute IRI with the scheme `semantica`, not `https://semantica.dev/ns#entity_...`. It parses, so nothing complains, but it never matches the declared prefix expansion used elsewhere in the same file, including the N-Triples serializer.

### Expected

A deterministic mint, for example `hashlib.sha256(text.encode()).hexdigest()[:16]`, written as a proper IRI in the declared namespace (`https://semantica.dev/ns#entity_<hash>`) or as the prefixed name `semantica:entity_<hash>` without angle brackets.

Point 2 applies to the default type as well: `entity.get("type", "semantica:Entity")` reaches `<semantica:Entity>`, so entities that fall back to the default type get a type IRI that no SHACL `sh:targetClass` or OWL class in the `https://semantica.dev/ns#` namespace will ever match. That is the same failure mode as #1082, reached from the data side rather than the shape side.

### LINKED ISSUE #1107: [BUG] The namespace every export mints terms in (https://semantica.dev/ns#) returns 404 and no vocabulary ships with the package
### Summary

Every RDF export mints terms in `https://semantica.dev/ns#`: `semantica:text`, `semantica:confidence`, `semantica:openEndedInterval`, the synthesised entity IRIs, and the default type `semantica:Entity`. The provenance exporter mints under the same base by design (`DEFAULT_BASE_URI`, so KG and PROV URIs co-resolve).

That namespace has no vocabulary behind it.

### Evidence

```
$ curl -sSL -H 'Accept: text/turtle, application/rdf+xml' -o /dev/null -w '%{http_code} %{content_type}\n' https://semantica.dev/ns
404 text/html;charset=utf-8
```

and the distributed package contains no RDF vocabulary of any kind:

```
$ find $(python -c 'import semantica,os;print(os.path.dirname(semantica.__file__))') \
    -name '*.ttl' -o -name '*.owl' -o -name '*.rdf' -o -name '*.jsonld'
(no results)
```

semantica 0.6.5, installed from PyPI.

### Why this matters beyond tidiness

1. **Nothing declares what the terms mean.** A consumer receiving a Semantica export cannot determine whether `semantica:confidence` is an annotation property, a datatype property, what its range is, or whether `semantica:text` is intended as an `rdfs:label` equivalent. There is no document to consult and no document to fetch.
2. **No consumer can tell a real term from a typo.** In the open world an undeclared IRI is unknown, not wrong, so `semantica:txet` and `semantica:text` are equally valid to every RDF tool. Closed-world checking is the only thing that separates them, and it requires a published vocabulary to check against. With none, that check cannot be run at all: pointed at a Semantica export, the closed-world checker in [open-ontologies](https://github.com/fabio-rovai/open-ontologies) reports `no ontology vocabulary found (0 declared terms), nothing was checked` rather than a verdict.
3. **It undercuts the interoperability claim.** The project describes itself as W3C standards compliant with full provenance. Emitting PROV-O alongside terms from an unresolvable namespace means the PROV half is checkable and the payload half is not.

### What resolving this would look like

Publishing a small vocabulary at `https://semantica.dev/ns#` declaring the handful of terms the exporters actually emit, with `rdfs:domain`, `rdfs:range` and labels, and content negotiation so the IRIs dereference. The set is small: I count six terms across the RDF, OWL-Time and provenance paths.

Related: #1101 covers a second problem with the same namespace, where the fallback path writes `<semantica:entity_N>`, an IRI in the scheme `semantica` rather than the prefix expansion, so those terms would not resolve even once the namespace does.

fabio-rovai:
Closes #1107. Closes #1101.

Taking these together as suggested in #1107: the vocabulary on its own would not have made the minted IRIs resolve, because the fallback path was writing IRIs that could never resolve under any namespace.

## The vocabulary

`semantica/ontology/vocabulary/semantica-ns.ttl` declares the fourteen terms the exporters actually emit, taken from the emitting call sites in `export/rdf_exporter.py`, `export/json_exporter.py` and `provenance/manager.py` rather than from what a vocabulary ought to contain. Nothing is included for completeness.

- **Classes**: `sem:Entity`, `sem:Relationship`
- **Entity terms**: `sem:text`, `sem:confidence`, `sem:metadata`
- **Relationship terms**: `sem:related_to`, `sem:source`, `sem:target`, `sem:type`
- **Document terms**: `sem:entities`, `sem:relationships`, `sem:exportedAt`
- **Temporal**: `sem:openEndedInterval`
- **Provenance**: `sem:role_generator`, typed `prov:Role`

The three judgement calls agreed in #1107 are implemented as agreed: `sem:confidence` and `sem:metadata` carry no `rdfs:domain` since relationship exports need them too, `sem:exportedAt` is ranged `xsd:dateTime` rather than `xsd:dateTimeStamp` because `datetime.now().isoformat()` is timezone-naive, and `sem:text` stays distinct from `rdfs:label` rather than aliasing it.

It ships inside the package so it can be loaded without a network round trip, with a small accessor:

```python
from semantica.ontology.vocabulary import vocabulary_turtle
```

and is the same document intended to be served at the namespace IRI. Hosting and content negotiation at `semantica.dev/ns#` remain yours to arrange; this PR does not assume anything about how.

`tests/ontology/test_vocabulary.py` ties the document to the code. Every term the serializers can write must be declared, so adding a term to an exporter without declaring it fails the build rather than shipping an undeclared IRI. That is the part that keeps this from drifting six months from now.

## Deterministic entity IRIs

An entity arriving without an id had one minted from Python's builtin `hash()`, which is randomised per process. The same entity therefore received a different IRI on every run:

```
run 1: <semantica:entity_1966390833113473535>
run 2: <semantica:entity_-1569031364158313747>
```

so an export could not be diffed against an earlier one, deduplicated against an earlier load, or joined to a provenance record written by an earlier process.

Minting now uses SHA-256 and writes a full IRI in the declared namespace. The prefixed form mattered as much as the randomness: `<semantica:entity_N>` inside angle brackets is an IRI in the scheme `semantica`, not the expansion of the declared prefix, so those nodes never joined with anything written through it. The same applied to the default entity and relationship types in the Turtle path, which now write full IRIs too.

`tests/export/test_rdf_exporter_iri_minting.py` covers stability across processes by running the mint under three different `PYTHONHASHSEED` values in subprocesses, which is the form the regression actually took.

Deliberately left alone: the unvalidated IRIs in #1099, which is assigned elsewhere, and the unescaped literals in #1098. This PR touches only the fallback path.

## Verification

134 export tests and 91 ontology tests pass.


KaifAhmad1 (approved):
Reviewed the vocabulary + deterministic-IRI changes end to end, including the two rounds of prior review (temporal IRI collisions, the confidence-range contradiction) — both correctly fixed.

Found and pushed a follow-up commit closing the remaining gaps:

- `serialize_to_rdfxml` still defaulted the entity type to the bare string `"semantica:Entity"` written into an `rdf:resource` attribute — unlike a Turtle angle-bracket or an XML element name, that attribute value isn't namespace-expanded, so it hit the exact #1101 failure mode on a path the original tests didn't cover. Now uses the full-IRI `DEFAULT_ENTITY_TYPE`, with a regression test.
- `json_exporter.py` emits `semantica:format` and `@type: "semantica:KnowledgeGraph"`, neither declared in the vocabulary nor covered by the `EMITTED_TERMS` drift guard — both now declared and guarded, so the "undeclared terms fail the build" claim actually holds.
- `MANIFEST.in` didn't mirror the `pyproject.toml` package-data addition, so a source-distribution install could ship without the vocabulary file.
- The cross-process minting-stability test replaced the subprocess's entire environment with a POSIX-only `PATH`, breaking it on Windows — now overrides only `PYTHONHASHSEED` on top of the inherited environment.
- Minor: folded the hand-rolled `hashlib.sha256(...).hexdigest()` calls into the existing `hash_data()` helper.

229 export/ontology tests pass locally. Approving.

Thanks for the thorough work here, @fabio-rovai — the vocabulary grounded in actual call sites rather than a wishlist, the drift-guard tests tying the doc to the code, and the fast, precise turnaround on both review rounds made this an easy PR to review and land.

PR #1112, opened 2026-08-19
fix(export): normalize Turtle resource IRIs

### LINKED ISSUE #1099: [BUG] Entity ids and types are written into <> without IRI validation, so GraphBuilder's own defaults emit invalid Turtle

mikemikimike:
Closes #1099.

Normalize resource identifiers at the RDF export boundary so GraphBuilder defaults cannot produce invalid or unintended relative IRIs.

This change applies the shared normalization path across Turtle (including temporal Turtle), RDF/XML, and N-Triples serialization:

- Bare and relative identifiers are minted beneath the Semantica namespace with safe percent-encoding.
- Absolute IRIs remain unchanged.
- Configured and input-context prefixes are expanded using the effective namespace mapping.
- Malformed URI-like values are handled safely instead of aborting export.
- Entity IDs, entity types, relationship endpoints, relationship types, and optional/None values are covered.
- Existing export behavior for valid absolute IRIs is preserved.

Regression coverage includes the original GraphBuilder reproduction, deterministic Turtle output parsing with rdflib, temporal Turtle, RDF/XML, N-Triples, context-prefix expansion, malformed identifiers, Windows-style paths, and existing absolute IRIs.

Validation:

- 38 focused regression tests passed
- 175 export tests plus 46 subtests passed
- `git diff --check` passed

The implementation also preserves the existing namespace declaration behavior; that pre-existing concern is outside this issue's scope.

### DIFF
--- a/semantica/export/rdf_exporter.py
+++ b/semantica/export/rdf_exporter.py
@@ -32,6 +32,7 @@ License: MIT
 from pathlib import Path
 from decimal import Decimal, InvalidOperation
 from typing import Any, Dict, List, Optional, Set, Union
+from urllib.parse import quote, urlsplit
 
 from ..utils.exceptions import ProcessingError, ValidationError
 from ..utils.helpers import ensure_directory, hash_data
@@ -395,6 +396,29 @@ class RDFSerializer:
 
     # OWL-Time namespace URI
     _OWL_TIME_NS = "http://www.w3.org/2006/time#"
+    _SEMANTICA_NS = "https://semantica.dev/ns#"
+
+    def _as_turtle_iri(
+        self, value: Any, namespaces: Optional[Dict[str, str]] = None
+    ) -> str:
+        """Return an absolute, safely encoded IRI for a Turtle resource."""
+        value = str(value)
+        try:
+            parsed = urlsplit(value)
+        except ValueError:
+            parsed = urlsplit("")
+        if parsed.scheme:
+            prefix, separator, local_name = value.partition(":")
+            namespace = (namespaces or self.namespace_manager.namespaces).get(prefix)
+            if namespace and separator:
+                return quote(namespace + local_name, safe=":/?#[]@!$&'()*+,;=%")
+            remainder = value[len(prefix) + 1 :]
+            if len(prefix) >= 2 and (
+                remainder.startswith(("//", "/"))
+                or any(token in remainder for token in (":", "/", "@"))
+            ):
+                return quote(value, safe=":/?#[]@!$&'()*+,;=%")
+        return self._SEMANTICA_NS + quote(value, safe="")
 
     # Design decision — TemporalBound.OPEN in RDF:
     # OWL-Time has no standard predicate for "no known end date." We use
@@ -467,7 +491,10 @@ class RDFSerializer:
             text = entity.get("text") or entity.get("label", "")
             confidence = normalize_confidence(entity.get("confidence", 1.0))
 
-            lines.append(f"<{entity_id}> a <{entity_type}> ;")
+            lines.append(
+                f"<{self._as_turtle_iri(entity_id, merged_namespaces)}> a "
+                f"<{self._as_turtle_iri(entity_type, merged_namespaces)}> ;"
+            )
             if confidence is None:
                 self.logger.warning(
                     f"Entity {entity_id} has a confidence that is not a number "
@@ -488,7 +515,11 @@ class RDFSerializer:
             target_id = rel.get("target_id") or rel.get("target")
             rel_type = rel.get("type", DEFAULT_RELATION_TYPE)
 
-            lines.append(f"<{source_id}> <{rel_type}> <{target_id}> .")
+            lines.append(
+                f"<{self._as_turtle_iri(source_id)}> "
+                f"<{self._as_turtle_iri(rel_type)}> "
+                f"<{self._as_turtle_iri(target_id)}> ."
+            )
 
             if include_temporal:
                 owl_lines = self._owl_time_triples_for_rel(rel, idx, time_axis)
--- a/semantica/export/rdf_exporter.py
+++ b/semantica/export/rdf_exporter.py
@@ -516,9 +516,9 @@ class RDFSerializer:
             rel_type = rel.get("type", DEFAULT_RELATION_TYPE)
 
             lines.append(
-                f"<{self._as_turtle_iri(source_id)}> "
-                f"<{self._as_turtle_iri(rel_type)}> "
-                f"<{self._as_turtle_iri(target_id)}> ."
+                f"<{self._as_turtle_iri(source_id, merged_namespaces)}> "
+                f"<{self._as_turtle_iri(rel_type, merged_namespaces)}> "
+                f"<{self._as_turtle_iri(target_id, merged_namespaces)}> ."
             )
 
             if include_temporal:
--- a/semantica/export/rdf_exporter.py
+++ b/semantica/export/rdf_exporter.py
@@ -29,6 +29,7 @@ Author: Semantica Contributors
 License: MIT
 """
 
+import re
 from pathlib import Path
 from decimal import Decimal, InvalidOperation
 from html import escape as xml_escape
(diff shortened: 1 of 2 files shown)


Sameer6305 (dismissed):
Thanks @mikemikimike for the thorough implementation and for addressing the review findings.

During the review, we found a few gaps beyond the original #1099 reproduction:

* Temporal OWL-Time IRIs were bypassing the new normalization helper.
* Unknown `prefix:value` identifiers could still become relative IRIs.
* Bare identifiers needed correct percent-encoding semantics.
* RDF/XML and N-Triples had the same underlying unsafe IRI handling.
* `None` types and Windows-style path identifiers needed safer handling.
* We also added regression coverage for these edge cases and verified the existing export behavior.

These were addressed and pushed to the PR in **commit `9831a4bd`** (`fix(export): close IRI correctness gaps identified in PR #1112`).

The final verification looks solid:

* **38 focused regression tests passed**
* **175 export tests + 46 subtests passed**
* `git diff --check` is clean
* Exact #1099 GraphBuilder reproduction now produces valid, deterministic Turtle.
* No remaining merge-blocking issues were found.

The namespace declaration concern is pre-existing and can be handled separately.

@KaifAhmad1  **Approved from my side. Ready to merge and close #1099.** 🚀

KaifAhmad1 (changes requested):
Thanks for the work here — the Turtle fix itself is solid and the `_as_turtle_iri` approach is the right shape. But I checked this branch out and ran the issue's exact repro against all three formats the PR description claims to cover, and found the description (and Sameer's approval) no longer match what's actually in the diff.

**The PR body and Sameer's review don't match the current diff.** The current diff (`gh pr diff`, checked against fresh `origin/main`) touches exactly one method — `serialize_to_turtle()`. But the description still says *"applies the shared normalization path across Turtle (including temporal Turtle), RDF/XML, and N-Triples serialization,"* and Sameer's now-stale/dismissed review explicitly says *"RDF/XML and N-Triples had the same underlying unsafe IRI handling... These were addressed."* That broader work existed at commit `9831a4b`, but the later force-push (`1a25aae`, "handle contextual turtle iris") narrowed the branch back to Turtle-only — matching your own later comment asking whether RDF/XML and N-Triples should be included here or tracked separately. Nobody updated the description or approval status to reflect the narrower scope, so a reviewer skimming the thread (or using the bypass-merge button) would reasonably believe this is fixed across formats. It isn't.

**Verified: the original bug is still live in N-Triples and RDF/XML.** Running the issue's exact `Acme Corp` / `Jane Doe` repro on this branch:

```
N-Triples: <https://semantica.dev/ns#Acme Corp> ...
  -> rdflib strict parser: ParserError: Invalid line (literal space inside the IRI)

RDF/XML: <rdf:Description rdf:about="Acme Corp"> ...
  -> rdflib resolves it relative to the file location and warns:
     "file:///.../Acme Corp does not look like a valid URI, trying to serialize this will break"
```

That RDF/XML result is the *exact* "more dangerous" scenario the issue called out by name — the graph silently acquires `file://` identity. `serialize_to_rdfxml()` (`rdf:about`/`rdf:resource`, ~lines 692-693, 716-717) and `serialize_to_ntriples()`'s `expand_uri()` helper are untouched by this PR.

**The two Qodo findings are also still open** — I reproduced both directly against `_as_turtle_iri`:
- `_as_turtle_iri("mailto:foo")` → `https://semantica.dev/ns#mailto%3Afoo` — valid opaque absolute IRIs (`mailto:`, `isbn:`, `geo:`, etc.) get incorrectly re-minted under the Semantica namespace, silently changing node identity for previously-valid identifiers.
- `_as_turtle_iri("http://example.org/bad%zz")` → passed through unchanged — malformed percent-encoding isn't validated or repaired, so invalid-Turtle output is still reachable through a different input shape.

**Ask:** either finish RDF/XML + N-Triples normalization in this PR (reusing `_as_turtle_iri` or an equivalent shared helper, as the description already claims) and resolve the two opaque-IRI/percent-encoding cases, or rewrite the PR description to Turtle-only scope, drop the "Closes #1099" claim (the issue's own repro still fails for two of three formats), and open a tracked follow-up for the rest.

KaifAhmad1 (approved):
Re-reviewed against `cf6c9b7b`. This closes out the review cycle.

Thanks @mikemikimike for the persistence here — the `_as_turtle_iri` boundary is the right shape, and the earlier rounds (temporal Turtle, RDF/XML, N-Triples, opaque schemes, malformed `%zz`) got this most of the way there.

Two correctness gaps were still open in `52ba7b6` that neither the PR description nor the prior Qodo passes fully closed:

1. **Percent-escapes double-encoded** — removing `%` from the `quote()` safe-set to fix the `%zz` case also broke every *already-valid* percent-escape (`%20` → `%2520`), silently changing identity for any real-world absolute IRI containing one.
2. **Built-in prefixes shadowed by `@context`** — `serialize_to_ntriples`/`serialize_to_rdfxml` passed only `extract_namespaces()` (context-only) into `_as_turtle_iri`, whose fallback to the built-in namespace map only triggered when that dict was *empty*. The moment any `@context` was present, `semantica:`/`rdf:`/`rdfs:`/`owl:` stopped resolving in those two formats — same input KG produced `<https://semantica.dev/ns#Entity>` in Turtle but literal `<semantica:Entity>` in N-Triples/RDF-XML.

I verified both were reproducible on `52ba7b6` (checked out the branch, exercised `_as_turtle_iri` and the three serializers directly), then pushed `cf6c9b7b` to this branch with the fix plus two new regression tests covering exactly these cases:

- A smart-quote helper that leaves existing valid `%XX` escapes alone (still repairing malformed ones like `%zz`).
- `_as_turtle_iri` now always merges the built-in namespace map with whatever the caller passes, instead of only falling back when the passed dict is empty.

Full suite: 234 passed, 46 subtests, `git diff --check` clean. All CI checks are green on `cf6c9b7b`.

Approving — this now closes #1099 across Turtle, RDF/XML, and N-Triples with no remaining known gaps.

PR #1113, opened 2026-08-19
fix(export): normalize entity name to label on all RDF paths

### LINKED ISSUE #1097: [BUG] convert_kg_to_rdf is never called: every entity exports as semantica:text "" on all RDF paths
### Summary

`RDFSerializer.convert_kg_to_rdf()` (`semantica/export/rdf_exporter.py:249`) is the method that normalises an entity's `name` into `label`/`text`. It is defined but never called anywhere in the package, so no public export path ever runs it. `serialize_to_turtle()` then reads `entity.get("text") or entity.get("label", "")`, finds neither key on graphs produced by `GraphBuilder` (which emits `name`), and writes an empty string.

Result: every entity in every RDF export has an empty label. The human-readable content of the graph is silently dropped.

### Version

semantica 0.6.5 (pip, `semantica[shacl,tripletstore-oxigraph]`), Python 3.13.

### Reproduction

```python

source = {
    "entities": [
        {"id": "https://example.org/acme", "name": "Acme Corp",
         "type": "https://example.org/Org", "confidence": 0.91},
    ],
    "relationships": [],
}
print(open("out.ttl").read())
```

### Actual

```turtle
<https://example.org/acme> a <https://example.org/Org> ;
    semantica:text "" ;
    semantica:confidence 0.91 .
```

### Expected

```turtle
<https://example.org/acme> a <https://example.org/Org> ;
    semantica:text "Acme Corp" ;
    semantica:confidence 0.91 .
```

### Why it is easy to miss

The output is valid Turtle and parses cleanly in both rdflib and a strict RDF 1.1 parser (Oxigraph). Nothing errors, nothing warns. The triple count is correct. Only the values are gone, so a round-trip test that counts triples or checks conformance will pass.

### Suggested fix

Call `convert_kg_to_rdf()` from `RDFExporter.export_to_rdf()` before dispatching to a serializer, so every format benefits, rather than fixing the key lookup in each `serialize_to_*` method separately. A regression test asserting a non-empty `semantica:text` for a `GraphBuilder` graph would have caught this.

Verified against the same input with two independent parsers. Happy to open a PR if the maintainers prefer the fix at the `export_to_rdf` boundary.

cxzg007:
## Summary

Fixes #1097: entities that only carry a `name` field were exported with empty
`semantica:text` / `rdfs:label` values across all RDF formats.

## Root cause

`RDFSerializer.convert_kg_to_rdf()` already implemented `name` -> `label`/`text`
normalization, but it was **never called** by `export_to_rdf()`. As a result the
four serialization paths (turtle / ntriples / rdfxml / jsonld) read
`entity.get("text") or entity.get("label", "")` and produced empty literals for
name-only entities.

## Fix

Call the existing normalizer once at the `export_to_rdf()` boundary so every
format benefits from a single code path:

```python
data = self.serializer.convert_kg_to_rdf(data)
```

Properties:
- **Idempotent** — guarded by `if "text" not in ... and "label" not in ...`
- **Non-destructive** — never overwrites an existing `text`/`label`
- **Fallback** — when no `name` is present, derives a label from the id suffix

## Tests

Added 7 regression tests in `tests/export/test_rdf_exporter.py`:
- `test_name_only_entity_exports_nonempty_label` parametrized over all 4 formats
- `test_name_only_export_to_file_contains_label` (file write path)
- `test_existing_text_not_overwritten_by_name`
- `test_id_fallback_label_when_no_name`

`pytest tests/export/test_rdf_exporter.py` -> 17 passed.

Closes #1097


pkupt (commented):
Reviewed this thoroughly since my own (#1141) was just closed as a duplicate of this work — agreed, this is the right home for it. The `_escape_turtle_literal`/`_escape_xml_text` + `export_to_rdf` normalization approach is clean and covered.

Two small observations that are genuinely out of scope here but worth noting for a possible follow-up (not blocking):
- `convert_kg_to_rdf()` runs at the `export_to_rdf()` layer only; calling `RDFSerializer.serialize_to_jsonld()` (or the other serialize_* methods) directly still yields empty `semantica:text` for `name`-only entities. A future hardening could route normalization through every public serialize_* entry point.
- N-Triples `safe_text` previously escaped only quotes/newlines; your unified escaper now handles backslash/tab — nice improvement.

Happy to pick this up as a separate minimal PR if that would help. Otherwise, no complaints from me — ship it.

KaifAhmad1 (approved):
Verified the fix against current main (the actual RDFSerializer/Neo4j-equivalent export paths, not just the diff): convert_kg_to_rdf() genuinely had zero callers before this, the id-fallback and escaping logic are correct, and the merge-conflict resolution I pushed reconciles cleanly with main's own escaping consolidation while also closing a live RDF/XML injection gap main had missed. Full tests/export/ suite (323 tests) passes, all CI checks green. LGTM.

PR #1115, opened 2026-08-19
fix(kg): keep relationship endpoint aliases in sync

### LINKED ISSUE #1110: [BUG] Keep relationship endpoint aliases in sync after entity merging
## Bug Description

After entity resolution merges duplicate entities, `GraphBuilder` remaps the canonical `source`/`target` relationship fields but leaves the original `source_id`/`target_id` aliases unchanged. The same relationship can therefore contain two different endpoint IDs.

## Steps to Reproduce

```python

graph = GraphBuilder(
    merge_entities=True,
    resolve_conflicts=False,
).build({
    "entities": [
    ],
    "relationships": [
        {
            "source_id": "alice:2",
            "target_id": "org:1",
            "type": "WORKS_FOR",
        }
    ],
})

print(graph["relationships"][0])
```

## Expected Behavior

All endpoint representations for one relationship should refer to the same canonical entity after merging:

- `source` and `source_id` should both be `alice:1`.
- `target` and `target_id` should both be `org:1`.

## Actual Behavior

The relationship contains contradictory values:

```python
{
    "source_id": "alice:2",
    "target_id": "org:1",
    "source": "alice:1",
    "target": "org:1",
    "type": "WORKS_FOR",
}
```

`alice:2` is no longer a canonical entity ID.

## Environment

- **OS**: macOS
- **Python Version**: 3.11.15
- **Semantica Version**: current `main` (`b77e3e8c`)

## Impact

Consumers reading `source_id`/`target_id` can observe stale or dangling endpoints, while consumers reading `source`/`target` see the remapped graph. This can make validation, analysis, and persistence disagree.

## Possible Solution

When remapping an endpoint, update every present alias for that endpoint together. Add a regression test covering a merged entity and a relationship supplied with `source_id`/`target_id`.

I have a reproduction and a proposed fix and would be happy to implement it.

T1mn:

## Description

After entity resolution, relationship endpoints were remapped only in the `source` and `target` fields. Existing `source_id` and `target_id` aliases could still point to the pre-merge IDs.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #1110

## Changes Made

- Keep `source` and `source_id` synchronized during entity-resolution remapping.
- Keep `target` and `target_id` synchronized during entity-resolution remapping.
- Extend the KG regression test to cover both endpoint aliases.

## Testing

- [x] Tested locally

### Test Commands

```bash
pytest -q tests/kg/test_entity_pipeline.py tests/kg/test_entity_resolver_exact.py
```

Result: `9 passed`.

## Documentation

## Breaking Changes

**Breaking Changes**: No

## Checklist

## Additional Notes

The local package-build check was not run because the environment does not have the `build` module installed.


ZohaibHassan16 (commented):
Two things I think we should fix before approving:

---

**1. Remove the debug prints from the test file**

These will show up in CI output on every test run:

```python
print(f"DEBUG: graph_data keys: ...")
print(f"DEBUG: Entities count: ...")
print(f"DEBUG: Relationships count: ...")
print(f"DEBUG: First entity: ...")
```

Can we remove these before merging?

---

**2. Add one more test case**

Right now there's a test where the alias is stale but the primary endpoint is already correct.

I don't think we have a case where both `source` and `source_id` start with the merged-away ID, something like:

```python
{
    "source": "alice:2",
    "source_id": "alice:2",
    "target": "org:1",
    "type": "WORKS_FOR",
}
```

Would be good to add a test for this and assert that both `source` and `source_id` get updated to `"alice:1"` after building.

---

Other than those two things, the core fix looks good to me. Once they're addressed, I'd be good with approving this.

ZohaibHassan16 (approved):
All good, approving.

PR #1116, opened 2026-08-19
fix(kg): validate entity_id aliases

### LINKED ISSUE #1111: [BUG] GraphValidator should honor entity_id aliases
## Bug Description

`GraphValidator` only indexes entities by `id`, even though the KG pipeline accepts `entity_id` aliases and `get_entity_id()` already defines the shared identifier contract. Entity-id-only graphs are therefore reported as invalid.

## Steps to Reproduce

```python
from semantica.kg.graph_validator import GraphValidator

graph = GraphBuilder(
    merge_entities=True,
    resolve_conflicts=False,
).build({
    "entities": [
    ],
    "relationships": [
        {"source_id": "alice:2", "target_id": "org:1", "type": "WORKS_FOR"}
    ],
})

result = GraphValidator().validate(graph)
print(result)
```

## Expected Behavior

The validator should recognize `entity_id` as the entity identifier, resolve the relationship endpoints against the same identifier contract, and avoid reporting the valid `org:1` node as missing or orphaned.

## Actual Behavior

The graph is reported invalid with errors including:

- the Acme entity is missing `id`;
- the relationship target `org:1` is dangling;
- orphan-node detection does not account for the entity keyed by `entity_id`.

## Environment

- **OS**: macOS
- **Python Version**: 3.11.15
- **Semantica Version**: current `main` (`b77e3e8c`)

## Impact

Graphs accepted by the KG builder can fail validation, and downstream validation results can disagree with entity resolution and relationship handling.

## Possible Solution

Use the shared `get_entity_id()` helper when building the validator's entity index and structural graph. Accept either `id` or `entity_id` without changing the existing required name/type checks, and add regression tests for entity-id-only entities and relationships.

I have a reproduction and a proposed fix and would be happy to implement it.

T1mn:

## Description

`GraphValidator` indexed entities only by `id`, while the KG pipeline also accepts the `entity_id` alias. Entity-id-only graphs could therefore be reported as invalid even when their relationships were valid.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #1111

## Changes Made

- Use the shared `get_entity_id()` helper for entity validation and ID indexing.
- Accept `id` and `entity_id` consistently for relationship endpoint validation.
- Use the same endpoint resolution for structural checks and add regression coverage.

## Testing

- [x] Tested locally

### Test Commands

```bash
pytest -q tests/kg/test_graph_validator.py tests/kg/test_entity_pipeline.py
python -m compileall -q semantica/kg/graph_validator.py
```

Result: `5 passed`.

## Documentation

## Breaking Changes

**Breaking Changes**: No

## Checklist

## Additional Notes

The local package-build check was not run because the environment does not have the `build` module installed.


KaifAhmad1 (approved):
Reviewed by checking out the branch into a worktree and running it, not just reading the diff. This branch was BEHIND main, so I merged origin/main into it locally first — clean merge, no conflicts.

**Verified the regression is real**: reverted `graph_validator.py` to the parent commit (`4a886d97`) — all 3 new tests fail, confirming `GraphValidator` really did index entities by `id` only, making `entity_id`-only entities invisible to uniqueness tracking, dangling-edge checks, and structural (cycle/orphan) analysis.

**Verified the fix**: on the merged branch, all 3 new tests in `tests/kg/test_graph_validator.py` pass, and the full `tests/kg/` suite passes clean: 499 passed, no regressions.

**Code read**: correctly reuses the pre-existing shared `get_entity_id()` helper rather than reinventing resolution logic. The unhashable-ID handling is correct — the `try`/`except TypeError` wraps both the `in` check and the `.add()`, so a genuinely unhashable ID (e.g. a list) can't partially succeed; verified with a direct probe. Beyond the entity-id aliasing itself, this also fixes a real second defect: the old cycle/orphan-detection code built the structural `nx_graph` from raw `r["source"]`/`r["target"]` with no `source_id`/`target_id` fallback, so relationships using the canonical alias keys were silently excluded from cycle and orphan detection too, not just entity indexing. The PR unifies this through the same `get_relationship_endpoint`/`is_valid_id` helpers already used for dangling-edge checks, which also fixes the third bullet point in the original issue (orphan detection not accounting for entity_id-keyed entities). `is_valid_id` is also hoisted out of the per-relationship loop it used to be redefined in — a reasonable cleanup alongside the fix.

**One side effect I found by probing beyond the PR's own suite, not a regression**: an entity with an explicit `id: None` (key present, value null) now gets a new `MISSING_FIELD` issue where the old code silently accepted it. This is arguably a correctness improvement (a null ID was never usable), but it is a behavior change to the existing `id` check that the PR description says it isn't making. Confirmed it doesn't break anything in the full suite — flagging for awareness, not blocking.

Both Qodo findings are resolved. No blocking issues found. Approving.

Issue #1114, opened 2026-08-19
[BUG] Every timestamp is written timezone-naive, in two idioms that mean different things, so a timezone-qualified SPARQL filter silently drops them


fabio-rovai:
Filing this as @KaifAhmad1 asked on #1107, where the missing timezone forced the shipped vocabulary to weaken `sem:exportedAt`'s range. It turned out to be wider than the one call site, so the report covers what I could verify through the public API.

## Bug Description

Every timestamp Semantica writes is timezone-naive, in two different idioms that mean two different things and serialize identically:

- `datetime.now().isoformat()` — 106 calls across 41 files, 18 of them in `semantica/export/`, including `json_exporter.py:505` which writes `sem:exportedAt` into a JSON-LD export. This reads the **local** clock of whatever machine ran the export.
- `datetime.utcnow().isoformat()` — 70 calls across 23 files, including `provenance/schemas.py:94`, whose value reaches RDF as `prov:generatedAtTime`, `prov:startedAtTime`, `prov:endedAtTime` and `prov:atTime`, typed `xsd:dateTime` (`provenance/manager.py:1239`, `1283`, `1286`, `1294`). This means **UTC**.

Neither writes an offset, so once the value is in RDF nothing distinguishes them. `2026-08-19T14:19:04` from the PROV-O exporter is 14:19 UTC; the same string from the JSON-LD exporter is 14:19 in an unrecorded local zone, anywhere from 00:19Z to 02:19Z the following day.

`datetime.utcnow()` is also deprecated. On Python 3.13, constructing a `ProvenanceEntry` under `-W error::DeprecationWarning` raises:

```
DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a
future version. Use timezone-aware objects to represent datetimes in UTC:
datetime.datetime.now(datetime.UTC).
```

## Why it matters

**A timezone-qualified SPARQL filter silently excludes every Semantica-written timestamp.** Under XSD 1.1, comparing a value with no timezone against one with a timezone is indeterminate whenever they fall within the ±14-hour window, and SPARQL turns an indeterminate comparison into an error, which `FILTER` discards as a non-match. Loading three exports into Oxigraph — one stamped the way Semantica stamps it, two stamped correctly:

```sparql
SELECT ?e WHERE {
  ?e sem:exportedAt ?t .
  FILTER (?t < "2026-08-19T18:00:00Z"^^xsd:dateTime)
}
```

```
[...]

PR #1120, opened 2026-08-19
Mint JSON-LD @ids the same way the RDF serializers do (#1101, missed by #1109)

fabio-rovai:
#1109 merged six minutes before I pushed this commit to its branch, so it went in without the last fix and #1101 closed with the defect still live on the JSON-LD paths. Same commit, rebased onto main. Detail and evidence are in [my comment on #1109](https://github.com/semantica-agi/semantica/pull/1109#issuecomment-5343326604); reopening #1101 or filing a fresh issue is your call, whichever you prefer for the changelog.

#1109 fixed `serialize_to_turtle`, `serialize_to_ntriples` and `serialize_to_rdfxml`. Both JSON-LD writers still interpolate the entity's own text into `f"semantica:entity/{text}"` (`json_exporter.py:533`, `rdf_exporter.py:606`) and the endpoints into `f"semantica:rel/{source}_{target}"` (`json_exporter.py:572`, `rdf_exporter.py:625`). Three consequences, all reproducible on main through the public API.

**An entity whose text contains a space disappears from the export.** `semantica:entity/Acme Corp` expands to an IRI containing a space, which is invalid, and a JSON-LD parser drops the node rather than reporting it. One character apart:

```
text='Acme Corp'  triples= 2  entity node present: []
text='AcmeCorp'   triples= 6  entity node present: ['https://semantica.dev/ns#entity/AcmeCorp']
```

Most organisation and person names an extractor produces contain a space, so this is the common case, and it fails silently.

**Relationships collapse onto one node.** `serialize_to_jsonld` resolved endpoints from `source_id`/`target_id` only, while the rest of the module accepts `source`/`target` too — the asymmetry review caught on the temporal path in #1109, still present here. Every relationship carrying the second form minted the identical `semantica:rel/_`. Two unrelated relationships in, one node out, carrying both types:

```
S https://semantica.dev/ns#rel/_
  P https://semantica.dev/ns#type   O works_for
S https://semantica.dev/ns#rel/_
  P https://semantica.dev/ns#type   O employs
```

**The two serializations of one graph disagree on identity.** Turtle writes `<https://semantica.dev/ns#entity_a73cb4563ee2e72c>`, JSON-LD writes `semantica:entity/Acme Corp` for the same entity, so an export in one format cannot be joined to an export in the other. That is what #1101 is about, on the paths its fix did not reach.

Both writers now call `mint_entity_iri`/`mint_relationship_iri`, resolving endpoints both ways and passing the list index the RDF paths pass, so one knowledge graph carries one node identity whichever serializer wrote it.

Separately, `JSONExporter.export_entities` and `export_relationships` write `"@id": "semantica:entities"` into a `@context` that declares only `@vocab`, never the `semantica` prefix. A processor reads that as an IRI in the scheme `semantica`; rdflib returns the predicate literally as `semantica:entities`. That is the original #1101 defect on a third path, so the prefix is declared there now.

`tests/export/test_jsonld_iri_minting.py` parses each export with a real JSON-LD processor instead of asserting on the JSON text, and covers all of the above. Every test in it fails on main and passes here. 228 export and ontology tests pass.

Two things I found while probing and deliberately did not touch, since both are modelling calls rather than bug fixes:

1. `@vocab` is `https://semantica.dev/vocab/`, a second namespace distinct from `https://semantica.dev/ns#` and equally undeclared. Unprefixed keys land there — `vocab/metadata`, `vocab/entity_count`, `vocab/text`, and an entity's own type, so `{"type": "ORG"}` becomes `https://semantica.dev/vocab/ORG`. The vocabulary that just landed says nothing about that namespace, and the `EMITTED_TERMS` drift guard does not cover it.
2. `_convert_kg_to_jsonld` mints the document node as `https://semantica.dev/graph/{datetime.now().isoformat()}`, so the same graph exported twice gives two document IRIs and the identity depends on the exporter's timezone. Filed as #1114 along with the timezone-naive stamps you asked me to report.


KaifAhmad1 (approved):
Ports #1109's IRI-minting fix to the two paths it missed — both JSON-LD writers now go through `mint_entity_iri`/`mint_relationship_iri` instead of interpolating raw entity text and relationship endpoints into an f-string IRI, and `export_entities`/`export_relationships` declare the `semantica` prefix their `@context` was already writing `semantica:entities` against.

**Verified independently, not just read:**
- Checked out the branch and ran `tests/export/test_jsonld_iri_minting.py` — all 7 pass.
- Reverted only the two touched source files to `main` and reran the same tests — all 7 fail, confirming these are real regressions on `main` today, not a test file asserting nothing.
- Ran the full `tests/export/` + `tests/ontology/` suite on the branch — 236 passed, matching the PR description.
- Traced the `index` argument through `mint_relationship_iri` across `serialize_to_turtle` (temporal path only — Turtle has no reified relationship node outside `include_temporal`) and `serialize_to_jsonld` — both enumerate the same `relationships` list from 0, so the entity-IRI agreement the tests assert between JSON-LD and Turtle is on solid ground, and correctly doesn't extend that assertion to relationship IRIs, which Turtle has no analog for.
- Confirmed the new `json_exporter.py → rdf_exporter.py` import doesn't create a cycle (`rdf_exporter.py` has no back-reference).
- All 10 CI checks pass.

**Scope check:** the two things flagged but deliberately left untouched — the undeclared `@vocab` namespace, and `_convert_kg_to_jsonld`'s non-deterministic `datetime.now()` document IRI (tracked separately in #1114) — are correctly out of scope; both are modeling decisions, not instances of this bug.

No changes requested.

PR #1121, opened 2026-08-19
Write timestamps with an explicit UTC offset, and tighten sem:exportedAt to xsd:dateTimeStamp (#1114)

### LINKED ISSUE #1114: [BUG] Every timestamp is written timezone-naive, in two idioms that mean different things, so a timezone-qualified SPARQL filter silently drops them

fabio-rovai:
Closes #1114, the timezone report you asked me to file on #1107.

`semantica/export/` stamped every value with `datetime.now().isoformat()`, which reads the machine's **local** clock. `semantica/provenance/` stamped its own with `datetime.utcnow().isoformat()`, which reads **UTC**. Both return a naive datetime and both serialize identically, so once the value leaves the process nothing distinguishes them: the same string means two different instants depending on which module wrote it.

In RDF the consequence is silent rather than loud. Under XSD 1.1, comparing a value with no timezone against one with a timezone is indeterminate whenever the two fall inside the ±14 hour window; SPARQL turns an indeterminate comparison into an error, and `FILTER` discards errors as non-matches. Three timestamps in an Oxigraph store, one stamped the way Semantica stamped it and two stamped correctly:

```sparql
SELECT ?e WHERE { ?e sem:exportedAt ?t . FILTER (?t < "2026-08-19T18:00:00Z"^^xsd:dateTime) }
```

```
    https://example.org/exportB          <- "2026-08-19T15:00:00+05:30"
```

`exportA`, carrying `"2026-08-19T14:19:04"` exactly as `_convert_kg_to_jsonld` wrote it, is simply absent. `prov:generatedAtTime`, `prov:startedAtTime`, `prov:endedAtTime` and `prov:atTime` all carried values written that way, so an audit trail could not be ordered against timestamps from any other system.

The same export run under four timezones at one instant, before the fix:

```
  TZ=UTC                  sem:exportedAt=2026-08-19T14:19:02.821750
  TZ=Europe/London        sem:exportedAt=2026-08-19T15:19:03.169252
  TZ=Asia/Kolkata         sem:exportedAt=2026-08-19T19:49:03.511919
  TZ=America/Los_Angeles  sem:exportedAt=2026-08-19T07:19:03.867268
```

## What this changes

`utc_now()` and `utc_now_iso()` in `semantica/utils/helpers.py`, exported from `semantica.utils`, used at all 29 call sites in `export/` (`json_exporter`, `yaml_exporter`, `report_generator`, `export_provenance`) and `provenance/` (`manager`, `schemas`, `bridge_axiom`). Values now read `2026-08-19T14:19:04.229937+00:00`.

`sem:exportedAt`'s range in the vocabulary that landed with #1109 tightens from `xsd:dateTime` to `xsd:dateTimeStamp`. The term's comment had to explain why the weaker range was necessary; that reason is gone.

`datetime.utcnow()` is deprecated as of Python 3.12 and scheduled for removal — constructing a `ProvenanceEntry` under `-W error::DeprecationWarning` on 3.13 raised, and no longer does.

## What this deliberately does not change

The other **147** naive call sites, in `context/`, `vector_store/`, `seed/` and elsewhere. Those timestamps are compared against values parsed back from previously stored naive strings (`datetime.fromisoformat(data["timestamp"])` in `context/decision_models.py`, `context/causal_analyzer.py`, `provenance/schemas.py` and 35 other places), so converting the write side alone would raise `TypeError: can't compare offset-naive and offset-aware datetimes` on data written by an earlier version. That sweep needs a read side that treats a missing offset as UTC, and it is a different change with a migration story attached. Happy to take it next if you want it.

One thing worth your eye: `ProvenanceManager.query_recorded_between` compares ISO strings lexicographically. Adding the `+00:00` suffix does not reorder anything, because the comparison diverges long before the offset in every realistic case, but a bound written to match a stored value character-for-character would now sit fractionally below it. Existing behaviour was already lexical on ISO strings, so I have left it alone rather than change the semantics of a query method in a timestamp PR.

## Verification

`tests/export/test_timestamp_timezones.py` and `tests/provenance/test_timestamp_timezones.py`: offset presence on every export and provenance path, PROV-O literals valid as `xsd:dateTimeStamp`, comparison against a timezone-aware instant without `TypeError`, the declared range matching what the exporter writes, the document `@id` still a valid IRI with `+00:00` in it, and the Oxigraph filter that dropped the naive value.

That last test picks a bound one hour ahead rather than a round year, on purpose: a bound years away is determinate even for a naive value, so the version of the test I wrote first passed without the fix. 13 of the 14 tests fail with this commit's `semantica/export`, `semantica/provenance` and vocabulary reverted.

564 export, ontology and provenance tests pass. Across the full suite the failure set is identical before and after — 329 failures either way, all from optional dependencies missing in my environment.


KaifAhmad1 (approved):
Fixes #1114 — every export/provenance timestamp was naive (`datetime.now()` local time in `export/`, `datetime.utcnow()` UTC in `provenance/`, indistinguishable once serialized), so a timezone-qualified SPARQL `FILTER` silently dropped Semantica-written records instead of erroring. Adds `utc_now()`/`utc_now_iso()`, ports all 29 call sites, tightens `sem:exportedAt`'s vocabulary range to `xsd:dateTimeStamp`, and incidentally fixes the Python 3.12 `datetime.utcnow()` deprecation. Second commit fixes a real regression this PR introduced and Qodo caught: `query_recorded_between`/`audit_log` compared timestamps as raw strings, which breaks once some entries carry `+00:00` and others don't — now compares instants via a new `to_utc_datetime()` that reads a missing offset as UTC (correct, since that's what the pre-fix values actually meant).

**Verified independently:**
- Ran both new test files (19 tests) on the branch — all pass.
- Reverted the touched source files to `main` while keeping the new tests — collection fails outright (`ImportError: cannot import name 'utc_now'`), confirming the fix is load-bearing, stronger evidence than the PR's own "11/13, 5/5 fail" claim.
- Ran the full `tests/export/ tests/ontology/ tests/provenance/` suite: 579 passed, 1 failure. That failure (`test_minted_entity_iri_is_stable_across_processes`) passes in isolation and — checked — fails identically on unmodified `main`, so it's pre-existing subprocess-test flakiness, not something this PR broke.
- Grepped `export/` and `provenance/` for any remaining naive `datetime.now()`/`datetime.utcnow()` call — none found.
- Confirmed `schemas.py`'s retained `datetime` import is still used elsewhere (`SourceReference.from_dict`), not dead code.
- All 10 CI checks pass.

**Design calls, all reasonable:** leaving the other 147 naive call sites for a separate migration (they compare against previously-stored naive data, so converting the write side alone would raise `TypeError`); `to_utc_datetime` returning `None` rather than raising on unparseable input, with both range-query methods falling back to the historical string comparison on an unreadable bound — verified by `test_an_unreadable_bound_falls_back_to_the_previous_behaviour`.

No changes requested.

PR #1123, opened 2026-08-19
Read the ontology shape the generator actually emits, and stop minting empty class IRIs (#1103)

### LINKED ISSUE #1103: [BUG] OWLExporter and OntologyGenerator disagree on the ontology dict schema: every class exports as <> and all properties are dropped
### Summary

`OntologyGenerator` emits ontology dicts shaped `{"classes": [...], "properties": [...]}` (see `ontology/ontology_generator.py:409, 458-459, 482-483`), and `SHACLGenerator` consumes exactly that shape. `OWLExporter` reads a different schema:

- class IRI from `cls.get("uri") or cls.get("id", "")` (`export/owl_exporter.py:392`), never `name`
- properties from `ontology.get("object_properties", [])` and the datatype equivalent, never `properties`

So feeding the OWL exporter the output of the project's own ontology generator produces a file in which every class has an empty IRI and no property is exported at all.

### Version

semantica 0.6.5, Python 3.13.

### Reproduction

```python
from semantica.export.methods import export_owl

ontology = {
    "classes": [{"name": "Person", "label": "Person"},
                {"name": "Organization", "label": "Organization"}],
    "properties": [
        {"name": "fullName", "type": "datatype", "domain": "Person", "range": "string", "required": True},
        {"name": "foundedOn", "type": "datatype", "domain": "Organization", "range": "date"},
    ],
}
export_owl(ontology, "onto.ttl", format="turtle")
print(open("onto.ttl").read())
```

### Actual

```turtle
<https://semantica.dev/ns#> a owl:Ontology ;
    rdfs:label "SemanticaOntology" ;
    owl:versionInfo "1.0" .

<> a owl:Class ;
    rdfs:label "Person" .

<> a owl:Class ;
    rdfs:label "Organization" .
```

Three consequences:

1. `<>` is the empty relative IRI, which denotes the document itself. Both classes are therefore the **same node**, and that node is also the ontology document.
2. A strict RDF 1.1 parser rejects the file outright: Oxigraph gives `Parser error at line 11 between columns 1 and 3: No scheme found in an absolute IRI`.
3. All three properties are missing. The output is silent about this; no warning is logged.

### Expected

`export_owl` should accept the shape `OntologyGenerator` produces: mint class IRIs from `name` under the ontology namespace when `uri`/`id` is absent, and read `properties` (splitting on `type`) in addition to `object_properties`/`datatype_properties`. Failing that, it should raise rather than emit `<>`.

### Note

#478 fixed an earlier invalid-Turtle-and-dropped-properties bug in this same exporter. The underlying schema mismatch between the generator and the exporter was not addressed, so the same symptom returns through the generator path.

fabio-rovai:
Closes #1103.

## What was wrong

Two independent faults met in the OWL export path, and together they meant a generated ontology could not round trip at all.

**Properties were dropped.** `OWLExporter` reads `object_properties` and `data_properties`. `OntologyGenerator` emits a single `properties` list tagged with `type` and `@type`. Nothing bridged the two, so every generated property was silently discarded.

**Classes collapsed onto one node.** `ClassInferrer` writes `"uri": None` when it is constructed without a namespace manager. The stage 3 guard is `if "uri" not in cls`, which never fires, because the key is present and only its value is missing. The exporter then evaluated `cls.get("uri") or cls.get("id", "")` to the empty string and wrote `<>`.

`<>` is a relative IRI. It resolves against the parser's base, and rdflib uses the current working directory:

```
rdflib.term.URIRef('file:///Users/fabio/projects/semantica/')
```

So two classes became one subject carrying two `rdfs:label` values, and the identity of that subject changed with the directory the export ran from. Oxigraph does not resolve it at all:

```
SyntaxError: Parser error at line 2 between columns 1 and 3:
No scheme found in an absolute IRI
```

Here is the full Turtle export of a two class, three property ontology on `main`:

```turtle
<https://semantica.dev/ontology/> a owl:Ontology ;
    rdfs:label "GeneratedOntology" ;
    owl:versionInfo "1.0" .

<> a owl:Class ;
    rdfs:label "Person" ;
    rdfs:comment "Class representing person entities" .

<> a owl:Class ;
    rdfs:label "Organization" ;
    rdfs:comment "Class representing organization entities" .
```

Three properties are missing, and the two classes are the same node.

## What this changes

- `_split_properties` classifies the combined `properties` list by `type`/`@type` and merges it with any explicit `object_properties` and `data_properties`, so both dict shapes work and cannot drift apart again.
- `_term_iri` resolves a class or property IRI, falling back from `uri` to `iri` to `id` to a name joined onto the ontology base. A term with none of those is skipped with a warning instead of being written as `<>`.
- Domain and range references resolve through a class index, so a bare name such as `"Person"` lands on the IRI that class was actually exported under rather than on a re-derived guess.
- Data property ranges resolve properly. `"string"`, `"xsd:string"` and a full IRI all give one well formed datatype. The previous `rdfs:range xsd:{range}` produced `xsd:xsd:string` for generator output. Turtle keeps the compact `xsd:` form the module already used, so existing expectations still hold.
- Both `not in` guards in the generator become truthiness checks, and the stage 4 fallback mints an absolute IRI rather than assigning a bare class name.
- XML text and attribute values are escaped. They were interpolated raw, so a label containing `&` or `<` broke the document.

Same input, after:

```turtle
<https://semantica.dev/ontology/Person> a owl:Class ;
    rdfs:label "Person" ;
    rdfs:comment "Class representing person entities" .

<https://semantica.dev/ontology/worksAt> a owl:ObjectProperty ;
    rdfs:label "worksAt" ;
    rdfs:domain <https://semantica.dev/ontology/Person> ;
    rdfs:range <https://semantica.dev/ontology/Organization> .

<https://semantica.dev/ontology/age> a owl:DatatypeProperty ;
    rdfs:label "age" ;
    rdfs:domain <https://semantica.dev/ontology/Person> ;
    rdfs:range xsd:integer .
```

## Verification

Turtle and RDF/XML now serialise the same 25 triples for the same ontology, and both are accepted by rdflib and by Oxigraph.

`tests/export/test_owl_exporter_generator_schema.py` adds 10 regression tests. They drive the exporter from a real `OntologyGenerator` run rather than a hand written stand in, and assert on the parsed graph rather than on serialised text, so they cannot pass on output that merely looks right. All 10 fail on the parent commit.

The existing export and ontology suites pass, 231 tests. `tests/ontology/test_notebook_14.py` was excluded from that run because it imports sklearn, which is not installed here, and it is untouched by this change.


KaifAhmad1 (approved):

**Verified the regression is real**: reverted `owl_exporter.py`/`ontology_generator.py` to the parent commit (`2a303cf4`) with the new test file applied — 15/16 new tests fail, confirming classes really do collapse onto `<>` and generator-shaped `properties` really are dropped on `main`.

**Verified the fix**: on this branch, all 16 new tests in `tests/export/test_owl_exporter_generator_schema.py` pass (including the strict Oxigraph-parser check), and the full `tests/export/` + `tests/ontology/` suites pass clean: 253 passed, 46 subtests, no regressions.

**Code read**: `_term_iri`/`_join_iri`/`_iri_safe` correctly fall back uri→iri→id→name with percent-encoding, and skip-with-warning instead of ever re-emitting `<>`. `_class_iri_index`/`_resolve_class_ref` correctly route domain/range through the actual exported class IRI, and correctly distinguish prefixed names like `owl:Thing` from real absolute IRIs. `_split_properties` correctly merges the generator's combined list with explicit `object_properties`/`data_properties` and warns on malformed/untyped entries instead of silently dropping them. The `"uri" not in cls"` → `not cls.get("uri")` fix is correct given `ClassInferrer` sets the key to `None`. `improve_coherence` no longer reaches for a `namespace_manager` that `OntologyOptimizer` doesn't have.

**Manually probed edge cases** not in the test suite — blank/whitespace-only class names, empty/`None` domain-range lists — both handled correctly (clean skip + warning, no crash, no garbage IRI).

No correctness issues found. CI is green. Approving.

PR #1124, opened 2026-08-19
Target the namespace the data uses, and stop attaching domain-less properties to every class (#1104, #1105)

### LINKED ISSUE #1104: [BUG] SHACLGenerator mints targets under /shapes/ while every other module uses /ns#, so generated shapes never match generated data and always report conforms
### Summary

`SHACLGenerator` defaults `base_uri` to `https://semantica.dev/shapes/` and mints **both** the shape IRIs and the `sh:targetClass` / `sh:path` IRIs under it. Every other module mints terms under `https://semantica.dev/ns#` (`RDFSerializer`'s `semantica` prefix, `OWLExporter`'s `ont` prefix, `ProvenanceManager.DEFAULT_BASE_URI`).

The generated shapes therefore target classes that no other part of the package ever produces, so validation matches zero focus nodes and reports conformance regardless of the data.

### Version

semantica 0.6.5, Python 3.13.

### Reproduction

```python
from semantica.ontology.ontology_generator import SHACLGenerator

onto = {
    "classes": [{"name": "Person", "label": "Person"}],
    "properties": [{"name": "fullName", "type": "datatype", "domain": "Person",
                    "range": "string", "required": True}],
}
gen = SHACLGenerator()
open("shapes.ttl", "w").write(gen.serialize(gen.generate(onto), "turtle"))
```

produces

```turtle
@prefix ex: <https://semantica.dev/shapes/> .
<https://semantica.dev/shapes/PersonShape>
    a sh:NodeShape ;
    sh:targetClass ex:Person ;
    sh:property [ sh:path ex:fullName ; sh:datatype xsd:string ; sh:minCount 1 ; ... ] .
```

Now validate an instance of the class as any other Semantica module would write it:

```turtle
@prefix ont: <https://semantica.dev/ns#> .
<https://example.org/p1> a ont:Person .          # missing the required fullName
```

```python
from pyshacl import validate
import rdflib
d = rdflib.Graph().parse("data.ttl", format="turtle")
s = rdflib.Graph().parse("shapes.ttl", format="turtle")
print(validate(d, shacl_graph=s)[0])
```

### Actual

`True` (conforms). The instance is missing a property the shape declares `sh:minCount 1`, and validation passes, because `ex:Person` (`https://semantica.dev/shapes/Person`) and `ont:Person` (`https://semantica.dev/ns#Person`) are different IRIs.

Changing only the data namespace to `https://semantica.dev/shapes/` makes the same instance report `False` with the expected `MinCountConstraintComponent` violation, confirming the shapes themselves are fine and it is purely the namespace that does not line up.

### Why this is worse than a normal false negative

A SHACL run that matches nothing is indistinguishable, in its report, from a SHACL run where everything passed. `conforms: True` with zero focus nodes is the output either way, so a pipeline that validates before publishing gets a green light on unvalidated data.

### Expected

One namespace default across the package, or `SHACLGenerator` taking the target namespace from the ontology dict rather than from its own `base_uri`. Separately, it would be worth reporting the focus-node count in the validation report so an empty match is visible.

### Related

#1082 is the same end state (shapes that never match) arriving from a different cause, namespace mangling rather than namespace disagreement. Fixing #1082 alone will not make generated shapes match generated data.

### LINKED ISSUE #1105: [BUG] SHACLGenerator attaches domain-less properties to every node shape, so a required property with no domain becomes mandatory on every class
### Summary

In `SHACLGenerator._attach_property_shapes()` (`ontology/ontology_generator.py`), a property with no declared `domain` is attached to **every** node shape:

```python
else:
    # No domain declared → attach to all shapes
    self.logger.debug(f"Property '{pname}' has no domain — attaching to all node shapes")
    for node_shape in graph.node_shapes:
        node_shape.property_shapes.append(self._build_property_shape(prop))
```

When that property is `required`, the generator emits `sh:minCount 1` on every class in the ontology. Missing domain information is treated as "applies universally" rather than "unknown", so the generated shapes reject data that the ontology permits.

### Version

semantica 0.6.5, Python 3.13.

### Reproduction

```python
from semantica.ontology.ontology_generator import SHACLGenerator

onto = {
    "classes": [{"name": "Person"}, {"name": "Organization"}],
    "properties": [
        {"name": "fullName", "type": "datatype", "domain": "Person", "range": "string", "required": True},
        {"name": "taxIdentifier", "type": "datatype", "range": "string", "required": True},  # no domain
    ],
}
gen = SHACLGenerator()
print(gen.serialize(gen.generate(onto), "turtle"))
```

### Actual

`taxIdentifier` appears with `sh:minCount 1` on **both** `PersonShape` and `OrganizationShape`. Every Person instance is now required to carry a tax identifier.

### Expected

An undeclared domain should produce no property shape, or a property shape with no `sh:minCount`, or an explicit opt-in flag such as `attach_undomained=True`. A visible warning rather than a debug-level log would also help, since the current behaviour silently invents constraints that the source ontology never stated.

### Interaction with quality_tier="strict"

`_apply_quality_tier()` sets `closed=True` on all shapes in the strict tier. Combined with the above, strict-tier shapes both require properties that were never scoped to the class and forbid everything not enumerated. That combination will reject essentially any real dataset.

fabio-rovai:
Closes #1104. Closes #1105.

## #1104: the shapes matched nothing, and the validator called that conforming

`SHACLGenerator` used one namespace for two different jobs. `base_uri` says where the shape resources live, and it was also used to expand every `sh:targetClass` and `sh:path`. With the default `https://semantica.dev/shapes/`, this is what came out:

```turtle
<https://semantica.dev/shapes/PersonShape>
    a sh:NodeShape ;
    sh:targetClass ex:Person ;          # ex: = https://semantica.dev/shapes/
    sh:property [
        sh:path ex:fullName ;
        sh:minCount 1 ;
    ] .
```

No data graph anywhere in the package uses `https://semantica.dev/shapes/Person`. `RDFSerializer` writes the `ns#` vocabulary, `OWLExporter` writes the ontology's own class IRIs. The shapes selected zero focus nodes.

The part that makes this worth fixing rather than documenting is that the failure is invisible. A shape with no focus nodes is vacuously satisfied, so on data that plainly breaks the stated constraint:

```turtle
ex:alice a ex:Person .   # required fullName is absent
```

pySHACL 0.25, the `shacl` extra this package ships, reports:

```
Validation Report
Conforms: True
```

A user validating extraction output against generated shapes gets a clean report from a shapes file that checked nothing.

### What changed

The two namespaces are now separate. `target_namespace` resolves in this order:

1. an explicit `target_namespace=` argument,
2. the ontology's declared `namespace.base_uri`,
3. the namespace of any absolute IRI a class or property already carries, since a term's own IRI is the most reliable evidence of where the data lives,
4. the ontology `uri`,
5. the vocabulary namespace the package ships, rather than the shapes namespace.

`SHACLGraph` gains a `term_iris` index mapping each class and property name to the IRI it expands to, and `_uri` resolves through it. Shapes are still indexed internally by name, so domain matching and inheritance propagation are untouched.

Same ontology, after:

```turtle
<https://semantica.dev/shapes/PersonShape>
    a sh:NodeShape ;
    sh:targetClass <https://example.org/onto#Person> ;
    sh:property [
        sh:path <https://example.org/onto#fullName> ;
        sh:datatype xsd:string ;
        sh:minCount 1 ;
    ] .
```

```
Conforms: False
Constraint Violation in MinCountConstraintComponent
	Focus Node: ex:alice
	Result Path: ex:fullName
	Message: Less than 1 values on ex:alice->ex:fullName
```

The shape resources keep living in the shapes namespace, so they stay distinct from the classes they target.

## #1105: a domain-less property was asserted on every class

`_attach_property_shapes` ended with "No domain declared, attach to all shapes". That states a constraint the ontology does not. With `minCount 1` it makes every instance of every class invalid, and it was logged at debug, so nothing surfaced.

A property with no declared domain is now left unattached, with a warning naming the property and pointing at the fix. `attach_domainless_properties=True` restores the previous behaviour for anyone relying on it.

## On the changed test

`tests/ontology/test_ontology_advanced.py::test_no_domain_property_attaches_to_all_shapes` asserted the #1105 behaviour directly, so it pinned the defect in place. Rather than delete it I split it in two: the original expectation now runs against the explicit opt-in, and a new sibling asserts the default. Flagging it because editing an existing test is worth a second pair of eyes.

## Verification

`tests/ontology/test_shacl_target_namespace.py` adds 17 tests. They validate real data through pySHACL rather than reading the shapes text, which matters here specifically because the old output could pass a text-level check while matching nothing. They are parametrised over three ontology shapes: one carrying class URIs and a declared namespace, one declaring only a namespace, and one carrying only URIs.

Export and ontology suites pass, 239 tests. `tests/ontology/test_notebook_14.py` was excluded from that run because it imports sklearn, which is not installed here, and it is untouched by this change.


KaifAhmad1 (approved):

**Verified the regression is real**: reverted `ontology_generator.py`/`engine.py` to the parent commit (`2a303cf4`) with the new test file applied — 19/27 new tests fail, confirming shapes really did target the wrong namespace (`https://semantica.dev/shapes/...` instead of the data namespace) and domain-less properties really were force-attached to every class.

**Verified the fix**: on this branch, all 27 new tests in `tests/ontology/test_shacl_target_namespace.py` pass, and the full `tests/ontology/` + `tests/export/` suites pass clean: 281 passed, 46 subtests, no regressions. The `target_namespace` resolution order, the split `class_iris`/`property_iris` indices, the `attach_domainless_properties` opt-in, and the `to_shacl` parameter forwarding all check out against the described behavior.

**Two non-blocking gaps found by testing beyond the PR's own suite**, noting them for a follow-up rather than blocking on them:

1. `_term_iri`'s final fallback (when a name isn't in either index) resolves through `graph.base_uri` rather than the resolved `target_ns`. When no `namespace.base_uri` is declared and a domain/range names a class the ontology doesn't itself declare (a dangling/typo'd reference), this falls back to the *shapes* namespace again — reproduced with an object property `range="Organization"` where no such class exists in `classes`, giving `sh:class <https://semantica.dev/shapes/Organization>`.
2. N-Triples output doesn't expand well-known prefixed names the way Turtle/JSON-LD do. `range="owl:Thing"` (which is what the generator's own stage-4 fallback assigns per #1123) serializes correctly in Turtle and JSON-LD but as the literal wrong term `owl:Thing` in N-Triples — parses to a different resource than the real `http://www.w3.org/2002/07/owl#Thing`.

Both sit in the same "silently wrong, then vacuously conforms" failure class this PR closes for the main path, just on paths the current tests don't reach. Worth a fast follow-up. Approving — the core fix is correct, tested, and a clear improvement over `main`.

PR #1125, opened 2026-08-19
Write confidence as one typed decimal on every serialization path (#1100, #1102)

### LINKED ISSUE #1100: [BUG] Turtle and N-Triples exports of the same KG are different graphs (relative vs expanded IRIs, xsd:decimal vs xsd:float)
### Summary

Exporting one knowledge graph to `turtle` and to `ntriples` produces two graphs that are not isomorphic. The serializers are independent string builders that disagree on both subject identity and literal datatype.

### Version

semantica 0.6.5, Python 3.13.

### Reproduction

```python

source = {"entities": [{"id": "Acme Corp", "name": "Acme Corp", "type": "ORG", "confidence": 0.91}],
          "relationships": []}
export_rdf(kg, "out.nt", format="ntriples")
```

### Actual

`out.ttl`:

```turtle
<Acme Corp> a <ORG> ;
    semantica:confidence 0.91 .
```

`out.nt`:

```
<https://semantica.dev/ns#Acme Corp> <...#type> <https://semantica.dev/ns#ORG> .
<https://semantica.dev/ns#Acme Corp> <https://semantica.dev/ns#confidence> "0.91"^^<http://www.w3.org/2001/XMLSchema#float> .
```

Two divergences from the same input:

1. **Subject and type identity.** Turtle emits the raw value as a relative reference; N-Triples prefix-expands it into `https://semantica.dev/ns#`. These are different nodes. Anything that exports Turtle for review and N-Triples for loading gets two different graphs.
2. **Literal datatype.** `0.91` bare in Turtle is `xsd:decimal` by the Turtle grammar; N-Triples types it `xsd:float`. `xsd:decimal` and `xsd:float` are disjoint in the XSD value space, so a SHACL `sh:datatype` constraint or an OWL range that holds for one export fails for the other.

### Expected

All four serializers should be projections of one graph model. A property-based test asserting `rdflib.compare.isomorphic()` across `turtle`, `ntriples`, `jsonld` and `rdfxml` for the same input would pin this down permanently.

### Suggested fix

Build an `rdflib.Graph` once in `export_to_rdf()` and let rdflib serialize to each format. rdflib is already a required dependency. That removes four hand-written serializers and this whole class of divergence at the same time.

### LINKED ISSUE #1102: [BUG] confidence is interpolated without a type check: a non-numeric value emits 'semantica:confidence high .' (invalid Turtle)
### Summary

`serialize_to_turtle()` interpolates `confidence` as a bare token with no type check:

```python
confidence = entity.get("confidence", 1.0)
...
lines.append(f"    semantica:confidence {confidence} .")
```

Nothing in the dict-based API validates the type of `confidence`, so a string, `None` or a bool reaches the serializer unchanged and produces invalid Turtle.

### Version

semantica 0.6.5, Python 3.13.

### Reproduction

```python
kg = {"entities": [{"id": "https://example.org/e2", "name": "X",
                    "type": "https://example.org/Person",
                    "confidence": "high"}],
      "relationships": []}
```

### Actual

```turtle
<https://example.org/e2> a <https://example.org/Person> ;
    semantica:text "" ;
    semantica:confidence high .
```

- Oxigraph: `Parser error at line 7 between columns 26 and 30: high is not a valid RDF object`
- rdflib 7.x: `BadSyntax (objectList expected)`

`None` produces `semantica:confidence None .` and `True` produces `semantica:confidence True .`, both invalid for the same reason. Python's `True` is also not the Turtle boolean literal `true`.

### Expected

Coerce to a typed literal and validate: `"0.91"^^xsd:decimal` for numbers, an explicit error or an omitted triple for non-numeric values. The N-Triples serializer already emits `"0.91"^^xsd:float` for the same field, so the two paths should be reconciled at the same time (see the companion report on cross-format divergence).

### Note

`RDFExporter.export_to_rdf()` does call `RDFValidator.validate_rdf_syntax()` before serializing, and it logs `Continuing with export, but data may be invalid` rather than failing. In this case it does not flag the value at all, so the warning never fires and an unparseable file is written silently.

fabio-rovai:
Closes #1100. Closes #1102.

## #1100: one value, four different RDF terms

I filed this as a Turtle versus N-Triples disagreement. Probing the other two paths, it is worse than that. All four serializers render the same confidence differently. Exporting one entity with `confidence: 0.9` and parsing each output back:

| format | value | datatype |
|---|---|---|
| Turtle | `0.9` | `xsd:decimal` |
| N-Triples | `0.9` | `xsd:float` |
| RDF/XML | `0.9` | none, a plain literal |
| JSON-LD | `0.9` | `xsd:double` |

Turtle wrote the value bare, and the Turtle grammar reads a bare decimal as `xsd:decimal`. N-Triples typed it `xsd:float` explicitly. RDF/XML emitted `<semantica:confidence>0.9</semantica:confidence>` with no `rdf:datatype`. JSON-LD emitted a native JSON number, which expands to `xsd:double`.

Those are four distinct RDF terms. `FILTER(?c = "0.9"^^xsd:decimal)` matches one of them. Loading a Turtle export and an N-Triples export of the same graph into one store gives the entity two different confidence values.

There was a second divergence in the same property. N-Triples omitted the triple entirely when `confidence` was absent, while the other three wrote the `1.0` default, so the two serializations differed in triple count as well as in datatype.

### The choice of datatype

`xsd:decimal`, for two reasons. It is what the Turtle path already produced, so the most used output is unchanged and no existing user's data shifts. And it is exact, where `xsd:float` is 32 bit binary and cannot represent 0.9 or 0.95 at all.

`normalize_confidence` produces one canonical lexical form, and all four paths write it with `CONFIDENCE_DATATYPE`. Values that arrive in exponent notation are reformatted, because `1e-05` is a valid Python repr and an invalid `xsd:decimal` lexical form.

I kept the `1.0` default rather than removing it, since three of the four paths already wrote it and this issue is about the paths agreeing. Fabricating a confidence for an entity that never declared one is arguably wrong on its own terms, but that is a behaviour change nobody asked for, so I have left it alone rather than bundle it in. Happy to file it separately if you want it changed.

## #1102: one bad field cost the whole export

The Turtle path interpolated the value with no type check:

```python
lines.append(f"    semantica:confidence {confidence} .")
```

A confidence of `"high"` produced:

```turtle
<...#e1> a <...#ORG> ;
    semantica:text "x" ;
    semantica:confidence high .
```

```
BadSyntax: Bad syntax (objectList expected) at ^ in:
"...semantica:confidence'^b' high .\n"
```

The document does not parse at all, so one bad field from one extractor loses every entity in the export.

A value that cannot be a decimal is now omitted, with a warning naming the entity, rather than written as something the vocabulary contradicts. Numeric strings such as `"0.85"` are still accepted, since that is a usable value in an awkward wrapper. Booleans are rejected: `bool` subclasses `int` in Python, so `True` would otherwise silently become a confidence of 1.

## Vocabulary

`sem:confidence` shipped in #1109 with no `rdfs:range`, deliberately, and with a comment saying so, because declaring either candidate would have contradicted the other exporters. It now declares `xsd:decimal`, and a drift guard test asserts the vocabulary and `CONFIDENCE_DATATYPE` stay in agreement.

## Verification

`tests/export/test_confidence_literal_typing.py` adds 20 tests. They compare the parsed graphs of all four formats rather than their text, assert Turtle and N-Triples are isomorphic, and check the unusable-value cases through an actual parse so a malformed document fails the test rather than passing a substring check.

Export and ontology suites pass, 240 tests. `tests/ontology/test_notebook_14.py` was excluded from that run because it imports sklearn, which is not installed here, and it is untouched by this change.

## Note on overlap

This touches `rdf_exporter.py`, as do #1112 and #1122. The lines differ (IRI normalisation and literal escaping there, the confidence literal here), so it should merge cleanly in any order, but flagging it.


KaifAhmad1 (approved):

**Verified the regression is real**: reverted `rdf_exporter.py`/`semantica-ns.ttl` to the parent commit (`2a303cf4`) with the new test file applied — all 26 new tests fail, confirming the four serializers really did disagree on datatype (`xsd:decimal` / `xsd:float` / untyped / `xsd:double` for the identical value) and a non-numeric confidence really did produce unparseable Turtle.

**Verified the fix**: on this branch, all 26 new tests in `tests/export/test_confidence_literal_typing.py` pass, and the full `tests/export/` + `tests/ontology/` suites pass clean: 307 passed, 46 subtests, no regressions.

**Code read of `normalize_confidence`**: hand-traced the logic — bool rejection (`bool` subclasses `int`), NaN/Infinity rejection via `is_finite()`, the exponent-magnitude guard (`adjusted()` bounded to ±100, closing the `1e100000000`-style memory blowup the automated review caught), the signed-zero collapse (`-0.0` → `"0"`), and the trailing-zero/dot trim — all correct, including on cases like `"0.05"` where the zero/dot stripping could plausibly misfire but doesn't.

**Probed one path the PR's own suite doesn't directly cover**: an entity with the `confidence` key omitted entirely (not just `None`) — the second half of #1100, since N-Triples used to skip the triple while the other three wrote the `1.0` default. Manually verified all four serializers now emit the identical `"1"^^xsd:decimal` term.

No correctness issues found. CI is green. Approving.

PR #1126, opened 2026-08-19
Give the OWL-Time interval a subject the graph can reach (#1106)

### LINKED ISSUE #1106: [BUG] include_temporal emits OWL-Time intervals attached to a dangling identifier with no path from the relationship they date
### Summary

With `include_temporal=True`, `serialize_to_turtle()` emits an OWL-Time interval for a relationship, but attaches it to a node that nothing else in the graph references. The relationship itself is written as a plain triple, so there is no path from the subject, the predicate or the object to the interval that is supposed to date it.

### Version

semantica 0.6.5, Python 3.13.

### Reproduction

```python

kg = {
    "entities": [{"id": "https://example.org/e1", "text": "A",
                  "type": "https://example.org/Org", "confidence": 1.0}],
    "relationships": [{"source": "https://example.org/e1",
                       "target": "https://example.org/e2",
                       "type": "https://example.org/worksFor",
                       "valid_from": "2020-01-01", "valid_until": "2023-06-30"}],
}
export_rdf(kg, "temporal.ttl", format="turtle", include_temporal=True)
```

### Actual

```turtle
<https://example.org/e1> <https://example.org/worksFor> <https://example.org/e2> .

<semantica:rel_0_0> time:hasTime <semantica:rel_0_0__valid_interval> .
<semantica:rel_0_0__valid_interval> a time:Interval ;
    time:hasBeginning <semantica:rel_0_0__valid_begin> ;
    time:hasEnd <semantica:rel_0_0__valid_end> .
<semantica:rel_0_0__valid_end> a time:Instant ;
    time:inXSDDateTimeStamp "2023-06-30"^^xsd:dateTimeStamp .
<semantica:rel_0_0__valid_begin> a time:Instant ;
    time:inXSDDateTimeStamp "2020-01-01"^^xsd:dateTimeStamp .
```

Checked with rdflib against the parsed graph:

```
triples with semantica:rel_0_0 as subject: 1
triples with semantica:rel_0_0 as object:  0
any triple connecting e1 to semantica:rel_0_0: False
```

`semantica:rel_0_0` is a dangling identifier. A consumer reading this file can see that some interval exists and cannot determine what it dates. Querying "when did e1 work for e2" returns nothing, which is the only question the temporal export exists to answer.

### Two further observations from the same output

1. `<semantica:rel_0_0>` is an IRI in the scheme `semantica`, not the expansion of the declared `semantica:` prefix (`https://semantica.dev/ns#`). It parses, so no error is raised, and it will not join with anything written through the prefix.
2. `"2020-01-01"^^xsd:dateTimeStamp` is not a valid lexical form for that datatype, which requires a time component and a timezone offset. The values passed in are dates. An OWL-DL reasoner or a `sh:datatype` check treats these as ill-typed literals; rdflib does not flag them because it registers no lexical validator for `xsd:dateTimeStamp`, so the defect survives a round trip through rdflib unnoticed.

### Expected

Reifying the relationship, or attaching the interval to the relationship's own identifier, would let a consumer trace an interval back to what it dates. As emitted, the temporal triples are unreachable from the data they describe.

fabio-rovai:
Closes #1106.

## What was wrong

`include_temporal=True` emits a well formed OWL-Time interval. It just hangs off nothing:

```turtle
<...ns#e1> <...ns#employs> <...ns#e2> .

<...ns#rel_0_0940a860853f71df> time:hasTime <...ns#rel_0_0940a860853f71df__valid_interval> .
<...ns#rel_0_0940a860853f71df__valid_interval> a time:Interval ;
    time:hasBeginning <...ns#rel_0_0940a860853f71df__valid_begin> ;
    time:hasEnd <...ns#rel_0_0940a860853f71df__valid_end> .
```

`rel_0_0940a860853f71df` appears in exactly one place: as the subject of `time:hasTime`. Counting inbound arcs to it gives zero.

A relationship is serialised as a single triple, `<e1> <employs> <e2>`, so there is no node in the graph for the time to attach to. The interval is minted against an IRI that describes a relationship the graph never names. Every triple parses, every datatype validates, and there is no path from the `employs` edge to its own validity interval, which is the only reason to export it.

## What changed

The JSON-LD path already reifies relationships:

```json
{"@id": "...", "@type": "semantica:Relationship",
 "semantica:source": {...}, "semantica:target": {...}, "semantica:type": "employs"}
```

and `sem:Relationship`, `sem:source`, `sem:target` and `sem:type` are all declared in the vocabulary shipped in #1109. So the shape already exists and is already documented; Turtle simply was not writing it.

Turtle now emits that same node when it has temporal data to attach:

```turtle
<...ns#e1> <...ns#employs> <...ns#e2> .

<...ns#rel_0_0940a860853f71df> a semantica:Relationship ;
    semantica:source <...ns#e1> ;
    semantica:target <...ns#e2> ;
    semantica:type "employs" .

<...ns#rel_0_0940a860853f71df> time:hasTime <...ns#rel_0_0940a860853f71df__valid_interval> .
```

This has the side benefit of bringing the Turtle and JSON-LD descriptions of a relationship into agreement.

Scope kept deliberately tight:

- The direct triple is still written. Reification is added alongside the edge, not in place of it.
- Nothing is reified when a relationship carries no temporal data, so default output and `include_temporal=False` output are byte for byte unchanged.
- The relationship IRI comes from the same `mint_relationship_iri` the interval was already using, so the interval keeps its existing subject rather than moving.

## Verification

`tests/export/test_owl_time_reachability.py` adds 7 tests. The one that matters is a SPARQL walk from the edge to its interval, which is exactly what the dangling node made impossible:

```sparql
SELECT ?begin WHERE {
    ?s ?p ?o .
    ?rel sem:source ?s ;
         sem:target ?o ;
         time:hasTime/time:hasBeginning/time:inXSDDateTimeStamp ?begin .
}
```

There is also a check that every term the reified node emits is declared in the shipped vocabulary, since a reification nobody declared would only move the problem.

Export and ontology suites pass, 228 tests. `tests/ontology/test_notebook_14.py` was excluded from that run because it imports sklearn, which is not installed here, and it is untouched by this change.

## One thing I deliberately did not do

`semantica:type` is written as a literal, matching the JSON-LD export and the `owl:DatatypeProperty` declaration in the vocabulary. An object property pointing at the predicate IRI would be stronger for querying, but that is a vocabulary change affecting the JSON-LD path too, so it belongs in its own issue rather than here. Say the word and I will file it.


KaifAhmad1 (approved):
Reviewed by checking out the branch into a worktree and running it, not just reading the diff. This branch was BEHIND main (which now includes #1123/1124/1125, also touching rdf_exporter.py), so I merged origin/main into it locally first to check for conflicts before evaluating.

**Verified the regression is real**: reverted `rdf_exporter.py` to the parent commit (`2a303cf4`) — 7/9 new tests fail, confirming the OWL-Time interval really did hang off an IRI with zero inbound arcs and no path from the relationship edge to its own validity interval.

**Merges cleanly with current main, no conflicts.** After merging origin/main locally: all 9 new tests in `tests/export/test_owl_time_reachability.py` pass, and the full `tests/export/` + `tests/ontology/` suites pass clean: 316 passed, 46 subtests, no regressions. Output composes correctly with #1125's confidence-typing changes now on main.

**Code read**: `_reified_relationship_triples` mints `rel_id` via the exact same `rel.get("id") or mint_relationship_iri(idx, source_id or "", target_id or "")` expression `_owl_time_triples_for_rel` already uses, which is the detail that makes the fix actually work (the reified node and the interval's subject have to be the same IRI) — verified this holds, including when source_id/target_id are absent. The literal-escaping for `semantica:type` covers everything Turtle's STRING_LITERAL_QUOTE grammar actually forbids raw (`"`, `\`, `\n`, `\r`) — confirmed by hand with a string containing all four. Reification is correctly gated on there being actual temporal data to attach, so default output and `include_temporal=False` stay untouched.

One pre-existing, out-of-scope observation: the *direct* triple line (`<{source_id}> <{rel_type}> <{target_id}> .`, untouched by this PR) does no escaping when `rel_type` contains characters illegal inside a Turtle IRI ref. This predates #1126 and isn't something this PR should be blocked on — the new reified-literal path this PR adds does escape correctly. Worth its own issue if you want it tracked.

PR #1127, opened 2026-08-19
Let a registered custom method refuse (#1108)

### LINKED ISSUE #1108: [BUG] method_registry swallows exceptions from custom methods, so a registered gate cannot refuse an export
### Summary

`method_registry` is documented as the surface for "community-contributed extensions", but a registered method cannot fail. `export_rdf` (and its siblings) wrap the custom call in a bare `except Exception` and fall back to the default implementation:

```python
custom_method = method_registry.get("rdf", method)
if custom_method and custom_method is not export_rdf:
    try:
        return custom_method(data, file_path, format=format, **kwargs)
    except Exception as e:
        logger.warning(f"Custom method {method} failed: {e}, falling back to default")
```

So any extension whose purpose is to *refuse* something (validation, policy, schema gating, a pre-publication check) is silently downgraded to the behaviour it was registered to prevent. The failure is logged at warning level and the caller receives a normal return.

### Version

semantica 0.6.5, Python 3.13.

### Reproduction

Register a method that verifies the exported RDF and raises when it is unsound:

```python
from semantica.export.registry import method_registry
from semantica.export.rdf_exporter import RDFExporter
from pathlib import Path

class VerificationError(RuntimeError): pass

def verified(data, file_path, format="turtle", **kw):
    path = Path(file_path)
    RDFExporter().export(data, path, format=format, **kw)
    # any independent RDF 1.1 parser will do here
    from open_ontologies_lite import OntologyEngine
    result = OntologyEngine.validate(path.read_text(), format)
    if not result.ok:
        path.unlink(missing_ok=True)
        raise VerificationError(result.error)
    return result

method_registry.register("rdf", "verified", verified)

kg = GraphBuilder(resolve_conflicts=False).build(sources=[{
    "entities": [{"id": "Acme Corp", "name": "Acme Corp", "type": "ORG", "confidence": 0.9}],
    "relationships": [],
}])
export_rdf(kg, "out.ttl", format="turtle", method="verified")
```

### Actual

```
WARNING  Custom method verified failed: Parser error at line 5 between columns 1 and 12:
         Invalid IRI code point ' ', falling back to default
```

`out.ttl` exists, 358 bytes, containing `<Acme Corp> a <ORG> ;`, which no RDF 1.1 parser accepts. The gate ran, correctly identified the problem, deleted the file, and the fallback wrote it back. `export_rdf` returned normally, so nothing downstream can tell that verification failed.

### Why the fallback is the wrong default here

Falling back makes sense for a method that is merely *unavailable*, say an optional dependency that is not installed. It is the opposite of what is wanted for a method that is *asserting something is wrong*. As written the two cases are indistinguishable, and the more dangerous one wins: the export that a check just rejected is the one that ends up on disk.

The same pattern appears in `export_json`, `export_csv`, `export_owl` and the other `export_*` wrappers, so the behaviour is consistent across the registry rather than specific to RDF.

### Effect on what can be built

This forecloses a category of extension rather than making one awkward: anything that gates, validates, enforces a policy, or refuses to emit. Those are the extensions most worth contributing to an accountability platform, and today the registry cannot express them.

I hit this building a verification adapter against 0.6.5, which is how the reproduction above was found rather than constructed.

fabio-rovai:
Closes #1108.

## What was wrong

Every module that supports custom methods wrapped the registered callable like this:

```python
custom_method = method_registry.get("rdf", method)
if custom_method and custom_method is not export_rdf:
    try:
        return custom_method(data, file_path, format=format, **kwargs)
    except Exception as e:
        logger.warning(f"Custom method {method} failed: {e}, falling back to default")

# ... default implementation runs anyway
```

That makes a registered method advisory. It can add behaviour, but it cannot decline.

For a gate, a validator, or a policy check, declining is the entire purpose. Raising is how such a method says "do not produce this output". Swallowing the exception and running the default produces exactly the output the method was registered to prevent, and the only trace is a warning line.

The demonstration from the issue: a verifier rejects invalid RDF and deletes the file. The fallback writes it straight back. The caller is told nothing beyond a warning, and ends up with a file that failed validation.

This forecloses the whole category of validate, gate and policy extensions that a method registry otherwise makes possible.

## What changed

`semantica/utils/custom_methods.py` holds the policy in one place:

```python
result = call_custom_method(logger, method, custom_method, data, file_path, format=format, **kwargs)
if result is not CUSTOM_METHOD_FELL_BACK:
    return result
```

- An exception from a registered method propagates to the caller.
- `fallback_on_custom_error=True` restores the previous warn-and-continue behaviour for that call. It is consumed by the policy and never forwarded to the custom method.
- The sentinel is compared by identity, so a custom method that legitimately returns `None`, `0`, `""` or `[]` is not mistaken for a failure. The old code returned inside the `try`, so this needed care.

## Scope, and a note on size

The issue was filed against the export path, but the same swallow is in six modules, so leaving the others would not close it. All 58 sites are converted:

| module | sites |
|---|---|
| `export/methods.py` | 13 |
| `ingest/methods.py` | 13 |
| `normalize/methods.py` | 13 |
| `parse/methods.py` | 12 |
| `embeddings/methods.py` | 4 |
| `kg/methods.py` | 3 |

The rewrite is mechanical: 57 sites were byte-identical in shape and one had a multi-line call. Every site now reads the same way.

This is a behaviour change in error handling across six modules, which is a bigger blast radius than the other fixes I have open. I would rather show you the whole thing than half-fix it, but if you would prefer this split (say, `export/` first to settle the policy, then a follow-up for the other 45), I am happy to do that. Equally, if you would rather the default stayed warn-and-continue and refusal were opt-in, that is a one-line flip of the default in `call_custom_method` and I will make it.

## Verification

`tests/utils/test_custom_method_can_refuse.py` adds 13 tests, including:

- the issue's own demonstration: a gate deletes the target and raises, and the file must not exist afterwards,
- a custom method that succeeds is unaffected,
- the opt-in still falls through to the default,
- falsy return values are not treated as failures,
- a guard, parametrised over all six modules, that none still carries the swallow.

Across the six affected modules the failure set is identical to `upstream/main`: 37 pre-existing failures before and after, none new, with 869 passing against 856 on the baseline. The 22 collection errors and 37 failures are all missing optional dependencies in this environment (`chardet`, `sklearn` and similar) and are present unchanged on `main`.


KaifAhmad1 (approved):
Reviewed by checking out the branch into a worktree and running it, not just reading the diff. This is a larger PR (58 call sites across 6 modules), so I read every conversion rather than sampling.

**Verified the fix is real and complete**: removing the new `custom_methods.py` module entirely breaks test collection immediately, confirming there was no escape hatch in the old swallow-and-fallback pattern. All 21 new tests in `tests/utils/test_custom_method_can_refuse.py` pass on the branch.

**All 58 call sites checked by hand**, not sampled — every one pops `fallback_on_custom_error` from its own kwargs first, passes it through explicitly, and checks the sentinel by identity. This includes the two easiest sites to get wrong: the `**context`-named bag in `normalize/methods.py::disambiguate_entity`, and the one multi-line call site (`clean_data`) — both converted correctly. No site missed the pattern.

**Full test suite across all six affected modules + utils**: 1172 passed, 10 failed. Compared the 10 failures directly against `main` — identical failures, same errors (missing `defusedxml`/`pdfplumber`/etc. in this environment), none caused by this PR.

**Reproduced the second Qodo finding's exact scenario** (`pool_embeddings(method="mean", fallback_on_custom_error=True)` after a registered custom method raises) — confirmed it now falls back cleanly to real mean-pooling with no leaked-kwarg `TypeError`.

**On the first Qodo finding ("opt-in swallows custom exceptions")**: I read this independently and agree with the author's pushback, not the bot. `fallback_on_custom_error=True` swallowing the exception is the explicit, named contract of that flag — an opt-in escape hatch for callers who want the old behavior, not a silent failure (the warning names both the method and the exception). The default path with no flag now correctly propagates, which is the entire point of the PR. Resolving that thread as working-as-designed rather than a defect.

PR #1132, opened 2026-08-19
test(provenance): assert stored records in KG provenance suites (#946)

### LINKED ISSUE #946: test/provenance: strengthen test assertions and remove deprecated UTC datetime usage
## Description

During the review of PR #941, the affected `kg` test suite completed successfully with **473 passed, 0 failures**. The review also surfaced a few pre-existing maintenance issues unrelated to the GraphBuilder changes.

These issues do not block PR #941 and should be addressed separately to keep the current PR focused.

### 1. Tests returning values instead of asserting results

Several tests in:

* `tests/test_provenance_workflows.py`
* `tests/test_integration_comprehensive.py`

appear to return values such as:

```python
return "some string"
```

instead of making explicit assertions.

Pytest reports these as `PytestReturnNotNoneWarning`.

This weakens the test suite because executing a test without asserting its expected result can allow incorrect behavior to pass unnoticed.

**Expected action:**

* Identify all affected tests.
* Replace unintended return statements with meaningful assertions.
* Ensure each test actually validates the behavior it is intended to cover.
* Confirm the resulting tests fail appropriately when the expected behavior is broken.

### 2. Deprecated `datetime.utcnow()` usage in provenance code

The review also identified deprecated `datetime.utcnow()` usage in provenance-related code, including:

* `semantica/provenance/manager.py`
* `kg_provenance.py`

This currently produces deprecation warnings and may create future Python compatibility concerns.

**Expected action:**

* Identify all relevant `datetime.utcnow()` usages.
* Replace them with an appropriate timezone-aware UTC implementation.
* Preserve existing timestamp semantics, serialization behavior, and API compatibility.
* Update affected tests where necessary.
* Verify that the related deprecation warnings are eliminated.

## Scope

These issues were verified as **pre-existing and unrelated to PR #941**. They were discovered during the review and are being tracked separately rather than expanding the scope of the GraphBuilder PR.

## Acceptance Criteria

* [ ] Affected tests use explicit, meaningful assertions.
* [ ] `PytestReturnNotNoneWarning` is eliminated from the affected tests.
* [ ] Relevant provenance code no longer uses deprecated `datetime.utcnow()`.
* [ ] Existing UTC timestamp semantics are preserved.
* [ ] Relevant provenance and integration tests pass.
* [ ] No new warnings or regressions are introduced.

pravit-amp:
## Description

Two pre-existing maintenance defects in the provenance layer, surfaced during review of #941 and tracked separately to keep that PR focused.

**Deprecated `datetime.utcnow()`.** `ProvenanceManager`, `ProvenanceEntry`, `BridgeAxiom`, and `GraphBuilderWithProvenance` stamped UTC with `datetime.utcnow()`, which is deprecated and scheduled for removal. They now use `datetime.now(timezone.utc).isoformat()`, matching the form `PipelineWithProvenance` already uses (`pipeline/pipeline_provenance.py:79`).

**Tests that asserted on generated IDs rather than stored state.** The KG workflow and integration suites checked that a tracker call returned an ID matching a prefix — `assert cent_id.startswith('centrality_')` — without ever reading the record back. An ID generator that returned a well-formed string and wrote nothing would pass. Worse, some of those calls named tracker methods that do not exist anywhere in `semantica/` (`track_layer_analysis`, `track_centrality_score`), so the assertions were being satisfied without a real interaction behind them. These tests now read provenance back through `get_provenance()` and assert on algorithm metadata, and call the methods that actually persist records.

Several tests in the same files also ended in `return <value>` instead of asserting, which pytest reports as `PytestReturnNotNoneWarning` — 16 occurrences across the affected suites, now zero.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #946

## Changes Made

- `semantica/provenance/manager.py`, `semantica/provenance/schemas.py`, `semantica/provenance/bridge_axiom.py`, `semantica/kg/kg_provenance.py` — replace `datetime.utcnow()` with `datetime.now(timezone.utc)` for `timestamp`, `first_seen`, `last_updated`, `invalidated_at_time`, and the `activity_started_at_time` / `activity_ended_at_time` pair
- `tests/kg/test_provenance_workflows.py`, `tests/kg/test_provenance_workflows_simple.py`, `tests/kg/test_integration_comprehensive.py` — assert against records read back via `get_provenance()` instead of against generated ID strings; replace calls to the non-existent `track_layer_analysis` / `track_centrality_score` helpers with the methods that persist records; remove `return`-instead-of-assert endings
- `tests/provenance/test_manager.py` — new `TestTimezoneAwareUtcTimestamps` covering `track_entity` and `invalidate`: parses each stamp, asserts `tzinfo is not None` and a zero UTC offset, bounds it between two `datetime.now(timezone.utc)` reads, round-trips it through storage, and asserts no `utcnow` `DeprecationWarning` is emitted
- `docs/guides/provenance.md`, `docs/reference/provenance.md` — update the PROV-O mapping table and the `ProvenanceEntry` example to the new call and the offset-bearing timestamp format

## Testing

- [x] Tested locally

### Test Commands

```bash
pytest tests/provenance/ tests/kg/ -q
# 814 passed, 30 skipped, 0 failed
```

Affected suites specifically, against the parent commit as a baseline:

| | parent | this PR |
|---|---|---|
| passed | 131 | 134 |
| warnings | 1,639 | 4 |
| `PytestReturnNotNoneWarning` | 16 | 0 |

The four remaining warnings are pre-existing `utcnow` deprecations from outside this PR's scope — see Additional Notes.

The strengthened assertions were checked by mutation rather than by a green run alone: neutering the manager's storage write (`self.storage.store(...)` → no-op) fails 10 tests, and reverting the UTC change fails the two new timezone tests. `python -m build` produces `semantica-0.6.5.tar.gz` and `semantica-0.6.5-py3-none-any.whl`.

## Documentation

## Breaking Changes

**Breaking Changes**: No (behavior) / format-level change worth noting

The stored timestamp string gains a UTC offset suffix:

```
before: 2026-08-19T22:35:38.501697
after:  2026-08-19T22:35:38.501801+00:00
```

The instant recorded is unchanged and still UTC — only the serialized form differs. `datetime.fromisoformat()` parses both, and the one in-tree consumer that parses these fields (`change_management/change_log.py:73`) already normalizes offsets. The migration-relevant case is a store holding records written before and after this change: lexicographic sorting or string equality across the two formats will not behave, so compare parsed datetimes rather than raw strings. Both doc pages were updated to show the new form.

## Checklist

## Additional Notes

**The `utcnow` sweep is deliberately scoped to the modules the issue names, and is not exhaustive.** `datetime.utcnow()` remains in roughly 29 other modules, including 20+ sibling `*_provenance.py` wrappers (`embeddings`, `conflicts`, `vector_store`, `llms`, `ontology`, `reasoning`, `export`, and others) that follow the identical `activity_started_at_time` / `activity_ended_at_time` shape.

This is visible in CI: the 4 warnings remaining in the affected suites come from `embeddings_provenance.py:48,50` and `conflicts_provenance.py:47,62`, both reached through `tests/provenance/test_manager.py`. I left them out because converting them changes the serialized format for every subsystem at once, which is a wider compatibility question than this issue scoped — but I'm happy to fold the full sweep into this PR or open a follow-up, whichever maintainers prefer.

`CHANGELOG.md` may conflict — this branch is behind `main` and the conflict is limited to adjacent entries at the top of the Fixed section.


ZohaibHassan16 (approved):
Everything looks good.

One small thing before merge: could you rename the PR to something like:

`test(provenance): assert stored records in KG provenance suites (#946)`

PR #1143, opened 2026-08-20
fix(kg): compute real relationship duration for temporal stability metric

### LINKED ISSUE #1142: [BUG] Temporal-evolution stability metric is a hardcoded placeholder, not a duration
## Description

`TemporalGraphQuery.analyze_evolution()` documents the `stability` metric as a
"relationship duration/stability measure", but the implementation appends a
constant `1` for every relationship that has both `valid_from` and
`valid_until` set:

```python
if valid_from and valid_until:
    # Simplified duration calculation
    durations.append(1)  # Placeholder
result["stability"] = sum(durations) / len(durations) if durations else 0
```

## Impact

`stability` is therefore always `1.0` when any bounded relationship exists and
`0` otherwise. It never reflects how long relationships actually stay valid, so
it cannot distinguish a graph of decade-long relationships from one of
one-second relationships — the metric is effectively meaningless.

## Expected

`stability` should compute the mean valid-time duration (e.g. in seconds) across
relationships that have both bounds set.

## Location

`semantica/kg/temporal_query.py`, `analyze_evolution()`

I have a fix ready and would like to submit a PR.

cxzg007:
Closes #1142

## Problem

`TemporalGraphQuery.analyze_evolution()` documents the `stability` metric as a
relationship duration/stability measure, but the implementation appended a
constant `1` for every relationship with both bounds set
(`durations.append(1)  # Placeholder`). The result was always `1.0` when any
bounded relationship existed and `0` otherwise — it never reflected how long
relationships actually stayed valid.

## Fix

`stability` now computes the mean valid-time duration in seconds
(`(valid_until - valid_from).total_seconds()`) across relationships that have
both bounds set. Relationships with a missing or open `valid_from`/`valid_until`
are skipped (their duration is unbounded), and non-positive intervals are
clamped to `0`; an empty set still reports `0`. The docstring is updated to
describe this accurately.

## Tests

Adds 3 tests in `tests/kg/test_kg.py`:

- mean-duration result across two bounded relationships
- skipping of unbounded / half-open intervals
- empty-graph zero case

`tests/kg/test_kg.py` passes (39 passed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **The temporal-evolution `stability` metric was a hardcoded placeholder, not a duration**
+  - `TemporalGraphQuery.analyze_evolution()` documents `stability` as a "relationship duration/stability measure", but the implementation appended a constant `1` for every relationship with both `valid_from` and `valid_until` set (`durations.append(1)  # Placeholder`). The reported stability was therefore always `1.0` when any bounded relationship existed and `0` otherwise — it never reflected how long relationships actually stayed valid, so it could not distinguish a graph of decade-long relationships from one of one-second relationships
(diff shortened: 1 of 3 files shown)


ZohaibHassan16 (approved):
LGTM. Verified the math fix on main vs this branch. Tested edge cases around missing/inverted bounds and checked that downstream tests aren't broken by the updated return values. All tests passing.

Issue #1129, opened 2026-08-19
[BUG] OntologyIngestor drops every class in a JSON-LD named graph and reports success


13g4d0:
## Summary

`POST /api/ontology/load` returns `status: "success"` for a JSON-LD ontology whose terms live in a **named graph**, while loading none of them. The registry then reports `class_count: 0`, `property_count: 0`, and `/api/ontology/search` finds nothing.

A JSON-LD document with a top-level `@id` **and** `@graph` puts its `@graph` contents into a *named* graph. `rdflib.Graph.parse()` loads only the default graph and discards the rest — silently. `semantica/ingest/ontology_ingestor.py:109` uses `Graph()`.

This is the same `Graph` → `Dataset` migration that #756 identified and #757 merged for `JenaStore`. The ingest path was not covered by it.

## Reproduction

Twelve lines, two classes:

```json
{
  "@context": {"ex": "https://example.org/ns#", "owl": "http://www.w3.org/2002/07/owl#", "rdfs": "http://www.w3.org/2000/01/rdf-schema#"},
  "@id": "https://example.org/ns",
  "@type": "owl:Ontology",
  "@graph": [
    {"@id": "ex:Thing", "@type": "owl:Class", "rdfs:label": "Thing"},
    {"@id": "ex:Other", "@type": "owl:Class", "rdfs:label": "Other"}
  ]
}
```

```bash
curl -X POST http://127.0.0.1:8000/api/ontology/load \
  -H 'Content-Type: application/json' \
  -d "{\"content\": $(jq -Rs . minimal.jsonld), \"format\": \"json-ld\", \"name\": \"repro\"}"
```

Observed:

```
{"status":"success","uri":"https://example.org/ns","nodes_added":1,"edges_added":1,"format":"json-ld"}
```

```bash
curl -s http://127.0.0.1:8000/api/ontology/registry
# ... "uri": "https://example.org/ns", "class_count": 0, "property_count": 0, "concept_count": 0
```

Expected: `class_count: 2`, or a failure that says the document could not be read.

## Why the parser sees nothing

```python
from rdflib import Graph, Dataset, RDF, OWL

g = Graph(); g.parse("minimal.jsonld", format="json-ld")
len(g)                                  # 5   (document header only)
len(set(g.subjects(RDF.type, OWL.Class)))   # 0

ds = Dataset(); ds.parse("minimal.jsonld", format="json-ld")
len(ds)                                 # includes the named graph
```

[...]

PR #1145, opened 2026-08-20
Keep JSON-LD payloads in the default graph (#1144)

### LINKED ISSUE #1144: [BUG] Every JSON-LD export hides its payload in a named graph, so a plain Graph().parse() reads 2 of 21 statements
## Summary

Every JSON-LD document `JSONExporter` writes puts its payload in a **named graph**, so `rdflib.Graph.parse()` reads the document header and discards the data. Nothing reports a problem.

A JSON-LD document carrying a top-level `@id` *and* a top-level `@graph` is a named graph: the members of `@graph` become quads named by that `@id`, and the default graph is left holding only the statements written beside them. `semantica/export/json_exporter.py:415` stamps that `@id` on the same object that `json_exporter.py:405` and `json_exporter.py:408` filled with the payload.

This is the write side of #1129. Semantica produces exports that Semantica cannot read back.

## Reproduction

```python
from rdflib import Dataset, Graph
from semantica.export import JSONExporter

kg = {
    "entities": [
        {"id": "https://example.org/e1", "text": "Acme Corp", "type": "ORG"},
        {"id": "https://example.org/e2", "text": "Jane Roe", "type": "PERSON"},
    ],
    "relationships": [
        {"source_id": "https://example.org/e1",
         "target_id": "https://example.org/e2", "type": "employs"}
    ],
    "metadata": {"source_document": "contract.pdf"},
}

JSONExporter().export_knowledge_graph(kg, "kg.jsonld", format="json-ld")

g = Graph(); g.parse("kg.jsonld", format="json-ld")
print(len(g))                                    # 2

ds = Dataset(); ds.parse("kg.jsonld", format="json-ld")
print(sum(1 for _ in ds.quads((None, None, None, None))))   # 21
```

Observed on v0.6.6: **2 of 21 statements survive a default-graph read.** The 19 that vanish are the entire knowledge graph, entities, relationships, types, confidences and all. The two that survive are `semantica:exportedAt` and `semantica:format`, which describe the export rather than the data.

Where they went:

```
urn:x-rdflib:default                                          2 triples
https://semantica.dev/data/2026-08-20T09:07:24.038997+00:00  19 triples
```

The name of that graph is the wall clock at export time, so it is not stably addressable either: a `GRAPH <...>` query needs an IRI that no consumer can predict, and two exports of the same input produce two disjoint named graphs.

## Two separate triggers

**1. Any list or generic-dict export.** `_convert_to_jsonld` puts the payload in `@graph` and then adds a document `@id` beside it:

```python
jsonld["@graph"] = data              # json_exporter.py:408
...
if include_metadata:
    jsonld["@id"] = f"https://semantica.dev/data/{utc_now_iso()}"   # :415
```

```python
JSONExporter().export(
    [{"@id": "https://example.org/a",
      "http://www.w3.org/2000/01/rdf-schema#label": "A"}],
    "list.jsonld", format="json-ld")
# Graph(): 2 triples, Dataset(): 4 quads
```

**2. `export_knowledge_graph`, which converts twice.** `json_exporter.py:230` converts the graph to JSON-LD, then passes the finished document to `export()`, which converts it again at `json_exporter.py:155`. The converted document no longer has `entities` or `relationships` keys, so the second pass falls through to the generic branch and wraps it in `@graph`. The result carries two `@context` blocks, two document nodes minted a few hundred microseconds apart, and the knowledge graph one level down inside a named graph:

```json
{
  "@context": {...},
  "@graph": [
    { "@context": {...},
      "@id": "https://semantica.dev/graph/2026-08-20T09:07:24.038864+00:00",
      "@type": "semantica:KnowledgeGraph",
      "semantica:entities": [...] }
  ],
  "@id": "https://semantica.dev/data/2026-08-20T09:07:24.038997+00:00",
  "semantica:exportedAt": "...",
  "semantica:format": "json-ld"
}
```

`RDFExporter.serialize_to_jsonld` is not affected: it builds `@context` plus `@graph` and never adds a top-level `@id`.

## Why this is quiet

Nothing on the path can see it. The file is valid JSON-LD, it round-trips through a JSON-LD processor, and a quad-aware reader finds everything. Only a default-graph reader loses the data, and that is the common case: `Graph().parse(...)`, most SPARQL loaders, and any query without a `GRAPH` clause. It is the same failure mode as #1129, from the opposite direction.

## Expected

An export whose payload a plain `Graph().parse()` can read in full. The export's own metadata is worth keeping; it just has to be one more node in the default graph rather than a name wrapped around everything else.

## Fix

PR opens shortly. `_attach_document_metadata` puts the document node inside `@graph` when the payload lives there, and makes it the document itself otherwise, so no export names its own graph by accident. An already-converted document is merged rather than nested, which also removes the duplicate `@context` and the second document node.

Nine tests, all failing before the change: a parametrised check that no export emits a top-level `@id` beside `@graph`, a check that `Graph()` and `Dataset()` return the same count for every payload shape, and a check that the entities and the relationship are actually there after a plain parse. Full suite failure sets are identical before and after (539 pre-existing, mostly missing optional dependencies in this environment), so nothing else moves.

Two things this deliberately leaves alone, both worth their own issue:

* `@vocab` is `https://semantica.dev/vocab/`, a second namespace that does not resolve and is not the one the package ships. Every entity type and every user metadata key lands in it: `ORG` becomes `https://semantica.dev/vocab/ORG`.
* The document IRI is minted from the wall clock, so re-exporting identical input produces a different subject each time.

fabio-rovai:
Fixes #1144.

`JSONExporter` wrote every JSON-LD document as a **named graph**, so a plain `rdflib.Graph.parse()` read the document header and discarded the data. On v0.6.6 a two-entity, one-relationship knowledge graph parsed as **2 triples** with `Graph()` and 21 quads with `Dataset()`. The 19 that vanished were the entire graph. Nothing reported a problem: the file is valid JSON-LD and a quad-aware reader finds everything.

This is the write side of #1129. Semantica produced exports that Semantica could not read back.

## Cause

A top-level `@id` beside a top-level `@graph` names the graph. `_convert_to_jsonld` put the payload in `@graph` and then stamped the document `@id` next to it.

`export_knowledge_graph` compounded it. It converted the graph to JSON-LD and handed the finished document to `export()`, which converted it a second time. A converted document no longer carries `entities`/`relationships` keys, so the second pass fell through to the generic branch and wrapped the whole knowledge graph in `@graph`, producing two `@context` blocks and two document nodes minted microseconds apart.

## Change

`_attach_document_metadata` decides where the export's own metadata goes. When the payload lives in `@graph` the document node goes in beside it as one more node; otherwise the document node is the document itself. Either way nothing names the graph, and the metadata is still there.

`_is_jsonld_document` stops the second conversion: a dictionary that already declares a `@context` is merged rather than nested. That removes the duplicate context and the second document node as well.

An `@id` the payload already carries is not overwritten, so the knowledge-graph conversion keeps naming its own document node.

`RDFExporter.serialize_to_jsonld` needed no change: it never added a top-level `@id`.

## Result

Same graph, same export call:

| | before | after |
|---|---|---|
| `Graph().parse()` | 2 triples | 20 triples |
| statements outside the default graph | 19 | 0 |
| `@context` blocks | 2 | 1 |
| document nodes | 2 | 1 |

## Tests

`tests/export/test_jsonld_default_graph.py`, 9 tests, 7 failing before the change and the other 2 guarding the one path that was already correct:

* no export emits a top-level `@id` beside `@graph`, parametrised over a knowledge graph, a list and a generic dict;
* `Graph()` and `Dataset()` return the same count for each of those, which is the assertion that would have caught this;
* the entities, their labels and the relationship are present after a plain parse;
* no nested `@context`, the signature of the double conversion;
* the export's own provenance still reaches the default graph.

The counts come from parsing, never from the serialized text. As text the old output looked entirely reasonable, which is why it survived this long.

Full suite failure sets are identical before and after: 539 pre-existing failures on `main` and 539 on this branch in the same environment, most of them missing optional dependencies. No new failure, nothing fixed by accident.

## Not in this PR

Both were visible while probing and both deserve their own issue rather than being folded in here:

* `@vocab` is `https://semantica.dev/vocab/`, which returns 404 and is not the namespace the package ships at `ns#`. Every entity type and every user metadata key is swallowed into it: `{"type": "ORG"}` becomes `https://semantica.dev/vocab/ORG`.
* The document IRI is minted from the wall clock, so exporting identical input twice produces two different subjects.

Happy to take either next, or to split this one further if you would rather review the two triggers separately.

### DIFF
--- a/semantica/export/json_exporter.py
+++ b/semantica/export/json_exporter.py
@@ -416,7 +416,13 @@ class JSONExporter:
 
         # Convert data based on type
         if isinstance(data, dict):
-            if _is_jsonld_document(data):
+            # A knowledge graph is converted even when it carries a context of
+            # its own: the specialized conversion is what mints entity ids and
+            # relationship endpoints, and skipping it leaves them raw keys.
+            if "entities" in data or "relationships" in data:
+                # Knowledge graph structure - use specialized conversion
+                jsonld.update(self._convert_kg_to_jsonld(data, **options))
+            elif _is_jsonld_document(data):
                 # Already JSON-LD: merge it rather than nesting it. Wrapping a
                 # converted document in @graph re-typed the payload as a named
                 # graph and doubled the @context, which is what happened when
@@ -424,10 +430,15 @@ class JSONExporter:
                 context = data.get("@context")
                 if isinstance(context, dict):
                     jsonld["@context"].update(context)
+                elif context is not None:
+                    # A context may also be a URL or an array of them, which
+                    # cannot be merged key by key. Keeping both as an array
+                    # preserves the caller's term expansion, which wins over
+                    # ours, while still defining the semantica prefix. An
+                    # explicit null is left alone: in an array it would reset
+                    # the active context and take our own terms with it.
+                    jsonld["@context"] = [jsonld["@context"], context]
                 jsonld.update({k: v for k, v in data.items() if k != "@context"})
-            elif "entities" in data or "relationships" in data:
-                # Knowledge graph structure - use specialized conversion
-                jsonld.update(self._convert_kg_to_jsonld(data, **options))
             else:
                 # Generic dictionary - wrap in @graph
                 jsonld["@graph"] = [data]
@@ -463,20 +474,31 @@ class JSONExporter:
             jsonld: Document being built, modified in place
             include_provenance: Whether to record how and when it was exported
         """
+        # A caller may hand us a document that is deliberately a named graph.
+        # That name is theirs to keep, but our own statements must not end up
+        # inside it, where a default-graph reader would never see them.
+        payload_is_named_graph = "@id" in jsonld and "@graph" in jsonld
+
         document: Dict[str, Any] = {}
         # Do not overwrite an identifier the payload already carries: the
         # knowledge-graph conversion names its own document node.
-        if "@id" not in jsonld:
+        if "@id" not in jsonld or payload_is_named_graph:
             document["@id"] = f"https://semantica.dev/data/{utc_now_iso()}"
         if include_provenance:
             document["semantica:exportedAt"] = utc_now_iso()
             document["semantica:format"] = "json-ld"
 
-        if "@graph" in jsonld:
-            document.setdefault(
-                "@id", f"https://semantica.dev/data/{utc_now_iso()}"
-            )
-            jsonld["@graph"] = list(jsonld["@graph"]) + [document]
+        if payload_is_named_graph:
+            named = {key: value for key, value in jsonld.items() if key != "@context"}
+            for key in [key for key in jsonld if key != "@context"]:
+                del jsonld[key]
+            jsonld["@graph"] = [named, document]
+        elif "@graph" in jsonld:
+            # @graph may be a single node object as well as an array. list() on
+            # a dictionary yields its keys, which would discard the node.
+            members = jsonld["@graph"]
+            members = list(members) if isinstance(members, list) else [members]
+            jsonld["@graph"] = members + [document]
         else:
             jsonld.update(document)
 
--- a/semantica/export/json_exporter.py
+++ b/semantica/export/json_exporter.py
@@ -34,6 +34,25 @@ from ..utils.progress_tracker import get_progress_tracker
 from .rdf_exporter import SEMANTICA_NS, mint_entity_iri, mint_relationship_iri
 
 
+def _is_jsonld_document(data: Dict[str, Any]) -> bool:
+    """
+    Report whether a dictionary is already a JSON-LD document.
+
+    ``export_knowledge_graph`` converts a knowledge graph to JSON-LD and then
+    hands the finished document to ``export()``, which converted it a second
+    time. The converted document no longer carries ``entities``/
+    ``relationships`` keys, so the second pass treated it as an opaque value and
+    buried it inside ``@graph``.
+
+    Args:
+        data: Dictionary to test
+
+    Returns:
+        True when the dictionary declares a JSON-LD context
+    """
+    return "@context" in data
+
+
 class JSONExporter:
     """
     JSON exporter for knowledge graphs and semantic data.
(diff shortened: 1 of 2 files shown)


KaifAhmad1 (approved):
Reviewed by checking out the branch into a worktree and running it, not just reading the diff. This branch was BEHIND main, so I merged origin/main into it locally first — clean merge, no conflicts.

**Verified the regression is real**: reverted `json_exporter.py` to the parent commit (`78fc9028`) — 10/13 new tests fail, confirming a two-entity knowledge graph really did split into 2 default-graph triples plus 19 triples hidden inside a wall-clock-named graph, with nothing reporting a problem.

**Verified the fix**: on the merged branch, all 13 new tests in `tests/export/test_jsonld_default_graph.py` pass, and the full `tests/export/` suite passes clean: 217 passed, 46 subtests, no regressions. All 4 Qodo review threads are already resolved.

**Traced `_attach_document_metadata`'s four cases by hand and verified each empirically** (triples == quads in every case, confirmed via both `Graph()` and `Dataset()` parses):
- plain list/dict export → document node appended as one more `@graph` member, no top-level `@id` reintroduced;
- knowledge graph → the KG conversion's own `@id` is preserved, provenance merged directly onto that single node (confirmed `_convert_kg_to_jsonld` never uses `@graph`, so there's no named-graph risk on this path even before the fix touches it);
- the double-conversion case (`export_knowledge_graph` handing its output back through `export()`) → single `@context`, single document node, no duplication;
- a caller-supplied deliberate named graph → correctly nested one level down inside the outer default graph, with our own provenance as a sibling node — verified `Dataset()` still finds the caller's graph name and `Graph()` finds our provenance.

Also hand-probed `include_metadata=False` and `include_provenance=False` (not in the test suite) — both stay fully default-graph-readable, no crash.

**One thing worth flagging, not a defect in this PR**: `export_entities`/`export_relationships` build their own custom `@context` (with a `@vocab` and an `entities: {@container: @list}` mapping) and hand it to `export()`, but the pre-existing `entities`-key check fires before the "already JSON-LD" branch this PR adds, so that custom context gets silently replaced by the KG-conversion's own context. Verified this live. It predates this PR entirely — the old code never even reached the named-graph bug on this path since KG-conversion output has no `@graph` key — so it's out of scope here, but worth its own issue if you want it tracked.

PR #1148, opened 2026-08-20
fix(export): escape Turtle/N-Triples string literals (closes #1098)

### LINKED ISSUE #1097: [BUG] convert_kg_to_rdf is never called: every entity exports as semantica:text "" on all RDF paths

### LINKED ISSUE #1098: [BUG] serialize_to_turtle interpolates literals unescaped: a quote or newline in entity text emits invalid Turtle

pkupt:
## Summary

Fixes #1098: `serialize_to_turtle()` (and the N-Triples path) interpolate
entity `text` into double-quoted string literals **unescaped**. A `text`
value containing a `"`, backslash, newline, carriage return, or tab emits
invalid Turtle/N-Triples that other parsers reject.

## Fix

- Add `RDFSerializer._escape_turtle_literal()` — escapes backslash first,
  then double quote, newline, CR, tab (per the RDF 1.1 Turtle
  `STRING_LITERAL_QUOTE` grammar).
- Apply it to the `semantica:text` literal in `serialize_to_turtle()`.
- Reuse it for the N-Triples `semantica:text` triple, which previously
  escaped only quotes and newlines (backslashes and tabs leaked through).

Scoped strictly to literal escaping, per maintainer guidance to cover
#1098 separately from the `name`->`label` normalization (#1113, which
closes #1097). No overlap: this PR does not touch `convert_kg_to_rdf`.

## Tests

`tests/export/test_rdf_literal_escaping.py`:
- Turtle + N-Triples: quote, backslash, newline, tab each escaped
- Plain text unchanged (no over-escaping)

`tests/export/` -> 161 passed, 1 skipped.

Closes #1098


ZohaibHassan16 (approved):
LGTM

PR #1151, opened 2026-08-20
fix(mcp): export_graph failed on every format — convert kg dict, disable progress

Arasz:
## Problem

The MCP server's `export_graph` tool is broken on **every** format in 0.6.5 and 0.6.6:

1. **`json` branch**: `_tool_export_graph` calls `JSONExporter().export(graph)`, but that method requires a `file_path` argument → `TypeError` surfaced as `{"error": "JSONExporter.export() missing 1 required positional argument: 'file_path'"}`.
2. **RDF branches** (`json-ld`, `turtle`, `ttl`, `nt`, `xml`): `RDFExporter().export_to_rdf(graph, ...)` receives the `ContextGraph` **object** instead of the canonical kg dict → `AttributeError: 'ContextGraph' object has no attribute 'get'`.
3. **All branches**: the RDF path prints a rich progress bar to **stdout** (`🔄 Semantica is exporting: ...`). Over stdio MCP, stdout *is* the JSON-RPC framing, so the interleaved progress corrupts the protocol and the client hangs. Observed live: 300s MCP timeout on an empty graph while the same call returns in <1s directly.

Net effect: no agent can export the graph, so downstream persistence bridges (e.g. ai-badger's `.semantica/` autosave + AiRaccoon watch) never receive data.

## Fix

- Convert the graph with `ContextGraph.to_kg_dict()` before handing it to the exporters (available since 0.6.6).
- Serialize the `json` branch to a string via `json.dumps` (the tool's contract returns `{"format": ..., "data": <string>}` like the RDF branches).
- Force `SEMANTICA_DISABLE_PROGRESS=1` for the server process — stdout is the protocol channel, not a console. (The env var already exists and is honored by the progress tracker.)

## Tests

`tests/test_mcp_server_export_graph.py` (unittest, matching the repo's MCP test style):
- json branch returns parseable `entities`/`relationships` payload, no error
- every RDF format returns a non-empty string, no error
- the server process has progress disabled

All 5 pass; verified the failures against 0.6.6 before the fix.

### DIFF
--- a/mcp/__init__.py
+++ b/mcp/__init__.py
@@ -21,6 +21,17 @@ Configure in Claude Desktop, Windsurf, Cline, Continue, VS Code:
     }
 """
 
+import os
+
+# MCP stdio framing IS stdout: any progress bar or console renderer that writes
+# to stdout would interleave with the JSON-RPC stream and corrupt framing for
+# every client.  This package is always used as an MCP stdio server, so force
+# progress tracking off for the entire process.  Set before importing server /
+# tools so the Semantica progress-tracker singleton is never created with
+# output enabled (the singleton reads this variable at construction time and
+# the enabled.setter re-checks it, so later re-enable attempts are also blocked).
+os.environ["SEMANTICA_DISABLE_PROGRESS"] = "1"
+
 # `semantica.__version__` is the authoritative package version — see
 # semantica/mcp_server/__init__.py for why it is used directly rather than
 # importlib.metadata.version("semantica").
--- a/mcp/tools/export.py
+++ b/mcp/tools/export.py
@@ -80,7 +80,12 @@ def handle_export_graph(args: dict) -> dict:
         if rdf_fmt:
             try:
                 from semantica.export import RDFExporter
-                rdf_str = RDFExporter().export_to_rdf(graph, format=rdf_fmt)
+                # RDFExporter.export_to_rdf() expects the canonical kg dict
+                # {"entities": [...], "relationships": [...]}, not a ContextGraph
+                # object.  Convert before handing off; passing the raw graph
+                # caused AttributeError: 'ContextGraph' object has no attribute
+                # 'get' on every RDF format.
+                rdf_str = RDFExporter().export_to_rdf(graph.to_kg_dict(), format=rdf_fmt)
                 return {"format": rdf_fmt, "data": rdf_str}
             except Exception as exc:
                 return {"error": f"RDF export failed: {exc}"}
--- a/semantica/mcp_server/__init__.py
+++ b/semantica/mcp_server/__init__.py
@@ -267,9 +267,16 @@ def _tool_get_graph_analytics(args: dict) -> dict:
         return {"error": str(exc)}
 
 
+_EXPORT_GRAPH_FORMATS = ("turtle", "ttl", "nt", "xml", "json-ld", "json")
+
+
 def _tool_export_graph(args: dict) -> dict:
     """Export the current knowledge graph to a serialised format."""
     fmt = args.get("format", "json-ld")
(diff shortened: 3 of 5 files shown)


Sameer6305 (dismissed):
Thanks, @Arasz, for the solid initial implementation. The core diagnosis was correct, and the fixes for the `semantica/mcp_server` path were well targeted.

### What was solid

* Correctly identified the JSON exporter `file_path` contract issue.
* Correctly converted `ContextGraph` to the canonical KG dict before RDF export.
* Correctly addressed MCP stdio progress output on stdout.
* Added focused regression tests for the affected formats and behavior.

### What I found missing during review

During the deeper review, I found that the repository has a **second active MCP implementation under `mcp/`** with the same RDF `ContextGraph` → dict bug and the same stdout/progress risk. These were not covered by the original PR changes.

I also verified the newer `progress_tracker.py` changes and checked for regressions and additional `export_graph` paths.

### What I fixed

I addressed the missing standalone `mcp/` issues in commit:

`9b30c8af — fix(mcp): repair standalone export_graph`

This:

* converts `ContextGraph` with `to_kg_dict()` in `mcp/tools/export.py`
* disables Semantica progress for the standalone MCP server
* adds regression coverage for the standalone MCP export and stdout behavior

### Validation

The final targeted review found:

* **42/42 directly relevant MCP/progress tests passing**
* **892 tests passed** in the broader targeted export/context run
* **0 PR-caused test failures**
* The single remaining failure is the pre-existing `pyoxigraph` optional-dependency issue on `main`.

I also verified both active `export_graph` implementations and found no remaining blocking issues.

**@KaifAhmad1 I approve this from my side. The PR is ready for your final review and merge.**

KaifAhmad1 (approved):
## Final review

Thanks @Sameer6305 for catching the standalone `mcp/` package — that was a real gap in the original PR and your fix there (`9b30c8a`) is correct: `graph.to_kg_dict()` conversion and progress-disable are both right, and `mcp/tools/export.py` already had proper unsupported-format handling even before your change.

One thing slipped through the second review round: Qodo's original finding #1 ("unsupported formats return JSON") was about `semantica/mcp_server/__init__.py` specifically, and it was never actually fixed — only replicated correctly into the new `mcp/` package. I reproduced it live against the PR branch: `_tool_export_graph({"format": "yaml"})` returned `{"format": "yaml", "data": "{...JSON...}"}` with no error, because this server never validates tool-call args against its own declared `inputSchema` enum before dispatch.

Pushed `e2fc76c` to close that:
- `_tool_export_graph` now validates `format` against an explicit `_EXPORT_GRAPH_FORMATS` tuple up front and returns a proper error for anything outside it, instead of falling through to `json.dumps` mislabeled with the requested format.
- The tool's `inputSchema` enum now reuses that same constant so the two can't drift apart again.
- 2 new regression tests in `tests/test_mcp_server_export_graph.py` (verified red before the fix, green after): one for the `format="yaml"` case, one asserting the schema enum matches the handler's accepted formats.

Findings #2 (env var unconditionally overwritten at import) and #3 (a pre-existing progress-tracker singleton isn't retroactively disabled) I'm leaving as-is — I traced both and they're non-blocking: #2 looks like a deliberate protocol-safety call (an MCP stdio process should never emit progress to stdout regardless of any other setting) rather than a bug, and #3 doesn't affect the primary `python -m mcp.server` / `python -m semantica.mcp_server` entry points given `semantica`'s lazy top-level imports — it only matters if a host embeds these modules after already using Semantica directly in the same process, which is a narrow enough case not to hold up this fix.

All 10 CI checks pass on the latest commit, and the full `tests/test_mcp_server_export_graph.py` suite (7 tests) passes locally. Approving — the original bug report (export_graph broken on every format) is fully fixed across both MCP implementations.

PR #1153, opened 2026-08-20
fix(context): make auto_generate_id a real InitVar in decision models

### LINKED ISSUE #1152: [BUG]Dead code: auto_generate_id validation branch is unreachable in decision_models dataclasses
## Description

In `semantica/context/decision_models.py`, all six dataclasses (`Decision`, `DecisionContext`, `Policy`, `PolicyException`, `Precedent`, `ApprovalChain`) declare:

```python
def __post_init__(self, auto_generate_id: bool = True):
```

However, `auto_generate_id` is neither a dataclass field nor an `InitVar`, so the dataclass-generated `__init__` calls `__post_init__()` with **no arguments**. The parameter therefore always falls back to its `True` default, which makes the `auto_generate_id=False` validation branch permanently unreachable dead code:

```python
elif not self.decision_id and not auto_generate_id:  # never reached
    raise ValueError("decision_id is required when auto_generate_id=False")
```

## Impact

The intended contract — "require a caller-supplied id when auto-generation is disabled" — can never be enforced. Any caller that tries to opt out via `auto_generate_id=False` instead hits:

```
TypeError: __init__() got an unexpected keyword argument 'auto_generate_id'
```

## Root cause

Leftover from an incomplete cleanup (see #328, which removed `auto_generate_id=False` from the construction call sites but left behind the orphaned `__post_init__` parameter and its dead validation branch).

## Proposed fix

Declare `auto_generate_id: InitVar[bool] = True` on each of the six dataclasses so the generated `__init__` forwards it to `__post_init__`, restoring the required-id contract. Serialization is unaffected (`InitVar` is not a real field, so `fields()`, `to_dict()` and `from_dict()` all stay unchanged). Add regression tests covering both the auto-generate and the required-id paths.

I'd like to work on this.

cxzg007:
## Summary

The six decision-model dataclasses in `semantica/context/decision_models.py` (`Decision`, `DecisionContext`, `Policy`, `PolicyException`, `Precedent`, `ApprovalChain`) declared `auto_generate_id` only as a plain `__post_init__` parameter. Because it was neither a dataclass field nor an `InitVar`, the generated `__init__` never forwarded it, so it was always its `True` default and the `auto_generate_id=False` validation branch was **unreachable dead code**.

This PR declares `auto_generate_id: InitVar[bool] = True` on each dataclass so the generated `__init__` forwards it to `__post_init__`, restoring the intended "require a caller-supplied id when auto-generation is disabled" contract.

## Why this is safe

- `InitVar` is not a real field, so `dataclasses.fields()`, `to_dict()` and `from_dict()` are all unchanged — serialization behavior is identical.
- Default remains `True`, so existing callers that rely on auto-generated ids are unaffected.
- No production call site passes `auto_generate_id` (verified by grep); the previous cleanup in #328 already removed those call sites.

## Changes

- `semantica/context/decision_models.py`: add `auto_generate_id: InitVar[bool] = True` to the six dataclasses and update each `__post_init__` signature to receive it.
- `tests/context/test_decision_models.py`: add `TestAutoGenerateIdContract` covering:
  - `auto_generate_id` is not exposed as a field / instance attribute
  - default path auto-generates an id
  - `auto_generate_id=False` with empty id raises `ValueError`
  - explicit id is honored when `auto_generate_id=False`

## Testing

- `pytest tests/context/test_decision_models.py` → 38 passed
- Downstream suites (decision recorder / query / policy engine / to_kg_dict) → 108 passed, no regressions

Fixes #1152


Sameer6305 (approved):
Thanks for the clean and focused fix! 🙌

I reviewed the implementation and tests in detail, including all six decision-model dataclasses, `InitVar` behavior, the restored `auto_generate_id=False` validation path, serialization/deserialization, backward compatibility, and relevant existing test coverage.

Everything looks solid:
- `auto_generate_id` is correctly implemented as an `InitVar`.
- The previously unreachable validation branch is now correctly reachable.
- Existing default behavior remains unchanged.
- Serialization behavior is preserved as expected.
- The added regression tests cover the important contract and edge cases.
- Relevant test suites pass with no regressions.
- I also checked for potential bugs, unintended behavior, and problematic/infinite-loop scenarios and found none.

**Approved ✅**

Great work keeping the change minimal and focused. This looks ready to merge.

Issue #756, opened 2026-07-19
Add named-graph support to JenaStore (requires Graph -> ConjunctiveGraph/Dataset migration)


Sameer6305:
### Summary
Currently, `JenaStore` wraps a single `rdflib.Graph` instance to handle its underlying state and connection. Because `rdflib.Graph` fundamentally represents a single (default) graph and only accepts 3-tuples, it has no concept of named graphs. As a result, the `JenaStore.add_triplets` method currently ignores any `graph=` option silently, continuing to write all triples to the default graph.

### Architectural Finding
Implementing named-graph support for Jena is not just a matter of passing a query parameter (as it was for RDF4J). It requires a structural migration of the core wrapper:
* `self.graph` must be upgraded from `rdflib.Graph` to `rdflib.ConjunctiveGraph` or `rdflib.Dataset` (which natively support `(s, p, o, graph_uri)` 4-tuples).
* This migration affects `_initialize_graph()` and specifically how the remote `rdflib.plugins.stores.sparqlstore.SPARQLStore` is bound, configured, and instantiated.
* `add_triplets` will then need to be updated to detect the `graph=` option and append the 4th tuple element before calling `self.graph.add()`.

### Scoping
Because this requires a structural change to how `JenaStore` initializes and binds to `rdflib`—carrying potential ripple effects for existing `JenaStore` operations—it is a significantly bigger change than the native CONSTRUCT support introduced in PR #754. 

To maintain strict stability and avoid scope creep, this migration is being tracked as its own dedicated body of work requiring a proper scoping discussion before implementation. This follows the same disciplined pattern we used when splitting Blazegraph CONSTRUCT support (#322) from the RDF4J/Jena CONSTRUCT implementations (#754).

*(Note: #754 implements CONSTRUCT support for Jena and RDF4J, as well as named-graph support for RDF4J. This issue tracks the final outstanding piece—named-graph support for Jena—and will fully close out the cross-backend parity effort for named graphs once implemented.)*

Sameer6305:
@KaifAhmad1 is this issue good to go for implementation?
Some architectural decisions are present in this issue.

KaifAhmad1:
> [@KaifAhmad1](https://github.com/KaifAhmad1) is this issue good to go for implementation? Some architectural decisions are present in this issue.

[...]

PR #1156, opened 2026-08-20
Read JSON-LD named graphs in OntologyIngestor (#1129)

### LINKED ISSUE #1129: [BUG] OntologyIngestor drops every class in a JSON-LD named graph and reports success

13g4d0:
Fixes #1129.

## What was happening

A JSON-LD document with a top-level `@id` **and** `@graph` places its terms in a named graph. `rdflib.Graph.parse()` loads only the default graph and discards the rest — without raising — so every class and property in such a document was dropped while `POST /api/ontology/load` returned `status: "success"` and the registry showed `class_count: 0`.

## The change

`OntologyIngestor.ingest_ontology` parses into a `Dataset` and flattens the quads into the working `Graph`, so both the default and the named graphs survive. Everything downstream (`_convert_to_dict`, the fallback format guessing) is untouched.

This is the same `Graph` → `Dataset` migration #757 made for `JenaStore` (#756). The ingest path was not covered by it.

## Measured

On the 12-line reproduction from the issue:

```
before   classes=0  properties=0
after    classes=2  properties=0
```

On the real ontology that surfaced this: 25 triples / 1 subject with `Graph`, **719 triples / 45 classes / 40 object properties** with `Dataset`.

## Tests

`tests/ingest/test_ontology_named_graph.py`:

- the named-graph document yields its two classes and one object property;
- a **canary** on the default-graph document, so the fix cannot trade one blind spot for another;
- the reported result matches the terms actually returned.

Reverting `Dataset()` to `Graph()` turns all four red — the tests fail for the reason they claim to guard.

`tests/ontology`, `tests/export` and `tests/ingest` pass apart from six failures in web/feed/database/API ingestion. Those are unrelated to this change and fail the same way on an unmodified checkout of `main` in this environment (no outbound network).

## Not included

Making a load that yields **zero classes** stop returning `status: "success"`. That success value is what made this take an afternoon to find, and I offered it in the issue — but it is a behaviour change on a different layer, and reviewing it separately seemed better than bundling. Happy to add it here or open a follow-up, whichever you prefer.

## Possibly the same blind spot elsewhere

`semantica/explorer/utils/rdf_parser.py:87` and `semantica/explorer/routes/sparql.py:125` construct a plain `Graph()` on the same kind of input. I have not reproduced a failure on either and did not touch them.

---

@cxzg007 — thank you for offering to take this and then handing it back; the plan you posted was the same one, down to the regression test.

### DIFF
--- a/semantica/ingest/ontology_ingestor.py
+++ b/semantica/ingest/ontology_ingestor.py
@@ -40,7 +40,7 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional, Union
 
 import rdflib
-from rdflib import RDF, RDFS, OWL, Graph
+from rdflib import RDF, RDFS, OWL, Dataset, Graph
 
 from ..utils.exceptions import ProcessingError, ValidationError
 from ..utils.logging import get_logger
@@ -106,15 +106,21 @@ class OntologyIngestor:
                 raise ValidationError(f"File not found: {file_path}")
 
             self.progress.update_tracking(tracking_id, message="Parsing RDF graph...")
-            g = Graph()
-            
+            # `Dataset`, not `Graph`: a JSON-LD document with a top-level `@id` *and*
+            # `@graph` places its terms in a NAMED graph. `Graph.parse()` loads only the
+            # default graph and discards the rest without an error, so every class and
+            # property in such a document was dropped while the load reported success.
+            # Parsing into a Dataset and flattening the quads keeps both. Same migration
+            # #757 made for JenaStore; the ingest path was not covered by it.
+            ds = Dataset()
+
             # Use provided format or let rdflib guess based on extension
             parse_kwargs = kwargs.copy()
             if format:
                 parse_kwargs['format'] = format
-            
+
             try:
-                g.parse(file_path, **parse_kwargs)
+                ds.parse(file_path, **parse_kwargs)
             except Exception as e:
                 # Fallback: try to guess format from extension if not provided and initial parse failed
                 if not format:
@@ -130,11 +136,15 @@ class OntologyIngestor:
                     guessed_fmt = fmt_map.get(ext)
                     if guessed_fmt:
                         self.logger.info(f"Retrying with guessed format: {guessed_fmt}")
-                        g.parse(file_path, format=guessed_fmt, **kwargs)
+                        ds.parse(file_path, format=guessed_fmt, **kwargs)
                     else:
                         raise e
                 else:
                     raise e
+
+            g = Graph()
+            for subject, predicate, obj, _context in ds.quads((None, None, None, None)):
+                g.add((subject, predicate, obj))
 
             self.progress.update_tracking(tracking_id, message="Converting to internal format...")
             
--- a/semantica/ingest/ontology_ingestor.py
+++ b/semantica/ingest/ontology_ingestor.py
@@ -110,9 +110,12 @@ class OntologyIngestor:
             # `@graph` places its terms in a NAMED graph. `Graph.parse()` loads only the
             # default graph and discards the rest without an error, so every class and
             # property in such a document was dropped while the load reported success.
-            # Parsing into a Dataset and flattening the quads keeps both. Same migration
-            # #757 made for JenaStore; the ingest path was not covered by it.
-            ds = Dataset()
+            # Same migration #757 made for JenaStore; the ingest path was not covered by it.
+            # `default_union=True` makes the Dataset itself present triples from every
+            # graph as one merged view (it is an rdflib.Graph subclass, so it satisfies
+            # _convert_to_dict()'s Graph-typed contract directly) instead of copying every
+            # quad into a second in-memory Graph.
+            ds = Dataset(default_union=True)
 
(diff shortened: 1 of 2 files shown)


KaifAhmad1 (approved):
## Final review

Verified independently rather than taking the description at face value:

- **Root cause and regression tests are real.** I manually reverted the `Dataset()` parse back to plain-`Graph()`-style behavior on this branch: all 4 tests in `tests/ingest/test_ontology_named_graph.py` go red for exactly the claimed reason (named-graph terms vanish). Restored the fix, they go green.
- **No regressions.** Ran `tests/ontology` + `tests/ingest` in full: 403 passed, 10 failed — and I confirmed the identical 10 failures (all network-dependent: web/feed/database/sitemap fetching, no outbound network in this sandbox) occur on an unmodified `main` too. Same set, same reasons, on both branches.
- **`_convert_to_dict` and the format-guessing fallback are genuinely untouched**, as claimed.

On Qodo's one finding (Dataset→Graph copy, tagged perf/remediation-recommended, not action-required): I pushed `d05ef9d0` closing it properly rather than leaving it as a documented tradeoff — `Dataset(default_union=True)` gives a Graph-compatible merged view for free (it's an `rdflib.Graph` subclass at runtime, confirmed with `isinstance`), so the manual `for quad in ds.quads(...): g.add(...)` copy loop is unnecessary. Same tests, same full suite, all still passing after the change — zero behavior difference, one less O(n) copy.

Also worth tracking, not blocking: the author's own callout that `semantica/explorer/utils/rdf_parser.py:87` does the same `Graph()`-on-external-file-parse pattern on user-uploaded RDF in the explorer/web UI. I looked — plausible, unverified, correctly left out of scope here. Worth a follow-up issue.

Approving. Thanks for a tightly-scoped fix with real regression coverage — this is what a good bug-fix PR looks like.

PR #1158, opened 2026-08-20
SHACL guide: warn that RDFS inference makes range checks unfalsifiable (#1130)

### LINKED ISSUE #1130: [DOCS] SHACL guide: rdfs:range + RDFS inference makes every sh:class range check unfalsifiable
## Summary

If a property declares `rdfs:range` and validation runs with RDFS entailment, a `sh:class` constraint on that property **can never fail**. RDFS is an *entailment* rule, not a constraint: everything in the object position of the property is inferred to be of the range class, so the inference manufactures exactly the type the shape checks for.

The SHACL guide (`docs/guides/shacl-validation.md`) does not mention this, and its **Common Pitfalls** section is where a reader would look. The failure is silent and in the dangerous direction: the report says `conforms: True` on data that does not conform.

## Reproduction

```python
from pyshacl import validate
from rdflib import Graph

data = Graph(); data.parse(data='''
@prefix ex: <https://example.org/ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
ex:contains rdfs:domain ex:Container ; rdfs:range ex:Item .
ex:box a ex:Container ; ex:contains ex:notAnItem .
ex:notAnItem a ex:Fish .
''', format='turtle')

shapes = Graph(); shapes.parse(data='''
@prefix ex: <https://example.org/ns#> .
@prefix sh: <http://www.w3.org/ns/shacl#> .
ex:ContainerShape a sh:NodeShape ; sh:targetClass ex:Container ;
    sh:property [ sh:path ex:contains ; sh:class ex:Item ] .
''', format='turtle')

for inf in ("none", "rdfs"):
    conforms, _, _ = validate(data, shacl_graph=shapes, inference=inf)
    print(inf, conforms)
```

```
none  False     <- correct: ex:notAnItem is a Fish, not an Item
rdfs  True      <- the entailment made it an Item
```

## Workarounds that do not work

Checking the *declared* type instead of the inferred class does not help, because the entailment materialises the type as a real triple:

```python
g = Graph(); g += data
validate(g, shacl_graph=shapes, inference="rdfs", inplace=True)
(EX.notAnItem, RDF.type, EX.Item) in g   # True
```

So `sh:node` over a shape requiring an explicit `rdf:type` is satisfied too. Under RDFS entailment the constraint is not expressible in SHACL at all — the only fixes are to validate without inference, or to not declare `rdfs:range` on properties you intend to constrain.

## Suggested doc change

A bullet in **Common Pitfalls**, roughly:

> **Running range checks under RDFS inference.** `rdfs:range` is an entailment rule, not a constraint: with `inference="rdfs"` every object of the property is inferred into the range class, so `sh:class` and `sh:node` range checks on it always pass. Validate those shapes with `inference="none"`, or omit `rdfs:range` for properties you want SHACL to police. The trade-off is that `sh:targetClass` stops reaching subclasses.

Worth noting because turning inference *on* reads like making validation stricter, and it does the opposite.

## Context

Found while running our own shapes against three real graphs (107, 465 and 601 nodes). A shape that had been in our repo since 13 Aug reported zero violations under `inference="rdfs"` and one violation, correctly, under `inference="none"` — with a violation injected by hand to test the instrument.

Environment: `pyshacl==0.40.1`, `rdflib==7.6.0`.

Happy to open the docs PR if you want it phrased differently.

13g4d0:
Closes #1130.

Adds one bullet to **Common Pitfalls** in `docs/guides/shacl-validation.md`.

## Why

`rdfs:range` is an *entailment* rule, not a constraint. With `inference="rdfs"`, every object of the property is inferred into the range class — so the inference manufactures exactly the type a `sh:class` range check looks for, and the check can never fail. The report says `conforms: True` on data that does not conform.

Eight lines of Turtle, from the issue:

```
inference="none"  ->  conforms=False    # correct: the object is a Fish, not an Item
inference="rdfs"  ->  conforms=True
```

Common Pitfalls seemed the right place because that is where a reader goes when a validation run came back green and they do not believe it.

## The part that is easy to get wrong

The note also records what does **not** work, because it looks like it should: checking the *declared* type with `sh:node` over a shape requiring an explicit `rdf:type`. The entailment materialises the type as a real triple —

```python
validate(g, shacl_graph=shapes, inference="rdfs", inplace=True)
(EX.notAnItem, RDF.type, EX.Item) in g   # True
```

— so nothing downstream can distinguish a declared type from an inferred one. Under RDFS entailment the constraint is not expressible in SHACL at all; the only fixes are to validate without inference, or to not declare `rdfs:range` on properties you intend to constrain.

I spent an afternoon writing a cleverer shape before measuring that. Seemed worth saving the next person the detour.

## How it surfaced

Running our own shapes against three real graphs (107, 465 and 601 nodes). A shape that had been in the repo for a week reported zero violations under `inference="rdfs"` and one violation, correctly, under `inference="none"` — with a violation injected by hand to check that the shape could fail at all.

Docs only; no code touched.


Sameer6305:
Closing this PR as superseded since the same content is already covered in another PR.

PR #1160, opened 2026-08-21
fix(context): raise instead of silently returning False when compliance check fails

### LINKED ISSUE #1159: [BUG] check_compliance() silently returns False on internal errors, masking failures as policy violations
## Bug Description

`PolicyEngine.check_compliance()` wraps its evaluation logic in a broad
`except Exception: ... return False`. As a result, an *internal failure* while
performing the check (e.g. malformed `policy.rules`, unexpected metadata) is
indistinguishable from a legitimate *"decision is non-compliant"* result. Both
surface to the caller as `False`.

In an audit/compliance context this is dangerous: evaluation bugs get silently
reported as policy violations, and a check that could not actually be performed
is recorded as a definitive non-compliant verdict.

## Steps to Reproduce

1. Call `check_compliance(decision, policy_id)` where the underlying
   `_evaluate_compliance` raises (for instance, malformed `policy.rules`).
2. Observe the method returns `False` instead of surfacing the error.

## Expected Behavior

A check that *cannot be performed* should raise (e.g. `ProcessingError`), while
`False` is reserved strictly for *"evaluated and found non-compliant"*. The
existing `ValueError` for a missing policy should be preserved.

## Actual Behavior

Any exception raised during evaluation is swallowed and reported as `False`
(non-compliant).

## Environment

- **Semantica Version**: `main` branch (`semantica/context/policy_engine.py`)

## Possible Solution

Raise `ProcessingError` from the `except` branch, keep the `ValueError`
re-raise, and document the semantic distinction in the docstring. I have a fix
plus a regression test ready and am willing to submit a PR.

cxzg007:
## Summary

Fixes #1159.

`PolicyEngine.check_compliance()` wrapped its evaluation in a broad
`except Exception: ... return False`. An internal failure during the check was
therefore indistinguishable from a legitimate "non-compliant" verdict — both
returned `False`. In audit/compliance scenarios this silently masks bugs as
policy violations.

## Changes

- The `except` branch now raises `ProcessingError` (with the original error
  chained via `from e`) instead of returning `False`.
- The existing `ValueError` re-raise (missing policy) is preserved.
- `False` now strictly means "evaluated and found non-compliant".
- Docstring updated to document the `Returns` / `Raises` distinction.
- Added regression test `test_check_compliance_raises_on_evaluation_error`.

## Testing

`pytest tests/context/test_policy_engine.py -q` → 44 passed.

## Notes

Diff is scoped to the fix only (+55/-2 across 2 files); no unrelated
formatting/import changes.


zeekay (dismissed):
I agree with the principle. "Evaluated and found non-compliant" and "could not determine compliance" are different facts, and collapsing them into one `False` is the wrong default for an audit surface — a `TypeError` from a malformed rule shouldn't read as a policy violation. Removing the swallow from `_check_compliance_with_rules` so the failure reaches `check_compliance` is the right shape.

Ran it merged onto current `main`: `pytest tests/context/ -m "not integration"` — 805 passed, 2 skipped, against 803 on `main`. Both new tests pass, and the malformed-rule one exercises the real path rather than mocking `_evaluate_compliance`, which is what makes it worth having.

Three things to settle before this lands.

**1. The docs now say something untrue.** `docs/guides/policy-engine.md:135` reads:

> `check_compliance` takes a `Decision` object and the policy ID and returns a boolean.

and eight call sites across `docs/guides/policy-engine.md` (162, 383, 472, 545, 622) and `docs/guides/decision-intelligence.md` (283, 610) are written as `if engine.check_compliance(...)`. After this change a malformed policy raises through all of them. The guide needs the `ProcessingError` case documented — that's the whole point of the change, and a reader who doesn't know it can't write correct code against the new contract.

**2. It needs a CHANGELOG `[Unreleased]` entry.** This is a user-visible behavior change to a public method: code that today gets `False` will get an exception. `[Unreleased]` is empty on `main` right now, so this would be the first entry.

**3. The rationale you wrote covers cases the change doesn't touch, and the gap should be deliberate rather than accidental.** The docstring says a raise means "could not determine compliance". But `_evaluate_compliance` still returns `False` for several genuinely-undetermined cases:

```python
field_value = self._get_metadata_field(metadata, decision, field)
if field_value is None:
    return False          # missing metadata field
...
else:
    if rule_key not in metadata:
        return False      # unknown rule key, field absent
```

A missing field is "the decision didn't carry the evidence" — reading that as non-compliant is defensible policy semantics, and different from "the rule itself is malformed". But that line isn't drawn anywhere in the code or the docstring, so as written the stated principle appears half-applied. One sentence in the docstring — malformed rules raise, absent evidence is non-compliance by policy — turns an apparent inconsistency into a decision.

Worth thinking about while you're there: `_evaluate_compliance` compares `field_value < rule_value` against decision metadata, which is caller data. A decision carrying a string where the policy expects a number now raises instead of returning `False`. For a single check that's correct and desirable. If anything ever loops this over a batch of decisions, one bad record will abort the batch rather than mark that record non-compliant. Nothing in the tree does that today — `check_compliance` has no callers in `semantica/`, `mcp/` or `integrations/` — so it's not a blocker, just the consequence to keep in mind if a batch path shows up.

Add the docs update and the changelog entry and I'm happy to approve.

Sameer6305 (approved):
Approved — PR #1160

This PR fixes issue #1159, where PolicyEngine.check_compliance() could silently return False when an internal evaluation or policy-checking error occurred, making operational failures indistinguishable from legitimate policy non-compliance.

The fix ensures:

False is reserved for a successfully evaluated non-compliant result.
Internal evaluation failures now raise ProcessingError and preserve the original exception via exception chaining.
Lower-level rule-evaluation failures are no longer silently swallowed.
The existing ValueError behavior for a missing policy is preserved.
Documentation and changelog entries clearly describe the new error semantics.

Validation performed from my side:

Reviewed the implementation and downstream check_compliance() callers.
Identified that the original regression test did not actually exercise the lower-level rule helper, and added coverage for that path.
Added coverage for policy-store lookup/infrastructure failures.
Strengthened the evaluation-error test to verify ProcessingError.__cause__.
Rebased the PR onto the current origin/main.
Resolved the single CHANGELOG.md conflict while preserving both upstream changes and the PR changes.
Verified the final diff contains only the intended PR changes.
tests/context/test_policy_engine.py: 47 passed
tests/context/: 818 passed, 2 skipped
Full-suite validation: 7,080 passed, 50 failed, 239 skipped, 6 xfailed. The same 50 failures were reproduced on origin/main and were confirmed unrelated/pre-existing.
Verified the local and remote PR branch point to the same final SHA: aaccca4edc5167b67a04789d4f13b6d6ec0c7673.

The PR is clean, scoped to the reported issue, and the identified regression paths are covered. Approved for merge.

Issue #1147, opened 2026-08-20
[BUG] The document IRI of an export is minted from the wall clock, so re-exporting an unchanged graph creates a new graph every time


fabio-rovai:
## Summary

The subject a JSON-LD export names its knowledge graph with is minted from the wall clock:

```python
# semantica/export/json_exporter.py:470
"@id": f"https://semantica.dev/graph/{utc_now_iso()}",
"@type": "semantica:KnowledgeGraph",
```

```python
# semantica/export/json_exporter.py:415
jsonld["@id"] = f"https://semantica.dev/data/{utc_now_iso()}"
```

So the identity of a graph is the moment it was serialized. Export the same graph twice and RDF says they are two different graphs.

This is independent of #1144 and survives its fix. I left it out of #1145 deliberately because it is a modelling question rather than a data-loss bug.

## Reproduction

One entity, no relationships, exported three times with no change in between:

```python
KG = {"entities": [{"id": "https://example.org/e1", "text": "Acme Corp", "type": "ORG"}],
      "relationships": []}

merged = Graph()
for i in range(3):
    JSONExporter().export_knowledge_graph(KG, f"churn{i}.jsonld", format="json-ld")
    merged.parse(f"churn{i}.jsonld", format="json-ld")
```

```
https://semantica.dev/graph/2026-08-20T09:16:24.096823+00:00
https://semantica.dev/graph/2026-08-20T09:16:24.100514+00:00
https://semantica.dev/graph/2026-08-20T09:16:24.101131+00:00

merging the three exports yields 3 semantica:KnowledgeGraph nodes
the one entity is claimed by 3 of them
merged graph: 15 triples for a 1-entity graph exported 3 times
```

Three identifiers, microseconds apart, for one graph that never changed.

## Why it matters

**Merging is the normal thing to do with RDF, and it does not work here.** Loading a day of exports into a store gives one `semantica:KnowledgeGraph` node per export rather than per graph. Counting graphs counts export runs. There is no way to ask whether two exports are of the same graph, because the only thing distinguishing them is when they were written.

**Re-exporting is not idempotent.** A pipeline that exports on every run accumulates subjects without bound, each one claiming the same entities through `semantica:entities`. Nothing dedupes, because at the RDF level nothing is duplicated.

[...]

Issue #1146, opened 2026-08-20
[BUG] @vocab points at https://semantica.dev/vocab/, which 404s and is not the namespace the package ships: every entity type and every user metadata key is minted into it


fabio-rovai:
## Summary

Every JSON-LD context Semantica writes sets

```json
"@vocab": "https://semantica.dev/vocab/"
```

`https://semantica.dev/vocab/` returns **404**. It is not the namespace the package ships, which is `https://semantica.dev/ns#` (`semantica/ontology/vocabulary/semantica-ns.ttl`, added in #1109 for #1107). The string appears in five places and is mentioned in no document and asserted in no test:

* `semantica/export/json_exporter.py:267, 297, 393, 481`
* `semantica/export/rdf_exporter.py:588`

`@vocab` is not decoration. It is the rule a JSON-LD processor applies to every key and every `@type` value that is not otherwise defined, so it decides the IRI of a great deal of user data.

## What lands in it

```python
kg = {
    "entities": [{"id": "https://example.org/e1", "text": "Acme Corp", "type": "ORG"}],
    "relationships": [],
    "metadata": {"format": "pdf", "source": "contract"},
}
JSONExporter().export_knowledge_graph(kg, "t.jsonld", format="json-ld")
```

Parsed:

```
type IRI                 https://semantica.dev/vocab/ORG
metadata predicates      https://semantica.dev/vocab/format
                         https://semantica.dev/vocab/source
```

Two distinct problems.

**1. Entity types.** Every type a user extracts becomes an IRI in a namespace that does not resolve and that the project does not publish. `ORG`, `PERSON`, and whatever a domain pipeline produces all get minted on `semantica.dev` as though they were Semantica's terms. This is #1107 again, one namespace over.

**2. Arbitrary user keys.** Anything unprefixed is swallowed, including every key of the user's own `metadata` dictionary. Two unrelated pipelines that both write a `source` key produce the same IRI for two different predicates, and that IRI is on `semantica.dev`. A key that happens to match one of Semantica's own terms produces a near-twin:

```
Semantica's own term      https://semantica.dev/ns#format
a user metadata key       https://semantica.dev/vocab/format
```

Nothing distinguishes them to a reader, and they mean different things.

## The serializers also disagree

The same knowledge graph, the same type, two exports:

```
JSON-LD   https://semantica.dev/vocab/ORG
Turtle    file:///private/tmp/scratch/ORG
```

[...]

PR #1165, opened 2026-08-21
Carry metadata through every RDF serialization (#1154)

### LINKED ISSUE #1154: [BUG] Every RDFExporter serialization drops entity and graph metadata: convert_kg_to_rdf carries it, no serializer ever reads it
## Summary

`metadata` appears exactly once as data in `semantica/export/rdf_exporter.py`:

```python
# rdf_exporter.py:302, in convert_kg_to_rdf
"metadata": knowledge_graph.get("metadata", {}),
```

It is carried into `rdf_data` and then **never read again**. None of the four serializers looks at `rdf_data["metadata"]`, and none of them looks at an entity's own `metadata` key either. The four are `serialize_to_turtle` (:335), `serialize_to_rdfxml` (:490), `serialize_to_jsonld` (:558) and `serialize_to_ntriples` (:644). Every entity loop reads exactly `id`, `text`/`label`, `type` and `confidence` and discards the rest of the dictionary.

So a knowledge graph exported through `RDFExporter` loses all of its metadata, at both graph and entity level, in every format, with no warning and a log line reporting success.

`JSONExporter` keeps the same metadata. Two writers in the same package give two different answers for the same input.

## Reproduction

```python
KG = {
    "entities": [{
        "id": "https://example.org/e1",
        "text": "Acme Corp",
        "type": "https://example.org/ORG",
        "metadata": {"page": 3, "extractor": "layoutlm"},
    }],
    "relationships": [],
    "metadata": {"source_document": "contract.pdf", "reviewed_by": "jane"},
}
```

On v0.6.6 (78fc9028). "in file" is a substring search of the bytes written, so it is independent of any reader:

```
writer / format          triples  metadata in file                         metadata parsed
----------------------------------------------------------------------------------------
RDFExporter  turtle            3  NONE                                     NONE
RDFExporter  ntriples          2  NONE                                     NONE
RDFExporter  rdfxml            3  NONE                                     NONE
RDFExporter  jsonld            3  NONE                                     NONE
JSONExporter json-ld           2  [source_document, reviewed_by, page, extractor]  NONE
```

`RDFExporter` never writes it. `JSONExporter` writes it and #1144 then hides it from `Graph().parse()`, which is why that last row parses nothing today. With #1145 applied the divergence is unambiguous:

```
RDFExporter  turtle            3  NONE                                     NONE
JSONExporter json-ld          13  [source_document, reviewed_by, page, extractor]  [all four]
```

Ten triples of user data, present through one exporter and absent through the other.

(The `turtle` 3 vs `ntriples` 2 disagreement in the same table is #1100, not this issue.)

## Why it matters

**It is the provenance that gets dropped.** The fields that say where an extracted statement came from and how much to trust it are exactly what goes in `metadata`: `source_document`, `page`, `extractor`, `reviewed_by`. An entity keeps its label and its confidence score and loses the record of which document produced it. A confidence number with no provenance behind it is the least useful half of the pair.

**The loss is silent and total.** No warning, no partial write, no log line that differs from a successful export. The Turtle file is well-formed and looks complete; there is nothing in it to indicate that anything was omitted. Any check that the export "worked" passes.

**Choosing a serialization format silently changes what your data means.** A pipeline that writes Turtle for a triplestore and JSON-LD for an API ships two different graphs from one `KnowledgeGraph`, and only the JSON-LD one is complete. Nothing in the API suggests format is a data-retention decision.

**`convert_kg_to_rdf` looks like it handles this.** It explicitly copies `metadata` into its return value, so a reader of that function reasonably concludes metadata is supported. The key is populated and then dropped downstream, which is harder to notice than never having been there.

## A related modelling question, if metadata does get serialized

`JSONExporter` writes each metadata dictionary as an *anonymous* node:

```
blank node  source_document  contract.pdf
blank node  page             3
```

A blank node is relabelled on every parse, so merging repeated exports of an unchanged graph produces a fresh metadata node each time: three exports give three metadata nodes hanging off one subject. That is the same non-idempotence as #1147, one level down, and it will not be fixed by fixing #1147. Whatever `RDFExporter` ends up emitting is worth giving a stable subject at the same time, rather than reproducing the anonymous form in four more places.

Worth noting these predicates are minted under `@vocab`, so they are also the user metadata keys discussed in #1146. The three questions interact: what to emit (this issue), what namespace to emit it in (#1146), and what subject to hang it on (above).

## What I would do

1. Serialize graph-level metadata as statements about the graph subject, and entity-level metadata as statements about the entity, in all four `RDFExporter` formats, matching what `JSONExporter` already produces so the two writers agree.
2. Give the metadata node a stable subject rather than an anonymous one, so re-export stays idempotent.
3. Settle the predicate namespace as part of #1146 rather than minting a second answer here.

Happy to open the PR. Point 3 is the one I would rather have your call on first, since it is the same decision #1146 is waiting on. I can either follow whatever you pick there, or emit metadata under the caller's `NamespaceManager` base and leave `@vocab` alone entirely.

fabio-rovai:
Closes #1154.

`convert_kg_to_rdf` copies `metadata` into the RDF-ready dictionary at `rdf_exporter.py:302`, and no serializer reads it back out. Turtle, N-Triples, RDF/XML and `RDFExporter.serialize_to_jsonld` each write an entity's id, type, text and confidence, and nothing else. `JSONExporter`'s json-ld path keeps the metadata, so the same knowledge graph exported two ways carries the user's data through one exporter and not the other.

The fields that go missing are the provenance ones. An entity keeps its confidence score and loses the source document, the page, the extractor and the reviewer that produced it.

### Measured

An entity carrying four metadata keys, exported on `e3405ebc` and on this branch, parsed with rdflib rather than read as text:

| format | before | after | of which metadata |
|---|---|---|---|
| turtle | 3 triples | 7 | 4 |
| ntriples | 3 | 7 | 4 |
| rdfxml | 3 | 7 | 4 |
| jsonld | 3 | 7 | 4 |

The same four statements in all four formats, same terms, same datatypes.

### What is exported, and what deliberately is not

The keys Semantica itself writes are mapped to declared terms in `DEFAULT_METADATA_TERMS` and declared in `semantica-ns.ttl`: `GraphBuilder.build_graph` writes `num_entities`, `num_relationships`, `temporal_enabled`, `timestamp` and `entity_resolution_applied`; `create_snapshot` writes `snapshot_time`; `load_from_neo4j` writes `source`, `uri` and `database`. Those are the package's own vocabulary, so declaring them is not a guess.

A key the caller supplied is a different matter, and I have not touched it here. Which namespace an arbitrary metadata key belongs in is #1146, which is open on your modelling call, and I said in that issue I would wait for it. So an unmapped key is not exported, and it warns rather than disappearing quietly:

```
Metadata key 'reviewed_by' has no term and was not exported. Which namespace a
caller-supplied key belongs in is issue #1146; pass
metadata_terms={'reviewed_by': '<iri>'} to export it now.
```

A caller who already knows the answer passes `metadata_terms={key: iri}` and gets the statement today. Whatever #1146 settles on becomes the default for unmapped keys, and nothing here has to be unwound to get there.

### Three modelling notes worth your eye

**Two keys cannot keep their own name.** `sem:source` is already the `ObjectProperty` carrying the subject of a reified relationship. Reusing it for the Neo4j loader's `source`, which is the string `"neo4j"`, would put a literal where an entity belongs. The metadata key stays `source` and the term is `sem:sourceSystem`; likewise `uri` maps to `sem:sourceUri`, the one metadata term whose value is a node rather than a literal. This is why the map is key to term rather than key to namespace.

**`sem:builtAt` and `sem:snapshotAt` have range `xsd:string`, not `xsd:dateTime`.** `GraphBuilder._get_timestamp` returns `datetime.now().isoformat()`, timezone-naive. #1114 is the demonstration of what typing such a value as `xsd:dateTime` costs: a timezone-qualified SPARQL filter over it hits XSD 1.1's indeterminate comparison and silently drops the row. #1121 swept `export/` and `provenance/` to an explicit offset and deliberately left `kg/` alone, because `context/` and `vector_store/` compare against naive values already on disk. Until that sweep reaches `GraphBuilder`, a string that looks like a timestamp is more honest than a type that invites arithmetic it cannot support.

**Graph-level metadata is written only when the caller names the graph** with `graph_uri`. This serializer has never minted a document node, and #1147 is where that default belongs once it lands; I would rather wire the two together than mint a second document identity here.

### Consistency

`_typed_literal_parts` chooses the lexical form and datatype once, so the four serializers cannot come to disagree about them the way they disagreed about `confidence` in #1100. The JSON-LD path writes explicit `@value`/`@type` rather than JSON's native numbers, which would otherwise have made an integer `xsd:double` there and `xsd:integer` in the other three.

### Verification

21 tests in `tests/export/test_metadata_passthrough.py`, every one asserting on the parsed graph rather than on serialized text, parametrised across all four formats where the behaviour is shared. Covered: the statement arriving in each format, datatype agreement across formats, the caller-named graph subject, the refusal-with-a-warning path, the `metadata_terms` override, a metadata literal containing a quote and a newline, an IRI-valued term written as a node, output being unchanged when no metadata is present, and a drift guard asserting every default term is declared in the shipped vocabulary.

`tests/export` and `tests/ontology`: 274 pass. Full-suite failure set is identical to the parent commit, 512 before and 512 after.

Happy to fold the caller-key half in as soon as you have called #1146, or to split anything here if you would rather take it in smaller pieces.

### DIFF
--- a/semantica/export/rdf_exporter.py
+++ b/semantica/export/rdf_exporter.py
@@ -193,6 +193,24 @@ def _escape_literal(value: str) -> str:
     )
 
 
+#: Turtle/N-Triples IRIREF grammar excludes these unescaped between `<` and
+#: `>`: control characters, space, and <>"{}|^`\. An IRI-valued metadata
+#: value (currently only sem:sourceUri, from the caller-controlled "uri"
+#: metadata key) is written as `<{value}>` with no other quoting, so a value
+#: containing one of these characters — a ">" followed by a full triple, for
+#: instance — closes the IRIREF early and lets the rest of the string be
+#: parsed as further RDF statements. This is the same shape of defect the
+#: entity/relationship IRIs were hardened against; that hardening resolves
+#: prefixes as well, which a metadata value never needs, so this stays a
+#: narrower, dedicated guard rather than reusing _as_turtle_iri.
+_IRI_REF_UNSAFE_RE = re.compile(r'[\x00-\x20<>"{}|^`\\]')
+
+
+def _safe_iri_ref(value: str) -> str:
+    """Percent-encode the characters an IRIREF may not contain unescaped."""
+    return _IRI_REF_UNSAFE_RE.sub(lambda m: quote(m.group(0), safe=""), value)
+
+
 def _escape_xml(value: str) -> str:
     """Escape a string for either XML element text or an attribute value.
 
@@ -328,7 +346,7 @@ def _typed_literal_parts(term: str, value: Any) -> tuple:
 def _turtle_object(term: str, value: Any) -> str:
     kind, lexical, datatype = _typed_literal_parts(term, value)
     if kind == "iri":
-        return f"<{lexical}>"
+        return f"<{_safe_iri_ref(lexical)}>"
     if datatype is None:
         return f'"{_escape_literal(lexical)}"'
     return f'"{lexical}"^^<{datatype}>'
--- a/semantica/export/rdf_exporter.py
+++ b/semantica/export/rdf_exporter.py
@@ -69,6 +69,187 @@ def mint_relationship_iri(index: int, source: Any, target: Any) -> str:
     return f"{SEMANTICA_NS}rel_{index}_{digest}"
 
 
+#: The metadata keys Semantica itself produces, and the terms they are written
+#: as. GraphBuilder.build_graph writes the first five, create_snapshot writes
+#: snapshot_time, and load_from_neo4j writes source / uri / database. These are
+#: Semantica's own vocabulary, so they are minted in the declared namespace and
+#: declared in semantica-ns.ttl.
+#:
+#: A key the caller supplied is a different matter. Which namespace an
+#: arbitrary metadata key belongs in is issue #1146, and until that is settled
+#: the exporter refuses to guess: it warns and skips, and a caller who already
+#: knows the answer passes ``metadata_terms``.
+#:
+#: The map is key -> term rather than key -> namespace because two of the keys
+#: cannot keep their own name. ``source`` on a graph loaded from Neo4j is the
+#: system it came from, while sem:source is already the ObjectProperty holding
+#: the subject of a reified relationship; reusing it would put a string where
+#: an entity belongs.
+DEFAULT_METADATA_TERMS: Dict[str, str] = {
+    "num_entities": f"{SEMANTICA_NS}numEntities",
+    "num_relationships": f"{SEMANTICA_NS}numRelationships",
+    "temporal_enabled": f"{SEMANTICA_NS}temporalEnabled",
+    "entity_resolution_applied": f"{SEMANTICA_NS}entityResolutionApplied",
+    "timestamp": f"{SEMANTICA_NS}builtAt",
+    "snapshot_time": f"{SEMANTICA_NS}snapshotAt",
+    "source": f"{SEMANTICA_NS}sourceSystem",
+    "uri": f"{SEMANTICA_NS}sourceUri",
+    "database": f"{SEMANTICA_NS}sourceDatabase",
+}
+
+#: Terms whose value is a node rather than a string. Everything else stays a
+#: literal: a metadata value that merely looks like a URL is not thereby a
+#: reference to one.
+IRI_VALUED_METADATA_TERMS: Set[str] = {f"{SEMANTICA_NS}sourceUri"}
+
+_XSD_NS = "http://www.w3.org/2001/XMLSchema#"
+
+
+def _escape_literal(value: str) -> str:
+    """Escape a string for a Turtle or N-Triples quoted literal."""
+    return (
+        value.replace("\\", "\\\\")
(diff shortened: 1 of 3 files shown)


KaifAhmad1 (approved):
Approving. Summary of where this landed, for the record:

- Metadata now genuinely round-trips through all four RDF serializers (Turtle, N-Triples, RDF/XML, JSON-LD), consistently typed, with unmapped caller keys refused loudly rather than dropped silently — exactly what #1154 asked for.
- Merge conflict against main's IRI-normalization hardening resolved (fb02c868), keeping both the metadata clauses and the IRI-safety normalization.
- The Turtle/N-Triples injection gap in the IRI-valued metadata path (the "uri" → sem:sourceUri case) is fixed (220fb10e), with regression tests.

Thanks for the thorough original work here, @fabio-rovai — the modelling notes (sourceSystem vs. source, string vs. dateTime timestamps, the drift-guard test) made this easy to review and verify. LGTM.

PR #1166, opened 2026-08-21
Verified RDF export: integrations/open_ontologies (#1108)

fabio-rovai:
Adds `integrations/open_ontologies`, alongside the existing `agno`, `crewai` and `openclaw` integrations. It registers a `"verified"` export method through Semantica's own `method_registry`, so nothing in the core changes and an existing pipeline behaves exactly as it does today until it asks for `method="verified"`.

The method exports exactly as the default does, then hands the bytes to an engine that had no part in writing them.

### Why an independent engine, and not more checks in the same code

A generator cannot check its own output: it shares the assumptions that produced the bug. If the code that writes an identifier has the wrong idea of what a valid IRI is, the code that reads it back has the same wrong idea and the tests pass.

Concretely, on the documented path. `GraphBuilder` defaults an entity's id to its surface text, so a two-entity export emits:

```turtle
<Acme Corp> a <ORG> ;
```

rdflib parses that. It resolves the subject against the current working directory, hands back `file:///…/Acme%20Corp`, and logs a warning nothing in a batch job is reading. Oxigraph refuses the file: `Invalid IRI code point ' '`. One reader gives you a graph whose identifiers depend on which directory the job ran in and reports success; the other tells you there is no graph.

Three checks run, in the order in which a failure in one makes the next meaningless: strict RDF 1.1 syntax, then a closed-world vocabulary check, then SHACL. The middle one is the one open-world tooling cannot do. An invented predicate is *unknown* rather than wrong in RDF, so it passes parsing and SHACL untouched, and inventing plausible terms is exactly what an extractor does when it is unsure. The third carries the focus-node count behind every verdict, because a shapes graph whose targets are absent reports conformance having examined nothing, and that report is indistinguishable from a real pass.

### It depends on #1127, and the test proves it

Before that fix `method_registry` caught every exception from a custom method and fell back to the default. A gate that rejected invalid RDF and deleted the file was followed by the default writing it straight back, so a registered policy hook could not refuse anything. That is why I held this until #1127 landed rather than opening it in the first wave.

Restoring the old behaviour, by flipping the `fallback_on_custom_error` default at `export/methods.py:225`, fails `test_registration_routes_export_rdf_through_the_gate` and nothing else. Measured on merged main:

```
--- default export ---   wrote 450 bytes, no error raised
--- verified export ---  refused: not valid RDF: Invalid IRI code point ' '
                         file on disk: False
```

### One thing to know before merging: the extra needs 0.5.0, not 0.4.0

I found this while testing and it is the reason the floor in `pyproject.toml` is where it is.

Up to 0.4.0, `open-ontologies-lite` declared `mcp` as a core dependency, though it imports it in exactly one module, its MCP server. `mcp` pulls a starlette major that `fastapi>=0.109.2` does not accept, so `pip install open-ontologies-lite` into this project **uninstalls fastapi**, and every import of `semantica/explorer` then fails with `ModuleNotFoundError: No module named 'fastapi'`. Three tests in `tests/test_security_regression_pr2.py` go from passing to erroring on that alone.

I have fixed it on that side rather than working around it here: `mcp` is now behind a `server` extra and the library API does not touch it. That is released as **0.5.0**, and this PR's extra requires `>=0.5.0` so the conflict cannot be reintroduced by an older pin. Given your existing note about `crewai` and the dependency-audit gates, I have also kept this extra **out of `all`** for the same reason, and I would rather you decided that than have me assume it.

### Verification

8 tests in `tests/integrations/open_ontologies/`, skipped cleanly when the package is not installed, covering: the unsound export raising with no file left behind, the identical export succeeding without verification so the bytes are visibly unchanged, a sound export keeping its file, `raise_on_failure=False` keeping the file for corpus surveys, an undeclared term being reported, shapes that match nothing withholding the verdict rather than passing, the registry route refusing, and the default method being left alone.

Full-suite failure set is identical to the parent commit, 514 before and 514 after, measured with the extra installed.

Happy to split the adapter from the extra, or to move it out of tree entirely if you would rather integrations stay to agent frameworks.

### DIFF
--- a/integrations/open_ontologies/__init__.py
+++ b/integrations/open_ontologies/__init__.py
@@ -1,0 +1,133 @@
+"""
+Semantica × Open Ontologies Integration
+=======================================
+
+Verified RDF export: write the graph, then have an independent engine read it
+back before the file is trusted.
+
+Public surface
+--------------
+verify_rdf            — check an RDF string, return a ``VerificationReport``
+verified_export_rdf   — export, verify, and refuse to leave an unsound file
+VerificationReport    — what the independent engine found
+VerificationError     — raised when a verified export would have written junk
+register              — install the ``"verified"`` method in the export registry
+
+Quick start
+-----------
+    pip install semantica[open-ontologies]
+
+    >>> from integrations.open_ontologies import register
+    >>> register()
+    >>> from semantica.export.methods import export_rdf
+    >>> export_rdf(kg, "graph.ttl", method="verified", ontology=vocab_ttl)
+
+Nothing in the core changes. The method is registered through Semantica's own
+``method_registry`` and is opt-in per call, so an existing pipeline behaves
+exactly as it did until it asks for ``method="verified"``.
+
+Why an independent engine
+-------------------------
+A generator cannot check its own output: it shares the assumptions that
+produced the bug. If the code that writes an identifier has the wrong idea of
+what a valid IRI is, the code that reads it back has the same wrong idea, and
+the tests pass.
+
+Concretely, and this is why the check earns its place. rdflib is lenient by
+design: given a subject written as ``<Acme Corp>``, which the IRI grammar
+forbids, it resolves the reference against the current working directory,
+yields ``file:///…/Acme%20Corp``, and logs a warning. Oxigraph refuses the
+file outright. One reader hands you a graph whose identifiers depend on which
+directory the job ran in and reports success; the other tells you there is no
+graph. This integration puts the strict reader after the export.
+
+Three checks run in the order in which a failure in one makes the next
+meaningless:
+
+1. **Syntax, strictly.** Is it RDF 1.1 at all.
+2. **Vocabulary, closed-world.** Which terms were used and declared nowhere.
+   RDF is open-world, so an invented predicate is unknown rather than wrong and
+   passes both parsing and SHACL untouched. Closing the world against a
+   declared vocabulary is the only way to tell a real term from a plausible
+   one, which matters here because inventing plausible terms is what an
+   extractor does when it is unsure.
+3. **Shapes, with an audit of what was examined.** Not only whether constraints
+   held, but over how many nodes. A shapes graph whose targets are absent from
+   the data reports conformance having checked nothing, and that report is
+   indistinguishable from a real pass.
+
+Compatibility
+-------------
+Requires ``open-ontologies-lite >= 0.4.0``, a pure-Python package over the
+Oxigraph engine. No Rust toolchain, no service to run, no network. SHACL needs
+the ``open-ontologies-lite[shacl]`` extra. Every symbol imports without it
+installed; the import happens inside the call.
+
+This integration depends on #1108, fixed in #1127. Before that, a custom method
+could not refuse anything: ``method_registry`` caught every exception and fell
+back to the default, so a gate that rejected invalid RDF and deleted the file
+was followed by the fallback writing it straight back.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any, Optional
+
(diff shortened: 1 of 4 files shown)


ZohaibHassan16 (approved):
Thanks again @fabio-rovai for the excellent work on verified RDF export for #1108.

Since it had been a few weeks, I went ahead and used maintainer permissions to push the 4 contained follow-ups noted by @KaifAhmad1 so we can land this:

1. **Cleanup on Exception:** Wrapped the verify call in `try...except` so unverified artifacts are deleted when validation raises an exception under `raise_on_failure=True`.
2. **Encoding Passthrough:** Forwarded caller-supplied `encoding` to `read_text` so verified exports match the encoding written by `RDFExporter`.
3. **Vocabulary Gating:** Gated closed-world checks on `ontology and policed_namespaces` to ensure partial configurations skip the check as documented rather than passing against an empty string.
4. **Version Floor:** Aligned module docstrings with `open-ontologies-lite >= 0.5.0`.
5. **Main Sync & Fixture Update:** Merged with latest `main` and updated `UNSOUND` to use an invalid IRI (`http://example.com/##`) to reflect recent IRI normalization changes in `RDFExporter`.

All 12 unit and regression tests pass cleanly. Ready to merge.

fabio-rovai:
Thanks for picking this up and pushing the four fixes. I checked all of them against the tree rather than taking them on trust, and they are right, including the fixture swap, which I want to confirm explicitly because it changes what the first test proves. `RDFExporter` now percent-encodes a bare entity id into the semantica namespace, so the original `Acme Corp` case emits `<https://semantica.dev/ns#Acme%20Corp>`, which is a valid IRI. The case from the issue is genuinely closed upstream, so moving `UNSOUND` to an id that survives normalisation was necessary rather than a weakening.

I also re-ran the falsification from the description on your merge commit, since it is the only thing that shows the gate can still refuse. Flipping the `fallback_on_custom_error` default at `semantica/export/methods.py:225` fails exactly one test, `test_registration_routes_export_rdf_through_the_gate`, with the original #1108 symptom in the log: `falling back to default because fallback_on_custom_error was set`. Nothing else moves.

I have pushed one more commit, because while verifying the above I found that none of these tests run in CI.

`requirements-ci.txt` is compiled with `--extra all`, and on your advice I kept `open-ontologies` out of `all`, so `open-ontologies-lite` is never installed by any job. There is also no `pytest tests/integrations/open_ontologies/` step anywhere in the workflows. The twelve tests are therefore not skipped, they are never collected, and the green checks on this PR carry no information about the integration they are meant to cover. That is an uncomfortable thing to merge on a PR whose entire argument is that a generator cannot check its own output.

Rather than move the extra into `all`, which is the thing you did not want for dependency-audit reasons, I followed the `explorer-extra` precedent already in `ci.yml`: a pinned, hashed requirements file for the extra plus a build step that installs it and runs the tests. It is resolved under `--constraint requirements-ci.txt`, so it shares 44 packages with the existing CI environment at identical versions, with no conflicts, and adds exactly one package. It runs last, so it cannot perturb anything before it.

The same commit corrects the test module docstring, which still described the pre-normalisation `<Acme Corp>` behaviour and so no longer matched the fixture directly beneath it.

One note for anyone installing the extra locally on an older pin. `open-ontologies-lite` 0.4.0 declared `mcp` as a core dependency, which drags in a starlette major that `fastapi>=0.109.2` rejects, so installing it uninstalls fastapi and every `semantica/explorer` import then fails. That is fixed on the library side in 0.5.0, which is why the floor is where it is, but it is worth knowing if you hit it.

Since this is a fork branch, the push has parked the checks pending your approval, so they will need a click before they go green again. The new step is the last one in the build job and adds a single package to the environment, so it cannot disturb anything ahead of it.

Happy to drop the CI commit if you would rather land this as it was and track the gap separately.

PR #1181, opened 2026-08-22
fix(export): JSON-LD document @id is now content-derived, not wall-clock

### LINKED ISSUE #1147: [BUG] The document IRI of an export is minted from the wall clock, so re-exporting an unchanged graph creates a new graph every time

reddynitish:
## Summary

`JSONExporter` minted the JSON-LD document/graph `@id` from `utc_now_iso()`
in two places (`_convert_kg_to_jsonld`, `_attach_document_metadata`).
Re-exporting an unchanged knowledge graph therefore produced a new subject
every time, so merging repeated exports duplicated graph identity instead
of recognizing the exports as the same graph.

## Fix

- Mint the default graph/document IRI from content rather than wall-clock time.
- Add optional `graph_uri` and `document_uri` overrides for callers that already have a stable graph name.
- Preserve `semantica:exportedAt` so export time is still recorded separately.
- Apply the fix to both JSON-LD export paths affected by #1147.

## Tests

Added regression coverage for:

- identical re-exports producing the same `@id`
- changed content producing a different `@id`
- repeated exports merging to one graph identity
- `exportedAt` still changing between exports
- caller-supplied `graph_uri`
- caller-supplied `document_uri`
- generic JSON-LD export path

Relevant export tests:
223 passed, 1 skipped, 1 pre-existing unrelated failure due to missing `pyoxigraph`.

Full repository suite also completed:
4726 passed, 189 skipped, 167 failed, 8 errors.
The failures/errors are in unrelated modules and environment/dependency paths; none reference `export`, `json_exporter`, or the changed IRI behavior.

Closes #1147

### DIFF
--- a/semantica/export/json_exporter.py
+++ b/semantica/export/json_exporter.py
@@ -28,10 +28,33 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional, Union
 
 from ..utils.exceptions import ProcessingError, ValidationError
-from ..utils.helpers import ensure_directory, utc_now_iso, write_json_file
+from ..utils.helpers import ensure_directory, hash_data, utc_now_iso, write_json_file
 from ..utils.logging import get_logger
 from ..utils.progress_tracker import get_progress_tracker
 from .rdf_exporter import SEMANTICA_NS, mint_entity_iri, mint_relationship_iri
+
+
+def _content_iri(prefix: str, payload: Any) -> str:
+    """Mint a document IRI from what was exported, not when.
+
+    Minting from ``utc_now_iso()`` gave every export of the same graph a new
+    identity a few microseconds apart, so re-exporting an unchanged graph was
+    never idempotent and merging exports duplicated every node (#1147). This
+    mirrors ``mint_entity_iri`` (#1109): identical content hashes to the same
+    IRI, and any change to the content changes it too. ``default=str`` keeps
+    the hash defined for values ``json.dumps`` would otherwise reject, such as
+    ``datetime`` objects a caller may have left in the graph.
+
+    Args:
+        prefix: IRI prefix the digest is appended to
+        payload: JSON-serializable value whose content determines the digest
+
+    Returns:
+        A stable IRI of the form ``{prefix}{16-hex-char digest}``
+    """
+    canonical = json.dumps(payload, sort_keys=True, default=str)
+    digest = hash_data(canonical)[:16]
+    return f"{prefix}{digest}"
 
 
 def _is_jsonld_document(data: Dict[str, Any]) -> bool:
@@ -230,7 +253,10 @@ class JSONExporter:
                 - statistics: Statistics dictionary (optional)
             file_path: Output JSON file path
             format: Export format - 'json' or 'json-ld' (default: self.format)
-            **options: Additional options passed to conversion methods
+            **options: Additional options passed to conversion methods:
+                - graph_uri: Caller-supplied IRI for the graph node when
+                  format='json-ld', overriding the default content-derived
+                  IRI (see #1147)
 
         Example:
             >>> kg = {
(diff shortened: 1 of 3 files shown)


ZohaibHassan16 (approved):
Approved.

Reproduced the bug on main first using the repro from the issue. Exported the same one-entity graph 3 times and got 3 different timestamp-based `@id`s. Merging them gives 15 triples, same as the issue.

Applied the fix and ran the same repro again. The `@id` is now the same across all 3 exports. `exportedAt` still changes like expected, and merging now gives 9 triples.

Also checked `_content_iri`. It does its own `json.dumps(sort_keys=True, default=str)` before hashing. This is needed because `hash_data` doesn't use `default=str` for dicts, so something like a datetime in the payload could fail there.

Checked the updated `test_document_iri_carrying_an_offset_is_a_valid_iri` too. The old assertion depended on the timestamp being part of the `@id`, so it doesn't make sense after this change. The new test still checks the important parts: valid IRI and RDF round trip.

Ran:

`tests/export/test_jsonld_document_iri.py`
8 passed.


LGTM.

PR #1189, opened 2026-08-22
feat(ontology): expose public SHACL validation API

### LINKED ISSUE #1186: [FEATURE] Promote _run_pyshacl to a public validation API
## Summary

The SHACL validation entry point that the official guide tells users to call is private: `docs/guides/shacl-validation.md` repeatedly calls `_run_pyshacl(...)` (a module-level function in `semantica/ontology/ontology_validator.py` with a leading underscore), and `OntologyValidator.validate()` covers structure validation only. Users following the guide therefore depend on a private function for the documented workflow.

## Expected Behavior

A public, stable entry point for "validate a serialized data graph against serialized SHACL shapes", e.g. `run_shacl_validation()` (or making `_run_pyshacl` public under its real name), with the guide updated to match and the private alias kept for compatibility if needed.

## Impact

- API polish / stability: private helpers have no compatibility guarantee, and this one is the documented path.
- New users following the guide learn an underscore-prefixed call as the canonical pattern.

Found while wiring SHACL validation into a demo built on the library.

mikemikimike:
## Summary

Closes #1186.

The SHACL guide documented the private `_run_pyshacl` helper as the canonical validation entry point. This change exposes `run_shacl_validation` through `semantica.ontology`, keeps `_run_pyshacl` as a compatibility alias, and updates all guide examples.

## Root cause and scope

The implementation existed only under a private module-level name, while users following the guide had to import a private symbol. The public API and compatibility alias share the same implementation, so behavior and supported RDF formats are unchanged. This affects the direct ontology validation entry point; Explorer routes and `OntologyValidator.validate()` are separate paths and are unchanged.

## Tests and verification

- Installed the documented SHACL dependency: `python -m pip install "pyshacl>=0.25.0"` (`pyshacl 0.40.1`, `rdflib 7.6.0`).
- `python -m pytest tests/ontology/test_ontology_advanced.py -q` — **33 passed**.
- The regression coverage now exercises both non-conforming and conforming graphs through the real pySHACL path.
- It verifies the new public API and the legacy alias return equivalent stable semantic results. During validation, comparing complete serialized dictionaries exposed that pySHACL generates different blank-node shape identifiers on separate calls; the test now compares stable report fields instead of nondeterministic identifiers.
- `git diff --check` — passed.

Black/isort/flake8 were attempted on the changed Python files, but the repository's legacy files contain pre-existing formatting/import violations outside this change. No unrelated formatter rewrite was applied. A mypy run was attempted but did not complete in the local environment.

No Docker-based verification was needed after the local optional dependency installation succeeded.

### DIFF
--- a/docs/guides/shacl-validation.md
+++ b/docs/guides/shacl-validation.md
@@ -8,7 +8,7 @@ icon: "shield-check"
 
 SHACL (Shapes Constraint Language) is a standard for validating graph-based data. While an ontology defines the conceptual *schema* (the "what" exists in your domain), SHACL defines the structural *rules and constraints* (the "how" it should be structured). 
 
-In Semantica, `SHACLGenerator` produces constraint rules (shapes) based on your ontology, and the public `run_shacl_validation` function evaluates your actual data against these rules. If a node violates a rule (e.g., missing a required property or using the wrong datatype), a detailed violation report is generated. The historical `run_shacl_validation` name remains available as a compatibility alias.
+In Semantica, `SHACLGenerator` produces constraint rules (shapes) based on your ontology, and the public `run_shacl_validation` function evaluates your actual data against these rules. If a node violates a rule (e.g., missing a required property or using the wrong datatype), a detailed violation report is generated. The historical `_run_pyshacl` name remains available as a compatibility alias.
 
 ## Why Use SHACL Validation?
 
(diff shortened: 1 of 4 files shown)


KaifAhmad1 (approved):
Verified directly against the branch (not just the diff): ran `tests/ontology/test_ontology_advanced.py` with real `pyshacl 0.40.0`/`rdflib 7.6.0` installed — all 33 tests pass, including both new regression tests. The `_run_pyshacl` compatibility alias is correctly one-directional (delegates to `run_shacl_validation`, no duplicated logic), the internal caller in `engine.py` still works unchanged, and `run_shacl_validation` is properly exported from `semantica.ontology`. No functional issues — one cosmetic nit (a missing blank line between two test methods) that doesn't affect collection or execution. LGTM.

PR #1192, opened 2026-08-23
fix(triplet_store): honor RDF4J repository id

### LINKED ISSUE #1191: bug(triplet_store): RDF4JStore ignores its repository_id constructor argument
## Summary

`RDF4JStore.__init__` declares `repository_id` as a named parameter, but the constructor body never reads that parameter — it reads `repository_id` from `**config` instead, which is always empty for a value passed as an explicit keyword argument. The result: `RDF4JStore(endpoint=..., repository_id="my_repo")` silently connects to the `"default"` repository instead of `"my_repo"`, with no error or warning.

Found while verifying the connection examples in `docs/storage-backends.md` (#899) against the real adapter APIs.

## Location

`semantica/triplet_store/rdf4j_store.py`

```python
def __init__(
    self, endpoint: Optional[str] = None, repository_id: Optional[str] = None, **config
):
    ...
    self.endpoint = endpoint.rstrip("/")
    self.repository_id = config.get("repository_id", "default")  # <- bug: ignores the `repository_id` parameter
```

Compare with `ApacheAgeStore`, which handles the equivalent case correctly:

```python
self.connection_string = connection_string or config.get("connection_string", ...)
self.graph_name = graph_name or config.get("graph_name", "semantica")
```

## Reproduction

```python
from unittest.mock import patch, MagicMock
from semantica.triplet_store.rdf4j_store import RDF4JStore

with patch("semantica.triplet_store.rdf4j_store.requests.get") as mock_get:
    mock_get.return_value = MagicMock(status_code=200)
    store = RDF4JStore(endpoint="http://localhost:8080/rdf4j-server", repository_id="semantica")
    print(store.repository_id)  # prints "default", not "semantica"
```

There is no workaround via the public constructor: since `repository_id` is a named parameter, any keyword argument with that name binds to it directly and never reaches `**config`, so `config.get("repository_id", ...)` can never see a caller-supplied value through the documented `__init__` signature.

## Impact

Any caller that follows the documented/docstring usage (`RDF4JStore(endpoint=..., repository_id="repo1")`, as shown in the class's own docstring example) silently ends up talking to the wrong repository. This is a silent-wrong-behavior bug, not a crash, so it can be easy to miss in testing against a server that only has a `default` repository.

## Suggested fix

```python
self.repository_id = repository_id or config.get("repository_id", "default")
```

matching the pattern already used in `age_store.py` for `connection_string`/`graph_name`.

## Suggested scope

- Fix the one-line bug above.
- Add a regression test asserting `RDF4JStore(endpoint=..., repository_id="x").repository_id == "x"`.
- Once fixed, drop the "currently has no effect" caveat added to `docs/storage-backends.md` in #899.

Freakz2z:
## Summary

- honor the explicit `repository_id` constructor argument when selecting the RDF4J repository
- add a regression test that verifies both the stored identifier and connection URL
- remove the temporary documentation caveats that said the argument had no effect

Closes #1191

## Validation

- `pytest tests/triplet_store/test_rdf4j_store.py tests/triplet_store/test_construct_query_detection.py tests/triplet_store/test_sparql_injection.py -q` — 65 passed
- focused Black checks for the changed implementation and test ranges
- `git diff --check`

A broader `tests/triplet_store` run reached 339 passed and 1 skipped; the remaining pre-existing Jena concurrency test expects five unique thread IDs, but short-lived threads reuse IDs locally and produce fewer unique triples. It is unrelated to the RDF4J constructor path changed here.

### DIFF
--- a/docs/reference/triplet_store.md
+++ b/docs/reference/triplet_store.md
@@ -182,7 +182,7 @@ for row in result.bindings:
     store = TripletStore(
         backend="rdf4j",
         endpoint="http://localhost:8080/rdf4j-server",
-        repository_id="semantica",   # passed through **config
+        repository_id="semantica",   # selects the remote repository
     )
     ```
 
--- a/docs/storage-backends.md
+++ b/docs/storage-backends.md
@@ -35,7 +35,7 @@ This page is intentionally conservative: it distinguishes between an adapter exi
 | FalkorDB | LPG | Yes | Yes | Partial | Partial | Redis-based; provenance depends on node/edge properties, and multi-graph isolation depends on the selected graph name. |
 | Amazon Neptune | LPG | Yes | Yes | Partial | Partial | Use the property-graph endpoint; AWS auth, VPC, and endpoint configuration can affect local tests. Provenance depends on node/edge properties. |
 | Apache AGE | LPG | Yes | Yes | Partial | Partial | Runs through PostgreSQL/AGE; Cypher compatibility and property handling can differ from standalone LPG engines. |
-| RDF4J | RDF | Yes | Partial | Partial | Partial | Context separation relies on named graphs; triple-level provenance may require reification or graph-level metadata. `RDF4JStore(repository_id=...)` currently has no effect — the constructor always connects to the `"default"` repository regardless of the value passed; track a fix separately. |
+| RDF4J | RDF | Yes | Partial | Partial | Partial | Context separation relies on named graphs; triple-level provenance may require reification or graph-level metadata. |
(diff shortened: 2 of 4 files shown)


KaifAhmad1 (approved):
Verified locally: checked out the branch, ran the full test_rdf4j_store.py suite (38 passed), and confirmed the encoding fix covers all four URL builders (`_connect`, `_get_sparql_endpoint`, `_get_update_endpoint`, `begin_transaction`) with a regression test against a nasty repo id (`team/repo ?#`). The `black --check` failure is pre-existing on `main`, not introduced here. LGTM.

PR #1193, opened 2026-08-23
fix(utils): write console progress only to an interactive stdout

### LINKED ISSUE #1185: [BUG] Progress tracker floods non-TTY stdout; SEMANTICA_DISABLE_PROGRESS is undocumented
## Summary

Every stage of a non-interactive script that uses the library prints rich progress bars (emoji, spinner, ETA) to stdout. `ProgressTracker` respects a `SEMANTICA_DISABLE_PROGRESS` environment variable (read in `semantica/utils/progress_tracker.py`), but that escape hatch is undocumented and must be discovered by reading source. Building a plain `python demo.py` on top of the library floods the output with one progress bar per stage, drowning the program's own output.

## Steps to Reproduce

```bash
pip install semantica
python - <<'PY'
from semantica.ingest import FileIngestor
obj = FileIngestor().ingest_file("sample.txt")
PY
```

Run in a plain terminal or with output redirected to a file: the progress bar renders to stdout (including escape sequences in redirected output).

## Expected Behavior

Progress bars are for interactive use. Standard practice: when stdout is not a TTY, either disable the tracker automatically or degrade to plain one-line logs. At minimum, `SEMANTICA_DISABLE_PROGRESS` should be documented (README / docs) so script authors know it exists.

## Impact

- Scripts and CI logs carry progress-bar escape-sequence noise.
- Users of the library in pipelines need source-diving to find the env var.
- (Minor) NERExtractor and several reasoners explicitly re-enable the tracker in `__init__`, so per-instance disabling is not enough.

## Suggested Fix

Auto-disable the tracker when `sys.stdout.isatty()` is false (with the env var remaining as an override), and document the env var.

Found while building a CLI demo on top of the library.

ALDRIN121:
## Description

Non-interactive consumers of the library — scripts, CI jobs, anything with piped or redirected stdout — get one rich progress bar per pipeline stage written straight into their output, escape sequences included. A plain `python demo.py > out.txt` captured 173 bytes of progress-bar noise around 10 bytes of the program's own output. The escape hatch, `SEMANTICA_DISABLE_PROGRESS`, exists but is only documented in the utils reference page, where script authors won't find it.

This PR stops the flood at the display level and documents both progress switches.

## Type of Change

## Related Issues

Fixes #1185

## Changes Made

- `ProgressTracker` attaches `ConsoleProgressDisplay` only when stdout is an interactive terminal, under Jupyter, or when `SEMANTICA_FORCE_PROGRESS` is set. `FileProgressDisplay` is unchanged, so progress logging still works in pipelines.
- New `SEMANTICA_FORCE_PROGRESS` env var, following the `NO_COLOR`/`FORCE_COLOR` convention. `SEMANTICA_DISABLE_PROGRESS` keeps its existing meaning and takes precedence when both are set (verified empirically, and covered by a test).
- Both variables are now documented in the README Quick Start and `docs/reference/utils.md`.

## Deviations from the issue

The issue suggested disabling the tracker on non-TTY stdout. This gates the display instead, because:

1. Disabling the tracker would short-circuit before `FileProgressDisplay` and take file progress logging down with it — exactly the output pipelines *do* want.
2. ~20 modules set `progress_tracker.enabled = True` in `__init__`; fixing this at the tracker level would require teaching that property setter about TTY state too. Gating the display leaves both alone.

The reported bug — progress-bar noise in redirected stdout — is fixed either way. The issue's "undocumented" and "re-enable" claims were also corrected in the issue thread; the env var was documented but undiscoverable, and the re-enable sites are already neutralized by the property setter.

## Design note

The claim comment on the issue proposed an `enabled: Optional[bool] = None` constructor opt-in; during implementation the opt-in became `SEMANTICA_FORCE_PROGRESS`, which needs no signature change and follows the `NO_COLOR`/`FORCE_COLOR` convention.

**Known limitation:** TTY detection runs once at tracker construction (the tracker is a process-wide singleton), so a process that redirects stdout after first use needs the env vars to change behaviour.

## Testing

- [x] Tested locally

### Test Commands

- `pytest tests/test_progress_tracker_regressions.py` — 11 passed (6 new)
- `pytest tests/split tests/pipeline tests/visualization tests/ingest tests/core tests/utils -m "not integration"` — 539 passed, 7 failed. The 7 failures are pre-existing network-blocked tests that also fail on `main` with this change reverted.
- Style: `black --check` and `flake8` (project settings from `.pre-commit-config.yaml`) report no findings on any changed lines. `mypy` not run — unavailable in the dev environment.

### Before / after (issue repro, stdout redirected to a file)

- `main`: 173 bytes — progress bar with escape sequences wrapped around 10 bytes of program output
- this branch: 10 bytes — program output only

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,8 @@ semantica doctor
 # Config file           pass    ~/.semantica/config.yaml
 ```
 
+**Running in a script or CI?** Progress bars are written only when stdout is an interactive terminal (or a Jupyter notebook), so piping and redirecting stay clean by default. Override with `SEMANTICA_DISABLE_PROGRESS=1` to silence progress everywhere, or `SEMANTICA_FORCE_PROGRESS=1` to keep it when stdout is redirected. `SEMANTICA_DISABLE_PROGRESS` takes precedence.
+
 <div align="center">
 
 If Semantica solves a real problem for you, a star helps others find it.
--- a/docs/reference/utils.md
+++ b/docs/reference/utils.md
@@ -77,7 +77,17 @@ Most users won't call utils directly: it's the **shared foundation** for all mod
     export SEMANTICA_LOG_LEVEL=DEBUG
     export SEMANTICA_LOG_FORMAT=json     # "json" | "text"
     export SEMANTICA_DISABLE_PROGRESS=true
+    export SEMANTICA_FORCE_PROGRESS=true
     ```
+
+    <Tip>
+      **Progress bars follow your terminal.** Console progress is written only when
+      stdout is an interactive terminal (or a Jupyter notebook), so piping or
+      redirecting output no longer fills logs with progress bars and escape
+      sequences. Set `SEMANTICA_DISABLE_PROGRESS` to silence progress even in a
+      terminal, or `SEMANTICA_FORCE_PROGRESS` to keep it when stdout is redirected.
+      `SEMANTICA_DISABLE_PROGRESS` wins if both are set.
+    </Tip>
   </Step>
 </Steps>
 
--- a/semantica/utils/progress_tracker.py
+++ b/semantica/utils/progress_tracker.py
@@ -64,6 +64,29 @@ def _progress_disabled_from_env() -> bool:
         "on",
     )
 
+
+def _progress_forced_from_env() -> bool:
+    """Return whether console progress is forced on despite a non-interactive stdout."""
+    return os.getenv("SEMANTICA_FORCE_PROGRESS", "").strip().lower() in (
+        "1",
+        "true",
+        "yes",
+        "on",
+    )
+
+
+def _stdout_is_tty() -> bool:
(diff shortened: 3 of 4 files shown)


KaifAhmad1 (approved):
Verified locally: checked out the branch, confirmed the TTY gate (`_stdout_is_tty() or is_jupyter or SEMANTICA_FORCE_PROGRESS`) covers both `ConsoleProgressDisplay` construction sites and nowhere else, and that `FileProgressDisplay`/Jupyter dynamic detection are unaffected. Ran the regression suite — 10/11 passed; the one failure (subprocess timeout in the knowledge-base build test) reproduces identically on unpatched `main`, so it's pre-existing and unrelated. `black --check` "fails" too, but same drift exists on `main` already. Good test coverage for the TTY/non-TTY/FORCE/DISABLE-precedence matrix. LGTM.

Issue #1119, opened 2026-08-19
[TECH DEBT] Move Markdown URL sanitization helper out of React component module


Sameer6305:
## Problem Statement

The Markdown viewer currently exports the `isSafeUrl` helper directly from `MarkdownContentViewer.tsx` so that it can be unit tested.

The Explorer lint configuration reports:

`react-refresh/only-export-components`

because the module exports a non-component helper alongside the React component.

This lint issue was pre-existing in the PR implementation and was intentionally left unchanged because fixing it was outside the scope of the Markdown viewer feature.

However, keeping utility functions in the React component module makes the module less clean and continues to trigger the React Refresh lint rule.

## Proposed Solution

Move the URL validation/sanitization helper into an appropriate shared utility module, for example:

```text
explorer/src/utils/...
````

or another location consistent with the existing Explorer architecture.

Then:

* export `isSafeUrl` from the utility module
* import it into `MarkdownContentViewer`
* update tests to import the helper from its new location
* remove the non-component export from `MarkdownContentViewer.tsx`

Avoid changing the lint configuration merely to suppress the warning.

## Acceptance Criteria

* `MarkdownContentViewer.tsx` exports only the React component(s) expected by the module.
* `isSafeUrl` remains directly unit-testable.
* The existing URL-safety behavior is unchanged.
* All existing URL-safety tests continue to pass.
* The `react-refresh/only-export-components` lint error is eliminated.
* No unrelated lint configuration changes are introduced.
* Explorer TypeScript, tests, lint, and build continue to pass.

## Security Requirement

This is primarily a code-organization/lint cleanup.

The URL-safety behavior itself must not be weakened or rewritten unnecessarily.

The existing protections against unsafe URL schemes must remain intact, including handling of:

* `javascript:`
* `data:`
* `vbscript:`
* unsafe/malformed URLs
* whitespace-only values

Any refactoring must preserve the existing security behavior exactly.

## Scope

This issue is limited to extracting the URL-safety helper from the React component module.

It does not change Markdown rendering behavior or introduce new URL policies.

```

**Priority-wise:** I’d make **Issue 1** a normal accessibility enhancement, **Issue 2** a performance investigation, and **Issue 3** low-priority technical debt. None should block the already-approved #1078.
```

pravit-amp:
I'd like to take this on — could it be assigned to me @Sameer6305?

Approach:

[...]

Issue #1118, opened 2026-08-19
[PERFORMANCE] Investigate large Markdown and GFM table rendering in Explorer


Sameer6305:
## Problem Statement

The Explorer Markdown viewer currently parses and renders the complete Markdown content into the browser DOM.

For normal node content this is acceptable, but large Markdown documents—especially large GFM tables—can become expensive to parse and render.

During review of the Markdown viewer, large GFM table content showed significantly higher rendering cost than normal Markdown content. The viewer's constrained visual height limits what is visible to the user, but it does not prevent the full Markdown document from being parsed and converted into DOM nodes.

This means a very large or pathological node can potentially cause noticeable browser main-thread work, UI freezes, or poor responsiveness.

## Areas to Investigate

Evaluate the behavior of:

- very large Markdown documents
- large paragraphs
- very large fenced code blocks
- large nested lists
- GFM tables containing hundreds/thousands of rows
- deeply nested Markdown structures
- rapid switching between nodes containing large content

Determine whether the current implementation has acceptable performance characteristics for realistic Explorer data.

## Proposed Investigation

Measure:

- Markdown parsing time
- React rendering time
- DOM node count
- memory usage
- interaction responsiveness
- performance when switching between large-content nodes

Consider possible mitigations only if justified by measurements, such as:

- virtualization for very large tabular content
- progressive rendering
- content-size thresholds
- lazy rendering
- more efficient Markdown rendering
- rendering safeguards for pathological input

Do not introduce an arbitrary content limit without establishing an appropriate product/technical threshold.

## Acceptance Criteria

- Establish a reasonable maximum/expected Markdown content size for Explorer.
- Benchmark representative large Markdown and GFM table cases.
- Identify whether large content can materially degrade Explorer responsiveness.
- If a meaningful performance problem exists, implement an appropriate mitigation.
- Normal Markdown rendering must remain unchanged.
- Source mode should remain exact and should not arbitrarily truncate user content unless explicitly designed and documented.
- Existing Explorer tests and build continue to pass.

## Scope

This issue is focused on client-side Markdown rendering performance.

It does not include Markdown editing, persistence, import/export, or revision-history functionality.

pravit-amp:
I'd like to take this on — could it be assigned to me @Sameer6305?

Since this is scoped as an investigation with mitigations "only if justified by measurements", here's what I'd measure plus one hypothesis I'd start from.

[...]

PR #1078, opened 2026-08-18
feat(explorer): add markdown content preview and source view

sakshi04-ui:
## Description

Adds a dedicated, read-only Markdown content viewer to the Semantica Explorer Graph Inspector with **Preview** and **Source** view modes. 

Previously, node content loaded into the client was primarily treated as an unstructured label without a dedicated reading pane. This PR introduces a focused `MarkdownContentViewer` component allowing users to read formatted Markdown (headings, lists, blockquotes, tables, code blocks) and inspect raw, unmodified source text directly within the graph inspector.

This is a focused, read-only first step toward the human-editable memory workflow (#765) and intentionally avoids write semantics until persistence and concurrency requirements are settled (#852).

## Type of Change

- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #900 

## Changes Made

- **`MarkdownContentViewer` Component (`explorer/src/workspaces/GraphWorkspace/MarkdownContentViewer.tsx`)**:
  - Implemented dual-mode tabs: **Preview** (rich GFM Markdown) and **Source** (monospace preformatted view).
  - Preserves exact source text, whitespace, indentation, and Unicode.
  - Added one-click copy-to-clipboard action with visual feedback.
  - Added empty and plain-text fallback handling.
  - **Security controls**: Raw HTML execution is disabled; safe URL scheme allowlist (`http:`, `https:`, `mailto:`, relative anchors `#`); remote images guarded with placeholder badges; external links open with `target="_blank"` and `rel="noopener noreferrer"`.
- **Inspector Integration (`explorer/src/workspaces/GraphWorkspace/GraphInspectorPanel.tsx`)**:
  - Rendered `<MarkdownContentViewer content={nodeContent} />` inside a native collapsible `<details className="node-panel-collapse">` matching the existing inspector visual language.
- **Dependencies (`explorer/package.json`)**:
  - Added maintained `react-markdown` (^10.1.0) and `remark-gfm` (^4.0.1).
  - Added `npm run test:markdown-viewer` and integrated it into the root `npm run test` script.
- **Unit Test Suite (`explorer/tests/markdownContentViewer.test.ts`)**:
[...]

Issue #1117, opened 2026-08-19
[ENHANCEMENT] Improve accessibility semantics for Markdown Preview/Source tabs


Sameer6305:
## Problem Statement

The Explorer Markdown content viewer provides `Preview` and `Source` controls that visually behave like tabs, but the current implementation does not provide the complete ARIA tab/tabpanel relationship expected by assistive technologies.

The current implementation includes some tab semantics such as `role="tablist"`, `role="tab"`, and `aria-selected`, but does not fully connect the selected tab to its corresponding content panel.

This can make the Preview/Source control less understandable and navigable for screen-reader and keyboard users.

## Proposed Solution

Improve the Markdown viewer's accessibility semantics by implementing a complete tab pattern.

The implementation should consider:

- `role="tablist"` on the tab container
- `role="tab"` on each tab
- `aria-selected` reflecting the active mode
- unique IDs for each tab
- `aria-controls` connecting each tab to its corresponding panel
- `role="tabpanel"` on the active content panel
- `aria-labelledby` connecting the panel back to its tab
- appropriate focus behavior
- visible focus indication
- sensible keyboard interaction for switching between Preview and Source

The implementation should remain consistent with the existing Explorer accessibility and UI conventions.

## Acceptance Criteria

- Preview and Source controls expose correct tab semantics to assistive technologies.
- The active tab is correctly represented through `aria-selected`.
- Each tab references its corresponding panel using `aria-controls`.
- Each panel references its controlling tab using `aria-labelledby`.
- Keyboard users can understand and operate the Preview/Source controls without relying solely on visual styling.
- Existing mouse/touch behavior remains unchanged.
- No regression to Markdown rendering or Source mode.
- Existing Explorer tests continue to pass.

## Scope

This issue is limited to accessibility improvements for the Markdown viewer controls.

It does not include changes to Markdown rendering, sanitization, content persistence, editing, or import/export workflows.

pravit-amp:
I'd like to take this on — could it be assigned to me @Sameer6305?

Approach:

- Wire the ids with `useId()` rather than hardcoded strings. `GraphWorkspace.tsx` already establishes this pattern for the search combobox — `const listboxId = useId()` with `aria-controls`, `aria-activedescendant` and derived `${listboxId}-${index}` ids — so following it keeps the Explorer consistent and, more practically, keeps the ids unique if the viewer is ever rendered in more than one inspector panel at a time. Hardcoded ids would collide silently in that case.
[...]

PR #1195, opened 2026-08-23
perf(explorer): stop re-parsing markdown on every viewer re-render (#1118)

pravit-amp:
## Description

Investigation and fix for #1118. The full measurements are posted [in the issue thread](https://github.com/semantica-agi/semantica/issues/1118); this PR lands the half that needs no product decision.

Profiling in headless Chromium against a production React build separated remark
parse time, React commit time, and DOM node count across large-prose,
large-code-block, deep-nested-list, and GFM-table fixtures. Two findings; one is
fixed here.

**Finding 1 (fixed here) — every re-render re-parsed the whole document and
remounted the whole subtree.** `remarkPlugins` and the ~20-entry `components` map
were inline literals, so each render allocated fresh arrow components; React saw
a new element type per mapped tag and replaced the DOM instead of updating it. A
DOM-identity probe (capture a deep node, force a parent re-render with identical
props, check `isConnected`) confirmed `remounted: true` on every fixture.

The remount wasn't the main cost, though — the re-parse was. `react-markdown`
runs the whole remark pipeline synchronously inside its own render, so any
unrelated state change (clicking Copy, toggling Preview/Source) re-paid the full
parse. Clicking Copy on a 2000-row table blocked the main thread for over a
second.

**Finding 2 (not fixed here) — large GFM table parse is quadratic, and it's
upstream.** The same table text parses in 12.5 ms without `remark-gfm` and
1155.9 ms with it at 2000 rows (~O(n^1.9)). That's the table extension in
micromark, not the viewer. Any mitigation is a product decision, so it stays on
the issue rather than being smuggled into this PR.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement
- [ ] Code refactoring

## Related Issues

Addresses #1118 (Finding 1). Leaving the issue open for Finding 2 — the upstream
GFM table parse cost and the threshold question — rather than auto-closing it.

Follow-up to #1078.

## Changes Made

- Hoisted `remarkPlugins` to a module-scope `REMARK_PLUGINS` constant and the
  `components` map to a module-scope `MARKDOWN_COMPONENTS: Components`, so both
  keep a stable identity across renders and React updates the rendered subtree
  instead of remounting it.
- Memoised the rendered `<ReactMarkdown>` element on `rawContent` via `useMemo`,
  so an unrelated re-render no longer re-enters the remark pipeline. Keyed on
  `rawContent` so a genuine node change still re-parses exactly once.
- Imported the `Components` type from `react-markdown` to type the hoisted map.
- The components map is moved **verbatim** — every override, the C-1 `node`-prop
  discard, the C-2 fragment-link handling, and the `isSafeUrl` gate are byte-for-byte
  unchanged, as are all styles. Rendered output is identical.

## Measurements

Median of 7 runs, headless Chromium, production React, Apple Silicon laptop.
Useful as ratios rather than absolutes.

Re-render cost (the path this PR fixes):

| Fixture | before | after |
|---|---|---|
| prose, 100 paragraphs (56 KB) | 9.5 ms | 0.1 ms |
| prose, 1000 paragraphs (563 KB) | 90.3 ms | 0.1 ms |
| code block, 2000 lines (99 KB) | 5.9 ms | 0.1 ms |
| nested list, depth 10 (69 KB) | 75.1 ms | 0.2 ms |
| GFM table, 500 rows (27 KB) | 112.5 ms | 0.1 ms |
| GFM table, 1000 rows (55 KB) | 364.2 ms | 0.1 ms |
| GFM table, 2000 rows (117 KB) | 1121.3 ms | 0.1 ms |

The DOM-identity probe now reports `remounted: false`. Initial mount and
node-switching are deliberately unchanged — those are genuine parses.

Initial parse, with and without `remark-gfm` (Finding 2, upstream, not fixed here):

| Rows | bytes | with `remark-gfm` | without |
|---|---|---|---|
| 100 | 5.0 KB | 12.8 ms | 1.2 ms |
| 250 | 13.4 KB | 33.6 ms | 2.3 ms |
| 500 | 27.4 KB | 92.8 ms | 3.2 ms |
| 1000 | 55.4 KB | 301.3 ms | 5.7 ms |
| 2000 | 117.4 KB | 1155.9 ms | 12.5 ms |

**On the issue's "establish a reasonable maximum content size" criterion:** document
size is the wrong threshold. 563 KB of prose parses in 85 ms and a 99 KB fenced
code block in 5.4 ms, while a 27 KB GFM table takes 102 ms. A byte-size limit would
reject a half-megabyte prose document that renders fine while admitting a 27 KB
table that doesn't. Row count predicts cost; bytes do not. If anything is gated, it
should be gated on table rows — around 500, using the ~100 ms perceptible-lag
boundary.

Two caveats on method, for the record: memory usage was not measured (main-thread
block duration turned out to be the binding constraint), and the parse-vs-React
split is approximate, since parse is a median of standalone runs while mount is a
single run. Neither affects the parse-dominates conclusion — parse is ~95% of mount
on the expensive cases.

## Testing

- [x] Tested locally
- [ ] Added tests for new functionality — see note below
- [ ] Package builds successfully (`python -m build`) — N/A, frontend-only change
      touching no Python packaging

### Test Commands

```bash
cd explorer && npm run test:graph-workspace
```

```bash
cd explorer && npm run build
```

Results:

- `npm run test:graph-workspace` — 66 passing, 0 failing.
- `npm run build` — `tsc -b && vite build` succeed.

**On the missing regression test.** This regression is invisible to the current
test setup: `node --test` renders via `renderToString` and cannot observe commit
behaviour or DOM identity, and a timing assertion would be flaky. So nothing in CI
currently stops this from regressing. `explorer` already depends on `playwright`, so
the profiling harness needs no new devDependency and I'm happy to contribute it as a
manually-run script (not CI) — but committing new tooling surface is a maintainer
call, so I've left it out of this PR rather than deciding unilaterally. Say the word
and I'll add it.

## Documentation

## Breaking Changes

**Breaking Changes**: No

Rendered output is unchanged, the component's props are unchanged, and Source mode
is untouched — it remains exact and untruncated, per the issue's acceptance criteria.
No content limit of any kind is introduced.

## Checklist

## Additional Notes

**Sequencing.** This branch is rebased onto `main` post-#1078. It touches the same
file as the #1119 PR (which moves `isSafeUrl` out into `markdownUrlSafety.ts`). The
two changes are in different regions of the file but will likely need a trivial
rebase depending on merge order — happy to rebase whichever lands second.

Relatedly, `eslint` on `MarkdownContentViewer.tsx` still reports one
`react-refresh/only-export-components` error at the `isSafeUrl` export. That is
pre-existing and is exactly what the #1119 PR removes; it is untouched here to keep
this PR scoped to performance.


Sameer6305 (approved):
Thanks @pravit-amp for taking this investigation on and for the thorough profiling work. The approach here was particularly solid because the performance claims were backed by measurements rather than assumptions, and the investigation correctly separated the viewer's own re-render problem from the upstream `remark-gfm` table-parsing cost.

### What was solid

* Correctly identified that inline `remarkPlugins` and the `components` map were causing unnecessary work during viewer re-renders.
* Used stable module-scope `REMARK_PLUGINS` and `MARKDOWN_COMPONENTS`.
* Correctly memoized the rendered Markdown on `rawContent`.
* Preserved the existing Markdown renderer implementations and security behavior.
* Correctly avoided introducing an arbitrary document-size limit without sufficient evidence.
* Distinguished the re-render regression from the separate large-GFM-table parsing problem.
* The profiling methodology and DOM-identity probe provided useful evidence rather than relying only on timing numbers.
* Scope remained appropriately limited to the performance issue.

### What I verified during review

I went through the implementation and the render/state flow in detail.

The memoization dependency is correct: `rawContent` is the only component-level value captured by the memoized Markdown element; the plugins and component map are now module-level constants. 

The renderer map was also checked for closures. The renderers don't capture component state, props, refs, or callbacks, so hoisting them to module scope is safe. Existing URL validation and Markdown security behavior remain unchanged. 

The performance behavior was also adversarially checked:

* Copy/parent re-render with unchanged content → cached Markdown element, no new parse.
* Content change → fresh parse.
* Preview → Source → no unnecessary parse.
* Source → Preview → one fresh parse, which is expected because the Markdown subtree was unmounted.
* The interaction with #1117's focus/ref/layout-effect changes remains compatible.  

Security checks also showed no regression across unsafe protocols, external links, fragments, images, and existing URL handling. 

### One clarification

There is a minor wording issue in the PR description: Preview → Source does not re-parse the Markdown; Source → Preview causes one fresh parse because the Preview subtree has been unmounted. The actual optimization is primarily eliminating unnecessary parsing/remounting during **Preview re-renders**, such as Copy or parent updates. This does not affect the correctness of the implementation. 

The large GFM-table scaling issue remains a legitimate follow-up/product decision rather than something that should be introduced into this PR. The review found no reason to add an arbitrary content threshold here.

### Final verification

* **66/66 tests passing**
* **0 TypeScript errors**
* **Lint clean**
* Security behavior unchanged
* No equivalent `ReactMarkdown` implementation elsewhere requiring the same fix
* No blockers identified
* No changes required from our side

The only worthwhile follow-ups are a browser-level regression test for the live re-render/DOM-identity behavior and a separate product decision around very large GFM tables. Neither is blocking for this PR. 

**Overall, this is a strong and well-reasoned performance fix.**

**Approved from my side.**

PR #1208, opened 2026-08-24
fix(conflicts): resolve TypeError on unhashable values in voting/credibility strategies

### LINKED ISSUE #1207: [BUG] TypeError: unhashable type when resolving VALUE_CONFLICT with dict/list values (voting & credibility strategies)
## Description
`ConflictResolver._resolve_by_voting` and `_resolve_by_credibility` assume all
`conflicting_values` are hashable. When a VALUE_CONFLICT contains dict or list
values, both strategies raise `TypeError: unhashable type: 'dict'` (or 'list'):

- voting uses `collections.Counter(conflicting_values)`
- credibility uses the values as dict keys in `value_weights`

## Reproduction
Create a VALUE_CONFLICT whose `conflicting_values` are dicts or lists, then call
`resolve_conflict(...)` with strategy `voting` or `credibility_weighted`.

## Expected
Resolution should handle unhashable values and return the original value
(preserving its type), without crashing.

## Impact
Any pipeline that surfaces structured (dict/list) values into conflict
resolution crashes. The common fully-string path is unaffected.

## Known limitation (of the proposed fix)
Unhashable values are keyed via `json.dumps(sort_keys=True, default=str)` with a
`repr()` fallback; extremely rare key collisions between a stringified structure
and an equal string value are possible but acceptable for de-duplication.

cxzg007:
## Summary
Fixes a `TypeError: unhashable type` crash in conflict resolution when a
`VALUE_CONFLICT` holds dict or list values.

`_resolve_by_voting` (via `collections.Counter`) and `_resolve_by_credibility`
(via dict keys) both required hashable values. Structured (dict/list) values
therefore crashed resolution.

## Changes
- Add module-level `_hashable_key(value)`: returns hashable values unchanged;
  for unhashable ones derives a stable key via
  `json.dumps(sort_keys=True, default=str)`, falling back to `repr()`.
- `_resolve_by_voting` and `_resolve_by_credibility` now aggregate by key while
  keeping a `key -> original value` map, returning the resolved value with its
  original type.
- The common all-string path is behaviourally unchanged.

## Tests
- `test_resolve_by_voting_with_unhashable_values` (dict values)
- `test_resolve_by_credibility_with_unhashable_values` (list values)
- Full `tests/conflicts/test_conflicts.py` suite: 21 passed.

## Known limitation
Extremely rare key collisions between a stringified structure and an equal
string value are possible but acceptable for de-duplication purposes.

Fixes #1207

### DIFF
--- a/semantica/conflicts/conflict_resolver.py
+++ b/semantica/conflicts/conflict_resolver.py
@@ -56,6 +56,7 @@ Author: Semantica Contributors
 License: MIT
 """
 
+import json
 from collections import Counter
 from dataclasses import dataclass, field
 from enum import Enum
@@ -65,6 +66,25 @@ from ..utils.logging import get_logger
 from ..utils.progress_tracker import get_progress_tracker
 from .conflict_detector import Conflict
 from .source_tracker import SourceTracker
+
+
+def _hashable_key(value: Any) -> Any:
+    """Return a stable, hashable key for a conflicting value.
+
+    Conflicting values may be unhashable (e.g. ``dict`` or ``list``), which
+    breaks :class:`collections.Counter` and dict-key based aggregation used by
+    the voting and credibility-weighted strategies. Hashable values are returned
+    unchanged so existing behavior is preserved; unhashable values are mapped to
+    a stable JSON string key.
+    """
+    try:
+        hash(value)
+        return value
+    except TypeError:
+        try:
+            return json.dumps(value, sort_keys=True, default=str)
+        except (TypeError, ValueError):
+            return repr(value)
 
 
 class ResolutionStrategy(str, Enum):
@@ -361,9 +381,18 @@ class ConflictResolver:
                 resolution_notes="No conflicting values to resolve",
             )
 
-        # Count value occurrences
-        value_counts = Counter(conflict.conflicting_values)
-        most_common_value, count = value_counts.most_common(1)[0]
+        # Count value occurrences (using stable hashable keys so that
+        # unhashable values such as dict/list do not break Counter)
+        key_to_value: Dict[Any, Any] = {}
+        keys: List[Any] = []
+        for value in conflict.conflicting_values:
+            key = _hashable_key(value)
+            key_to_value.setdefault(key, value)
+            keys.append(key)
+
+        value_counts = Counter(keys)
+        most_common_key, count = value_counts.most_common(1)[0]
+        most_common_value = key_to_value[most_common_key]
 
         # Calculate confidence based on vote ratio
         total_votes = len(conflict.conflicting_values)
@@ -391,8 +420,10 @@ class ConflictResolver:
                 resolution_notes="Insufficient data for credibility-based resolution",
             )
 
-        # Weight values by source credibility
+        # Weight values by source credibility (keyed by stable hashable keys so
+        # that unhashable values such as dict/list can be used as map keys)
         value_weights: Dict[Any, float] = {}
+        key_to_value: Dict[Any, Any] = {}
 
         for i, value in enumerate(conflict.conflicting_values):
             source = conflict.sources[i] if i < len(conflict.sources) else {}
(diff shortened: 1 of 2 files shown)


ZohaibHassan16 (commented):
Nice fix, but one thing i noticed while poking around though : not sure if intentional. the _hashable_key function tags hashable values with their type name too, like ("h", type(value).__name__, value). this actually changes how voting works for values that are equal but different types, not just the dict/string aliasing case Codex found.

e.g. on main right now:

```
Counter([1, 1.0, True, 1])  # merges them all, count = 4
```

but with this change it splits into int/float/bool buckets separately, so confidence goes from 1.0 down to 0.5 for the same input. seems like it could matter for real data since sources parsing numbers differently (int vs float) for the "same" value isn't that unusual.

also i think you don't even need the type name for the aliasing fix to work : the "h" vs "j" tags already keep dicts/strings separate from each other, so just ("h", value) without the type should be enough?

Could be intentional, but wanted to flag before approving since it wasn't mentioned in the PR. Lmk what you think

zeekay (dismissed):
The bug is real and the fix is nearly there — one detail in the key changes behavior that the PR doesn't mention, and it's a one-token fix.

Confirmed the crash on current `main`:

```
resolve_conflict(conflicting_values=[{"a":1},{"a":1},{"b":2}], strategy="voting")
-> TypeError: unhashable type: 'dict'
```

and it's gone on this branch, with the original dict (not a stringified key) coming back as `resolved_value`. `pytest tests/conflicts/ -m "not integration"` is 37 passed, 31 subtests on current `main` with this merged.

**The type name in the `"h"` key splits scalars that used to aggregate.**

`_hashable_key` returns `("h", type(value).__name__, value)`, so `1` and `1.0` become different keys. Measured, voting strategy, three sources:

| values | `origin/main` | this branch |
|---|---|---|
| `[1, 1.0]` | `1` @ conf 1.0 | `1` @ conf **0.5** |
| `[1, 1.0, 2]` | `1` @ conf 0.667 | `1` @ conf **0.333** |
| `[True, 1]` | `True` @ conf 1.0 | `True` @ conf **0.5** |

Two sources reporting the same population, one parsed as `int` and one as `float`, used to be two votes for one value and now read as a tie that `most_common` breaks by insertion order. For a conflict resolver over extracted values that's a common shape, and the confidence it reports is the number downstream consumers act on.

It also contradicts the docstring, which says "Hashable values keep their original identity within the `"h"` tag (so equal scalars still aggregate together)". With the type name in the key they don't.

The type name isn't needed for the aliasing case the docstring is defending against. `{"a": 1}` is unhashable and lands in `("j", '{"a": 1}')`; the string `'{"a": 1}'` is hashable and lands in `("h", '{"a": 1}')`. The `"h"`/`"j"` tag alone already keeps them apart. So:

```python
return ("h", value)
```

keeps `test_voting_does_not_alias_struct_with_its_json_text` passing and reproduces `Counter`'s existing semantics exactly for every hashable value, which is what a fix for "unhashable values crash" should do — leave the hashable path alone.

If the `int`/`float`/`bool` split is deliberate, that's a defensible position but it's a separate behavior change: it needs to be called out in the PR description and pinned by a test, because right now it's invisible and the docstring says the opposite.

One smaller thing, worth a comment rather than a change: `json.dumps(..., default=str)` means two distinct unhashable values whose `str()` happens to match will still merge. That's a much narrower corner than the one this fixes and the `repr()` fallback below it has the same property, so I wouldn't chase it — just don't let the docstring promise more than it delivers.

Drop the type name (or document and test the split) and I'm happy with this.

Sameer6305 (approved):
Approved.

This PR fixes the TypeError that occurred when VALUE_CONFLICT contained unhashable dict/list values in the voting and credibility-weighted strategies, while preserving the original value type in the resolution result.

During review, I identified that the initial fix added the Python type name to hashable keys, which unintentionally changed existing aggregation semantics for equal values such as `1`/`1.0` and `True`/`1`.

I pushed the follow-up fix in commit `146a4003`:
- changed the hashable key to `("h", value)` so existing Python equality/hash semantics are preserved
- retained the `"h"` vs `"j"`/`"r"` separation to prevent structured-value/string aliasing
- added regression tests for `1`/`1.0` and `True`/`1`

Validation:
- conflicts test suite: 24 passed
- explicit behavior checks: 13 passed
- `git diff --check`: clean

The changes are scoped to the conflict resolver and its tests. Approving PR #1208.

PR #1212, opened 2026-08-24
fix: guard Agno and OpenClaw integration requests against SSRF

Sameer6305:
## Summary

During a project-wide SSRF audit following the resolution of **#936**, we identified two integration paths that could make outbound HTTP requests without using the project's shared SSRF protection.

This PR hardens both paths while preserving their intended behavior.

### 1. Agno `AgnoKnowledgeGraph.load_urls()` — confirmed SSRF

`AgnoKnowledgeGraph.load_urls()` accepted caller-provided URLs and passed them directly to `urllib.request.urlopen()`.

The existing validation only checked the URL scheme (`http`/`https`). It did not protect against:

* Loopback/private IP addresses
* Cloud metadata endpoints such as `169.254.169.254`
* IPv6 internal addresses
* Hostnames resolving to private addresses
* Redirects to internal destinations

This was a genuine SSRF sink because the `urls` argument flows directly into the network request.

### 2. OpenClaw `OpenClawKGTool` — configuration hardening

`OpenClawKGTool` builds requests from its configured `base_url`. This is an operator-controlled value and is intentionally allowed to point to the local Semantica server (`http://localhost:8000`), so this was classified as a configuration-only risk rather than the same user-controlled SSRF vulnerability.

We added validation to prevent malformed or dangerous URL schemes while preserving the intended private/local deployment use case.

### Agno

* Replaced `urllib.request.urlopen()` with the shared `request_with_ssrf_guard()`.
* Removed the duplicated scheme-only validation.
* Added handling for SSRF validation failures so unsafe URLs are skipped without stopping the remaining ingestion process.
* Redirects and DNS resolution now receive the same SSRF protections used elsewhere in the project.

### OpenClaw

* Added `validate_url_for_request()` validation when `OpenClawKGTool` is initialized.
* Private IPs remain explicitly allowed because localhost is the documented/default deployment target.
* Non-HTTP(S) schemes and malformed URLs are rejected.

## Tests

Added focused security tests covering:

### Agno

* Loopback addresses
* Private IPv4 ranges
* Cloud metadata addresses
* IPv6 internal addresses
* `localhost` hostnames
* Hostnames resolving to private addresses
* Non-HTTP(S) schemes
* Redirects to private/metadata addresses
* Valid public URLs
* Mixed URL lists
* Network failures

### OpenClaw

* Valid localhost/private URLs
* Public URLs
* Invalid schemes
* Malformed URLs
* Default `localhost` configuration
* Trailing slash handling

All relevant tests pass:

* **29/29** new Agno SSRF tests
* **26/26** new OpenClaw tests
* **15/15** existing Agno integration tests

## Relationship to #936

This work was discovered as part of the broader SSRF audit performed after **#936** was resolved by **PR #959**.

#936 itself is already fixed: `SeedDataManager.load_from_api()` now uses the shared SSRF guard.

This PR addresses the **separate integration-level SSRF/hardening findings** discovered during that audit.

### DIFF
--- a/integrations/agno/knowledge_graph.py
+++ b/integrations/agno/knowledge_graph.py
@@ -277,25 +277,22 @@ class AgnoKnowledgeGraph(_KnowledgeBase):  # type: ignore[misc]
     def load_urls(self, urls: List[str]) -> None:
         """Fetch each URL and ingest the response body.
 
-        Only ``http`` and ``https`` schemes are permitted to prevent SSRF.
+        Uses the shared SSRF guard so that ``http`` and ``https`` are the only
+        permitted schemes, private/loopback/link-local/cloud-metadata addresses
+        are blocked by default, DNS resolution is validated, and every redirect
+        hop is re-checked before being followed.
         """
-        import urllib.request
-        from urllib.parse import urlparse
+        from semantica.ingest.ssrf import request_with_ssrf_guard
+        from semantica.utils.exceptions import ValidationError
 
         for url in urls:
-            parsed = urlparse(url)
-            if parsed.scheme not in ("http", "https"):
-                logger.warning(
-                    "Skipping URL with disallowed scheme '%s': %s",
-                    parsed.scheme,
-                    url,
-                )
-                continue
             try:
-                with urllib.request.urlopen(url, timeout=10) as resp:  # noqa: S310
-                    text = resp.read().decode("utf-8", errors="replace")
+                response = request_with_ssrf_guard("GET", url, timeout=10)
+                text = response.text
                 self._ingest_text(text, source=url)
                 logger.info("Loaded URL: %s", url)
+            except ValidationError as exc:
+                logger.warning("Skipping URL (SSRF check failed) %s: %s", url, exc)
             except Exception as exc:
                 logger.warning("Failed to fetch %s: %s", url, exc)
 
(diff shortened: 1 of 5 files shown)


KaifAhmad1 (approved):
Reviewed the diff, the shared SSRF guard module it hooks into, and ran the new + existing test suites locally in a worktree (72/72 new SSRF tests, 135/135 existing Agno integration tests, no regressions).

- Agno `load_urls()` now routes through `request_with_ssrf_guard()`, closing the confirmed SSRF sink (private/loopback/metadata IPs, DNS-rebind TOCTOU, unvalidated redirects were all previously unguarded).
- OpenClaw `OpenClawKGTool` validates `base_url` structurally at construction while correctly preserving `allow_private_ips=True` for its documented localhost/private deployment use case — confirmed all `_post`/`_get` endpoints are hardcoded literals, never caller-supplied, so this is rightly scoped as config hardening rather than a per-request guard.
- Both earlier Qodo findings (incomplete fallback validation, a test hitting real network) were properly fixed in the second commit — verified the fixes are correct, not just marked resolved.

Two minor non-blocking notes left in the conversation (session-per-URL in Agno's loop, effectively-dead `ImportError` fallback branch in OpenClaw) — neither affects correctness or security. LGTM.

PR #1213, opened 2026-08-24
fix: preserve generation kwargs in relation extraction

Sameer6305:
## Summary

Fixes LLM generation parameter propagation in relation extraction and ensures behavior-affecting generation parameters are correctly included in extraction cache keys.

This addresses two related issues:

1. `extract_relations_llm` was silently dropping caller-supplied generation parameters such as `max_tokens` before calling `generate_typed`.
2. Once those parameters are correctly forwarded, the extraction cache must distinguish calls that use different generation settings. Otherwise, a cached result generated with one configuration could be incorrectly returned for a later call using a different configuration.

The fix keeps generation behavior and cache identity consistent across entity, relation, and triplet extraction.

### 1. Generation parameters were dropped in relation extraction

`extract_relations_llm` previously constructed provider kwargs using a small allowlist:

```python
call_kwargs = {}
if "temperature" in kwargs:
    call_kwargs["temperature"] = kwargs["temperature"]
if "verbose" in kwargs:
    call_kwargs["verbose"] = kwargs["verbose"]
```

This meant supported generation parameters such as `max_tokens`, `top_p`, `seed`, and `max_completion_tokens` were silently discarded before reaching `generate_typed`.

For example, a caller could provide:

```python
max_tokens=128000
```

but the relation extraction path would not forward it to the provider.

The implementation now uses:

```python
call_kwargs = kwargs.copy()
call_kwargs["max_retries"] = max_retries
```

This preserves caller-supplied generation parameters while ensuring the explicit `max_retries` argument takes precedence.

The same `call_kwargs` is used by the structured-output fallback, so both provider paths receive the corrected parameters.

### 2. Generation parameters were missing from extraction cache keys

After fixing parameter propagation, a related cache correctness issue became apparent.

The extraction cache is enabled by default and its key is derived from the explicitly supplied cache parameters. Previously, generation parameters were not included in the cache identity.

This meant calls such as:

```text
max_tokens=4096
```

and:

```text
max_tokens=128000
```

could use the same cache entry when all other extraction inputs were identical.

The second call could therefore return the result generated by the first call without invoking the provider with its requested generation settings.

This issue was identified during review and has been addressed as part of this PR.

### Generation kwargs

* Preserve all caller-supplied generation kwargs in `extract_relations_llm`.
* Explicitly set `max_retries` after copying the kwargs.
* Keep the behavior consistent with the existing entity and triplet extraction implementations.
* Ensure both the normal `generate_typed` path and the structured-output fallback use the corrected kwargs.

### Cache keys

Added shared handling for behavior-affecting generation parameters used by the supported generation APIs.

The cache key now accounts for parameters including:

* `max_tokens`
* `max_completion_tokens`
* `temperature`
* `top_p`
* `top_k`
* `seed`
* `frequency_penalty`
* `presence_penalty`
* `stop`
* `logit_bias`
* `user`

Only supported generation parameters are included.

`None` values are omitted so that an unset parameter and an explicitly unset (`None`) parameter do not unnecessarily produce different cache entries.

Sensitive values such as API credentials are not included in the generation cache parameters, and the existing cache implementation continues to filter sensitive keys.

The cache-key handling is applied consistently to:

* `extract_entities_llm`
* `extract_relations_llm`
* `extract_triplets_llm`

This prevents the same class of cache collision across the three LLM extraction paths.

## Tests

Added focused regression coverage for cache behavior.

The tests verify that:

* Different `max_tokens` values bypass the existing cache entry.
* Identical `max_tokens` values reuse the cache.
* Different `temperature` values produce distinct cache behavior.
* Entity extraction distinguishes different `max_tokens` configurations.
* Triplet extraction distinguishes different `max_tokens` configurations.
* Existing `max_tokens` propagation tests continue to pass.

Focused test run:

```text
pytest tests/reproduce_issue_176.py -v
8 passed
```

Additional semantic extraction tests:

```text
32 passed, 3 failed
```

The three failures were verified as pre-existing on the base branch and are unrelated to these changes.

Also verified:

```text
git diff --check
```

with no whitespace errors.

## Scope

The changes are limited to the generation-parameter propagation and extraction-cache correctness issues described above.

Modified files:

```text
semantica/semantic_extract/methods.py
tests/reproduce_issue_176.py
```

No unrelated refactoring or provider-specific behavior changes are included.

## Why this is needed

The generation kwargs and cache identity must describe the same effective generation configuration.

Forwarding `max_tokens` and other generation settings without including them in the cache key would allow stale results generated under a different configuration to be returned.

This PR fixes both sides of that behavior so that:

**generation configuration → provider behavior → cache identity**

remain consistent.

### DIFF
--- a/semantica/semantic_extract/methods.py
+++ b/semantica/semantic_extract/methods.py
@@ -1906,13 +1906,10 @@ Entities found in text: {entities_str}"""
                 "[methods.extract_relations_llm] Calling llm.generate_typed (%s/%s)...",
                 provider, model,
             )
-        # Only forward minimal, safe parameters to provider calls
-        call_kwargs = {}
-        if "temperature" in kwargs:
-            call_kwargs["temperature"] = kwargs["temperature"]
-        if "verbose" in kwargs:
-            call_kwargs["verbose"] = kwargs["verbose"]
-
+        # Forward all caller-supplied generation kwargs so they reach
+        # generate_typed and the underlying provider API. max_retries is
+        # always set from the explicit parameter.
+        call_kwargs = kwargs.copy()
         call_kwargs["max_retries"] = max_retries
 
         # Select schema based on whether temporal extraction is requested
--- a/semantica/semantic_extract/methods.py
+++ b/semantica/semantic_extract/methods.py
@@ -140,6 +140,39 @@ _result_cache = ExtractionCache(
 if not config.get("cache_enabled", True):
     _result_cache.enabled = False
 
+# Generation kwargs that affect provider output and must therefore be part of
+# the cache key. This is the union of all parameters read by _add_if_set across
+# every provider in providers.py. Sensitive values (api_key, token, etc.) are
+# already filtered out by ExtractionCache._generate_key, so they need not be
+# excluded here.
+_GENERATION_CACHE_KEYS = frozenset({
+    "max_tokens",
+    "max_completion_tokens",
+    "temperature",
+    "top_p",
+    "top_k",
+    "seed",
+    "frequency_penalty",
+    "presence_penalty",
+    "stop",
+    "logit_bias",
+    "user",
+})
+
+
+def _generation_cache_params(kwargs: dict) -> dict:
+    """Return the subset of *kwargs* that affects generation output.
+
+    Only keys listed in ``_GENERATION_CACHE_KEYS`` are included so that
+    irrelevant or sensitive caller kwargs do not pollute the cache key.
+    Values that are ``None`` are omitted; a caller passing
+    ``temperature=None`` is equivalent to not passing it at all.
+    """
+    return {
+        k: v for k, v in kwargs.items()
+        if k in _GENERATION_CACHE_KEYS and v is not None
+    }
+
 # Try to import spaCy
 from ..utils.helpers import safe_import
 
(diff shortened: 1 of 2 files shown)


KaifAhmad1 (approved):
## Review

Verified against `providers.py`:

- The original bug is real: `extract_relations_llm`'s old allowlist (`temperature`/`verbose` only) silently dropped `max_tokens`, `top_p`, `seed`, etc. before calling `generate_typed`, while `extract_entities_llm`/`extract_triplets_llm` already forwarded full kwargs. The fix brings relations in line with the existing pattern.
- Forwarding all of `kwargs` into `generate_typed`/`generate_structured`/`generate` is safe — every provider implementation (`_add_if_set`, `.get()`) picks only known keys rather than spreading the dict straight into the underlying SDK call, so unrelated kwargs (`relation_types`, `api_key`, etc.) are silently ignored downstream, not errors.
- `ExtractionCache._generate_key` already strips `api_key`/`token`/`password`/`secret`/`auth`/`authorization` before hashing, so the new cache-key params don't leak credentials.
- Found and flagged one gap: the original `_GENERATION_CACHE_KEYS` only covered the common OpenAI-shaped params, so Anthropic's `system`/`stop_sequences`, Gemini's `candidate_count`, and Ollama's `repeat_penalty`/`num_ctx`/`context_window` were still forwarded to providers without being part of the cache key — reintroducing the same stale-cache bug for those providers (most notably Anthropic's `system` prompt). Fixed in the follow-up commit `d9836e70`, which extends the key set and adds 3 regression tests (verified red/green: they fail without the fix, pass with it).

All 11 tests in `tests/reproduce_issue_176.py` pass locally, and all 10 CI checks are green (`security-scan` finished after a brief pending state).

Approving.

PR #1217, opened 2026-08-25
fix(pipeline): preserve serializer round trips

### LINKED ISSUE #1216: [BUG] PipelineSerializer loses step state and rehydrates handlers as strings
## Bug Description

`PipelineSerializer.serialize_pipeline()` writes dependencies and delta-processing metadata as top-level step fields, but `PipelineBuilder.build_pipeline()` only reconstructs the nested `config`. A serialize/deserialize round trip therefore silently drops:

- `dependencies`
- `delta_mode`
- `base_version_id`
- `target_version_id`

There is a related runtime failure for handlers. `PipelineBuilder.add_step()` currently keeps an explicit handler inside `PipelineStep.config`; JSON serialization converts that callable to a string through `default=str`. Deserializing that payload restores the string as `step.handler`, and execution later fails with `TypeError: 'str' object is not callable`.

## Steps to Reproduce

```python
from semantica.pipeline import PipelineBuilder
from semantica.pipeline.pipeline_builder import PipelineSerializer

def handler(data):
    return data

builder = PipelineBuilder()
builder.add_step("extract", "source", handler=handler)
builder.add_step(
    "index",
    "sink",
    delta_mode=True,
    base_version_id="v1",
    target_version_id="v2",
)
builder.connect_steps("extract", "index")
pipeline = builder.build("incremental-index")

serializer = PipelineSerializer()
restored = serializer.deserialize_pipeline(
    serializer.serialize_pipeline(pipeline, format="json")
)

extract, index = restored.steps
print(extract.handler)
print(index.dependencies)
print(index.delta_mode, index.base_version_id, index.target_version_id)
```

## Expected Behavior

- Dependencies and delta-version fields survive dict and JSON round trips.
- Runtime handler callables are not persisted in serialized configuration.
- Legacy JSON containing a stringified handler is sanitized during deserialization.
- Ordinary business configuration remains intact.
- Passing a dict to deserialization does not mutate caller-owned data.

## Actual Behavior

- Dependencies become an empty list.
- Delta mode becomes `False`; base and target version IDs become `None`.
- A legacy/stringified handler can be restored as a truthy string and called at runtime.

## Environment

- **OS**: macOS 15.3.1
- **Python Version**: 3.12.14
- **Semantica Version**: `main` at `2075eca0`

## Possible Solution

Treat structural fields and runtime handlers as separate from user configuration:

1. Restore dependencies and delta-version fields from the serialized step schema.
2. Exclude runtime handlers and structural reserved keys from serialized `config`.
3. Sanitize legacy handler values before reconstruction without mutating input dictionaries.
4. Add dict/JSON round-trip and legacy-payload regression coverage.

I would like to work on this if the approach looks right. I have a focused fix and regression tests prepared on `fix/pipeline-serializer-roundtrip`.

cxzg007:
## Description

Fix `PipelineSerializer` round trips so step dependencies and delta-processing metadata are preserved. Runtime handler callables are now treated as process-local state: they are omitted from serialized configuration, and legacy stringified handlers are ignored during deserialization instead of being rehydrated as non-callable strings.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Fixes #1216

## Changes Made

- Restore step dependencies, delta mode, and base/target version IDs from the serialized schema.
- Exclude runtime handlers and structural reserved keys from serialized business configuration.
- Sanitize legacy handler values without mutating caller-owned dict inputs.
- Add dict and JSON regression tests for structural round trips, handler omission, business config preservation, and legacy payloads.

## Testing

- [x] Tested locally
- [x] Added tests for the fixed functionality

### Test Commands

```bash
.venv312/bin/python -m pytest tests/core tests/pipeline -q
# 52 passed

.venv312/bin/python -m compileall -q semantica/pipeline
.venv312/bin/python -m black --check --fast tests/pipeline/test_pipeline_serializer.py
.venv312/bin/python -m flake8 tests/pipeline/test_pipeline_serializer.py --max-line-length=88 --extend-ignore=E203,W503
git diff --check upstream/main...HEAD
```

An independent code review found one legacy-payload compatibility issue; it was fixed with a RED/GREEN regression test and the follow-up review reported no remaining Critical or Important issues.

## Documentation

## Breaking Changes

**Breaking Changes**: No

Serialized handlers were never portable callables; deserialization now consistently leaves them unbound. Applications should rebind runtime handlers after loading a pipeline.

## Checklist

- [x] I have commented the non-obvious legacy-handler sanitization behavior
- [x] My changes generate no new warnings in the affected tests

### DIFF
--- a/semantica/pipeline/pipeline_builder.py
+++ b/semantica/pipeline/pipeline_builder.py
@@ -272,7 +272,19 @@ class PipelineBuilder:
                 step_name = step_config.get("name")
                 step_type = step_config.get("type")
                 if step_name and step_type:
-                    self.add_step(step_name, step_type, **step_config.get("config", {}))
+                    step = self.add_step(
+                        step_name, step_type, **step_config.get("config", {})
+                    )
+                    step.dependencies = list(
+                        step_config.get("dependencies", step.dependencies)
+                    )
+                    step.delta_mode = step_config.get("delta_mode", step.delta_mode)
+                    step.base_version_id = step_config.get(
+                        "base_version_id", step.base_version_id
+                    )
+                    step.target_version_id = step_config.get(
+                        "target_version_id", step.target_version_id
+                    )
 
             # Set parallelism if specified
             if "parallelism" in pipeline_config:
@@ -398,14 +410,29 @@ class PipelineSerializer:
 
         Returns:
             Serialized pipeline
+
+        Notes:
+            Step handlers are runtime callables and are intentionally omitted from
+            the serialized representation. They must be rebound after deserialization.
         """
+        reserved_config_keys = {
+            "handler",
+            "dependencies",
+            "delta_mode",
+            "base_version_id",
+            "target_version_id",
+        }
         pipeline_data = {
             "name": pipeline.name,
             "steps": [
                 {
                     "name": step.name,
                     "type": step.step_type,
-                    "config": step.config,
+                    "config": {
+                        key: value
+                        for key, value in step.config.items()
+                        if key not in reserved_config_keys
+                    },
                     "dependencies": step.dependencies,
                     "delta_mode": getattr(step, "delta_mode", False),
                     "base_version_id": getattr(step, "base_version_id", None),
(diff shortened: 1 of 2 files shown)


ZohaibHassan16 (approved):
Reproduced this on main and the fix works for me. The metadata survives the round trip no and handlers no longer get restored as strings.

Ran the full suite as well : 53 passing. I also checked the orchestrator.py caller and nothing stood out there.

Nice fix. Approving.

PR #1220, opened 2026-08-25
refactor(ner): remove dead _extract_with_spacy method and unused self.nlp

### LINKED ISSUE #1058: refactor(semantic_extract): remove dead NERExtractor._extract_with_spacy() method
## Problem

`NERExtractor._extract_with_spacy()` in `semantica/semantic_extract/ner_extractor.py` is unreachable dead code.

The method contains a complete spaCy extraction implementation and uses `self.nlp`, but there are no callers anywhere in the repository.

## Current Behavior

The active `method="ml"` execution path is:

```text
NERExtractor.extract_entities()
  → get_entity_method("ml")
  → extract_entities_ml()
  → load_spacy_model(model)
  → nlp(text)
````

`_extract_with_spacy()` is not part of this execution path.

A repository-wide search confirmed:

* no production callers;
* no test callers;
* no subclasses overriding the method;
* no dynamic/reflection access;
* the method is not exposed through `__all__`;
* it is not documented as a public API.

The method operates on `self.nlp`, which is assigned during `NERExtractor` initialization, but the active extraction implementation does not consume that attribute.

## Impact

Keeping this orphaned implementation creates misleading internal architecture:

* developers may assume `_extract_with_spacy()` is an active alternative spaCy extraction path;
* `self.nlp` appears to be required by the extractor when its only known consumer is unreachable;
* the dead implementation must be maintained alongside the actual extraction path;
* future changes to entity extraction can leave this unused implementation increasingly inconsistent with the live code.

## Expected Behavior

The `NERExtractor` implementation should contain only active extraction paths and state that is actually required.

## Suggested Direction

1. Remove `_extract_with_spacy()` from `NERExtractor`.
2. Evaluate whether `self.nlp` in `__init__()` is still required after removing the method.
3. If `self.nlp` is intentionally retained as an inspectable/internal attribute, document that purpose explicitly.
4. Run the existing NER test suite after the cleanup.

## Acceptance Criteria

* `_extract_with_spacy()` is removed.
* A repository-wide search returns no references to `_extract_with_spacy`.
* Existing NER tests continue to pass.
* No active extraction path is changed.
* Any now-unused `self.nlp` initialization is either removed or has a documented purpose.

### PR DESCRIPTION
**Pull request by yunaremaia**
## Summary

Removes the unreachable `_extract_with_spacy()` method tracked in #1058, plus the `self.nlp` instance attribute that only that method ever read.

## Details

- The ML dispatch path goes through `get_entity_method("ml") -> extract_entities_ml()`, which loads the spaCy model lazily via the process-level cache in `methods.py` — `self.nlp` was never consulted.
- `__init__` still validates the spaCy runtime up front (preserving the `_ml_runtime_usable` gate) but no longer eagerly loads a model that nothing on the instance uses.
- Updated the runtime-failure test to assert the gate flag instead of the removed attribute.

## Verification

- `grep -rn '_extract_with_spacy' semantica/ tests/ apps/ examples/`: no callers or references remain.
- `tests/test_ner_configurations.py`: 8/8 passing locally.
- `tests/semantic_extract/`: 155 passed; the 3 failures in `test_performance.py` reproduce on a clean main checkout and are unrelated.

Fixes #1058

Signed-off-by: Yunare Maia <yunare@gmail.com>

### CODE REVIEW

**Review by ZohaibHassan16 — COMMENTED** [review]
Found something that's gonna break CI: tests/split/test_spacy_model_cache.py also uses NERExtractor.nlp : there's a whole test class in there (TestNERExtractorSpacyModelCache) checking cache reuse/retry behavior through .nlp identity comparisons. ran it against your branch and got:

```
FAILED test_ner_extractor_reuses_cached_model_across_instances — AttributeError: 'NERExtractor' object has no attribute 'nlp'
FAILED test_ner_extractor_distinct_model_names_load_separately — AttributeError: 'NERExtractor' object has no attribute 'nlp'
FAILED test_ner_extractor_failed_load_not_cached_and_retried — AttributeError: 'NERExtractor' object has no attribute 'nlp'
```

pretty sure this is why it slipped through : your grep was for _extract_with_spacy specifically, not .nlp, and the test run you listed was scoped to tests/semantic_extract/ but this file's in tests/split/ so it never got run.

Should be a small fix though, those tests are just checking cache identity/reuse , probably just needs rewriting to assert on the load-call counts (looks like they already track those with the attempts/calls lists) instead of comparing .nlp directly. Not saying keep the attribute around, just that those assertions need to change too.

lmk once that's sorted and i'll take another look

Sameer6305 (approved):
Thanks **@yunaremaia** for addressing the dead-code cleanup and for incorporating the review feedback.

Also, thanks **@ ZohaibHassan16** for the initial review and for catching the missed `NERExtractor.nlp` references in `tests/split/test_spacy_model_cache.py`. That was an important catch—the initial test run did not cover that directory, so those failures could otherwise have reached CI.

I completed a final pass over the updated PR, including:

* Reviewed the complete NER refactor and active ML extraction path.
* Verified `_extract_with_spacy()` has no remaining callers or dynamic references.
* Verified removing `NERExtractor.nlp` does not affect the active extraction path.
* Confirmed `_ml_runtime_usable` and the runtime-failure behavior remain intact.
* Reviewed the updated cache tests to ensure they still cover model reuse, model separation, and failed-load retry behavior.
* Performed a repository-wide search for remaining `.nlp` / `_extract_with_spacy` dependencies.
* Reviewed cache behavior for failure handling, model isolation, and thread-safety implications.
* Checked for security, robustness, and future-maintenance concerns.
* Confirmed no unintended production changes were introduced.

### Verification

* `tests/split/test_spacy_model_cache.py` — **12 passed**
* `tests/test_ner_configurations.py` — **8 passed**
* `tests/semantic_extract/ tests/split/` — **210 passed, 11 skipped**
* `tests/conflicts/test_config_bool_env.py` — **15 passed**
* `git diff --check` — **clean**

I also noted two minor test-quality considerations around cache identity assertions, but the relevant cache identity behavior is covered by the dedicated cache tests, so these are not blockers.

The additional config-related changes visible in the branch are from a commit already merged from `main` and are not part of the NER refactor itself.

**Final verdict: ✅ Ready to merge.**

PR #1171, opened 2026-08-21
fix(ontology): retain data properties for normalized class names

T1mn:
## Description

Keep data properties when raw entity types are normalized into ontology class names. For example, entities typed as `software engineer` produce `SoftwareEngineer`, and their `email` property is now emitted with the correct class domain and `xsd:string` range.

I had already prepared a local fix for this behavior and am sharing it for review. If a different scope or approach is preferred, please feel free to close this PR; I appreciate the guidance.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #1169

## Changes Made

- Match raw entity types to raw, normalized, or recorded class names.
- Group entity attributes by the inferred class definition.
- Use the normalized class name as each data property domain.
- Add regression tests for direct property inference and the full ontology pipeline.

## Testing

- [x] Tested locally
- [x] Added tests for new functionality
- [ ] Package builds successfully (`python -m build`)

### Test Commands

```bash
PYTHONPATH=. pytest -q tests/ontology/test_ontology_normalized_properties.py tests/ontology/test_ontology_comprehensive.py tests/ontology/test_ontology_classes.py tests/ontology/test_ontology_advanced.py
# 79 passed, 1 skipped
```

## Documentation

- [ ] Updated relevant documentation
- [ ] Added code examples if applicable
- [ ] Updated API reference if adding new APIs
- [ ] Updated cookbook if adding new examples
- [x] No documentation changes needed

## Breaking Changes

**Breaking Changes**: No

## Checklist

- [x] My code follows the project style guidelines
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] Package builds successfully
- [x] Commit messages follow conventions
- [x] No merge conflicts

## Additional Notes

The working tree was clean before this focused change.

PR #1226, opened 2026-08-27
fix(pipeline): set_parallelism now enables dependency-layer parallel execution (closes #1223)

### LINKED ISSUE #1223: [BUG] PipelineBuilder.set_parallelism() does not affect pipeline execution
## Bug Description

`PipelineBuilder.set_parallelism()` stores the requested level in `Pipeline.config`, but normal pipeline execution never consumes that value.

`ExecutionEngine._execute_steps()` topologically sorts the steps and then executes every step sequentially in a `for` loop. The engine creates a `ParallelismManager`, but it is not connected to the normal step execution path.

As a result, `builder.set_parallelism(2)` has no effect, despite `docs/guides/pipeline.md` stating that dependency-ready steps may run concurrently.

## Steps to Reproduce

```python
import threading
import time

from semantica.pipeline import ExecutionEngine, PipelineBuilder

active = 0
peak = 0
lock = threading.Lock()


def slow_branch(data):
    global active, peak
    with lock:
        active += 1
        peak = max(peak, active)

    time.sleep(0.1)

    with lock:
        active -= 1

    return {**data, "done": True}


builder = PipelineBuilder()
builder.add_step("source", "source", handler=lambda data: data)
builder.add_step(
    "branch_a",
    "worker",
    handler=slow_branch,
    dependencies=["source"],
)
builder.add_step(
    "branch_b",
    "worker",
    handler=slow_branch,
    dependencies=["source"],
)
builder.set_parallelism(2)

pipeline = builder.build("parallel-example")
ExecutionEngine(max_workers=2).execute_pipeline(pipeline, {})

print(peak)
```

## Actual Behavior

`peak` is `1`. Both dependency-ready sibling steps run sequentially.

## Expected Behavior

`set_parallelism(n)` should limit concurrent pipeline-step execution, while preserving the current serial behavior by default.

Parallel execution should be explicitly opt-in per step:

```python
builder.add_step(
    "branch_a",
    "worker",
    handler=slow_branch,
    dependencies=["source"],
    parallel_safe=True,
)
```

A step should run concurrently only when:

1. Its dependencies have completed.
2. It belongs to a dependency level containing only `parallel_safe=True` steps.
3. The pipeline parallelism is greater than one.
4. The shared input is a dictionary and can be copied safely.
5. The engine worker limit has not been reached.

### 1. Add an explicit parallel-safety contract

Add `parallel_safe: bool = False` to `PipelineStep`.

`PipelineBuilder.add_step()` should pop `parallel_safe` from the control fields so it is not forwarded to the handler through `step.config`.

The value must survive dict and JSON serialization round trips as a top-level step field.

### 2. Preserve serial behavior by default

Calculate the effective worker count as:

```text
min(pipeline.config.get("parallelism", 1), engine.max_workers)
```

Rules:

- No `set_parallelism()`: execute serially.
- `set_parallelism(1)`: execute serially.
- Linear dependency chains remain serial.
- `set_parallelism()` must reject booleans, non-integers, zero, and negative values with `ValidationError`.

### 3. Execute by dependency level

Reuse the existing dependency-level analysis.

For each level:

- Execute serially when it contains one step.
- Execute serially when any step is not `parallel_safe`.
- Execute serially when the current input is not a dictionary.
- Execute serially when input copies cannot be created before any handler starts.
- Otherwise, execute the entire level concurrently with a bounded thread pool.

The initial implementation should use an all-or-nothing rule for each dependency level. It should not mix serial and parallel mutations inside the same level.

Delta-mode steps should remain serial in the initial implementation.

### 4. Isolate inputs

Create a separate `copy.deepcopy(current_data)` for every parallel step before submitting any task.

If copying fails, fall back to the existing serial path before running any handler. This prevents shared mutable inputs and avoids executing handlers twice.

### 5. Merge dictionary deltas deterministically

Every `parallel_safe=True` handler must return a dictionary.

Compare each branch result with the common input and merge only keys that were added or changed.

Example:

```python
base = {"text": "Alice works at Acme"}

ner = {
    "text": "Alice works at Acme",
    "entities": ["Alice", "Acme"],
}

triplets = {
    "text": "Alice works at Acme",
    "triplets": [("Alice", "works_at", "Acme")],
}
```

Merged result:

```python
{
    "text": "Alice works at Acme",
    "entities": ["Alice", "Acme"],
    "triplets": [("Alice", "works_at", "Acme")],
}
```

Merge in pipeline declaration order rather than task completion order.

Conflict rules:

- Different keys: merge both.
- Same key and equal values: accept once.
- Same key and different values: raise `ProcessingError` containing the key and both step names.
- Missing input keys do not represent deletion.
- Ambiguous equality for complex values should be treated as a conflict.

If a parallel-safe handler returns a non-dictionary value, raise `ProcessingError`. Do not rerun it serially because handlers may have external side effects.

### 6. Reuse normal execution semantics

Parallel tasks must use the same single-step execution path as serial tasks.

Refactor the existing step/retry logic into a reusable helper rather than invoking handlers directly through `ParallelismManager.execute_pipeline_steps_parallel()`.

Parallel execution must preserve:

- delta interception for serial delta steps;
- retry policies;
- `StepStatus`, results, and errors;
- progress tracking;
- execution options;
- pipeline failure reporting.

If one parallel branch fails after retries:

1. Cancel tasks that have not started where possible.
2. Allow already-running threads to finish.
3. Discard the partial merged result.
4. Do not run downstream dependency levels.
5. Mark the pipeline as failed.

Pause and stop may take effect between dependency levels. Forcefully terminating running Python threads is out of scope.

## Acceptance Criteria

- [ ] Two eligible sibling steps overlap when pipeline parallelism is two.
- [ ] Active workers never exceed the effective worker limit.
- [ ] Pipelines without `set_parallelism()` behave exactly as before.
- [ ] `set_parallelism(1)` preserves the serial path.
- [ ] Linear dependency chains remain serial.
- [ ] A dependency level containing any unsafe step runs serially.
- [ ] `parallel_safe` is not forwarded to handler keyword arguments.
- [ ] Parallel handlers receive isolated input copies.
- [ ] Independent dictionary changes merge correctly.
- [ ] Equal writes to the same key are accepted.
- [ ] Conflicting writes raise a deterministic `ProcessingError`.
- [ ] Non-dictionary input falls back before handlers start.
- [ ] Non-dictionary parallel output fails without executing the handler twice.
- [ ] Retry, status, progress, and failure behavior remain consistent.
- [ ] A failed branch prevents downstream execution.
- [ ] Dict and JSON round trips preserve `parallel_safe`.
- [ ] Invalid parallelism values raise `ValidationError`.

## Testing Notes

Concurrency tests should use `threading.Barrier`, `threading.Event`, and a locked active-worker counter. They should not rely only on elapsed wall-clock time, which is unreliable in CI.

## Likely Files

- `semantica/pipeline/pipeline_builder.py`
- `semantica/pipeline/execution_engine.py`
- `semantica/pipeline/parallelism_manager.py`
- `tests/pipeline/`
- `docs/guides/pipeline.md`

## Risks

- Handler side effects may not be thread-safe.
- Deep-copying large inputs may increase memory use.
- Progress and retry components may expose thread-safety assumptions.
- Already-running Python threads cannot be forcibly cancelled.

These risks are why parallel execution should remain opt-in through `parallel_safe=True`.

## Out of Scope

- Process-based or distributed execution.
- User-defined reducers.
- Non-dictionary output merging.
- Deleting fields through branch outputs.
- Forcefully terminating running threads.
- Node-by-node dynamic scheduling; dependency-level barriers are sufficient for the initial implementation.

## Additional Context

This is not only a performance optimization. It also aligns the runtime with the documented `set_parallelism()` behavior while retaining a safe, backward-compatible serial default.

cxzg007:
## Summary

Fixes #1223.

`PipelineBuilder.set_parallelism()` accepted a value and stored it in the pipeline config, but the execution engine never read it — steps always ran strictly sequentially. This PR wires the config through the full stack (builder → serializer → engine) and adds dependency-layer parallel execution.

## Root cause

- `set_parallelism()` validated and stored `parallelism` in `config`, but `ExecutionEngine._execute_steps()` had no parallel code path.
- `parallelism` was also lost across a serialize/deserialize round trip (nested `config` key was never promoted to the top-level dict that `build_pipeline()` reads).

### `semantica/pipeline/pipeline_builder.py`
- `PipelineStep` gains an opt-in `parallel_safe: bool = False` field; `add_step()` pops it from kwargs so it never leaks into handler config.
- `set_parallelism()` validates input (bool / positive int, `ValidationError` otherwise).
- `build_pipeline()`, `PipelineBuilder.serialize()`, and `serialize_pipeline()` round-trip `parallel_safe`.
- `deserialize_pipeline()` promotes the nested `config` key to the top level so `parallelism` survives a round trip.

### `semantica/pipeline/execution_engine.py`
- `_execute_steps()` now groups steps into dependency layers (declaration order preserved within each layer).
- A layer runs in parallel only when ALL of: >1 step, dict-typed input, every step `parallel_safe`, no step in `delta_mode`. Otherwise the layer falls back to sequential execution.
- Parallel execution: each step's input is deep-copied for isolation; bounded via `ThreadPoolExecutor(max_workers=min(config parallelism, max_workers))`; retries / `StepStatus` / result / error / progress reporting are shared with the sequential path.
- On step failure, pending futures are cancelled and the error propagates so downstream layers never run.
- Layer results are merged back in declaration order; conflicting values for the same key raise `ProcessingError`.

### `docs/guides/pipeline.md`
- Documented effective parallelism cap, `parallel_safe` opt-in, fallback semantics, dict contract, isolation, merge order and conflict behavior.

## Design notes

- Parallelism is opt-in per step (`parallel_safe`) because handlers that share mutable state or rely on strict ordering are unsafe to run concurrently.
- `ParallelismManager.execute_pipeline_steps_parallel()` is intentionally NOT used; the engine implements its own bounded layer scheduler so retry/status semantics stay identical between sequential and parallel paths.

## Tests

- New `tests/pipeline/test_pipeline_parallel.py`: 22 tests, all passing.
- Full regression: `tests/pipeline` 61 passed, `tests/core` 19 passed, `tests/conflicts` 19 passed.
- Note: `tests/test_pipeline_orchestration.py` has 14 pre-existing failures on unmodified `main` (stale chained-API tests, e.g. `PipelineStep.build()`); verified via `git stash` baseline comparison that these are unrelated to this PR.

Closes #1223

### DIFF
--- a/docs/guides/pipeline.md
+++ b/docs/guides/pipeline.md
@@ -241,7 +241,18 @@ engine   = ExecutionEngine(max_workers=2, retry_on_failure=True)
 result   = engine.execute_pipeline(pipeline)
 ```
 
-`set_parallelism(n)` tells the engine how many steps it may run simultaneously. The topological sort guarantees that only steps whose dependencies are all completed are eligible for concurrent execution — you cannot accidentally run a step before its inputs are ready.
+`set_parallelism(n)` tells the engine how many steps it may run simultaneously; `n` must be a positive integer. The topological sort guarantees that only steps whose dependencies are all completed are eligible for concurrent execution — you cannot accidentally run a step before its inputs are ready. The effective concurrency is capped at `min(n, max_workers)`, so the engine's `max_workers` setting remains a hard resource ceiling.
+
+Concurrency is opt-in per step. A dependency layer only runs in parallel when every step in that layer is marked `parallel_safe`, the layer has more than one step, and the data flowing into the layer is a dict:
+
+```python
+builder.add_step("ner",      "ner_extract",     parallel_safe=True, confidence_threshold=0.75)
+builder.add_step("triplets", "triplet_extract", parallel_safe=True, include_temporal=True)
+```
+
+If any step in a layer is not marked `parallel_safe`, or if a step runs in delta mode, the entire layer falls back to sequential execution — parallelism never silently bypasses a step that was not declared safe. `parallel_safe` is a control field: like `dependencies`, it is consumed by the builder and never reaches your handler's config.
+
+Parallel-safe handlers must return a dict. Each step in a parallel layer receives an isolated deep copy of the layer's input, so steps cannot see each other's mutations. The per-step results are merged key by key in step declaration order: a key written by one step is added to the merged output, a key written by several steps with equal values is kept, and two steps writing different values for the same key fail the pipeline with a `ProcessingError` naming the conflicting key and both steps. Handlers that touch shared mutable resources — database connections, in-memory stores, global caches — should not be marked `parallel_safe`.
 
 ## Common Pitfalls
 
--- a/semantica/pipeline/execution_engine.py
+++ b/semantica/pipeline/execution_engine.py
@@ -61,6 +61,14 @@ class PipelineStatus(Enum):
     STOPPED = "stopped"
 
 
+class _ParallelResultContractError(ProcessingError):
+    """Raised when a parallel step violates the dict-result contract.
+
+    Contract violations are deterministic: re-running the handler cannot
+    change its return type, so the retry loop must be skipped entirely.
+    """
+
+
 @dataclass
 class ExecutionResult:
     """Pipeline execution result."""
@@ -271,7 +279,11 @@ class ExecutionEngine:
 
             if self._can_run_layer_in_parallel(layer, current_data):
                 merged = self._execute_parallel_group(
-                    layer, current_data, effective_parallelism, **options
+                    layer,
+                    current_data,
+                    effective_parallelism,
+                    pipeline_name=pipeline.name,
+                    **options,
                 )
                 if merged is None:
                     # Input isolation failed before any handler started;
@@ -316,6 +328,7 @@ class ExecutionEngine:
                 step,
                 current_data,
                 step_label=f"Step {step_idx + 1}/{total_steps}: {step.name}",
+                pipeline_name=pipeline.name,
                 **options,
             )
 
@@ -326,6 +339,8 @@ class ExecutionEngine:
         step: PipelineStep,
         data: Any,
         step_label: Optional[str] = None,
+        pipeline_name: Optional[str] = None,
+        require_dict_result: bool = False,
         **options,
     ) -> Any:
         """
@@ -335,16 +350,34 @@ class ExecutionEngine:
         policies, step status tracking and progress reporting behave
         identically. Returns the step result, or raises the final error
         after retries are exhausted.
+
+        When ``require_dict_result`` is set (parallel layers), a handler
+        returning a non-dict raises ProcessingError before the step is
+        marked completed, so status and progress reporting stay consistent.
+        Such contract violations are never retried: the handler's return
+        type cannot change between attempts.
         """
         step_tracking_id = self.progress_tracker.start_tracking(
             module="pipeline",
-            submodule=step.step_type or step.name,
+            # Pipeline identity + step name keep tracking IDs unique so
+            # concurrent steps of the same step_type cannot overwrite each
+            # other's progress records.
+            submodule=(
+                f"{pipeline_name or 'pipeline'}:"
+                f"{step.step_type or 'step'}:{step.name}"
+            ),
             message=step_label or f"Executing step: {step.name}",
         )
 
         try:
             step.status = StepStatus.RUNNING
             step_result = self._execute_step(step, data, **options)
+            if require_dict_result and not isinstance(step_result, dict):
+                raise _ParallelResultContractError(
+                    f"Step '{step.name}' is marked parallel_safe and must "
+                    f"return a dict so parallel results can be merged, got "
+                    f"{type(step_result).__name__}"
+                )
             step.status = StepStatus.COMPLETED
             step.result = step_result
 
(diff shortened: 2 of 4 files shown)


ZohaibHassan16 (changes requested):
There is one blocking issue before we merge this.

This branch currently includes `23baf21d fix(ontology): retain data properties for normalized class names (#1171)`, plus the ontology source and test files. That commit is unrelated to the pipeline parallelism change and has already been merged.

Could you rebase this branch onto current main and keep only the pipeline commits? 

Other than that, the scheduler changes and the follow-up fixes look good to me.

ZohaibHassan16 (approved):
I wrote my own tests on top of the already 22 to try and break the merge/isolation logic specifically:

- two parallel steps writing different values to the same new key → correctly raises with both step names and values
- two steps writing distinct keys while both echoing the same base key → merges fine, no false conflict
- one step mutating a nested dict, a sibling reading the same key → sibling sees None, so the deep-copy isolation is real, not just decorative
- circular deps → clean ValidationError, no stack overflow
- two steps fail, a downstream step depends on both → downstream never runs

All of it held up. The merge function is more careful than I expected too : it tells apart "step echoed the input unchanged" from "step actually wrote this", so two handlers both doing {**data, ...} don't fight over every untouched key.

Ran the full suite, 123 passed.

Good work, approving.

PR #1230, opened 2026-08-27
fix(ontology): reject normalized class name collisions

T1mn:
## Description
Class inference normalizes source entity types before emitting ontology classes. Distinct source types such as Person and person can therefore produce duplicate class names, while downstream property inference silently maps both sources to the first class.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Follow-up to merged PR #1171.

## Changes Made

- Detect source entity types that normalize to the same class name.
- Fail with a structured ValidationError before duplicate classes and misassigned properties are emitted.
- Add regression coverage for the Person/person collision.

## Testing

- [x] Tested locally
- [ ] Package builds successfully (python -m build)

### Test Commands

```bash
PYTHONPATH=. pytest -q tests/ontology/test_ontology_normalized_properties.py tests/ontology/test_ontology_comprehensive.py
python -m compileall -q semantica/ontology/class_inferrer.py tests/ontology/test_ontology_normalized_properties.py
```

## Documentation

## Breaking Changes

**Breaking Changes**: No

## Checklist

- [x] My code follows the project style guidelines

## Additional Notes

This is a focused follow-up that prevents ambiguous normalized class identifiers from producing a silently incorrect ontology. Thank you for reviewing it.


ZohaibHassan16 (approved):
The new check matches the exact `min_occurrences` filter used by the class-emission loop right below it, over the same `entity_types` dict. So it shouldn't flag anything that would've been filtered out anyway, and it shouldn't miss a collision for something that would actually get emitted. 

`ValidationError` usage looks good too. `validation_context` is actually stored and exposed in `exceptions.py`, so callers get the colliding types back rather than just a string error.

Ran both test files plus the full ontology suite:

* 44 passed across the two listed tests
* 95 passed / 4 skipped across the whole module


Only thing I'd mention in the PR description, not as a blocker: this propagates unconditionally through every caller (`OntologyEngine` and both `OntologyGenerator` overloads). There's no merge/rename recovery path, so if two extraction passes produce `Person` and `person`, generation now hard-fails instead of silently assigning properties to the wrong class.

Given the preference here seems to be loud failure over silently wrong output, that seems intentional. Just leaving it here.

Approving.

PR #1234, opened 2026-08-28
feat(ingest): add SAP OData ingestor (#1228)

### LINKED ISSUE #1228: [FEATURE] Add SAP OData ingestor for enterprise ERP data
## Problem Statement

`semantica/ingest/` has no connector for SAP, one of the most common enterprise systems-of-record (ERP: Financials, Materials Management, Sales & Distribution, HR via SuccessFactors). Teams wanting to build a knowledge graph over SAP master/transactional data (Business Partners, Sales Orders, Material Master, etc.) have no supported path today.

SAP exposes data through several very different surfaces, so scope needs to be picked deliberately:

| Surface | Access | Packaging |
|---|---|---|
| OData services (S/4HANA Cloud/on-prem, SuccessFactors, Ariba) | REST/OData v2/v4 over HTTP(S) | Pure Python, pip-installable |
| SAP RFC (BAPIs, older ECC systems) | `pyrfc` | Requires the proprietary **SAP NW RFC SDK** (not on PyPI, needs an SAP account to download, per-platform binaries) |
| SAP Business API / BTP APIs | REST, OAuth2 | Pure Python |

## Proposed Solution

Scope **phase 1 to the OData/REST surface only** — it covers the modern S/4HANA Cloud, SuccessFactors, and BTP-hosted APIs without forcing a proprietary SDK dependency on anyone who just `pip install`s the extra. RFC/BAPI support (older on-prem ECC) can be a follow-up behind a separate extra once someone has an SDK license to test against, and shouldn't block this issue.

```python
semantica/ingest/sap_ingestor.py
├── SAPODataConnector      # auth (OAuth2 client-credentials or Basic) + requests.Session, reuses ssrf.py guard like RESTIngestor
├── SAPODataEntity         # dataclass: records, entity_set, count, service, metadata, ingested_at
└── SAPIngestor
    ├── ingest_entity_set(service, entity_set, select=None, filter=None, top=None, skip=None)  # OData $select/$filter/$top paging
    ├── discover_service(service)                                                              # parse $metadata (EDMX) for entity sets/fields
    └── export_as_documents(data)                                                               # -> entity dicts for GraphBuilder
```

Auth (matching what SAP's OData gateway actually requires):
- OAuth2 client-credentials (SAP BTP / S/4HANA Cloud standard)
- Basic auth (common for on-prem NetWeaver Gateway in dev/test landscapes)

Wire-up (same as every other connector in this module):
- `pyproject.toml`: `ingest-sap = ["requests>=2.28.0"]` (no exotic dependency — reuse the `requests`-based pattern from `api_ingestor.py`/`ssrf.py` rather than pulling in an OData-specific client library, unless a contributor has a strong reason to prefer `pyodata`)
- `semantica/ingest/__init__.py`: register `SAPIngestor` / `SAPODataEntity` / `SAPODataConnector`
- `semantica/ingest/ingest_usage.md`: usage example
- Must route outbound requests through `ssrf.py`'s `request_with_ssrf_guard`, same as `api_ingestor.py` — SAP endpoints are user-supplied URLs like any other API ingestion target

## Use Cases

1. Ingesting Business Partner / Customer master data from an S/4HANA Cloud OData service into the knowledge graph.
2. Pulling Sales Order headers + line items (`$expand=to_Item`) to model order → line-item → material relationships.
3. Discovering an unfamiliar SAP service's entity sets and fields via `$metadata` before writing an extraction query, since SAP schemas are dense and vary heavily by module/customization.

## Impact Assessment

- **Who would benefit?** Enterprises with SAP as their ERP system-of-record — a very common ask alongside Salesforce for "connect our enterprise data."
- **Priority**: Medium (High demand, but the auth/landscape variability across SAP products means this needs more design care than a typical connector)
- **Breaking Changes**: No — additive, isolated behind the `ingest-sap` extra.
- **Dependencies**: `requests` only for phase 1 (already a core dependency); no proprietary SDK required.

## Implementation Ideas

```python
from semantica.ingest import SAPIngestor

ingestor = SAPIngestor(
    base_url="https://my-sap-system.example.com/sap/opu/odata/sap/API_BUSINESS_PARTNER",
    auth="oauth2",
    client_id="...",
    client_secret="...",
    token_url="https://my-sap-system.example.com/oauth/token",
)
partners = ingestor.ingest_entity_set(
    "API_BUSINESS_PARTNER", "A_BusinessPartner",
    select=["BusinessPartner", "BusinessPartnerFullName"],
    top=1000,
)
entities = ingestor.export_as_documents(partners)
```

## Additional Context

- This needs a short design discussion before implementation starts (hence `needs-discuss`): which auth flow(s) to prioritize, and whether to depend on a metadata-parsing library (`pyodata`) or hand-roll minimal EDMX parsing for `discover_service`.
- Related: #1227 (Salesforce connector) — same "enterprise data platform" motivation, independently scoped and mergeable separately.
- RFC/BAPI support for older on-prem ECC systems is explicitly out of scope for this issue; open a separate issue if/when a contributor has SAP NW RFC SDK access to build and test against it.

## Contribution

- [ ] I'm willing to help implement this feature
- [ ] I can help with documentation
- [ ] I can help with testing

pkupt:
## Summary

Adds a SAP OData ingestor to pull master/transactional data (Business
Partners, Sales Orders) from SAP OData services into the Context Graph. This
implements #1228.

Follows the existing three-layer connector pattern used by Snowflake and
Databricks (`XXXData` / `XXXConnector` / `XXXIngestor`) — here `SAPODataEntity`
/ `SAPODataConnector` / `SAPIngestor`, exposed lazily from `semantica.ingest`.

### Highlights

- **Dual auth** — OAuth2 client-credentials (BTP / S/4HANA Cloud) and Basic
  (on-prem NetWeaver). Every outbound request, **including the token
  exchange**, goes through `request_with_ssrf_guard` so user-supplied
  endpoints can never reach private/loopback/link-local address space.
- **EDMX without heavy deps** — `$metadata` is plain CSDL XML in both OData
  v2 and v4, so we hand-roll a minimal `xml.etree` parser instead of pulling
  in `pyodata`. Phase-1 is `requests`-only.
- **Server-driven pagination** — follows OData v2 `__next` (plain string or
  `__deferred` form) and v4 `@odata.nextLink`.
- **API symmetry** — `discover_service(service)` and
  `ingest_entity_set(..., service)` treat `service` the same way (a service
  root); `$metadata` is appended internally.

## Files changed

- `semantica/ingest/sap_ingestor.py` — new connector + ingestor
- `tests/ingest/test_sap_ingestor.py` — 22 tests (auth, EDMX parsing,
  v2/v4 pagination, SSRF routing, error paths, service-root normalization)
- `semantica/ingest/__init__.py` — lazy exports for the three symbols
- `pyproject.toml` — `ingest-sap = ["requests>=2.28.0"]` extra
- `docs/reference/ingest.md` / `docs/guides/ingest.md` — wiring notes

## Known limits (phase 1, called out in docstrings)

- OAuth2 token is cached and never refreshed (SAP token TTL is typically 1h);
  long jobs past the TTL should recreate the connector.
- No `max_pages` fuse on pagination; `top` bounds it when supplied.

Closes #1228

### DIFF
--- a/docs/guides/ingest.md
+++ b/docs/guides/ingest.md
@@ -382,6 +382,45 @@ For authentication details (PAT vs. OAuth M2M for Databricks; password vs. key-p
 
 > **Security Note:** Never hardcode credentials (`token`, `password`, `private_key`) in production code; pass them via environment variables (e.g., `DATABRICKS_TOKEN`, `SNOWFLAKE_PASSWORD`) or a secrets manager.
 
+## Source 7 — SAP OData
+
+`SAPIngestor` ingests an Entity Set from a SAP OData service (S/4HANA Cloud, SuccessFactors, or an on-prem NetWeaver Gateway over its REST surface). It speaks OData v2 and v4, follows server-driven pagination automatically, and flattens each record into a document dict via `export_as_documents()` — the same structured "transform to text, then store" pattern as the other sources.
+
+```python
+from semantica.ingest import SAPIngestor
+
+ing = SAPIngestor(
+    base_url="https://my-sap.example.com/sap/opu/odata/sap/API_BUSINESS_PARTNER",
+    client_id="...", client_secret="...",
+    token_url="https://my-sap.example.com/oauth/token",  # OAuth2 client-credentials (BTP/S/4HANA Cloud)
+    # On-prem NetWeaver often uses Basic auth instead — swap the block above for:
+    # username="erp_user", password="...",
+)
+
+# 1. Discover an unfamiliar service: entity sets + field types from $metadata
+sets = ing.discover_service()          # [{"name": "A_BusinessPartnerSet", "fields": [...]}, ...]
+
+# 2. Pull a page-walked Entity Set (v2/v4 next links handled for you)
+partners = ing.ingest_entity_set(
+    entity_set="A_BusinessPartnerSet",
+    select="BusinessPartner,BusinessPartnerFullName",   # $select
+    top=1000,                                           # cap on total rows
+)
+
+# 3. Flatten to document dicts, then build text for the graph
+docs = ing.export_as_documents(partners)
+partner_texts = [
+    f"Business Partner {d['BusinessPartner']}: {d['BusinessPartnerFullName']}"
+    for d in docs
+]
+```
+
+- Use `expand="to_Item"` (e.g. on a sales-order header set) to pull nested line items in one request — handy for modeling order → line-item → material relationships.
+- Every outbound request, including the OAuth2 token exchange, is routed through the SSRF guard, so a user-supplied SAP URL can never reach private/loopback/link-local address space.
+- Install with `pip install 'semantica[ingest-sap]'`.
+
+> **Security Note:** Never hardcode credentials (`client_secret`, `password`) in code; pass them via environment variables (e.g., `SAP_CLIENT_SECRET`, `SAP_PASSWORD`) or a secrets manager.
+
 ## Combining All Five Sources
 
 Once you have text from each source, `AgentContext.store()` accepts a flat list of strings. Semantica embeds and indexes them together — the context graph has no concept of which string came from which source unless you add metadata explicitly.
--- a/docs/reference/ingest.md
+++ b/docs/reference/ingest.md
@@ -28,6 +28,7 @@ icon: "database"
 | `DBIngestor` | SQL databases via SQLAlchemy: tables, views, and custom queries |
 | `SnowflakeIngestor` | Snowflake data warehouse queries and table exports |
 | `DatabricksIngestor` | Databricks Unity Catalog metadata, Delta table queries, and lineage |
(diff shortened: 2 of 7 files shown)


ZohaibHassan16 (approved):
Re-tested all of the Qodo findings against the fix and they look good now.

Real v4 OASIS CSDL parses correctly, including entity sets and fields.

Also forced a cross-schema type collision with two unqualified `Item` types and resolution stayed separated, no bleed between schemas.

For the credential leak case, I mocked an absolute next link to `evil.attacker.com`. It gets blocked before the second request, so credentials stay on the SAP host. That was the main one I wanted to verify.

Also ran a real `SAPODataEntity` through `GraphBuilder.build_single_source()` and it returns entities now instead of silently dropping them.

Other edge cases look fixed too:

* `top=0` short circuits
* bad row containers raise
* nullable defaults to `True` when omitted

Tests are at 30 now vs 22 before, and the new tests actually cover the reported issues. Full suite is green, lazy import works, branch is current with main, no conflicts.

Only thing to note is no token refresh / no `max_pages` guard yet, but both are documented and I don't think either should block this. Can follow up later if long-running SAP jobs become a real use case.

LGTM, approving.

PR #1236, opened 2026-08-28
fix(export): @vocab mints into the shipped ns# namespace

### LINKED ISSUE #1146: [BUG] @vocab points at https://semantica.dev/vocab/, which 404s and is not the namespace the package ships: every entity type and every user metadata key is minted into it

pkupt:
fix(export): @vocab mints into the shipped ns# namespace (#1146)

Every JSON-LD context set `@vocab` to `https://semantica.dev/vocab/`, which 404s — so entity types, relationship types and caller metadata keys all got minted under a namespace the package never ships. `SEMANTICA_NS` is the real one (`ns#`, the IRI `semantica-ns.ttl` declares).

**Follow-up after review**: moving `@vocab` to the live namespace turned out to be worse than the dead link, because every bare term in caller data then expanded into a term that *looks* official. `@vocab` is now removed from all five contexts (4 in `json_exporter.py`, 1 in `rdf_exporter.py`) — every document already uses explicit `semantica:` prefixes, so nothing else changes. Entity nodes are always `semantica:Entity` with the caller's label in `semantica:type` (the pattern relationship nodes already used), and `semantica:metadata` is typed `@json`, keeping the whole metadata dict as one `rdf:JSON` literal instead of expanding its keys. `sem:type`'s domain in the ttl opens up to cover entities, following the `sem:confidence` precedent. Regression tests now assert the negative space: no context declares `@vocab`, no caller label becomes an `ns#` class, no caller metadata key expands.

Fixes #1146.


ZohaibHassan16 (changes requested):
I exported an entity with:

```python
type: "ORG"
metadata: {"source": "crm_export_2024"}
```

Then parsed the generated output with rdflib and got:

```text
https://example.org/e1  rdf:type  https://semantica.dev/ns#ORG
_:b0                     https://semantica.dev/ns#source  "crm_export_2024"
```

There are two issues here:

1. `"ORG"` comes from the caller/extraction result, it's not a class defined by Semantica. The classes we actually define are `Entity`, `Relationship` and `KnowledgeGraph`.

   So right now any arbitrary extracted type gets turned into something under our official namespace, even though that term doesn't actually exist there.

2. `"source"` is just a metadata key from the caller, but `sem:source` is already a real `owl:ObjectProperty` in `semantica-ns.ttl`.

   In the example above, the string `"crm_export_2024"` ends up attached to that exact IRI. So this isn't only an unresolved vocab term anymore, it can collide with a real property that already has different semantics and expects a resource.

Before this change, `@vocab` was pointing at a dead URL. That's obviously wrong too, but at least arbitrary terms couldn't accidentally collide with the real Semantica vocabulary.

Changing `@vocab` to the live namespace means every bare/unscoped term can now expand into something that looks like an official Semantica term, which seems worse for anything actually consuming this output as RDF.

I don't think this is really a one-line URL fix.

Caller-controlled values like extracted types and arbitrary metadata keys probably shouldn't resolve through our `@vocab` at all. Either those fields need explicit context handling so they stay as caller data, or we remove `@vocab` and use the explicit `semantica:` prefix only for terms that are actually part of the Semantica vocabulary.

Requesting changes for now. Can you take another pass at both Qodo findings before this goes back up for review?

pkupt (commented):
You were right, this went deeper than a URL swap. Dropped `@vocab` from all five contexts — that was the actual door, since every document already uses explicit `semantica:` prefixes. Entity nodes are now always `semantica:Entity` with the caller's label in `semantica:type`, exactly how relationship nodes already worked, and `semantica:metadata` is typed `@json` so the whole dict rides as one `rdf:JSON` literal — your `crm_export_2024` example now round-trips without touching `sem:source`. Re-ran your repro locally: no `ns#ORG`, no literal on `sem:source`. Both `JSONExporter` and `serialize_to_jsonld` got the same treatment. Opened the domain on `sem:type` in the ttl to cover entities, following the `sem:confidence` precedent.

ZohaibHassan16 (approved):
Reran my original repro against this with the same entity: `type: "ORG"` and `metadata: {"source": "crm_export_2024"}`.

Parsed with rdflib and confirmed:

```text
https://example.org/e1  rdf:type     https://semantica.dev/ns#Entity
https://example.org/e1  ns#type      "ORG"
https://example.org/e1  ns#metadata  {"source":"crm_export_2024"}  (rdf:JSON literal)
```

No `ns#ORG`, no literal on `sem:source`, and no `sem:source` triples in the graph. This fixes the actual issue I was calling out.

Also checked `RDFExporter.serialize_to_jsonld` separately , same correct behavior there.

Tested an entity with no `type` field to make sure the `@type` rewrite didn’t regress that case. It still exports as `ns#Entity` with no stray `sem:type` triple.

The `sem:type` domain widening in the TTL looks fine too and is consistent with the existing `sem:confidence` pattern.

Export suite is green: 325 passed.

Looks good. Approving.

PR #1241, opened 2026-08-28
fix(explorer): dedupe temporal snapshot requests and apply latest-wins

### LINKED ISSUE #1128: GraphWorkspace polls /temporal/snapshot while idle, and the active-node chip lags the scrubber
Two issues in `explorer/src/workspaces/GraphWorkspace/GraphWorkspace.tsx`, seen against a backend implementing `TemporalSnapshotFastResponse` from `semantica/explorer/schemas.py`.

### 1. Idle polling of `/api/temporal/snapshot`

With the page loaded and untouched, `/api/temporal/snapshot` is re-requested roughly 10 times in 5 seconds with an identical `at` parameter.

It appears self-sustaining rather than timer-driven: handling the response calls `setGraphVersion(...)`, which re-renders, which recreates `summary`, which feeds back into `shouldFetchTemporalSnapshot`.

**Reproduction**

1. Point the Explorer at a backend serving `/api/graph/nodes`, `/api/graph/edges`, `/api/temporal/bounds` and `/api/temporal/snapshot` via `VITE_EXPLORER_API_TARGET`.
2. Open the graph workspace and leave it idle.
3. Watch the network panel filtered to `snapshot`.

**Observed:** ~10 identical requests in 5 seconds, indefinitely.
**Expected:** one request per scrubber position.

### 2. The active-node count lags the scrubber

After moving the scrubber, the active-node chip can keep showing the previous position's count -- scrubber at 2026/06 while the chip still reads the December value.

Position and count look like separate pieces of state updated by separate effects, with nothing tying a rendered count to the position that requested it. A monotonic sequence guard (latest-wins, discard out-of-order responses) would close it.

### Notes

Both were measured on 2026-08-19 against `main` at `e6b159e5`. The backend serves your exact response shape, and the temporal data does change across the range, so neither is an artefact of a flat or malformed dataset.

Happy to send a PR from `rajesh-ms/ontology-fde` if you would like a particular approach to either.

ALDRIN121:
## Description

`GraphWorkspace` polls `/api/temporal/snapshot` while idle, and the active-node chip lags the scrubber. The snapshot fetch/apply path had no idempotency or ordering protection.

**Root cause (investigated, not just hypothesized):** the snapshot effect re-requested the same scrubber position repeatedly and applied responses with no ordering protection.

1. **Duplicate requests for the same `at`.** The timeline's `timechange`/effect paths can call `onTimeChange` with the same timestamp more than once (timeline recreation while bounds settle, play ticks resetting the playhead, drag events), and each render of the effect with that `debouncedTime` fired a fresh fetch. Observed: 13+ identical-`at` requests at ~500 ms cadence, i.e. `n` requests for one scrubber position. The maintainer's `setGraphVersion` hypothesis is falsified — the react-query `summary` data reference is stable (`staleTime: Infinity`), so `setGraphVersion` does not recreate it — but the loop is real and reproduces via the effect's lack of dedup.
2. **Out-of-order responses clobbering newer state.** With variable network latency, an older position's response can land after a newer one's, overwriting the active-node count — the chip visibly lags the scrubber.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #1128

## Changes Made

- New pure module `explorer/src/workspaces/GraphWorkspace/temporalSnapshotGuards.ts`: a small stateful guard that enforces (a) at most one in-flight request per scrubber position — identical `at` values are deduplicated while a request is pending, breaking the idle/play polling loop; (b) successful snapshots are cached per position and re-applied when the scrubber returns (play wrap-around, back-scrubbing) without a refetch; (c) a response is applied only while the scrubber is still on its position, so out-of-order responses cannot clobber a newer position's count; (d) failed, cancelled, or superseded requests release their position so it can be fetched again on the next visit; (e) `reset()` drops all state when the graph summary changes (reload/retry), because cached snapshots describe the previous graph.
- `GraphWorkspace.tsx` snapshot effect wired through the guards (in-flight dedup, cached re-apply on revisit, position-checked application with release-on-reject, retry-able failures/cancels) plus a `useEffect` keyed on the `summary` query identity that resets guard state exactly when the graph data is replaced.
- 16 unit tests in `explorer/tests/temporalSnapshotGuards.test.ts` covering dedup, independent positions, revisit re-apply, play wrap-around, failure retry, stale-seq protection, position-returned races, reset-on-reload, and cache-bound eviction.

## Testing

- [x] Tested locally (stub backend, live page: mount now issues exactly one bounds request and one snapshot request — previously two of each; 28 s idle produced **zero** requests)
- [x] Added tests for new functionality (`explorer/tests/temporalSnapshotGuards.test.ts`, 16 tests)
- [x] `node --import tsx --test tests/*.test.ts` — 82/82 pass
- [x] `npx tsc -b` — clean

## Documentation

## Breaking Changes

**Breaking Changes**: No

## Checklist

## Additional Notes

Honest caveat on the issue's stated "polls while idle" loop: with the current code I could not reproduce continuous idle polling from a stable mount (idle produced a finite burst, not a steady loop). What is confirmed and fixed is the *duplicate-request* mechanism — the same scrubber position re-fetching in bursts (identical `at` values at ~500 ms cadence) — and the *out-of-order* clobbering. Both match the reported symptoms (wasted idle traffic; chip lagging the scrubber) and both are eliminated by the guards: any upstream churn that re-presents the same position now fetches at most once, applies at most the current position's data, and revisits re-apply cached snapshots.


ZohaibHassan16 (approved):
Wrote 3 of my own adversarial tests (stale apply, phantom cache entry, stale finish evicting a newer cache) , all green. Verified the reset()-on-reload claim directly: staleTime: Infinity + structural sharing really does keep summary stable except on real reloads. Full suite 82/82, tsc clean, no conflicts. Approving.

Issue #1083, opened 2026-08-18
SPARQLReasoner.execute_query is a stub: always returns empty bindings even with a triplet store configured


jiaqiwang969:
## Summary

`semantica/reasoning/sparql_reasoner.py` `execute_query()` never executes the query. Both branches return an empty `SPARQLQueryResult(bindings=[], variables=[])`:

```python
if self.triplet_store:
    # This would call the triplet store's query method
    # For now, return empty result
    result = SPARQLQueryResult(bindings=[], variables=[])
else:
    # Mock result for testing
    result = SPARQLQueryResult(bindings=[], variables=[])
```

The docstrings and the module docs present `SPARQLReasoner` as "reasoning over RDF/Triplet stores using SPARQL query expansion", so callers reasonably expect real results. Downstream code that trusts an empty result as "no matches" can silently draw wrong conclusions (e.g., a compliance query that appears to find no violations).

## Repro (semantica 0.6.5)

```python
from semantica.reasoning import SPARQLReasoner

r = SPARQLReasoner()
res = r.execute_query("SELECT ?s ?p ?o WHERE { ?s ?p ?o }")
print(res.bindings)   # [] — always, regardless of store/data
```

## Suggestion

Until the store-backed execution lands, raising `NotImplementedError` (or emitting a prominent warning) would be much safer than returning an empty result set — empty looks like a valid answer. If helpful, an interim implementation could delegate to rdflib for in-memory graphs.

Found while integrating semantica into an ontology-engineering teaching repo; we documented "use rdflib for SPARQL" as a workaround in our book↔code mapping table.

ALDRIN121:
Hi! I'd like to take this on. Following the issue's suggestion, I plan to make `execute_query()` raise `NotImplementedError` with a clear message instead of silently returning an empty result set, and add regression tests pinning that behavior. I'll open a PR shortly.

KaifAhmad1:
@ALDRIN121 Sure assing this to you

PR #1243, opened 2026-08-28
feat(reasoning): implement SPARQLReasoner.execute_query with store delegation and rdflib fallback

### LINKED ISSUE #1242: [FEATURE] SPARQLReasoner.execute_query: implement real execution instead of the NotImplementedError stub (follow-up to #1083)
### Background

Issue #1083 reported that `SPARQLReasoner.execute_query` always returned empty bindings even with a triplet store configured. It was closed via PR #1087, which replaced the silent empty result with raising `NotImplementedError` (with a regression test).

While that fix stops the silent data loss, the underlying capability is still missing: there is currently **no way to execute SPARQL queries through the reasoner at all**. Users with a configured triplet store or an rdflib-compatible graph still have to bypass the reasoner.

### Proposal

Implement `SPARQLReasoner.execute_query` with three execution paths:

1. **Delegation**: if the configured triplet store exposes `execute_query`, delegate with query + options and coerce the result into `SPARQLQueryResult`.
2. **rdflib fallback**: otherwise materialize an in-memory `rdflib.Graph` from the triplets and run the query there (SELECT/ASK/CONSTRUCT/DESCRIBE), loudly refusing with a `ProcessingError` when no graph data is available.
3. **Result-level reasoning**: the fallback runs after reasoning so inferred triplets participate in query results.

Plus a per-`(query, options)` result cache with copy-on-return isolation, and proper coercion of non-string RDF terms.

### Non-goals

- Cache invalidation on store/rule mutation (documented; callers use `clear_cache()`).
- Store-level options semantics in the rdflib fallback (warned loudly).

I'd like to take this on and will open a PR.

cxzg007:
Closes #1242
Follow-up to #1083 (closed by the `NotImplementedError` stub in #1087).

### What

Implements real execution for `SPARQLReasoner.execute_query` with three paths:

1. **Delegation** — when the configured triplet store exposes `execute_query`, the query and options are delegated and the store result is coerced into `SPARQLQueryResult` (dict-shaped, list-shaped, and typed-object store results are all supported).
2. **rdflib fallback** — otherwise an in-memory `rdflib.Graph` is materialized from the triplets (after reasoning, so inferred triplets participate in results) and SELECT / ASK / CONSTRUCT / DESCRIBE are executed there. A `ProcessingError` is raised when no graph data is available, instead of silently returning empty bindings.
3. **Per-(query, options) cache** with copy-on-return isolation: cached results cannot be mutated by callers, and cache entries are stored as copies. `clear_cache()` is the documented invalidation path.

Also: non-string RDF terms are coerced to strings, URI detection covers ftp/file/tag/doi schemes, and the fallback warns loudly when store-level `options` are not applicable.

### Tests

- 92 tests in `tests/reasoning/` pass.
- New regression tests: delegation round-trip, fallback (SELECT/ASK/CONSTRUCT), cache hit/isolation/key-discrimination, non-string coercion, `ProcessingError` on missing graph data, and inference-before-fallback ordering.

### Non-goals (documented)

- Automatic cache invalidation on store/rule mutation.
- Store-level options semantics in the rdflib fallback.

### DIFF
--- a/semantica/reasoning/sparql_reasoner.py
+++ b/semantica/reasoning/sparql_reasoner.py
@@ -29,14 +29,15 @@ Author: Semantica Contributors
 License: MIT
 """
 
-import re
+import json
+import time
 from dataclasses import dataclass, field
-from typing import Any, Dict, List, Optional, Set
+from typing import Any, Dict, List, Optional
 
 from ..utils.exceptions import ProcessingError, ValidationError
 from ..utils.logging import get_logger
 from ..utils.progress_tracker import get_progress_tracker
-from .reasoner import Rule, Reasoner
+from .reasoner import Reasoner, Rule
 
 
 @dataclass
@@ -84,9 +85,8 @@ class SPARQLReasoner:
         self.triplet_store = self.config.get("triplet_store")
         self.enable_inference = self.config.get("enable_inference", True)
 
-        # Reserved for query caching once a triplet-store execution path
-        # lands. execute_query() raises NotImplementedError until then, so
-        # the cache cannot be populated through any public path yet.
+        # Cache for executed queries, keyed by (query, options) and
+        # populated by execute_query(); cleared via clear_cache().
         self.query_cache: Dict[str, Any] = {}
 
     def expand_query(self, query: str, **options) -> str:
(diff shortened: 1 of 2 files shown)


cxzg007:
@codex review

ZohaibHassan16 (approved):
High quality implementation. Good to merge

PR #1245, opened 2026-08-28
fix(split): validate sliding window progress

### LINKED ISSUE #1244: [BUG] SlidingWindowChunker can fail to make progress with non positive stride
## Description

`SlidingWindowChunker` accepts an explicit `stride` value of `0` or a negative value because the constructor currently validates `chunk_size` and `overlap`, but does not validate that `stride` is positive.

This can break the forward progress invariant used by the sliding window loop. In the fixed size path, the cursor advances with `start += self.stride`, so a zero stride can cause the loop to stop making progress.

There are two closely related cases in `chunk_with_overlap()` as well.

First, `overlap_size` is applied without validating that it remains smaller than `chunk_size`. An overlap equal to `chunk_size` produces a zero stride, while a larger overlap produces a negative stride.

Second, `chunk_with_overlap()` temporarily mutates `overlap` and `stride`, but the original state is restored only after `self.chunk(text)` returns successfully. If chunking raises, the temporary state remains on the instance. The current restoration logic also recomputes `stride` from `chunk_size - overlap`, which can overwrite an explicitly configured custom stride.

## Steps to Reproduce

1. Create a `SlidingWindowChunker` with a zero stride.

```python
from semantica.split.sliding_window_chunker import SlidingWindowChunker

chunker = SlidingWindowChunker(
    chunk_size=100,
    stride=0,
)

print(chunker.stride)
```

2. The constructor succeeds and prints:

```text
0
```

3. Exercise the fixed size chunking path.

```python
chunker.chunk(
    "non empty input",
    preserve_boundaries=False,
)
```

4. The fixed size loop does not have a valid forward progress step because `start += self.stride` does not advance `start`.

A negative stride is also accepted:

```python
chunker = SlidingWindowChunker(
    chunk_size=100,
    stride=-1,
)

print(chunker.stride)
```

This also constructs successfully.

The helper can create the same invalid state through an overlap override:

```python
chunker = SlidingWindowChunker(chunk_size=100)

chunker.chunk_with_overlap(
    "some text",
    overlap_size=100,
)
```

Here the temporary stride becomes `0`.

## Expected Behavior

`SlidingWindowChunker` should reject configurations that cannot guarantee forward progress.

An explicit `stride` should be required to be greater than `0`.

`chunk_with_overlap()` should require:

```text
0 <= overlap_size < chunk_size
```

Temporary overlap and stride changes should also be restored after both successful and failed chunking calls.

If a custom stride was explicitly configured, that exact value should be restored after `chunk_with_overlap()` returns.

## Actual Behavior

The constructor currently accepts `stride=0` and negative stride values.

The fixed size path can therefore enter a loop where the cursor does not move forward.

`chunk_with_overlap()` can also derive a zero or negative stride from an invalid `overlap_size`.

In addition, if `self.chunk(text)` raises while a temporary overlap is active, the instance can retain the temporary overlap and stride. On successful calls, an explicitly configured stride can also be replaced by `chunk_size - overlap` during restoration.

## Environment

* **OS**: Windows
* **Python Version**: 3.13.15
* **Semantica Version**: 0.6.6
* **Installation Method**: Source checkout from the repository

## Code Example

```python
from semantica.split.sliding_window_chunker import SlidingWindowChunker

chunker = SlidingWindowChunker(
    chunk_size=100,
    stride=0,
)

print(f"constructed with stride={chunker.stride}")

chunker.chunk(
    "non empty input",
    preserve_boundaries=False,
)
```

A separate state restoration example is:

```python
chunker = SlidingWindowChunker(
    chunk_size=10,
    overlap=2,
    stride=3,
)

chunker.chunk_with_overlap(
    "abcdefghijklmnopqrstuvwxyz",
    overlap_size=4,
)

print(chunker.overlap)
print(chunker.stride)
```

The original custom stride should remain `3` after the helper call.

## Error Messages / Logs

There is no exception when `stride=0` or a negative stride is passed to the constructor.

For example, the current main branch accepts both configurations:

```text
BEFORE: invalid stride is accepted
stride=0  -> constructed successfully, stride = 0
stride=-1 -> constructed successfully, stride = -1
```

The main failure mode is lack of forward progress rather than a raised error.

## Additional Context

- The boundary preserving path has additional cursor correction logic, so the minimal reproduction above intentionally uses `preserve_boundaries=False` to exercise the fixed size path directly.

- I also checked the existing `TestSlidingWindowChunker` coverage in `tests/split/test_chunkers.py`. This looks like a focused correctness fix that can stay within the existing chunker implementation and test file.

- I have opened a focused #1245 with the proposed fix and regression coverage. If the approach looks good, please assign this issue to me.

## Possible Solution

Validate the progress invariant at construction time:

```python
if self.stride <= 0:
    raise ValidationError("stride must be positive")
```

Validate `overlap_size` before mutating the chunker state:

```python
if overlap_size < 0:
    raise ValidationError("overlap_size must be non-negative")

if overlap_size >= self.chunk_size:
    raise ValidationError("overlap_size must be less than chunk_size")
```

For temporary overlap overrides, preserve both the original overlap and the original stride, then restore both values with `try/finally`.

Regression coverage should include:

1. Zero stride is rejected.
2. Negative stride is rejected.
3. Negative overlap overrides are rejected.
4. Overlap equal to `chunk_size` is rejected.
5. Overlap greater than `chunk_size` is rejected.
6. The maximum valid overlap still works.
7. A custom stride is restored after a successful helper call.
8. Original state is restored if chunking raises.

This keeps the fix limited to `semantica/split/sliding_window_chunker.py` and the existing `TestSlidingWindowChunker` coverage in `tests/split/test_chunkers.py`.

HsienW:
## Description

This fixes a progress invariant issue in `SlidingWindowChunker`.

An explicit zero or negative `stride` could previously pass validation even though the fixed size chunking path depends on `stride` to move the cursor forward.

`chunk_with_overlap()` could also derive a zero or negative stride from an invalid temporary overlap. It temporarily mutates `overlap` and `stride`, but the original state was not guaranteed to be restored if chunking raised an exception. The previous restoration logic could also overwrite an explicitly configured custom stride.

This change validates these inputs before chunking and makes temporary overlap overrides exception safe.

### Before

`stride=0` and `stride=-1` are accepted by the constructor.

### After

Invalid stride and overlap values are rejected. The original overlap and custom stride are restored after both successful and failed `chunk_with_overlap()` calls.

## Type of Change

* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [ ] Documentation update
* [ ] Performance improvement
* [ ] Code refactoring

## Related Issues

Closes #1244 

## Changes Made

* Reject non positive `stride` values before chunking.
* Validate temporary `overlap_size` values before mutating chunker state.
* Restore the original overlap and custom stride with `try/finally`.
* Add regression coverage for invalid stride and overlap values, valid boundary behavior, and state restoration.

## Testing

* [x] Tested locally
* [x] Added tests for new functionality
* [x] Package builds successfully (`python -m build`)

### Test Commands

```bash
python -m pytest tests/split/test_chunkers.py::TestSlidingWindowChunker -v
python -m build
python -m py_compile semantica/split/sliding_window_chunker.py
git diff --check
```

Focused `SlidingWindowChunker` tests pass with 13 tests.

Package build passes and produces both the source distribution and wheel.

`py_compile` and `git diff --check` also pass.

Repository wide checks still expose existing baseline and environment failures outside this change:

* Full pytest collection stops because multiple optional dependencies are unavailable, resulting in 188 collection errors.
* Full mypy reports 1583 existing errors across 249 files.
* Black passes the changed runtime file. The existing `test_chunkers.py` baseline still requires formatting.
* isort reports existing import ordering issues in `test_chunkers.py`.
* flake8 reports 18 remaining violations, all reproduced on `upstream/main`.
* This PR introduces no new flake8 violations.

## Documentation

* [ ] Updated relevant documentation
* [ ] Added code examples if applicable
* [ ] Updated API reference if adding new APIs
* [ ] Updated cookbook if adding new examples
* [x] No documentation changes needed

## Breaking Changes

**Breaking Changes**: No

## Checklist

* [x] My code follows the project's style guidelines
* [x] I have performed a self-review of my code
* [x] I have commented my code, particularly in hard-to-understand areas
* [x] My changes generate no new warnings
* [x] Package builds successfully

## Additional Notes

The change is limited to:

* `semantica/split/sliding_window_chunker.py`
* `tests/split/test_chunkers.py`

No unrelated baseline formatting, lint, typing, or dependency issues are changed as part of this PR.


KaifAhmad1 (approved):
Verified locally: checked out the branch and ran the full `TestSlidingWindowChunker` suite (13 tests, all pass), and confirmed the underlying bug reproduces on `main` (`SlidingWindowChunker(chunk_size=100, stride=0)` currently constructs successfully). The fix correctly guarantees `stride > 0` at construction and within `chunk_with_overlap()`'s temporary override, the `try/finally` closes the exception-safety and custom-stride-clobbering bugs, and the boundary-preservation clamp path is safe given the new invariant. Matches the fix proposed in #1244 exactly. LGTM.

Issue #1029, opened 2026-08-16
[BUG] VectorStore.store_vectors() reissues live vector ids after a delete, silently overwriting an existing vector


pravit-amp:
## Bug Description

`VectorStore.store_vectors()` generates vector ids from the current store size, so after any
deletion it reissues an id that is **still in use** — overwriting a live vector and its metadata
with no error or warning.

On the `inmemory` backend (`vector_store.py:546`):

```python
start_idx = len(self.vectors)
for i, (vector, meta) in enumerate(zip(vectors, metadata)):
    vector_id = f"vec_{start_idx + i}"
```

The scheme is only correct while the id space is append-only. `delete_vectors()` shrinks it, so
`len(self.vectors)` stops tracking the highest id ever issued and starts pointing back into the
live set.

The consequence is silent data loss. Reached through `AgentMemory`, it is worse than losing a
vector: two live memory items end up sharing one embedding, so retrieving memory B by similarity
returns memory C's content, and deleting either one takes the other's vector with it.

Both `store_vectors()`/`delete_vectors()` and `AgentMemory.delete_memory()`'s vector cascade date
from the 2025-11-06 restructure, so this is long-standing rather than a recent regression.

## Steps to Reproduce

1. Create an `AgentMemory` bound to an in-memory `VectorStore`.
2. Store two memories, delete the **first**, then store a third.
3. Observe that the third memory is assigned the second memory's vector id.

## Expected Behavior

An id issued by the store is never reused while the original is live. After store → delete →
store, both the surviving memory and the new one should be independently retrievable, each with
its own embedding, and `count()` should report 2.

## Actual Behavior

The new vector is assigned `vec_1`, which memory B still owns. The write overwrites B's vector and
metadata; `count()` reports 1 for two live memories, and `_vector_ids` maps both B and C to the
same id.

## Environment

- **OS**: macOS 26.5.2 (arm64)
- **Python Version**: 3.12.13
- **Semantica Version**: 0.6.5 (`main` @ 8177d887)
- **Installation Method**: source

## Code Example

```python
from semantica.context import AgentMemory
from semantica.vector_store import VectorStore

store = VectorStore(backend="inmemory", dimension=8)
memory = AgentMemory(vector_store=store)

a = memory.store("memory A")     # -> vec_0
b = memory.store("memory B")     # -> vec_1
memory.delete_memory(a)          # vec_0 freed; vec_1 still live
[...]

PR #1250, opened 2026-08-29
feat(mcp): semantic retrieval tools (store, retrieve, update, remove)

### LINKED ISSUE #1235: [FEATURE] Add persistent semantic retrieval to the MCP context tools
## Feature description

The MCP server currently persists a knowledge graph and offers graph traversal / keyword matching, but it has no durable vector-backed retrieval path for document context. Add a configurable semantic retrieval backend and a bounded, attributable `retrieve_context` tool suitable for agent use.

## Scope

- Configure a supported persistent vector backend without hard-coding one vendor.
- Chunk and embed explicitly selected documents.
- Store stable chunk IDs plus source, document authority, status, version/hash, and project metadata.
- Retrieve bounded top-k passages and combine them with relevant graph relationships.
- Return original source citations and retrieval metadata.
- Handle updates, deleted/stale documents, unavailable backends, and empty results honestly.
- Preserve current graph persistence and existing MCP contracts.

## Non-goals

- Crawling or ingesting an entire filesystem by default.
- Treating retrieved text as product authority without source metadata.
- Replacing source documents with generated summaries.
- Shipping deployment-specific credentials or infrastructure configuration.

## Acceptance proof

1. An integration test starts a real supported vector backend, ingests two documents with distinct authority metadata, and retrieves the correct cited passage for each.
2. Retrieval preserves source path/URI, content hash/version, authority, and project metadata.
3. Updating a document replaces stale chunks deterministically.
4. Removing a document prevents its chunks from being returned.
5. Backend-unavailable and no-hit paths return explicit bounded responses, not fabricated context.
6. Existing MCP and graph-persistence tests remain green.

pkupt:
Closes #1235

Implements the four semantic retrieval tools from the issue on top of VectorStore, wired into the MCP server's tool registry.

`store_document(content, source, authority, version?, project?, metadata?, chunk_size?, chunk_overlap?)` chunks content with a fixed sliding window (1000 chars, 200 overlap, both configurable) and stores every chunk with full provenance metadata: chunk_id, source, authority, version, project, content hash, status, chunk index and character offsets. Re-storing identical content under the same (source, version) is a no-op keyed on the content hash, so retries never re-embed.

`retrieve_context(query, top_k?, project?)` embeds the query, returns the top chunks with scores and provenance (top_k capped at 10), and combines them with relevant graph relationships: for each hit source, nodes in the ContextGraph tagged with that source are looked up and their 1-hop neighbours attached. Graph failures degrade to an empty list; retrieval never depends on the graph being present.

`update_document(content, source, version?, ...)` and `remove_document(source, version?)` are keyed on (source, version).

On the delete side, in-memory vector ids come out as `vec_{len(self.vectors) + i}`, so after a delete the counter falls back and the next writes land on ids that are still in use (#1029). Deriving my own chunk ids does not fix that, it just hides it. Update and remove therefore rebuild the store: read everything, filter in memory, then rewrite from empty. Nothing to collide with once the counter starts at zero. If the re-store fails, the pre-rebuild state is restored, so a failed update cannot drop surviving documents. The real fix for #1029, ids that never get reused, is better as its own PR.

`SEMANTICA_VECTOR_BACKEND` picks the backend. The retrieval tools support `inmemory` (default) and `sqlite` (requires `SEMANTICA_VECTOR_DB_PATH`); other VectorStore backends such as faiss or pgvector are rejected at startup, because they have no metadata-scoped delete and update/remove could never work on them. The store dimension is always derived from the embedder, because VectorStore's 768 default matches no real model. `SEMANTICA_VECTOR_PATH` is treated as a directory, matching `VectorStore.save()`/`load()`; a persisted store whose dimension differs from the active embedder fails startup with a clear message instead of ranking across incompatible embedding spaces.

Provenance fields (source, authority, version, hash, status, chunk offsets, project) are owned by the tool: caller-supplied metadata can add extra keys but cannot overwrite them. Documents above 10,000 chunks are rejected at ingestion, which keeps update/remove exact against the persistent metadata lookup limit. Update is atomic: the replaced rows are snapshotted and re-stored if the new write fails, on both the in-memory and the persistent path. Results carry a `persisted` flag so a failed save to `SEMANTICA_VECTOR_PATH` is visible instead of silent.

Known limitations worth naming. Graph context does a full `find_nodes()` scan per query, fine for MVP-scale graphs but worth an index later. With a project filter, retrieval over-fetches 3x, so projects with lower hit rates may return fewer than top_k results. And score scales are backend-relative (in-memory cosine vs sqlite's 1/(1+distance)), so absolute values are not comparable across backends.

The test suite has 29 cases: chunking windows and stable ids, provenance metadata including the caller-metadata protection, idempotent re-store, ranked retrieval with project filter and graph context, update and remove (including write-failure rollback on both backends), the top_k cap, a regression for the #1029 interaction (remove-then-store must not overwrite a surviving chunk), persistence through SEMANTICA_VECTOR_PATH including the dimension-mismatch guard, backend rejection, the chunk-count cap, and sqlite roundtrip plus multi-document isolation (skipped when the sqlite_vec extension is missing). It patches both EmbeddingGenerator construction sites with a deterministic fake, so CI runs network-free; the real generators probe FastEmbed on init and block on a TCP timeout wherever models are not cached. Existing MCP tests still pass.

Just synced with upstream main to pick up the `semantica_mcp/` package move from #1312 — imports updated.
Flag anything you want different before this gets reviewed further.


ZohaibHassan16 (changes requested):
Nice work. Just a couple small things I caught:

i) In `update_document` if the document doesn't exist yet, `_upsert` currently creates it anyways with `status: "updated"` , and authority ends up as `"unknown"` if omitted. You should probably return `{"status": "not_found", ...}` here like `remove_document` does.

ii) In `get_vector_store()` if `store.load(vector_path)` blows up on existing directory, we just log a warning and carry on with an empty store. But if client later calls store/update `_persist` will run `store.save()` and overwrite the directory, wiping their corpus. Probably better to fail / raise on startup here, same as dimension check.

Once those two are sorted, we can merge.

ZohaibHassan16 (approved):
All tests pass cleanly, merging it.

Issue #1167, opened 2026-08-21
test: pytest tests/ collects 0 tests on a clean [dev] install — 10 modules import fastapi above their skip guards


dex0shubham:
## Summary

`pytest tests/` aborts during collection on a clean `pip install -e ".[dev]"` — the exact setup CONTRIBUTING documents. **No tests run at all**, not even the ones unrelated to the Explorer:

```
!!!!!!!!!!!!!!!!!! Interrupted: 10 errors during collection !!!!!!!!!!!!!!!!!!!
4 skipped, 10 errors in 47.36s
```

Excluding the ten offending modules, the same command runs **4,537 passing tests**. So ten unguarded imports currently gate the entire suite.

`fastapi` ships only in the `explorer` extra (`pyproject.toml`), not in `dev`, so a contributor following the documented setup does not have it.

## The guards exist but are unreachable

Five of the ten modules already carry a module-level skip whose message names the missing extra:

```python
pytest.skip(
    "starlette TestClient is required for explorer tests. Install semantica[explorer].",
    allow_module_level=True,
)
```

The intent to skip is therefore explicit — but the guard is placed **below** the import it is meant to protect, so the `ModuleNotFoundError` fires first and the guard never executes. It also tests `starlette`, while the module that actually fails to import is `fastapi`.

| File | Failing import | Guard |
|---|---|---|
| `tests/explorer/test_explorer_api.py` | L12 | L20 (unreachable) |
| `tests/explorer/test_explorer_auth.py` | L16 | L24 (unreachable) |
| `tests/explorer/test_ontology_subissue3.py` | L9 | L18 (unreachable) |
| `tests/explorer/test_sparql_route.py` | L17 | L25 (unreachable) |
| `tests/explorer/test_ontology_dns_pinning.py` | L30 | none |
| `tests/explorer/test_ontology_ssrf.py` | L24 | none |
| `tests/explorer/test_provenance_manager_wiring.py` | L11 | none |
| `tests/explorer/test_provenance_route.py` | L5 | none |
| `tests/explorer/test_vocabulary.py` | L5 (`from fastapi import FastAPI`) | none |
| `tests/test_security_regression.py` | L27 | none |

## Reproduction

```bash
python -m venv venv && . venv/bin/activate
pip install -e ".[dev]"
pytest tests/ -q
```

```
ERROR tests/explorer/test_explorer_api.py
ERROR tests/explorer/test_explorer_auth.py
[...]

PR #1252, opened 2026-08-29
test(integrations): make crewai and langchain test dirs packages

### LINKED ISSUE #1251: test(integrations): collection aborts — duplicate test_degradation.py basename with no __init__.py in either subpackage
## Summary

`pytest tests/integrations/` aborts during collection on current `main` (`30592b12`):

```
E   import file mismatch:
E   imported module 'test_degradation' has this __file__ attribute:
E     tests/integrations/crewai/test_degradation.py
E   which is not the same as the test file we want to collect:
E     tests/integrations/langchain/test_degradation.py
E   HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules

ERROR tests/integrations/langchain/test_degradation.py
!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
335 tests collected, 1 error
```

This is **not** a stale-`__pycache__` artifact — it reproduces after `find tests -name __pycache__ -type d -exec rm -rf {} +` on a fresh checkout.

## Cause

Two test files share a basename:

- `tests/integrations/crewai/test_degradation.py`
- `tests/integrations/langchain/test_degradation.py`

`tests/integrations/__init__.py` exists, but neither subdirectory has one. Under pytest's default `prepend` import mode the basedir is the first ancestor *without* `__init__.py`, so both modules are imported as plain `test_degradation` and the second one collides with the first.

19 of the 33 directories under `tests/` already carry an `__init__.py`, so the convention is established — these two subpackages were just missed.

## Effect

`tests/integrations/langchain/test_degradation.py` is never collected, so its graceful-degradation coverage does not run. Because a collection error aborts the run, it also stops everything collected after it in that directory.

## Suggested fix

Add an empty `__init__.py` to both subdirectories, matching the existing convention:

```
tests/integrations/crewai/__init__.py
tests/integrations/langchain/__init__.py
```

Verified locally — collection goes from `335 collected, 1 error` to `337 collected` (0 errors), and both modules then run: `3 passed`.

## Related, but independent

This is unrelated to #1167 (the `fastapi` collection errors). Both block collection, neither fixes the other — with #1167's guards applied this collision still errors, and with these `__init__.py` files added the `fastapi` errors remain.

## Note on the wider pattern

Seven basenames are duplicated under `tests/`. Only this one currently errors, because a collision needs *both* copies to sit in non-package directories:

| Basename | Status |
|---|---|
| `test_degradation.py` | both dirs lack `__init__.py` — **errors today** |
| `test_performance.py`, `test_seed_manager.py`, `test_spacy_model_cache.py`, `test_vocabulary.py` | one dir is a package — latent |
| `test_integration.py`, `test_timestamp_timezones.py` | both are packages — safe |

The latent four would start erroring if a sibling duplicate were ever added in another non-package directory. Adding `__init__.py` to the remaining 14 directories would close the class permanently, but that is a larger change and I have kept this report to the one that is actually broken. Happy to do the broader sweep instead if you would prefer it.

I would like to take this on if the approach looks right.

dex0shubham:
## Description

`pytest tests/integrations/` aborts during collection because two test files share a basename and neither subdirectory is a package:

```
import file mismatch:
imported module 'test_degradation' has this __file__ attribute:
  tests/integrations/crewai/test_degradation.py
which is not the same as the test file we want to collect:
  tests/integrations/langchain/test_degradation.py
```

Under pytest's default `prepend` import mode the basedir is the first ancestor without an `__init__.py`, so both files are imported as plain `test_degradation` and the second collides with the first. `tests/integrations/__init__.py` already exists — the two subpackages were simply missed.

Not a stale-`__pycache__` artifact: it reproduces after clearing `__pycache__` on a fresh checkout of `main`.

## Type of Change

## Related Issues

Closes #1251

## Changes Made

- Added `tests/integrations/crewai/__init__.py`
- Added `tests/integrations/langchain/__init__.py`

Both are one-line package markers, matching the convention already used by 19 of the 33 directories under `tests/`.

## Testing

- [x] Tested locally

Before:

```
$ python -m pytest tests/integrations/ -q --co
ERROR tests/integrations/langchain/test_degradation.py
!!!!! Interrupted: 1 error during collection !!!!!
335 tests collected, 1 error
```

After:

```
$ python -m pytest tests/integrations/ -q --co
337 tests collected

$ python -m pytest tests/integrations/crewai/test_degradation.py tests/integrations/langchain/test_degradation.py -q
3 passed
```

The langchain graceful-degradation tests were previously never collected; they now run and pass.

## Note

This is independent of #1167. Both block collection and neither fixes the other — the `fastapi` errors remain with this change applied, and this collision remains with #1167's guards applied.


ZohaibHassan16 (approved):
Applied the fix: 337 collected, 0 errors

LGTM

Issue #531, opened 2026-05-04
[BUG] Progress tracker Unicode output crashes on Windows consoles using cp1252


ZohaibHassan16:
## Summary

Semantica progress tracking writes Unicode symbols to stdout. On Windows consoles using `cp1252`, this can crash long-running operations such as chunking or extraction.

## Environment

- OS: Windows
- Python: 3.12
- Console encoding: cp1252
- Semantica version reported by pip: 0.0.5

## Reproduction

Run a Semantica operation that uses progress tracking, for example:

```python
from semantica.split import StructuralChunker

StructuralChunker().chunk("Some text")
```

## Actual Result

```text
UnicodeEncodeError: 'charmap' codec can't encode character '\U0001f9e0'
```

## Expected Result

Progress output should not crash the process.

## Workaround

Setting this environment variable avoids the crash:

```bash
PYTHONIOENCODING=utf-8
```

## Suggested Fix

Either detect stdout encoding and fall back to ASCII-safe progress text, or encode progress output with error handling.

PR #1254, opened 2026-08-29
fix(utils): write console progress to stderr instead of stdout

### LINKED ISSUE #1134: bug(mcp): progress bar writes to stdout and corrupts JSON-RPC; SEMANTICA_KG_PATH never persists; 3 broken setup steps in mcp/README.md
## Summary

Setting up the root-level `mcp/` server against Claude Code on a clean machine, I hit one **protocol-level bug** that corrupts the JSON-RPC stream, one **env var that doesn't do what it's documented to do**, and three **broken instructions in `mcp/README.md`**. All verified against `main` @ `2a303cf4`, semantica 0.6.5, Python 3.12.12, macOS arm64.

The server itself works well once configured — 17 tools, clean `initialize`/`tools/list`/`tools/call`. These are setup and stream-hygiene issues.

---

## 1. Progress bar writes to stdout, corrupting the JSON-RPC stream (protocol violation)

**Severity: high.** stdio MCP requires stdout carry *exclusively* newline-delimited JSON-RPC. `mcp/README.md` states this explicitly ("Log/debug output goes to `stderr` only"), but it isn't true.

`semantica/utils/progress_tracker.py:241` and `:247` call `sys.stdout.write(...)` unconditionally. Any tool whose code path emits progress corrupts the response.

Reproduce:

```bash
printf '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"t","version":"1"}}}\n{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"run_reasoning","arguments":{"facts":["Person(John)"],"rules":["IF Person(?x) THEN Mortal(?x)"]}}}\n' \
  | python -m mcp >out.txt 2>/dev/null
python -c "import json;[json.loads(l) for l in open('out.txt') if l.strip()]"
```

Line 2 of stdout is not valid JSON — the progress bar is prepended to the response body on the same line:

```
🔄 Semantica is reasoning: Inferring facts 🤔 reasoning Reasoner |░░░...░░| 0.0% ETA: - Rate: - Time: 0.00s Extracted: -{"jsonrpc": "2.0", "id": 2, "result": ...}
```

A strict MCP client fails to parse the response to `id=2`. Note this happens **with stderr sent to `/dev/null`**, which rules out stderr bleed.

Workaround (verified: 1 violation → 0, payload unchanged): set `SEMANTICA_DISABLE_PROGRESS=1` in the server env.

Suggested fix — don't rely on users knowing that env var. Either:
- give `progress_tracker` a configurable output stream defaulting to `sys.stderr` (progress is diagnostic output; stderr is the correct stream for it regardless of MCP), or
- have `mcp/server.py` and `semantica/mcp_server/` capture the real stdout at startup, then rebind `sys.stdout = sys.stderr` for the process, writing JSON-RPC only to the saved handle. That's the durable fix — it closes the whole class rather than this one tool, since any future `print()` on a tool path reopens the bug.

A regression test asserting every stdout line of a `tools/call` round-trip parses as JSON would lock this down.

## 2. `SEMANTICA_KG_PATH` is documented as "persist/load" but never persists

`mcp/README.md:219` documents it as `Path to persist/load the graph (JSON file)`. `mcp/session.py:33-39` only ever calls `_graph.load(kg_path)`. There is no `.save()` call anywhere in `mcp/` (`grep -rn "\.save(\|atexit\|persist" mcp/` returns only the docstring and the README line).

Consequence: every `record_decision` / `add_entity` made through the MCP server is silently lost when the client restarts the server. For a tool whose value proposition is a durable, auditable decision record, silent data loss is a bad default — and the README actively suggests it's handled.

Either implement the save side (on mutation or via `atexit`), or correct the README to say **load-only** and document `export_graph` as the way to get data out. `mcp/session.py:5-6` is more accurate than the README already ("Set SEMANTICA_KG_PATH to automatically load a persisted graph on start") — the README overpromises.

## 3. `pip install -e ".[mcp]"` — the `mcp` extra does not exist

`mcp/README.md:12`. `pyproject.toml` defines 50 extras; none is named `mcp`. The quick start fails at the first command.

```
extras: llm-openai, llm-groq, llm-gemini, llm-anthropic, llm-ollama, llm-deepseek,
llm-litellm, llm-instructor, llm-all, parse-docling, shacl, db-snowflake, db-databricks,
db-arrow, ingest-parquet, ingest-arrow, db-all, models-huggingface, graph-neo4j,
graph-falkordb, graph-amazon-neptune, graph-apache-age, graph-all, tripletstore-oxigraph,
vectorstore-*, infra, cloud, monitoring, viz, gpu, agno, crewai, watch, split-*, dev,
explorer, explorer-lite, all
```

The server needs no MCP SDK (it's hand-rolled JSON-RPC over stdio, and the official `mcp` package is **not** a dependency), so plain `pip install -e .` is correct. Either add an `mcp = []` alias extra or fix the README.

## 4. Wrong config file for Claude Code

`mcp/README.md:92` says `~/.claude/settings.json`. Claude Code reads MCP servers from `~/.claude.json` (user scope) or `.mcp.json` (project scope); `settings.json` holds permissions/hooks/env and has no `mcpServers` key. Following the README yields a server that silently never loads.

## 5. `claude mcp add` example doesn't parse

`mcp/README.md:108`:

```bash
claude mcp add semantica python -m mcp --cwd /path/to/semantica
```

Two problems: `-m mcp` needs a `--` separator or it's consumed as flags to `claude`, and `--cwd` is not a flag of `claude mcp add`.

Also worth noting: because `mcp/` is excluded from the wheel (`packages.find` includes only `semantica*`/`integrations*` — intentional per CHANGELOG, and the cause of #541), `python -m mcp` only resolves from a repo checkout. Setting `PYTHONPATH` rather than relying on cwd is what makes it robust, since the client controls the working directory.

Working invocation:

```bash
claude mcp add semantica \
  -e PYTHONPATH=/path/to/semantica \
  -e SEMANTICA_DISABLE_PROGRESS=1 \
  -- /path/to/venv/bin/python -m mcp
```

Verified end-to-end from an unrelated cwd: `initialize` → `serverInfo 0.6.5`; `record_decision` → `{"decision_id": "...", "status": "recorded"}`; `get_graph_summary` → `{"node_count": 3, "decision_count": 1, "graph_ready": true}`; `run_reasoning` → `{"derived_facts": ["Mortal(John)"]}`.

---

Happy to send a PR for any subset — the stdout fix (#1) is the one that matters; the rest are docs.

dex0shubham:
## Description

Point 1 of #1134, per @Sameer6305's direction on that thread: a configurable progress stream defaulting to `stderr`, rather than rebinding `sys.stdout` in the servers.

`ConsoleProgressDisplay` wrote every progress frame to `sys.stdout`. Progress is diagnostic output — `tqdm` and most progress renderers default to stderr for the same reason — and stdout has to stay clean for programs carrying a machine-readable protocol on it. The stdio MCP servers put newline-delimited JSON-RPC on stdout, where an interleaved progress bar makes a response body unparseable.

**Scope:** point 1 only. The `SEMANTICA_KG_PATH` persistence problem and the `mcp/README.md` corrections remain with @akaszubski, as agreed on the issue.

## Type of Change

## Related Issues

Closes #1134

## Changes Made

- `ConsoleProgressDisplay.__init__` takes an optional `stream`, defaulting to `sys.stderr`.
- The stream is resolved per write via a `stream` property rather than captured at construction, so a later rebinding of `sys.stderr` — pytest capture, for instance — is honoured.
- `_safe_write` and the four bare `sys.stdout.flush()` calls now use that stream. No `sys.stdout` references remain in the module.
- The emoji-capability probe inspects the target stream rather than stdout, so a cp1252 **stderr** still degrades correctly instead of raising.
- New regression tests in `tests/utils/test_progress_stream.py` (9 tests).

## Testing

- [x] Tested locally

New tests verify all 9 fail against the unmodified module and pass with the change:

```
$ git stash push semantica/utils/progress_tracker.py
$ python -m pytest tests/utils/test_progress_stream.py -q
9 failed

$ git stash pop
$ python -m pytest tests/utils/test_progress_stream.py -q
9 passed
```

Every test file touching the progress tracker (39 files):

```
main:      1 failed, 643 passed, 23 skipped
this PR:   1 failed, 652 passed, 23 skipped
```

The one failure, `test_backward_compatibility.py::test_progress_tracking_unchanged`, is pre-existing and order-dependent — it passes in isolation on `main` and on this branch alike.

## Notes for review

**Two existing cp1252 tests were retargeted.** `tests/deduplication/test_deduplication.py` patched `sys.stdout` to assert emoji auto-disables on a cp1252 console. Since progress now goes to stderr, they patch that stream instead. Their intent is unchanged — checking the encoding of the stream actually written to is what makes the degradation correct.

**The attach gate is deliberately untouched.** `_stdout_is_tty()` still decides *whether* a console display is created, based on stdout. Changing it to follow stderr is arguably more consistent now, and would restore progress for `python script.py > out.txt` (currently suppressed by #de31b436), but that is a separate behaviour change and it breaks `test_console_display_present_when_stdout_is_a_tty`. Happy to do it in a follow-up if you want it.

**On the current containments.** `semantica/mcp_server/__init__.py` and `mcp/__init__.py` both set `SEMANTICA_DISABLE_PROGRESS=1`, and #de31b436 suppresses progress when stdout is not a terminal, so the original corruption no longer reproduces on `main`. Those remain useful, but they are containments: both give up progress output entirely, and each new entry point has to remember them. This addresses the cause, so they stop being load-bearing.

**Unrelated observation:** `tests/test_extractors_dispatch.py:13` assigns a `MagicMock` over `sys.modules['semantica.utils.progress_tracker']` at import time and never restores it, so any later module-level import of that module receives mocks. The new tests import through a fixture that drops the poisoned entry, so they are unaffected, but it is the same class of problem as #859 and probably deserves its own issue.


KaifAhmad1 (approved):
Reviewed this locally — checked out the branch, ran the full new test file plus the retargeted cp1252 tests, and reproduced the description's own claims rather than taking them on faith.

**Verified:**
- All 9 new tests in `tests/utils/test_progress_stream.py` fail against unmodified `main` (`TypeError: unexpected keyword argument 'stream'`) and pass on this branch — reproduced exactly.
- The retargeted `TestProgressTrackerEncoding` tests in `tests/deduplication/test_deduplication.py` pass, and still assert against the stream progress is actually written to (now stderr), so the original #531 regression coverage is preserved, not weakened.
- `test_backward_compatibility.py::test_progress_tracking_unchanged` reproduces as failing in the full suite and passes in isolation on this branch — confirmed pre-existing/order-dependent, not introduced here.
- Grepped the full `progress_tracker.py`: no `sys.stdout` references remain outside the intentionally-untouched `_stdout_is_tty()` attach gate. No stray `print()`/`.write()` calls were missed.
- `JupyterProgressDisplay` and `FileProgressDisplay` don't touch stdout/stderr at all, so scoping the fix to `ConsoleProgressDisplay` only is correct and complete.

**Design:** resolving `stream` via a property instead of capturing it at `__init__` is the right call — it's what makes pytest's stderr capture and the rebinding tests actually work, at no cost. The second commit's fixture fix (restoring both the `sys.modules` entry and the parent-package attribute) is a genuinely easy-to-miss subtlety and it's handled correctly.

One non-blocking note: `_safe_write(...)` and the paired `.flush()` are two separate `self.stream` property reads, so they're not atomic against a concurrent `sys.stderr` rebind between them — but that's the same non-atomicity the original `sys.stdout` code already had, not a regression.

This is a clean, well-scoped, well-tested fix. Approving.

PR #1260, opened 2026-08-30
fix(explorer): surface server error detail in graph loading failures

### LINKED ISSUE #1256: Explorer UI masks actionable 503 detail (unconfigured SEMANTICA_API_KEY) as "Fetch failed: 503"
# Explorer UI masks actionable 503 detail (e.g. unconfigured `SEMANTICA_API_KEY`) as "Fetch failed: 503"

## Environment

- semantica 0.6.7 (pip wheel), Windows 11, Python 3.11.9
- Launched with `semantica-explorer --graph my_graph.json --port 8020` on default localhost, without `SEMANTICA_API_KEY` set

## Steps to reproduce

1. Start the Explorer without `SEMANTICA_API_KEY` (and without `SEMANTICA_ALLOW_ANONYMOUS=true`).
2. Open the UI in a browser.
3. The graph canvas fails with a red overlay: **"Could not load the graph — The Explorer API did not return graph data. Check that the backend is running and reachable, then try again."** and a detail line `Fetch failed: 503`.

## Actual behaviour

The backend's `require_auth` dependency already produces an excellent, actionable message
(`semantica/explorer/dependencies.py:59-67`):

> "Server is not configured for authentication. Set the SEMANTICA_API_KEY environment variable, or explicitly opt into unauthenticated access (development only) with SEMANTICA_ALLOW_ANONYMOUS=true."

…but the frontend discards it. In `explorer/src/workspaces/GraphWorkspace/useLoadGraph.ts` (lines 337-339 and 391-394) both fetch loops do:

```ts
const response = await fetch(url.toString(), { signal });
if (!response.ok) {
  throw new Error(`Fetch failed: ${response.status}`);
}
```

so the JSON body's `detail` never reaches `GraphLoadingOverlay` (which *does* render an `{error}` detail line at line 309). A first-time user is told to "check that the backend is running" while the server is up and literally telling them the one env var to set. I burned real time on exactly this loop before reading the source.

## Expected behaviour

On a non-OK response, read the response body and surface `detail` in the thrown error, e.g.:

```ts
if (!response.ok) {
  let detail = "";
  try {
    const body = await response.json();
    if (typeof body?.detail === "string") detail = ` — ${body.detail}`;
  } catch { /* non-JSON body */ }
  throw new Error(`Fetch failed: ${response.status}${detail}`);
}
```

Happy to open a PR for this (both fetch sites in `useLoadGraph.ts`).

## Related

- The same "silent failure" class affects the WebSocket: default `ALLOWED_ORIGINS` only contains the Vite dev ports (5173), so running the server on any other port gets `WS /ws/graph-updates → 403` and the UI doesn't surface that either. I'll file the missing `ALLOWED_ORIGINS` documentation separately.

wanglin1111111:
Closes #1256

## Problem

When the Explorer REST API returns a non-OK status, the graph loading hooks throw `Fetch failed: <status>` and discard the response body. The server's `detail` field often contains exactly the guidance the user needs — e.g. with an unconfigured `SEMANTICA_API_KEY`, `require_auth` returns:

> "Server is not configured for authentication. Set the SEMANTICA_API_KEY environment variable, or explicitly opt into unauthenticated access (development only) with SEMANTICA_ALLOW_ANONYMOUS=true."

… but the UI overlay only showed `Fetch failed: 503` plus a generic "check that the backend is running" hint, sending users down the wrong troubleshooting path (reported in #1256).

## Change

- `useLoadGraph.ts`: on non-OK responses in the nodes/edges fetch loops, read the JSON body and append its `detail` string to the thrown error. Non-JSON bodies degrade gracefully to the previous status-only message.
- `GraphLoadingOverlay` already renders the `{error}` detail line, so no UI changes were needed.

## Verification

- `bunx tsc --noEmit -p tsconfig.app.json` passes (clean, 0 errors).
- Manually reproduced the original failure (Explorer without `SEMANTICA_API_KEY`) and confirmed the server detail is now included in the error surfaced by the overlay.

Happy to adjust wording/format if maintainers prefer the detail surfaced differently (e.g. as a separate field in the progress/error state rather than appended to the message).

### DIFF
--- a/explorer/src/workspaces/GraphWorkspace/useLoadGraph.ts
+++ b/explorer/src/workspaces/GraphWorkspace/useLoadGraph.ts
@@ -318,6 +318,20 @@ interface EdgeListResponse {
 
 const PAGE_LIMIT = 1000;
 
+/** Surface the server's `detail` message (e.g. auth/setup guidance) on non-OK responses. */
+async function fetchErrorDetail(response: Response): Promise<string> {
+  try {
+    const body: unknown = await response.json();
+    const detail = (body as { detail?: unknown } | null)?.detail;
+    if (typeof detail === "string" && detail.trim()) {
+      return ` — ${detail.trim()}`;
+    }
+  } catch {
+    // Non-JSON or unreadable body: fall back to the status-only message.
+  }
+  return "";
+}
+
 async function fetchAllNodes(
   signal: AbortSignal,
   onProgress?: (progress: GraphLoadProgress) => void,
@@ -335,7 +349,7 @@ async function fetchAllNodes(
 
     const response = await fetch(url.toString(), { signal });
     if (!response.ok) {
-      throw new Error(`Fetch failed: ${response.status}`);
+      throw new Error(`Fetch failed: ${response.status}${await fetchErrorDetail(response)}`);
     }
 
     const data: NodeListResponse = await response.json();
@@ -390,7 +404,7 @@ async function fetchAllEdges(
 
     const response = await fetch(url.toString(), { signal });
     if (!response.ok) {
-      throw new Error(`Fetch failed: ${response.status}`);
+      throw new Error(`Fetch failed: ${response.status}${await fetchErrorDetail(response)}`);
     }
 
     const data: EdgeListResponse = await response.json();


ZohaibHassan16 (approved):
This should cover the error paths as well.

Approving.

Issue #1253, opened 2026-08-29
[FEATURE] Add Anthropic provider to semantica.llms


ZohaibHassan16:
### Description

I noticed that `semantica.llms` currently exposes `Groq`, `OpenAI`, `HuggingFaceLLM` and `LiteLLM`, but there isn't an `Anthropic` class.

This seems a little odd since Claude is mentioned a few times in `docs/guides/llm-integrations.md`, including as an example of a frontier model / something to validate accuracy with.

Right now Claude can still be used through LiteLLM:

```python
from semantica.llms import LiteLLM

llm = LiteLLM(model="anthropic/claude-sonnet-4-20250514")
```

but this doesn't work:

```python
from semantica.llms import Anthropic
```

I checked it directly:

```text
>>> from semantica.llms import Anthropic
ImportError: cannot import name 'Anthropic' from 'semantica.llms'

>>> import semantica.llms as m
>>> m.__all__
['Groq', 'OpenAI', 'HuggingFaceLLM', 'LiteLLM']
```

### What I found

It looks like most of the Anthropic implementation is actually already there.

`semantica/semantic_extract/providers.py` has an `AnthropicProvider(BaseProvider)` with the client setup, `is_available()`, `generate()`, structured output support, instructor integration, etc.

It is already being used internally for entity/relation extraction, so I don't think a new provider implementation is needed. It seems like the public `semantica.llms` wrapper was just never added.

I also tried initializing the provider without the package/API key and it seems to fail/degrade the same way as the other providers.

The optional dependency is already in `pyproject.toml` too:

```toml
llm-anthropic = ["anthropic>=0.122.0"]
```

and it's already part of the `all` extra.

I also checked the history of `semantica/llms/__init__.py`. It looks like the file has only had the original module creation commit, so my guess is Anthropic just wasn't added when this public module was introduced.

### Possible implementation

I think this could mostly follow the existing `openai.py` / `groq.py` wrappers.

Something like:

* add `semantica/llms/anthropic.py`
* wrap the existing `AnthropicProvider`
* use an Anthropic default model
* use `ANTHROPIC_API_KEY`
* export `Anthropic` from `semantica/llms/__init__.py`
* add it to `__all__`
* update the module docstring
* add the same kind of tests as the other LLM wrappers
* update `docs/guides/llm-integrations.md` so Anthropic can be shown directly instead of only through `LiteLLM`

### One related thing

While looking at this I noticed `BaseProvider` also has `generate_typed()`.

[...]

PR #1262, opened 2026-08-30
feat(llms): add Gemini, Ollama, DeepSeek, Novita provider wrappers

### LINKED ISSUE #1261: [FEATURE] Expose Gemini, Ollama, DeepSeek, and Novita as first-class semantica.llms providers
I noticed there are a few providers in `semantica/semantic_extract/providers.py` which aren't exposed through `semantica.llms`.

Right now `semantica.llms` has wrappers for `Groq`, `OpenAI`, `HuggingFaceLLM`, `LiteLLM`, and recently `Anthropic`, but there are also implementations for:

* `GeminiProvider`
* `OllamaProvider`
* `DeepSeekProvider`
* `NovitaProvider`

As far as I can tell these providers are already implemented. They have `__init__`, `is_available`, `generate`, and `generate_structured`, and also get `generate_typed` from `BaseProvider`.

The problem is that you have to import them from `semantic_extract.providers` directly, which seems like an internal module and isn't very obvious to users.

## Suggested change

Would it make sense to add wrappers for these under `semantica/llms/`, similar to the existing `Groq`, `OpenAI`, and `Anthropic` wrappers?

Like:

* `semantica/llms/gemini.py` -> `Gemini`
* `semantica/llms/ollama.py` -> `Ollama`
* `semantica/llms/deepseek.py` -> `DeepSeek`
* `semantica/llms/novita.py` -> `Novita`

They could just delegate to the existing provider classes and raise `ProcessingError` with a useful message if the required dependency isn't installed, same as the other wrappers.

They would also need to be exported from `semantica/llms/__init__.py`.

I think it would also be useful to add a small section for each one in `docs/guides/llm-integrations.md` and tests under `tests/`.

For the optional dependencies, `pyproject.toml` already has:

* `llm-gemini`
* `llm-ollama`
* `llm-deepseek`

There isn't an `llm-novita` extra at the moment. Since Novita uses the `openai` package, it could probably use `llm-openai`, or we could add `llm-novita` pointing to the same dependency.

## Why

Mainly this would make the existing providers available through the same public API as the other LLM integrations.

Ollama seems especially useful here because it lets people use local models without needing a hosted API. Gemini and DeepSeek are also fairly common providers.

At the moment someone looking through the documented `semantica.llms` API wouldn't know these integrations already exist unless they happened to look through `semantic_extract.providers`.

ZohaibHassan16:
Closes #1261

Yhe same issue as the Anthropic wrapper in #1253.

These four providers were already implemented in `semantic_extract/providers.py` but weren't exposed from `semantica.llms`, so they couldn't be used through the public API.

Changes:

* Added `Gemini`, `Ollama`, `DeepSeek` and `Novita` wrappers under `semantica/llms/`
* They follow the same pattern as `Groq`, `OpenAI` and `Anthropic` with `generate`, `generate_structured`, `generate_typed`, and `is_available`
* Added the missing `llm-novita` extra in `pyproject.toml`. It uses the `openai` dependency like DeepSeek, and is also included in `llm-all`
* Added docs for Gemini, Ollama and DeepSeek
* Updated the Novita docs to use the new wrapper instead of calling `create_provider` directly
* Added 8 tests for each provider, 32 tests total, following the same pattern as `test_llm_anthropic.py`

### DIFF
--- a/docs/guides/llm-integrations.md
+++ b/docs/guides/llm-integrations.md
@@ -62,7 +62,7 @@ Four factors drive provider selection, each optimized for different use cases:
 
 **Accuracy** matters most in high-stakes decisions: clinical contraindication checks, credit committee reasoning, and legal document analysis. Frontier models like Claude or GPT-4 available through `LiteLLM` provide the strongest reasoning capabilities.
 
-**Data residency** constraints eliminate cloud providers for classified or HIPAA-regulated workloads. `HuggingFaceLLM` with local model paths enables fully air-gapped deployments without network calls.
+**Data residency** constraints eliminate cloud providers for classified or HIPAA-regulated workloads. `HuggingFaceLLM` with local model paths, or `Ollama` pointed at a local server, both enable fully air-gapped deployments without network calls.
 
 **Cost at scale** favors high-throughput providers like Novita AI for bulk extraction pipelines processing thousands of documents per hour where per-token costs accumulate quickly.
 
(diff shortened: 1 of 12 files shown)


Sameer6305 (dismissed):
Thanks @ZohaibHassan16  for putting this together! I went through the implementation, existing provider patterns, tests, dependency changes, and the underlying provider code. Overall, I think the direction looks good and the wrappers follow the existing `semantica.llms` pattern nicely.

I did find a few things that I think are worth addressing before merging.

### 🔴 Blocking — Gemini legacy SDK credential isolation

The legacy `google-generativeai` path calls:

```python
old_genai.configure(api_key=self.api_key)
````

`configure()` uses module-level configuration. Because of that, when multiple `GeminiProvider` instances are created with different API keys in the same process, the later `configure()` call can overwrite the credentials used by an already-created `GenerativeModel`.

For example:

```python
provider_a = GeminiProvider(api_key="key_a")
provider_b = GeminiProvider(api_key="key_b")

# provider_a may now use key_b
provider_a.generate(...)
```

This could be particularly problematic for multi-tenant applications where different users/tenants supply their own credentials, since requests could be made using another tenant's API key.

Could we either:

* avoid the legacy SDK path in this wrapper,
* ensure credentials are isolated per provider instance, or
* explicitly prevent/warn about this unsafe multi-key usage?

I consider this the main blocking issue because of the potential for cross-tenant credential contamination.

### 🟡 Non-blocking — DeepSeek/Novita silently drop supported parameters

The DeepSeek and Novita implementations currently only forward:

```python
temperature
max_tokens
```

through `_add_if_set()`.

However, these are OpenAI-compatible APIs, and the existing Groq implementation forwards additional parameters such as:

```python
top_p
frequency_penalty
presence_penalty
seed
stop
user
```

This means something like:

```python
deepseek.generate(
    "prompt",
    temperature=0.7,
    top_p=0.9,
    seed=42,
)
```

will silently ignore `top_p` and `seed`.

Would it make sense to forward the supported parameters here as well? This would also make the behavior more consistent with the existing OpenAI-compatible provider wrappers.

### 🟡 Non-blocking — Duplicate `base_url` assignment

I noticed `DeepSeekProvider.__init__()` assigns:

```python
self.base_url = "https://api.deepseek.com/v1"
```

twice.

This doesn't appear to affect runtime behavior, but it looks accidental and would be good to clean up.

### 🟡 Non-blocking — Gemini legacy per-call model handling

In the legacy Gemini path, `generate_structured()` handles a per-call `model` override, but `generate()` appears to continue using the model/client created during initialization.

For consistency, could we make the two paths behave the same when a caller passes:

```python
model="..."
```

?

Also, when using the legacy SDK, an alternate model currently results in constructing a new `GenerativeModel` for that call. If practical, caching those model instances could avoid unnecessary repeated construction.

### 🟢 Minor API consistency

I also noticed that `Gemini.generate_structured()` and `Ollama.generate_structured()` are typed as returning `Dict`, while the underlying structured parsing can also produce a top-level list. The newer Anthropic/DeepSeek/Novita wrappers use the broader return type.

This isn't a blocker, but it may be worth aligning the annotations.

---

Other than the above, I think the overall implementation is in good shape. The public exports, optional dependency handling, wrapper delegation, documentation, and test coverage all look good, and I appreciate that the existing provider issues were addressed as part of making these providers public.

If the Gemini credential-isolation issue is addressed and the remaining items above are either fixed or intentionally tracked as follow-ups, I think this will be in a good state to merge.

Thanks!

Sameer6305 (approved):
Thanks @ZohaibHassan16 for addressing the review feedback.

### What looked solid
- Clean first-class wrappers for Gemini, Ollama, DeepSeek, and Novita.
- Good integration with the existing `semantica.llms` API pattern.
- Optional dependency handling, exports, docs, and test coverage look good.
- Error handling and provider delegation are consistent with the existing implementations.

### Review fixes
The issues identified during review, including the Gemini credential-isolation concern and provider parameter/model handling, were addressed in your `fix(llms): address review findings` commit (`2f8a98e`).

I re-reviewed the latest changes and confirmed the fixes are in place.

### Non-blocking follow-ups
We did identify a few minor API consistency/cleanup items during the review. They are not blockers for this PR, and we’ll raise separate issues for those so they can be tracked independently.

Everything looks good from my side. **Approving this PR.** 👍

PR #1264, opened 2026-08-30
feat(vector_store): add scan_vectors enumeration and wire up store mi…

ZohaibHassan16:
Vector stores currently only have `get_vector(id)` and `count()`, so there wasn't any way to loop over all vectors in a store.

Because of that, `semantica store migrate` couldn't actually migrate anything and always told the user to export/reindex manually.

This adds vector scanning for faiss, sqlite and pgvector for now. These backends can support normal pagination without too much extra backend-specific logic.

pinecone/qdrant/milvus/weaviate are not included yet since they each have different ways of listing data (scroll, list(), cursors, etc). Migrate will still show the existing manual export/reindex message for those.

Changes:

* added `scan_vectors(offset, limit)` to FAISSStore, SQLiteVecStore and PgVectorStore
* added `scan_vectors()` and `iter_vectors()` to `VectorStore`
* inmemory is handled directly, other backends delegate if they support it
* unsupported backends raise `NotImplementedError` instead of returning an empty result
* `store migrate` now works between faiss/sqlite/pgvector
* vectors and metadata are copied in batches
* `--namespace` is added to metadata when it doesn't already have one
* added 22 tests for backend scanning, facade behavior and the migrate CLI

### DIFF
--- a/semantica/cli.py
+++ b/semantica/cli.py
@@ -3718,6 +3718,36 @@ _MIGRATE_SUPPORTED_BACKENDS = {"faiss", "sqlite", "pgvector"}
 _MIGRATE_BATCH_SIZE = 500
 
 
+def _migrate_backend_config(vs_cfg: Dict[str, Any], backend: str) -> Dict[str, Any]:
+    """Resolve per-backend config out of the vector_store config section.
+
+    Supports both a per-backend nested shape (``vector_store.faiss.dimension``)
+    and the common flat single-backend shape (``vector_store.backend`` +
+    sibling keys), since either can appear depending on how many backends a
+    user has configured.
+    """
+    nested = vs_cfg.get(backend)
+    if isinstance(nested, dict):
+        return dict(nested)
+    if vs_cfg.get("backend") == backend:
+        return {k: v for k, v in vs_cfg.items() if k != "backend"}
+    return {}
+
+
+def _require_faiss_index_path(cfg: Dict[str, Any], role: str) -> str:
+    """FAISS has no server to hold state between commands: a fresh FAISSStore
+    starts empty and nothing outside the process persists it, so migration
+    needs an explicit on-disk index to read from or write to."""
+    index_path = cfg.get("index_path")
+    if not index_path:
+        raise click.ClickException(
+            f"faiss as migration {role} requires 'index_path' in the vector_store "
+            f"config (vector_store.faiss.index_path or vector_store.index_path "
+            f"when faiss is the configured backend)."
+        )
+    return index_path
+
+
 @store.command("migrate")
 @click.option("--from", "from_backend", required=True)
 @click.option("--to", "to_backend", required=True)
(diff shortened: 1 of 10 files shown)


Sameer6305 (dismissed):
@ZohaibHassan16 thanks for working on this! I went through the migration flow, the new scan/iteration APIs, the backend-specific implementations, the CLI path, and the new tests across multiple review passes.

Overall, the implementation looks solid. The pagination/iteration logic, metadata handling, ID propagation, SQL safety, batch handling, and error reporting all look good.

I did find one issue that I think should be addressed before merging:

### 🔴 Blocking — FAISS migration is not idempotent

The migration path passes the source IDs through to the destination:

`store migrate` → `iter_vectors()` → `dest.store_vectors(..., ids=...)` → `FAISSStore.add_vectors()`.

For SQLite and pgvector, writing an existing ID replaces/updates the existing record. FAISS, however, currently appends the vectors and IDs to the index without checking whether the ID already exists.

This means rerunning a migration into an existing FAISS destination can create duplicate vectors for the same logical ID.

For example:

```text
First migration:
A, B, C, D

Migration is interrupted after these vectors are written.

Rerunning the migration:
A, B, C, D
````

With FAISS, the second run can append another copy of those vectors rather than updating the existing entries.

This can lead to:

* `count()` becoming larger than the number of logical IDs
* duplicate vectors appearing in similarity search results
* `get_vector(id)` resolving to only one of the duplicate entries
* metadata being keyed by the same ID while multiple physical vectors exist
* repeated migrations accumulating duplicates

This is particularly relevant for the migration command because partial migrations are a realistic scenario, and users may reasonably retry a failed migration.

Could we make the FAISS destination behavior idempotent, or otherwise explicitly handle existing IDs during migration?

A fix could be implemented at the FAISS store/index layer or in the migration flow, depending on what fits the existing design best. The important part is that rerunning a migration should not silently create duplicate vectors for the same logical IDs.

### 🟢 Other areas checked

I also specifically checked the following during the review:

* `iter_vectors()` offset advancement and termination
* final partial batches
* SQLite/pgvector deterministic pagination
* ID propagation through `store_vectors()`
* metadata and namespace handling
* migration success/error reporting
* unsupported backend handling
* SQL parameterization and identifier handling
* connection/resource cleanup
* potential infinite/retry loops
* vector dimension handling

I didn't find additional high-confidence issues in these areas.

I also looked into FAISS indexes where `get_vector()` may not be able to reconstruct a vector (such as PQ). That appears to be a pre-existing FAISS limitation rather than something introduced by this PR, so I don't think it should block this change.

Once the FAISS duplicate-ID/idempotency issue is addressed, I think this PR will be in good shape to merge.

Thanks!

Sameer6305 (approved):
@ZohaibHassan16 Thanks for addressing the review feedback!

### What looked solid
- Clean `scan_vectors()` / `iter_vectors()` implementation across the supported backends.
- Migration batching, pagination, metadata, IDs, and namespace handling look correct.
- Unsupported backends are handled appropriately.
- Good test coverage for scanning, migration, and FAISS idempotency.
- No additional high-confidence security, data-integrity, or reliability issues found.

### Review fixes
The previously raised FAISS idempotency issue was addressed in the latest changes. I re-reviewed the implementation and confirmed that duplicate IDs are now handled correctly during migration retries within the same session.

### Non-blocking follow-up
One remaining FAISS persistence limitation is that `vector_ids` and metadata are not preserved across save/load, which can affect cross-session migration retries. This is non-blocking and we'll raise it as a separate issue.

Overall, the core migration functionality looks good and the requested changes have been addressed.

**Approving this PR.** 👍

PR #1266, opened 2026-08-30
ci: reusable install action, install-matrix, and release hardening

KaifAhmad1:
## Summary

Distribution and trust-signal infrastructure to make `pip install semantica` frictionless in downstream CI, and to bring the release pipeline in line with mature OSS practice.

- **`.github/actions/setup-semantica`** — reusable composite action other repos can call (`uses: semantica-agi/semantica/.github/actions/setup-semantica@main`) to install + verify `semantica` in one step
- **`install-matrix.yml`** — verifies the *published* PyPI package installs and imports cleanly on Ubuntu/macOS/Windows across Python 3.9-3.12, weekly + on every release; backs a new "pip install" README badge
- **`scorecard.yml`** — OpenSSF Scorecard analysis, weekly + on push to main, backing a new README badge (a concrete trust signal security/procurement teams check before adoption)
- **`release.yml`** — added a `twine check` gate before publish, catching a broken PyPI long-description render before it goes live (the existing Trusted Publishing/OIDC + SLSA attestation flow is unchanged)
- **`CITATION.cff`** — enables GitHub's native "Cite this repository" button, complementing `docs/citation.md`
- **`examples/ci/`** — copy-paste GitHub Actions, GitLab CI, and CircleCI templates for projects adopting `semantica`
- **`GROWTH.md`** — tracked checklist of distribution channels (what's done vs. outstanding), with explicit guardrails against artificially inflating download/install metrics

## Test plan

- [x] All new/modified YAML validated with `yaml.safe_load`
- [x] New action pins (`ossf/scorecard-action`, `github/codeql-action/upload-sarif`) resolved live against the GitHub API to satisfy `verify-action-pins.sh`
- [x] Local composite action reference (`./.github/actions/setup-semantica`) confirmed exempt from SHA-pin verification (script skips `uses: ./...`)
- [ ] Confirm `install-matrix.yml` and `scorecard.yml` run green after merge to `main` (both trigger on `push: branches: [main]`, plus `install-matrix` also on `release: published`)

### DIFF
--- a/.github/actions/setup-semantica/action.yml
+++ b/.github/actions/setup-semantica/action.yml
@@ -31,13 +31,17 @@ runs:
 
     - name: Install semantica
       shell: bash
+      env:
+        SEMANTICA_EXTRAS: ${{ inputs.extras }}
+        SEMANTICA_VERSION: ${{ inputs.version }}
       run: |
         python -m pip install --upgrade pip
-        if [ -n "${{ inputs.extras }}" ]; then
-          pip install "semantica[${{ inputs.extras }}]${{ inputs.version }}"
+        if [ -n "$SEMANTICA_EXTRAS" ]; then
+          spec="semantica[$SEMANTICA_EXTRAS]$SEMANTICA_VERSION"
         else
-          pip install "semantica${{ inputs.version }}"
+          spec="semantica$SEMANTICA_VERSION"
         fi
+        python -m pip install -- "$spec"
 
     - name: Verify install
       id: verify
--- a/.github/actions/setup-semantica/action.yml
+++ b/.github/actions/setup-semantica/action.yml
@@ -1,0 +1,48 @@
+name: 'Setup Semantica'
+description: 'Install Python, cache pip, and install the semantica package into a workflow'
+author: 'Semantica'
+
+inputs:
+  python-version:
+    description: 'Python version to set up'
+    required: false
+    default: '3.11'
+  version:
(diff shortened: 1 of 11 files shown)


ZohaibHassan16 (changes requested):
There’s still one issue from the latest Codex pass though: the `workflow_run` path in `install-matrix.yml` calls `setup-semantica` without passing a version. That means it just runs `pip install semantica`, so it installs whatever the latest version on PyPI is, then compares that installed version against the tag that triggered the workflow.

That doesn’t necessarily verify the release that just got published. If two releases happen close together, or the workflow is delayed/retried after a newer release is available, this can fail even though the original release is perfectly fine.

The fix should be pretty small: derive the expected version from the tag before calling the composite action, then pass `version: "==${expected}"` so it installs and verifies the exact release that triggered the run.

Everything else looks good. Requesting changes for this one issue.

ZohaibHassan16 (approved):
The constraint===$expected trick in the new "Pin expected version" step is deliberate : GITHUB_OUTPUT splits on the first =, so it writes constraint=="X.Y.Z", which correctly becomes pip install semantica==X.Y.Z for the actual triggering release. Non-workflow_run triggers skip the step and fall back to unconstrained install as before. Cleaner than my original suggestion too , no leftover post-hoc comparison step, pip install itself now fails naturally on propagation lag.

Pip-cache opt-in was also addressed, docs updated to match.

Re-verified everything else independently: all 4 touched YAML files parse clean with yaml.safe_load, and all 7 SHA pins across the diff resolve exactly against the live GitHub API for their tag comments.

Approving.

PR #1284, opened 2026-08-30
fix: make LanguageDetector minimum text length configurable

### LINKED ISSUE #1281: [BUG] LanguageDetector hard-coded length cutoff misreports short non-English text
## Description

`LanguageDetector` hard-codes a minimum stripped-text length of 10 characters in `detect()`, `detect_with_confidence()`, and `detect_multiple()`. Inputs below that threshold never reach `langdetect`; they silently return the configured `default_language`, which is `"en"` by default.

This makes short non-English text look like a successful English detection. A single character-count threshold is also a weak proxy for evidence across writing systems: a short Chinese phrase can contain substantially more linguistic information than an English string of the same length.

## Reproduction

```python
from semantica.normalize import LanguageDetector

text = "你好，这是中文文本"

print(len(text.strip()))
# 9

detector = LanguageDetector()
print(detector.detect(text))
# en
```

The `"en"` result is the default fallback; the underlying detector is not called. Calling `langdetect.detect(text)` directly identifies this example as `"zh-cn"`.

The same guard also affects confidence and multi-candidate APIs:

```python
detector.detect_with_confidence(text)
# ("en", 0.0)

detector.detect_multiple(text)
# [("en", 0.0)]
```

## Actual behavior

- The minimum length is fixed at 10 and cannot be configured.
- `**options` are accepted but cannot override the threshold.
- `detect()` conflates an insufficient-text fallback with a genuine detection of the default language.

## Expected behavior

Applications should be able to configure the minimum text length for their language mix, and the fallback behavior should be documented clearly so an insufficient-text result is not mistaken for a successful detection.

## Suggested scope

- Add a `min_text_length` configuration option and use it consistently in `detect()`, `detect_with_confidence()`, and `detect_multiple()`.
- Preserve the current default threshold if backward compatibility requires it.
- Document that inputs below the threshold return the configured fallback rather than a detected language.
- Add regression coverage for a short non-Latin input and for a caller-supplied threshold.

I would be happy to implement this after the desired fallback semantics are confirmed.

taoche:
## Summary

`LanguageDetector` hard-coded a 10-character minimum stripped-text length in its detection APIs. Short non-English inputs (e.g. the 9-character `"你好，这是中文文本"`) never reached `langdetect` and silently returned the fallback, indistinguishable from a genuine English detection.

- Add a `min_text_length` configuration option (default unchanged at `10`), with per-call override via `**options` in all detection APIs and through the `detect_language()` convenience wrapper
- **Make the fallback honest**: when no detection was performed (below-threshold text, empty text, langdetect unavailable) or detection failed, return `UNKNOWN_LANGUAGE` (`"unknown"`) instead of `"en"`. `"unknown"` is distinct from every ISO code, so a fallback can never be mistaken for a detected language. Callers that want the previous behavior configure `LanguageDetector(default_language="en")`
- **Single detection core**: `detect()` and `detect_with_confidence()` now delegate to `detect_multiple()`, so guard, fallback and error-handling semantics live in exactly one place instead of being triplicated
- **Typos are reported, not swallowed**: unknown per-call option names (e.g. `min_text_len`) trigger a warning (once per name per instance) instead of being silently ignored by `**options`
- **Hardened boundaries**: invalid `min_text_length` values (None, non-numeric strings) degrade to the fallback with a warning instead of raising `TypeError`; `detect_language()` copies the stored method config before merging kwargs so per-call overrides no longer mutate the global `normalize_config`
- Document the fallback semantics and the deliberate string-sentinel vs `Optional` trade-off (vs. entity confidence in #1285); export `UNKNOWN_LANGUAGE` from `semantica.normalize`

```python
detector = LanguageDetector()
detector.detect("你好，这是中文文本")                    # "unknown" (was a fake "en")
detector.detect("你好，这是中文文本", min_text_length=5)  # "zh-cn"
```

> **Behavior change**: the default fallback is now `"unknown"` instead of `"en"`. Inside this repo `LanguageDetector` is only consumed via the config-driven `normalize.methods.detect_language()`, and no code branches on the `"en"` fallback. If you prefer to keep `"en"` as the out-of-the-box default for compatibility, I'm happy to flip the default back — the `unknown` semantics would then remain available via `default_language="unknown"`.

## Test plan

- [x] Regression tests in `tests/normalize/test_language_detector.py` cover: short non-Latin fallback + configured/per-call thresholds, `"unknown"` semantics with `0.0` confidence, `default_language="en"` compatibility, invalid threshold degradation, global-config leak, unknown-option warning, and agreement across the three detection APIs
- [x] `pytest tests/normalize/` — 93 passed

Closes #1281

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Implements the deterministic core of ontology-based information extraction (OBIE; Wimalasuriya & Dou, 2010). No new runtime dependencies
   - New `tests/semantic_extract/test_schema_validator.py`
 
+### Changed
+
+- **`LanguageDetector`'s fallback is now `"unknown"` instead of `"en"`, and its minimum text length is configurable** (closes #1281) by @taoche
+  - **Breaking for anyone reading the fallback value.** `detect()`, `detect_with_confidence()`, `detect_multiple()`, `detect_batch()` and the `detect_language()` wrapper previously returned `"en"` whenever no detection was actually performed — text below the length threshold, empty text, `langdetect` not installed — or when detection raised. `"en"` is a valid ISO code, so a fabricated fallback was indistinguishable from a real English detection. They now return `UNKNOWN_LANGUAGE` (`"unknown"`), which is distinct from every ISO code. Nothing errors on this path, so callers relying on the old value get a different string silently: pass `LanguageDetector(default_language="en")` (or `detect_language(text, default_language="en")`) to restore the previous behavior
+  - `UNKNOWN_LANGUAGE` is exported from `semantica.normalize`. The out-of-band string is deliberate rather than the `Optional` treatment given to missing entity confidence in #1285: language codes flow into dict keys, filenames and `get_language_name()` lookups, so `None` here would trade an ambiguity for a crash surface
+  - **New `min_text_length` option** (default unchanged at 10 stripped characters), settable per instance via `LanguageDetector(min_text_length=...)`, per call via `**options` on every detection API, and through `detect_language()`. Ten characters is reasonable for Latin scripts and far too many for CJK — a 9-character Chinese string never reached `langdetect` at all
+  - `detect()` and `detect_with_confidence()` now delegate to `detect_multiple()`, so the length guard, the fallback and the error handling exist once rather than in three copies that can drift
+  - Invalid `min_text_length` values (`None`, non-numeric strings, negatives) degrade to the fallback with a warning instead of raising `TypeError` from the length comparison, which sits outside the detection exception handlers
+  - Unrecognized per-call option names (e.g. `min_text_len`) now warn once per name per instance instead of being silently absorbed by `**options`
+  - `detect_language()` copies the stored method config before merging per-call kwargs, so a one-call override no longer leaks into the shared `normalize_config`
+  - Docs: `docs/reference/normalize.md` and `semantica/normalize/normalize_usage.md`
+
 ## [0.7.0] - 2026-09-07
 
 ### Changed
--- a/docs/reference/normalize.md
+++ b/docs/reference/normalize.md
@@ -440,7 +440,19 @@ utf8_text = handle_encoding(raw_bytes, operation="convert")
     ```
 
     <Note>
-      `detect()` requires at least 10 characters for reliable detection. On shorter text it returns the `default_language` (default: `"en"`).
+      `detect()` skips detection on text shorter than `min_text_length` (default: 10 stripped characters) and returns the `default_language` instead. That fallback is `"unknown"` — the detector reports that it did not look rather than guessing a language. Both are configurable, per instance or per call:
+
+      ```python
+      # Lower the threshold for CJK, where 10 characters is far more than needed
+      detector = LanguageDetector(min_text_length=4)
+      detector.detect("短いテキスト")            # → "ja"
+
+      # Per-call override
+      LanguageDetector().detect("短いテキスト", min_text_length=4)   # → "ja"
+
+      # Opt back into assuming a language on text too short to examine
+      LanguageDetector(default_language="en").detect("Hi")          # → "en"
+      ```
     </Note>
 
     <Warning>
--- a/semantica/normalize/__init__.py
+++ b/semantica/normalize/__init__.py
@@ -143,7 +143,7 @@ from .entity_normalizer import (
     EntityNormalizer,
     NameVariantHandler,
 )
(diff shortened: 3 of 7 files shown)


zeekay (dismissed):
I'm with you on the substance. Returning `"en"` for text the detector never actually examined is a fabricated answer, and `"unknown"` is the honest one. The reasoning in the `UNKNOWN_LANGUAGE` comment for why this is an out-of-band string here while missing confidence is `None` in `semantic_extract` is genuinely good — language codes flow into dict keys and filenames, so `None` would trade an ambiguity for a crash surface. That distinction is worth having written down.

Making `min_text_length` configurable is right too. Ten characters is a reasonable default for Latin scripts and far too many for CJK, and it was hardcoded.

Test run against current `main` (merged, the branch is behind): `pytest tests/normalize/` — 88 passed, 5 skipped, against 81 passed on `main`. No in-tree consumer feeds `detect_language`'s result into anything that would choke on `"unknown"` — the `_detect_language` hits in `repo_ingestor.py` and `code_parser.py` are source-file language detection and unrelated.

**What blocks it: the change makes a doc line false, and that line isn't in the diff.**

`docs/reference/normalize.md:443` says:

> `detect()` requires at least 10 characters for reliable detection. On shorter text it returns the `default_language` (default: `"en"`).

Both halves are now wrong — the 10 is configurable and the default is `"unknown"`. Measured on the branch versus `main`:

```
                        main      this branch
detect_language("Hi")   'en'      'unknown'
LanguageDetector()
  .default_language     'en'      'unknown'
```

`normalize_usage.md` is updated, which is the right instinct, but `docs/` is the published reference and it now contradicts the code. Please update that Note in this PR.

**And it needs a CHANGELOG `[Unreleased]` entry.** Any caller relying on the `"en"` fallback gets a different string after this, with no error to tell them — precisely the change a reader scans the changelog for. `[Unreleased]` is empty on `main` right now.

One smaller thought, take it or leave it: `KNOWN_DETECTION_OPTIONS` plus `_warned_options` warns once per typo'd keyword, which is friendly, but it's a second mechanism arriving alongside the main change and no other class in `normalize/` does it. If it stays I'd want it consistent across the module eventually rather than living only here.

Fix the doc Note and add the changelog line and I'll approve.

KaifAhmad1 (dismissed):
@Sameer6305 @taoche — reviewed the full thread and current diff. My read: **needs changes before merge**, not ready yet.

1. **Review requirement isn't cleared.** @zeekay's `CHANGES_REQUESTED` from Sept 6 is still the standing review on GitHub — there's no approval since, so this can't merge as-is regardless of the rest.

2. **Stale docstrings from the revert.** In `semantica/normalize/language_detector.py`, both the module docstring and the `LanguageDetector` class docstring still say "Configurable minimum text length with an 'unknown' fallback." That's no longer accurate — `"unknown"` is opt-in now (`default_language=UNKNOWN_LANGUAGE`), not the default. Small fix, but it should match the code before this ships.

3. **The `en` vs `unknown` default needs an explicit decision, not an assumed one.** @Sameer6305's revert (`ea158c8a`) is mechanically clean — it's threaded consistently through the code, CHANGELOG, docs, and tests — but it also reverses the core fix `#1281` was filed for: short text below `min_text_length` goes back to silently returning `"en"`, indistinguishable from a real detection, unless a caller opts in to `UNKNOWN_LANGUAGE`. That's a legitimate backward-compat tradeoff, but it's a product decision, not a mechanical one, and it landed on the branch before we'd actually agreed on it. Let's confirm which default we want before this merges.

Everything else — the `min_text_length` config surface, the single-detection-core refactor, the config-mutation fix, the typo-warning on unknown options, test coverage — looks solid. @taoche, good work threading this through; just want the above settled first.

Sameer6305:
@KaifAhmad1 
Thanks for the review.

1. **Review requirement** — the requested changes have been addressed; the `CHANGES_REQUESTED` status itself has not been dismissed yet, which is why I’m responding here before clearing it.

2. **Stale docstrings** — agreed, this has been fixed in the latest commit. Both docstrings now reflect `"en"` as the default and `"unknown"` as opt-in.

3. **`en` vs `unknown`** — I suggest keeping `"en"` as the default for this PR to preserve backward compatibility. `min_text_length` being configurable addresses #1281 without introducing a separate breaking API change. If we want `"unknown"` as the default, I suggest handling that separately so it can be discussed as an intentional behavior change.

If you’re okay with this direction, please re-review and approve the PR, and we can proceed with the merge.

Sameer6305 (approved):
Thanks @taoche, for the work on this and for addressing the review feedback. I’ve re-reviewed the current diff, tests, and docs, and everything looks good. **Ready to merge.**

PR #1285, opened 2026-08-30
fix: represent unavailable entity confidence as None instead of 1.0

### LINKED ISSUE #1282: [BUG] spaCy NER reports unavailable entity confidence as 1.0
## Description

The spaCy NER adapter represents unavailable per-entity confidence as `1.0`. Standard spaCy `Span` objects in `doc.ents` do not normally expose a `confidence` or `score` attribute, so every extracted entity appears to have perfect confidence even though no per-entity probability was produced.

This makes downstream filtering, review prioritization, and method comparison misleading. The same ambiguity also appears in the shared `Entity` model and `EntityConfidenceScorer`: `Entity.confidence` defaults to `1.0`, while the scorer treats `1.0` as the sentinel meaning "recalculate if needed." A genuine score of 1.0 and an unavailable score therefore cannot be distinguished.

## Reproduction

```python
from semantica.semantic_extract import NERExtractor

extractor = NERExtractor(method="ml", model="en_core_web_sm")
entities = extractor.extract("Apple Inc. was founded by Steve Jobs in Cupertino.")

for entity in entities:
    print(entity.text, entity.confidence, entity.metadata)
```

Typical output:

```text
Apple Inc. 1.0 {'extraction_method': 'ml', 'model': 'en_core_web_sm', ...}
Steve Jobs 1.0 {'extraction_method': 'ml', 'model': 'en_core_web_sm', ...}
Cupertino 1.0 {'extraction_method': 'ml', 'model': 'en_core_web_sm', ...}
```

In `extract_entities_ml()`, the value starts at `1.0` and changes only if the spaCy span happens to expose a score:

```python
confidence = 1.0
if hasattr(ent, "confidence"):
    confidence = ent.confidence
elif hasattr(ent, "score"):
    confidence = ent.score
```

The introduction cookbook compounds the issue by also displaying `1.0` whenever a result has no confidence field:

```python
confidence = entity.get("confidence", 1.0) if isinstance(entity, dict) \
    else getattr(entity, "confidence", 1.0)
```

## Actual behavior

- Absence of a model-provided score is reported as perfect confidence.
- Consumers cannot distinguish `unknown` from a genuine `1.0`.
- `EntityConfidenceScorer` overloads `1.0` as both a valid score and a recalculation sentinel.
- The cookbook presents the value as an ordinary confidence score.

## Expected behavior

Unavailable per-entity confidence should be represented explicitly and must not be reported as model certainty. Real scores supplied by extractors that support them should remain unchanged.

## Suggested approach

The cleanest model would make confidence optional:

```python
confidence: Optional[float] = None
```

The spaCy adapter could then leave it as `None` when the span exposes no score and record provenance such as `metadata["confidence_source"] = "unavailable"`. `EntityConfidenceScorer` should check for `None` rather than using `1.0` as a sentinel, and the cookbook should display `N/A` instead of defaulting to `1.00`.

If changing the field type is too disruptive, a backward-compatible transition could add explicit metadata such as `confidence_available=False` / `confidence_source="unavailable"`, update consumers to honor it, and deprecate the ambiguous default.

## Acceptance criteria

- A spaCy entity without a model-provided score is distinguishable from an entity with a genuine confidence of `1.0`.
- A score exposed by an extraction backend is preserved.
- `EntityConfidenceScorer` no longer uses `1.0` as an implicit missing-value sentinel.
- The entity-extraction cookbook renders unavailable confidence as `N/A` (or equivalent), not `1.00`.
- Regression tests cover both unavailable and backend-provided confidence.

I would like to work on this issue. If the preferred compatibility approach is confirmed, please assign it to @taoche.

taoche:
## Summary

The spaCy NER adapter fabricated `confidence=1.0` whenever a span exposed no per-entity probability — the normal case for standard spaCy models — so every `ml`-extracted entity reported perfect confidence. `EntityConfidenceScorer` compounded this by overloading `== 1.0` as a "recalculate" sentinel, which also silently overwrote genuine backend scores of 1.0.

**Root cause: an in-band sentinel.** "No measurement" was encoded as the maximum value of the measurement. This PR makes absence a first-class state:

- `Entity.confidence` is now `Optional[float]` defaulting to `None` ("unknown", not "certain"); producers record provenance under `metadata[CONFIDENCE_SOURCE_KEY]` using the `CONFIDENCE_SOURCE_*` constants owned by `types.py` (`model` / `heuristic` / `type_similarity` / `unavailable`)
- The spaCy adapter leaves `None` when the span exposes no score; backend-provided scores are preserved unchanged
- **The `extract()` pipeline scores unknown confidences (heuristic, labeled) before filtering**, so `extract()` guarantees every emitted entity carries a numeric confidence with provenance — downstream consumers never reason about `None`. Measured scores are never overwritten; `None` remains representable for code that bypasses the pipeline
- A single policy helper `meets_confidence_threshold()` defines the unknown-value semantics for the standalone filter APIs: unknown passes — absence of evidence is not low confidence (matches legacy behavior, no entities newly dropped)
- `EntityConfidenceScorer` recalculates only when `confidence is None`, so a genuine 1.0 survives scoring
- Ensemble voting averages only measured scores; `calculate_weighted_confidence()` falls back to similarity-only scoring when no measured confidence exists, and stays `None` when the caller explicitly disables similarity weighting
- `ExtractionValidator` counts unscored entities in a separate `unscored` metric and treats unknown as neutral, not zero
- The introduction cookbook and `semantic_extract_usage.md` document the layered semantics and render unavailable confidence as `N/A`

### Acceptance criteria from the issue

- [x] A spaCy entity without a model-provided score is distinguishable from a genuine `1.0` (provenance metadata + `None` at the adapter level)
- [x] Scores exposed by extraction backends are preserved
- [x] `EntityConfidenceScorer` no longer uses `1.0` as an implicit missing-value sentinel
- [x] Cookbook renders unavailable confidence as `N/A`
- [x] Regression tests cover both unavailable and backend-provided confidence

### Blast radius

**`extract()` callers see no behavioral change.** The pipeline scores unknown confidences before filtering, so every entity `extract()` emits still carries a numeric confidence, now with provenance. All numeric comparisons and aggregations inside `semantic_extract` were updated to be `None`-aware.

**`Entity` is not confined to `semantic_extract`, and the bypass path can now surface `None`.** `semantica/kg/graph_builder.py:157` and `:171` read `getattr(item, "confidence", 1.0)` and copy the value straight into the graph entity dicts. So an `Entity` constructed directly rather than obtained from `extract()` now lands in the graph as `confidence: None` where it previously landed as `1.0`:

```python
e = Entity(text="Acme", label="ORG", start_char=0, end_char=4)
g = GraphBuilder(resolve_conflicts=False).build([e], extract=False)
g["entities"][0]["confidence"]        # None (was 1.0)
g["entities"][0]["confidence"] > 0.5  # TypeError
```

Nothing in `semantica/` compares that value numerically — `graph_validator.py`, `entity_resolver.py` and the explorer routes only use `>` on the unrelated `confidence_decay` — so this is not a break in-tree. But a downstream consumer that builds `Entity` objects itself and then sorts or filters the resulting graph by confidence will get a `TypeError`, and that is stated plainly here rather than discovered from a traceback. (Credit to @zeekay for catching that the original wording of this section was wrong.)

`Relation`/`Triplet` and the separate `utils.types.Entity` are untouched. LLM/HuggingFace/pattern/regex paths always set explicit float scores and behave exactly as before.

## Test plan

- [x] `tests/semantic_extract/test_entity_confidence.py` (17 tests, mocked spaCy runtime): unavailable vs backend-provided scores, scorer sentinel removal, pipeline scoring guarantee, `None`-aware filtering/voting/metrics, similarity-only and disabled-weight edge cases
- [x] `pytest tests/semantic_extract/` — 175 passed
- [x] Pre-existing failures elsewhere reproduce identically on `main` (unrelated environment issues)

Closes #1282


zeekay (approved):
This is the right diagnosis and the right fix. Encoding "no measurement" as the maximum value of the measurement is the bug behind both symptoms, and the two-layer answer — `None` is representable at the type level, `extract()` guarantees a number with provenance before anything downstream sees it — is better than either "always a float" or "callers must handle None everywhere".

`EntityConfidenceScorer` keying recalculation off `== 1.0` was the part I'd have flagged if it weren't already fixed here: it meant a backend that genuinely reported 1.0 got silently overwritten, so the sentinel was destroying real data, not just mislabeling missing data.

Test runs against current `main` (merged, the branch is behind), unfiltered:
- `pytest tests/semantic_extract/ tests/kg/` — 682 passed, 11 skipped, 3 subtests, against 665 on `main`
- `pytest tests/export/ tests/conflicts/ tests/visualization/` — 417 passed, 1 failed, and that failure (`test_a_name_with_a_space_still_mints_a_valid_iri`, `ModuleNotFoundError`) is identical on `main`

Also exercised the paths the type change actually reaches, since a `None` where a float used to be is the kind of thing that surfaces three modules away: `GraphBuilder` with default settings, with `merge_entities=True`, and with conflict resolution on, plus `GraphValidator` over the result. All fine.

**Two things to fix before this merges.**

**The blast-radius claim isn't quite right.** The description says the `Entity` dataclass "is consumed only inside `semantic_extract`". It isn't — `semantica/kg/graph_builder.py:157` and `:171` read `getattr(item, "confidence", 1.0)` and copy it into the graph entity dicts. So an `Entity` constructed directly (not via `extract()`) now lands in the graph as `confidence: None` where it previously landed as `1.0`:

```python
e = Entity(text="Acme", label="ORG", start_char=0, end_char=4)
g = GraphBuilder(resolve_conflicts=False).build([e], extract=False)
g["entities"][0]["confidence"]        # None (was 1.0)
g["entities"][0]["confidence"] > 0.5  # TypeError
```

Nothing in `semantica/` compares that value numerically today — I checked `graph_validator.py`, `entity_resolver.py` and the explorer routes; the only `>` comparisons there are on `confidence_decay`, which is unrelated — so this isn't a break in-tree, and the pipeline guarantee covers everyone who goes through `extract()`. But a downstream consumer that builds `Entity` objects itself and then sorts or filters the resulting graph by confidence will get a `TypeError`, and they should be able to learn that from the PR rather than from a traceback. Please correct the sentence and say plainly that graph entity dicts can now carry `None` on the bypass path.

**It needs a CHANGELOG `[Unreleased]` entry.** A public dataclass field changing from `float` to `Optional[float]` is exactly what that section is for, and it's empty on `main` right now. The docs side is already handled — `semantic_extract_usage.md` and the cookbook rendering `N/A` are the right touches.

I don't have write access on this repo, so I can't merge it — approving the code so a maintainer can, once those two are in.

Sameer6305 (approved):
Reviewed this PR end-to-end.

The PR fixes spaCy NER reporting unavailable entity confidence as `1.0` by making `None` represent unavailable confidence and removing `1.0` as the recalculation sentinel, while preserving genuine backend-provided scores.

During review, we found and fixed four failing regression tests caused by mocking the wrong spaCy entry point, an `Agno` integration regression with `float(None)`, misleading `0.00` rendering for unavailable average confidence, and an outdated documentation example.

The focused confidence tests now pass, and the remaining broader test failures were verified as pre-existing on `main`.

The implementation and tests now look correct from my review. **Approved.**

PR #1312, opened 2026-08-30
This is resubmit of the pr for issue feat(integrations): add Google ADK support

### LINKED ISSUE #965: [FEATURE] Semantica Google ADK Integration
## Problem Statement

`integrations/__init__.py` already names Google ADK as a target for framework integrations, but no `integrations/google_adk/` package exists yet. Google's Agent Development Kit has its own tool and session conventions (`google.adk.tools.BaseTool`/`FunctionTool`, `google.adk.sessions.BaseSessionService`), and Semantica should provide the same kind of first-class adapter it already has for other frameworks.

## Proposed Solution

Add `integrations/google_adk/`:

```
integrations/google_adk/
├── __init__.py           # exports, __version__, ADK_AVAILABLE flag
├── kg_tools.py            # semantica_kg_tools() -> List[FunctionTool]: extract/add/query the ContextGraph
├── decision_tools.py       # semantica_decision_tools() -> List[FunctionTool]: record/query decisions
├── session_service.py       # SemanticaSessionService(BaseSessionService): graph-backed session state
└── README.md
```

### `kg_tools.py` / `decision_tools.py`

ADK's `FunctionTool` wraps a plain Python function, using type hints and the docstring to generate its schema. These should be thin functional wrappers around Semantica's extraction, graph, and decision-intelligence calls.

```python
def extract_entities(text: str) -> dict:
    """Extract named entities from text using Semantica's NER pipeline."""
    ...

def query_graph(query: str) -> dict:
    """Query the shared Semantica knowledge graph by keyword or Cypher."""
    ...

def semantica_kg_tools(graph: Optional["ContextGraph"] = None) -> List["FunctionTool"]:
    """Return ADK FunctionTools bound to a shared ContextGraph instance."""
    ...
```

### `SemanticaSessionService` (`session_service.py`)

Implements ADK's `BaseSessionService` (`create_session`, `get_session`, `append_event`, `list_sessions`) backed by `semantica.context.ContextGraph`, so ADK agent session/state history is persisted as graph nodes instead of ADK's default in-memory/Vertex-backed session store.

## Use Cases

1. An ADK agent's tool-calling loop uses `semantica_kg_tools()` to ground responses in a shared knowledge graph across multiple ADK agents/sub-agents in a workflow.
2. `SemanticaSessionService` persists ADK session state as a queryable graph, enabling cross-session precedent lookup ("has this user asked something like this before?").
3. A multi-agent ADK `SequentialAgent`/`ParallelAgent` pipeline shares a single `ContextGraph` across sub-agents via the session service, avoiding re-extraction of entities already known from an earlier stage.

## Impact Assessment

- **Who would benefit?** Teams building on Google's Agent Development Kit who want persistent, queryable shared context instead of ADK's default session backends.
- **Priority**: Medium
- **Breaking Changes**: No. Additive, isolated under `integrations/google_adk/`.
- **Dependencies**: `google-adk` as an optional extra only.

## Implementation Ideas / Contribution checklist

- [ ] `integrations/google_adk/__init__.py`: export `semantica_kg_tools`, `semantica_decision_tools`, `SemanticaSessionService`, `ADK_AVAILABLE`, `__version__`
- [ ] `integrations/google_adk/kg_tools.py`: implement functional tools and `semantica_kg_tools()`, lazy-import `google.adk.tools.FunctionTool`, fall back gracefully when absent
- [ ] `integrations/google_adk/decision_tools.py`: implement `semantica_decision_tools()`
- [ ] `integrations/google_adk/session_service.py`: implement `SemanticaSessionService`
- [ ] `integrations/google_adk/README.md`: install and quickstart docs
- [ ] `pyproject.toml`: add a `google-adk = ["google-adk>=1.0.0"]` extra alongside the existing integration extras, and add to the `all` extras bundle
- [ ] `tests/integrations/google_adk/`: unit tests using `pytest.importorskip("google.adk")` for the present-case tests

```python
# Example usage once implemented
from google.adk.agents import Agent
from integrations.google_adk import semantica_kg_tools

graph = ContextGraph()
agent = Agent(
    name="researcher",
    model="gemini-2.0-flash",
    tools=semantica_kg_tools(graph=graph),
)
```

## Additional Context

- ADK tools docs: https://google.github.io/adk-docs/tools/
- ADK sessions docs: https://google.github.io/adk-docs/sessions/

## Contribution

- [ ] I'm willing to help implement this feature
- [ ] I can help with documentation
- [ ] I can help with testing

Hitesh-XS:
for issue #965 

Changes Made
Restored pyproject.toml state: Reverted accidental rollbacks to keep the main branch packaging changes intact (restored version 0.6.7, correct anthropic and pyarrow bounds, ingest-sap, langchain, and package-data fixes).

Added Google ADK dependency: Injected google-adk into the Agentic Framework Integrations section and properly bundled it into the all dependency group.

Fixed Pinecone Exception: Updated the vectorstore-pinecone dependency from the deprecated pinecone-client to the official pinecone package. This resolves the initialization crash that was preventing the context graph—and consequently all integration tests—from running.

Testing
Successfully installed the google-adk extras locally.

Ran the ADK test suite (pytest tests/integrations/google_adk -v).

All 36 tests are now passing seamlessly with the corrected Pinecone dependency.


ZohaibHassan16 (changes requested):
Most of it looks good, but there’s one blocker.

`extract_relations()` is currently broken. Running it directly gives:

```python
>>> extract_relations('Alice works at Acme Corp. Bob founded Acme Corp.')
{'relations': [], 'count': 0, 'error': "'RelationExtractor' object has no attribute 'extract_entities'"}
```

In `kg_tools.py`, `extract_relations()` calls:

```python
_get_relation_extractor().extract_entities(text)
```

but `RelationExtractor` doesn’t have an `extract_entities()` method. That belongs to `NERExtractor`. The standalone `extract_entities()` function a couple of functions above already uses `_get_ner_extractor()` correctly.

As written, every call to `extract_relations()` returns an empty relation list and puts the `AttributeError` in the `"error"` field.

This also looks like the issue described by Qodo finding #1, “Relation extraction always fails,” which is marked resolved. `_add_to_graph()` uses the NER and relation extractors correctly, so I think the standalone helper was just missed when the other path was fixed.

The existing `test_extract_relations_returns_dict` won’t catch this because it only checks the result types. The failure path still returns a dict with a list in `"relations"`, so the test passes even though extraction never succeeds.

The fix should just be to use `_get_ner_extractor().extract_entities(text)` here, matching the existing `extract_entities()` helper. Could you also make the test assert that extraction actually produced something and that there’s no `"error"` key? For example, checking `count > 0` would catch this regression.

Two small non-blocking things:

* I’d remove the `# FIX: renamed from _graph_lock to _default_graph_lock...` comments in `kg_tools.py`. They read more like implementation notes than comments we’d want to keep in the code.
* `.github/requirements/checkov.txt` changes `pywin32` from `312` to `307`, but I don’t see anything else in the PR that explains the downgrade. Could you check whether that was an accidental change from a local `uv pip compile` run?

The rest looked good to me.

Requesting changes for the `extract_relations()` bug.

ZohaibHassan16 (changes requested):
The two issues from the last round are fixed.

`extract_relations()` is working now:

```python
>>> extract_relations('Alice works at Acme Corp. Bob founded Acme Corp.')
{'relations': [{'source': 'Acme Corp.', 'relation': 'related_to', 'target': 'Acme Corp.', 'confidence': 0.5}], 'count': 1}
```

No more `AttributeError` or `error` key, and it returns a relation. `test_extract_relations_returns_dict` also checks `count > 0` and that `"error"` isn't present, which should catch this if it regresses.

The `# FIX:` comments are gone from `kg_tools.py`, and `.github/requirements/checkov.txt` is no longer part of the diff. The `pywin32` change is gone too.

One cleanup: `test_extract_relations_returns_dict` has `result=extract_relations(...)` without spaces around `=`, and `"Relation list should not ne empty"` has a typo.

There is a new blocker, though.

`ADK_AVAILABLE` ends up `False` when running from the repo root. The repo's top-level `mcp/` package shadows the third-party `mcp` package that `google-adk` imports internally.

I can reproduce it from this branch:

```python
>>> import mcp; mcp.__file__
'...\\mcp\\__init__.py'          # local package, not the PyPI package

>>> from google.adk.tools import FunctionTool
...
File ".../google/adk/agents/mcp_instruction_provider.py", line 25
    from mcp import types
ImportError: cannot import name 'types' from 'mcp' (.../mcp/__init__.py)
```

This happens during the normal `google.adk` import path: `agents` imports `mcp_instruction_provider`, which does `from mcp import types`. So anything that catches the failed ADK import and falls back to `ADK_AVAILABLE = False` is affected, including `kg_tools.py`, `decision_tools.py`, and `session_service.py`.

It also explains why `test_init.py::test_adk_available` fails instead of skipping: the test explicitly expects `ADK_AVAILABLE is True`.

I don't see anything in CI that would catch this. The only `pytest` invocation in `ci.yml` runs a specific explorer test, and `tests/integrations/google_adk/` doesn't appear to run in any workflow. As a result, this can merge with green CI even though the integration is broken from a normal repo-root checkout.

I'm not sure why you're seeing 36 passing tests locally. It may be a working-directory or `PYTHONPATH` difference, but the repo-root case needs to work.

I think this needs to be fixed at the import/package level rather than worked around in the tests. Either the third-party `mcp` package needs to resolve correctly when ADK imports it, or the local `mcp` package needs to be renamed/isolated so the two can't collide.

Could we also add the Google ADK integration tests to CI? That should catch this kind of breakage before merge.

Requesting changes for the `mcp` import collision. The relation extraction changes look good.

ZohaibHassan16 (approved):
Pushed 8 fixes on top of your branch. A couple of things marked as resolved on Qodo were still broken and I found a few more issues while going through it.

**Incomplete `mcp` rename**

The `mcp/` → `semantica_mcp/mcp/` move fixed the import collision with google-adk's `mcp` dependency, but a few references were left behind.

`cli.py` and four existing tests still imported from `mcp.*`, and files under `semantica_mcp/mcp` were also importing from bare `mcp` instead of using relative imports. That left `semantica mcp list-tools` and those tests failing with `ModuleNotFoundError`.

`semantica_mcp` also wasn't included in setuptools' package finder and didn't have an `__init__.py`, so it wouldn't have been included in an installed package.

I switched the internal imports to relative ones, updated `cli.py` and the tests, and added the package finder entry and `__init__.py`.

**`list_sessions()` return type**

`from google.adk.sessions import ListSessionsResponse` doesn't work with the installed ADK package. `ListSessionsResponse` lives in `base_session_service` and isn't re-exported from `sessions/__init__.py`.

Because of that, the import fell back to `Any` and `list_sessions()` returned a plain list rather than a `ListSessionsResponse`. Anything using the normal ADK API and accessing `resp.sessions` would fail with `AttributeError`.

`user_id` was also required here even though ADK defines it as `Optional[str] = None`.

Fixed both and added coverage for the actual return type and listing sessions across all users.

**Session node ID collisions**

`_node_id` was joining `app_name`, `user_id`, and `session_id` with plain colons. That makes values containing colons ambiguous. For example:

`app_name="tenant:A", user_id="alice"`

and

`app_name="tenant", user_id="A:alice"`

produce the same prefix with the old format.

The parts are now percent-encoded before joining. I also updated the legacy ID parsing fallback accordingly.

**`kg_tools` and `decision_tools` weren't sharing locks or the default graph**

Both modules had their own `_graph_locks` dict and their own default graph singleton.

That means using the same graph from both modules still gave you two different lock objects. Calling both factories without passing a graph also gave each module a different default graph, even though the README shows both tool sets sharing one `ContextGraph`.

Moved the shared graph and lock state into a new `_shared.py` used by both modules.

**Dead compatibility fallback in `add_to_graph`**

The `TypeError` fallback for extractors that supposedly don't accept `entities=` couldn't work. `RelationExtractor.extract_relations` requires `entities` as a positional argument with no default, so the fallback just raised another `TypeError` and could hide the original error.

Removed the fallback so the real exception is preserved.

**`append_event` persisted partial events**

ADK's base `append_event` ignores `event.partial`. This override called `super()`, which correctly did nothing for partial events, but then went on to persist the event node, edge, and state update anyway.

That meant every chunk of a streamed response could be stored as a permanent event.

Added the same early return used by the base implementation.

**`get_session` ignored `config`**

The method deleted `config` immediately, so `num_recent_events` and `after_timestamp` had no effect. Callers always got the full event history.

Implemented both options using the same behavior as `InMemorySessionService`.

**Session service blocked the event loop**

All five async methods were doing the graph scan synchronously while holding a `threading.RLock` on the event loop thread.

If that lock is contended, `Lock.acquire()` blocks the OS thread, so the whole event loop stops rather than just suspending the coroutine.

I moved the lock-protected work into synchronous helpers and run those with `asyncio.to_thread()`. `append_event` needs to call ADK's async `append_event` while holding the lock, so that call runs through `asyncio.run()` inside the worker thread; it doesn't do any real I/O.

I also added a regression test that makes `find_nodes` sleep and runs a heartbeat coroutine alongside a session call. The old code gets zero heartbeats while the call is running; the fixed version keeps the loop responsive.

All of these fixes have test coverage. The `google_adk` suite is now at 48 passing tests.

Issue #1272, opened 2026-08-30
FAISS: persist vector_ids and metadata across index save/load


Sameer6305:
## Description

`FAISSIndex.save()` currently persists only the underlying FAISS index, while the associated `vector_ids` and `metadata` are kept only in memory.

As a result, loading a previously saved FAISS index does not restore the logical IDs and metadata associated with the stored vectors.

### Reproduction

1. Create a FAISS store.
2. Add vectors with IDs and metadata.
3. Save the index.
4. Start a new process and load the saved index.
5. Inspect the store.

The loaded index can contain the physical vectors, but `vector_ids` and `metadata` are not restored.

This can cause:

- `count()` to report an incorrect value.
- `get_vector(id)` / `scan_vectors()` to behave incorrectly.
- Existing IDs to be invisible to duplicate detection.
- Re-running a migration after restarting the process to add duplicate physical vectors.

### Expected behavior

Saving and loading a FAISS store should preserve the complete logical state associated with the index:

- vector IDs
- metadata
- underlying FAISS vectors/index

### Suggested fix

Persist `vector_ids` and `metadata` alongside the FAISS index and restore them during `load()`.

Please also add save/load round-trip tests covering:
- vector IDs
- metadata
- count/get_vector/scan_vectors
- duplicate detection after reload
- migration retry after restarting the process

### Priority

Non-blocking follow-up.

libaojiang:
I'd like to take this on. I'll add a focused save/load round-trip test for vector IDs and metadata, then keep the implementation scoped to persisting that logical state alongside the FAISS index while preserving compatibility with existing index files that do not have metadata yet. Could you assign #1272 to me?

PR #1314, opened 2026-08-30
fix(vector_store): persist vector_ids and metadata across FAISS index save/load (#1272)

AhmadBilalDSA:
## Summary
Fixes silent data loss on FAISS index save/load by persisting `vector_ids`, `metadata`, `dimension`, and `index_type` to an atomic companion `.meta.json` file.

## Why This Matters (Addresses #1083 / Silent Migration Failure)
Previously, loading a saved FAISS index initialized `vector_ids = []` and `metadata = {}`. This caused `scan_vectors()` to return `[]`, `count()` to return `0`, and broke `semantica store migrate --from faiss` by copying zero records despite a valid binary index.

## Key Changes
- **Atomic Sidecar Persistence:** Writes metadata to `<index_path>.meta.json` atomically via temporary file replacement.
- **Missing Sidecar Warning:** Emits explicit `RuntimeWarning` and logger warnings when binary index exists without `.meta.json`.
- **Restoration:** Restores exact `vector_ids`, `metadata`, and config parameters on reload.
- **End-to-End Regression Test:** Added `test_faiss_store_save_load_scan_vectors_end_to_end` verifying `scan_vectors()` matches original records across fresh store instances.


ZohaibHassan16 (changes requested):
Conflict markers are resolve  and both things I asked for last time are included now : the missing sidecar warning and the end-to-end round-trip test. Those look good, thanks for fixing them quickly.

I checked Qodo's other two findings against the code as well, and I think both of them are valid.

## Finding 3 : valid metadata can break saving

I was able to reproduce this:

```
>>> json.dumps({"vector_ids": ["vec_0"], "metadata": {"vec_0": {"created_at": datetime.now()}}, ...})
TypeError: Object of type datetime is not JSON serializable
```

`FAISSStore.add_vectors()` accepts `metadata` as `Dict[str, Any]`, and there is no validation that says it has to be JSON serializable.

Before this PR, `save()` didn't serialize the metadata, so values like `datetime`, numpy scalars, sets, bytes, etc. didn't cause an issue. With this PR, `save()` now uses plain `json.dumps()`, so those values can make `save_index()` fail with a `TypeError`.

The current tests only use JSON-safe metadata, so this case isn't covered.

Could we either use:

```
json.dumps(..., default=str)
```

or catch the serialization error and raise a clearer `ProcessingError` that tells us which metadata key could not be serialized?

## Finding 2 : the sidecar can get out of sync

This one also looks valid.

`save()` writes the FAISS index first, then writes the sidecar after that. If the sidecar write fails, or the process stops between those two steps, the binary index has already been updated but the sidecar may still be old or missing.

Finding 3 can cause this too, since the FAISS file gets written before `json.dumps()` fails.

For a first save, the missing-sidecar warning handles this reasonably well. But when saving over an existing index, it can leave something like:

* FAISS index has N vectors
* old sidecar still has M vector IDs
* M < N

Then `load()` loads both without checking that the number of IDs matches the number of vectors in the index. That could lead to the wrong IDs being mapped silently.

I don't think we need the bigger generation-file approach suggested by Qodo. A small check in `load()` comparing `len(data["vector_ids"])` with the loaded FAISS index count should be enough. It can either warn or raise when they don't match.


Requesting changes for these two. Happy to review again once they're fixed.

ZohaibHassan16 (changes requested):
Nice, conflict markers are gone and both things from last round landed. Went through both fixes though and I don't think either one is actually done yet, sorry.

**default=str thing** : reproduced it, and yeah it stops the crash, but look what it does to the data:

    >>> json.dumps({"created_at": datetime(2026,1,1,12,0,0), "tags": {"a","b"}}, default=str)
    '{"created_at": "2026-01-01 12:00:00", "tags": "{\'b\', \'a\'}"}'

that `tags` value isn't coming back as a set, it's coming back as the literal string `"{'b', 'a'}"` forever. And a datetime becomes a plain str, so any code doing date comparisons on that metadata after a reload just breaks differently now instead of crashing at save time.

Your new test actually builds exactly this scenario (datetime, uuid, np.int64, a set) but only asserts `vector_ids`/`dimension`/`index_type` after loading, never `loaded_index.metadata[...]`. So it passes while the metadata silently changed type underneath it. Not saying default=str is wrong necessarily, just that we should either assert on what it actually does to the round trip, or do something less lossy for it.

**mismatch check** : same deal, I think a warning isn't enough here. Your own test proves it:

    assert loaded.vector_ids == ["vec_a", "vec_b"]
    assert loaded.index.ntotal == 3

That's 3 vectors physically in the index but only 2 come back through scan/count, and the test treats that as the expected passing outcome. checked cli.py, nothing there captures python warnings, so someone running `store migrate --from faiss` would never see this, it'd just quietly drop a vector and say "done". kind of the exact thing #1083 was about.

Also every other backend in this same batch of PRs (qdrant/weaviate/pinecone iter_all) raises ProcessingError for this exact class of "would return less than what's actually there" case, so warn-only here is inconsistent with the rest of the stack for no reason I can see. if there's a reason you want warn over raise here specifically I'm open to hearing it, just don't think "test shows truncated data returned as success" is where we want to land.

Requesting changes again, sorry to keep bouncing this, the core approach is solid though and this is close

ZohaibHassan16 (changes requested):
Both things from the last round are fixed. 

The count mismatch now raises `ProcessingError` correctly, and the lossless encoder does actually round-trip datetime/UUID/numpy/set with the original types intact. All 29 tests pass now. Nice work, and sorry for sending you around once more on the last one.

There is still one thing blocking this though, and unfortunately it hits the same bug we were trying to fix in the first place.

`bytes` still blows up `save()`.

It was in the original list along with datetime, numpy scalars, and sets, but there’s no handler for it yet:

```python
>>> index.metadata['x'] = {'blob': b'raw-embedding-hash'}
>>> index.save(path)
TypeError: Object of type bytes is not JSON serializable
```

The bad part isn’t just that `bytes` is unsupported. `save()` writes the FAISS file first and only then writes the metadata sidecar.

So in this case you end up with the index file on disk, but no `.meta.json`.

Then on the next `load()`, it goes through the no-sidecar path, warns, and gives you `vector_ids = []`.

So this is basically the #1083 failure mode again: the index looks valid, load succeeds, but you silently get zero vectors back. It’s just coming in through the one metadata type that wasn’t handled.

One other smaller thing I noticed: the special-key tagging can collide with normal user metadata.

```python
>>> json.dumps({'__set__': ['not','a','real','set']}, cls=_LosslessJSONEncoder)
>>> json.loads(_, object_hook=_lossless_object_hook)
{'set', 'a', 'not', 'real'}
```

So if someone has a normal dict with a key like `__set__`, `__datetime__`, `__date__`, or `__uuid__`, it gets reinterpreted on load instead of coming back as the same dict.

That’s definitely more of an edge case, and I’m not going to block hard on it, but it’s worth calling out since it’s still data changing shape across save/load.

Requesting changes for the `bytes` case.

Either support it, probably base64 under a `__bytes__` tag like the rest of the scheme, or reject unsupported metadata earlier with a clear error. What we should avoid is finding out halfway through `save()` after the FAISS file has already been written.

The tag collision would be good to fix too, but I’m fine with that being documented as a limitation for now if you don’t want to expand the scope.

ZohaibHassan16 (approved):
Both issues from the last round are fixed.

`bytes` round-trips correctly now. `{'blob': b'raw-embedding-hash'}` comes back as actual `bytes`, not a string, using the `__bytes__`/base64 tag.

The write-ordering fix also covers the broader failure case. I tested `save()` with a metadata type the encoder still can't serialize at all, a plain custom object, and it raises before writing anything to disk. No FAISS binary and no sidecar are left behind. That closes the orphaned-index case in general.

The tag-collision guard behaves as intended. A multi-key dict such as `{'__set__': [...], 'other': 1}` now round-trips unchanged. A dict that's exactly `{'__set__': [...]}` is still interpreted as a set, which is the narrower case you already called out as non-blocking.

Approving. Nice job sticking with this through all five rounds.

PR #1315, opened 2026-08-30
fix(parser): preserve indentation inside fenced code blocks during normalization

AhmadBilalDSA:
## Description
Fixes an issue where whitespace normalization collapsed spaces and stripped indentation inside fenced code blocks (` ``` `).

- Added `_split_code_blocks()` in `semantica/normalize/text_normalizer.py` to tokenize text into code and non-code segments, handling backtick/tilde fences, varying fence lengths, and language tags.
- Updated `normalize_whitespace()` so space-collapsing logic only targets non-code segments while code blocks remain preserved.
- Added 16 new unit and integration tests in `tests/normalize/test_text_normalizer.py` covering indented Python code, nested lists outside fences, mixed text, and unclosed fences.

## Type of Change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Checklist
- [x] I have followed the project's code style and contributing guidelines
- [x] I have added unit tests covering my changes
- [x] All existing and new tests pass locally (`pytest`)
- [x] I have updated documentation if necessary


ZohaibHassan16 (changes requested):
Thanks for working on this! However there are few bugs:

i) On line 538, `re.sub(r"\n{3,}", "\n\n", result)` runs across the joined text, so it collapses two blank lines inside code blocks down to one. That breaks PEP 8 formatting between functions and classes, so we should trim those excess newlines off the prose boundaries instead of regexing over the code blocks.

ii) Also in `TextNormalizer.normalize_text()` (line 190), there's still a return `normalized.strip()` at the end. `WhitespaceNormalizer` preserves leading fence indentation fine now, but the top-level method wipes it out if the file starts with an indented code block. You can probably just drop that `.strip()`.

A couple of other quick things: `handle_line_breaks` running at the start mangles Windows `\r\n` endings into `\r\r\n\n` after the regexes run (better to just keep it as \n internally and convert line breaks at the very end), and `text.replace("\t", " ")` turns tabs into a single space inside code blocks which breaks tab-indented code and Makefiles.

Once those newline and boundary strip bits are sorted out I'm good to approve this

ZohaibHassan16 (changes requested):
Thanks for the fixes, they look good.

One last thing on the line breaks: `line_break_type` in `normalize_whitespace()` is currently unused because of the original_ending check. If someone passes `line_break_type="windows"` on text that uses `\n`, it completely ignores the argument and keeps `\n` (and vice versa for CRLF with `line_break_type="unix"`).

Since everything is already normalized to `\n` internally during the regex passes, you can drop the `original_ending` logic and just do this at the end of `normalize_whitespace`():

```
return self.handle_line_breaks(result, line_break_type)
```

(You'll also just need to pass `line_break_type="windows"` in `test_crlf_line_endings_not_mangled` so it expects CRLF).

Once that is done we are good to merge.

ZohaibHassan16 (approved):
LGTM. Thanks for working on it.

Issue #1265, opened 2026-08-30
[FEATURE] add scan_vectors support for pinecone/qdrant/milvus/weaviate


ZohaibHassan16:
<html>
<body>
<html><head></head><body><p>Title: [FEATURE] add scan_vectors support for pinecone/qdrant/milvus/weaviate</p><h3>Problem</h3><p><code inline="">scan_vectors()</code> was added for faiss, sqlite and pgvector in #1264 , which also made <code inline="">store migrate</code> work between those backends.</p><p>The remaining backends still don't support it:</p><ul><li><p>pinecone</p></li><li><p>qdrant</p></li><li><p>milvus</p></li><li><p>weaviate</p></li></ul><p>Right now they raise <code inline="">NotImplementedError</code>, so <code inline="">store migrate</code> still can't be used if either the source or destination is one of these backends.</p><p>The main issue is that all four backends paginate differently.</p>
Backend | API
-- | --
Qdrant | client.scroll()
Pinecone | index.list() + fetch()
Milvus | collection.query_iterator()
Weaviate | collection.iterator() / cursor pagination

<p>Qdrant can return vectors and payload together.</p><p>Pinecone only returns IDs when listing, so those IDs have to be fetched after.</p><p>Milvus has a query iterator which is better than using large offsets with <code inline="">query()</code>.</p><p>Weaviate uses cursor-based pagination instead of normal numeric offsets.</p><p>These all need to return the same format that the existing backends use:</p><pre><code class="language-python">[
    {
        "id": "...",
        "vector": [...],
        "metadata": {...},
    }
]
</code></pre><h3>Proposed solution</h3><p>Implement <code inline="">scan_vectors()</code> separately for each backend and translate the backend response into the existing format.</p><h4>Qdrant</h4><p>Use <code inline="">client.scroll()</code> with vectors and payload enabled.</p><p>Something like:</p><pre><code class="language-python">client.scroll(
    collection_name=...,
    limit=limit,
    offset=offset,
    with_vectors=True,
    with_payload=True,
)
[...]

PR #1316, opened 2026-08-30
feat(vector_store): add iter_all enumeration for cursor-based backends

ZohaibHassan16:
Part 1 of #1265. This PR adds the facade dispatch and Qdrant support. Pinecone, Milvus and Weaviate will come in separate PRs.

I looked into the four remaining backend APIs before implementing this, and the approach from the issue doesn't really work as written. All four use cursor/keyset pagination and none of them can properly implement `scan_vectors(offset, limit)`:

* Qdrant's `scroll()` cursor is a point ID, not a row offset. `offset=1000` means "start at point ID 1000", not "skip 1000 rows"
* Pinecone uses an opaque continuation token, so there's no way to derive the token for logical page N
* Milvus has `query(offset=)`, but it's limited by the 16384 result window
* Weaviate uses the previous object's UUID as the cursor

There are basically two ways to fake positional offsets on top of those APIs, and neither is great.

One is keeping an `offset -> cursor` cache on the instance. That introduces mutable state, isn't thread safe, doesn't work for a fresh instance without walking there first, and the mapping can become invalid if writes happen during a scan.

The other is starting from zero on every call and discarding rows until the requested offset. With calls like `0, N, 2N, ...`, that means repeatedly scanning the same prefixes and makes the full scan quadratic in page count. For Pinecone it would also repeat list API calls.

The other libraries I checked don't expose this as a positional paging API either. LangChain's `VectorStore` doesn't have an enumerate-all primitive, LlamaIndex's core interface is based around IDs/filters, and Haystack keeps pagination internal. Iteration seems to be the portable abstraction here, not numeric offsets.

So instead of changing the meaning of `scan_vectors`, this adds a second optional primitive for backends that naturally support cursor iteration:

* Backends can implement `iter_all(batch_size)`, returning a generator that uses the backend's native paging API
* `VectorStore.iter_vectors()` uses `iter_all()` when available and otherwise falls back to the existing `scan_vectors()` offset loop
* Cursor-based backends keep raising `NotImplementedError` from `scan_vectors()`, since they don't actually support positional access
* `store migrate` only goes through `iter_vectors()`, so Qdrant migration works without changing the caller

Changes in this PR:

* `VectorStore.iter_vectors()` now dispatches to `iter_all()` when it's callable. This uses `callable()` rather than `hasattr()`, consistent with the existing `count()` behavior that `_MisShapedBackendStore` tests
* Added `QdrantStore.iter_all()`. It threads `scroll()`'s `next_page_offset` between requests instead of discarding it like `filter_by_metadata` currently does
* Added Qdrant to the backends accepted by `store migrate`
* Updated the migration docstring

There are two Qdrant details handled explicitly.

`scroll()` can return a final non-empty page while `next_page_offset` is already `None`. The implementation yields the records before checking whether the cursor is exhausted. That way the last page isn't dropped, and we don't accidentally call `scroll()` again with `offset=None`, which would restart the scan from the beginning.

The initialization guard also raises instead of doing the same log-and-empty-return behavior as `filter_by_metadata`. For migration, silently returning no records is dangerous because an uninitialized source looks exactly like an empty collection. `store migrate` could then report success after copying zero vectors, which is the failure mode from #1083.

This also fixes a dimension issue that would have broken Qdrant migration anyway.

`store migrate` currently reads `source._backend_store.dimension` when creating the destination. Qdrant, Milvus and Weaviate don't expose that attribute (Pinecone does), so migration falls back to the default dimension of 768. If the actual vectors have a different dimension, every insert fails.

Migration now reads the first source record, gets the dimension from `len(vector)`, creates the destination with that value, and then chains the first record back into the iterator so it still gets migrated.

Tests added:

* 6 facade dispatch tests, including fallback when `iter_all` exists but isn't callable, plus error propagation
* New `tests/vector_store/test_qdrant_store.py` with 10 tests. Qdrant didn't have a dedicated test file before. Covers cursor threading, the final-records-with-null-cursor case, the not-initialized guard, and propagation of backend errors
* 5 CLI tests around dimension inference, including a regression test to make sure the first record used for dimension detection is still migrated

One existing test was also changed. `test_migrate_refuses_unsupported_backend_pair` used `faiss -> qdrant` as the unsupported pair, which is no longer true, so it now uses `faiss -> weaviate`.

The mocks were written against the pinned client version shapes.

### DIFF
--- a/semantica/cli.py
+++ b/semantica/cli.py
@@ -5,6 +5,7 @@ This module provides the command-line interface for the Semantica framework,
 enabling users to interact with the framework via terminal commands.
 """
 
+import itertools
 import json
 import os
 import sys
@@ -3714,7 +3715,7 @@ def store_stats(cli_ctx: CLIContext, backend: str, fmt: str, local_json: bool) -
     _run_with_error_handling(_action)
 
 
-_MIGRATE_SUPPORTED_BACKENDS = {"faiss", "sqlite", "pgvector"}
+_MIGRATE_SUPPORTED_BACKENDS = {"faiss", "sqlite", "pgvector", "qdrant"}
 _MIGRATE_BATCH_SIZE = 500
 
 
@@ -3759,12 +3760,10 @@ def store_migrate(cli_ctx: CLIContext, from_backend: str, to_backend: str,
                   namespace: Optional[str], local_dry: bool, local_json: bool) -> None:
     """Migrate data between backends.
 
-    Direct migration is only wired up between faiss, sqlite, and pgvector -
-    these are the backends whose storage contract supports paging through
-    every stored vector. Migrating to or from qdrant, pinecone, milvus, or
-    weaviate still needs the export/reindex workaround below, since each of
-    those needs its own enumeration design (Qdrant scroll, Pinecone list,
-    etc.) that hasn't been built yet.
+    Wired up between faiss, sqlite, pgvector and qdrant. Migrating to or from
+    pinecone, milvus or weaviate still needs the export/reindex workaround
+    below, since each of those needs its own enumeration design (Pinecone
+    list, Milvus query_iterator, Weaviate cursor) that is not built yet.
 
     \b
     Example:
@@ -3804,7 +3803,18 @@ def store_migrate(cli_ctx: CLIContext, from_backend: str, to_backend: str,
         if source_index_path:
             source._backend_store.load_index(source_index_path)
 
+        # Pull the first record before building the destination: qdrant, milvus
+        # and weaviate expose no `dimension` attribute, so without this the
+        # destination falls back to the default 768 and rejects every insert.
+        # The data itself is the reliable source of truth.
+        source_iter = source.iter_vectors(batch_size=_MIGRATE_BATCH_SIZE)
+        first_item = next(source_iter, None)
+
         source_dimension = getattr(source._backend_store, "dimension", None)
+        if not source_dimension and first_item is not None:
+            first_vector = first_item.get("vector")
+            if first_vector is not None:
+                source_dimension = len(first_vector)
         if source_dimension and "dimension" not in dest_cfg:
             dest_cfg["dimension"] = source_dimension
 
@@ -3827,7 +3837,12 @@ def store_migrate(cli_ctx: CLIContext, from_backend: str, to_backend: str,
             metadata_batch.clear()
             ids_batch.clear()
 
-        for item in source.iter_vectors(batch_size=_MIGRATE_BATCH_SIZE):
+        items = (
+            source_iter
+            if first_item is None
+            else itertools.chain([first_item], source_iter)
+        )
+        for item in items:
             meta = dict(item.get("metadata") or {})
             if namespace and "namespace" not in meta:
                 meta["namespace"] = namespace
--- a/semantica/cli.py
+++ b/semantica/cli.py
@@ -5,7 +5,6 @@ This module provides the command-line interface for the Semantica framework,
 enabling users to interact with the framework via terminal commands.
 """
 
-import itertools
 import json
 import os
 import sys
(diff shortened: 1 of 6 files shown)


KaifAhmad1:
Reviewed this PR (`iter_all()` cursor enumeration for Qdrant + `iter_vectors()` dispatch). Small, well-tested diff — one correctness bug and one minor duplication.

**1. False-positive abort on a legitimate empty page (`qdrant_store.py`, `iter_all`, ~L655-664)**

The check treats "empty records + non-null `next_offset`" as truncation and raises `ProcessingError`:

```python
if not records:
    raise ProcessingError(
        "Qdrant returned an empty page alongside a continuation "
        "cursor, so the scan cannot advance. Refusing to return a "
        "truncated scan."
    )
```

This combination isn't necessarily truncation — it's Qdrant's documented behavior when a scroll batch lands entirely on tombstoned/soft-deleted points (common on any collection that's had deletes, which `delete_vectors()` here supports). Qdrant can legitimately return zero live records with a valid `next_page_offset` while there's more data ahead. As written, `iter_all()` would abort a perfectly healthy scan (e.g. mid `store migrate`) instead of just continuing with the returned cursor.

**2. Minor: duplicated record→dict conversion**

The `id`/`metadata`/`vector` extraction block in `iter_all()` (~L644-649) is copy-pasted from the existing `filter_by_metadata()` a few lines above, with no shared helper. Not a bug, but a future fix (e.g. handling named/multi-vector points) has to be applied in two places and is easy to miss in one.

Given #1317/#1326/#1328 are stacked on this branch, worth resolving here before it propagates further.

KaifAhmad1 (changes requested):
Reviewed this PR (`iter_all()` cursor enumeration for Qdrant + `iter_vectors()` dispatch). Small, well-tested diff — one correctness bug and one minor duplication.

**1. False-positive abort on a legitimate empty page (`qdrant_store.py`, `iter_all`, ~L655-664)**

The check treats "empty records + non-null `next_offset`" as truncation and raises `ProcessingError`:

```python
if not records:
    raise ProcessingError(
        "Qdrant returned an empty page alongside a continuation "
        "cursor, so the scan cannot advance. Refusing to return a "
        "truncated scan."
    )
```

This combination isn't necessarily truncation — it's Qdrant's documented behavior when a scroll batch lands entirely on tombstoned/soft-deleted points (common on any collection that's had deletes, which `delete_vectors()` here supports). Qdrant can legitimately return zero live records with a valid `next_page_offset` while there's more data ahead. As written, `iter_all()` would abort a perfectly healthy scan (e.g. mid `store migrate`) instead of just continuing with the returned cursor.

**2. Minor: duplicated record→dict conversion**

The `id`/`metadata`/`vector` extraction block in `iter_all()` (~L644-649) is copy-pasted from the existing `filter_by_metadata()` a few lines above, with no shared helper. Not a bug, but a future fix (e.g. handling named/multi-vector points) has to be applied in two places and is easy to miss in one.

Given #1317/#1326/#1328 are stacked on this branch, worth resolving here before it propagates further.

KaifAhmad1 (approved):
## Review

Verified both previously-flagged issues (duplicated record→dict conversion, false-positive abort on a legitimate empty page) against the current tip (fa967983):

- `_record_to_result()` is now a shared static method used by both `filter_by_metadata()` and `iter_all()` — no more copy-paste.
- The empty-page handling now uses a sentinel-based stuck-cursor check (`last_offset = object()` initially, compared against `next_offset` each iteration) instead of unconditionally raising on any empty page. Traced this by hand against all the new test cases (threading across pages, final page with null cursor, empty page with an advancing cursor, and a genuinely stuck cursor) — behaves correctly in each.

Ran the full test suite locally (not just CI): `test_qdrant_store.py` + `test_vector_manager_persistent.py`, 55/55 passing.

Scope is also correctly narrowed — `_MIGRATE_SUPPORTED_BACKENDS` still excludes qdrant (confirmed directly against the branch), so the facade init/write-dispatch gaps that blocked `store migrate` support are correctly deferred rather than silently broken. `iter_all()` itself is a self-contained, well-tested addition.

`verify-action-pins.sh`-equivalent isn't relevant here, but all 13 CI checks are green.

Approving.

PR #1317, opened 2026-08-30
feat(vector_store): add weaviate iter_all

ZohaibHassan16:
Part 2 of #1265. This is stacked on PR #1316, so that one should be reviewed first.

This PR adds `WeaviateStore.iter_all()` using Weaviate's UUID cursor and `VectorStore.iter_vectors()` dispatch for it.

**Edit:** wiring Weaviate into `store migrate` (as this description originally said) has been dropped for the same reason it was dropped from #1316 for Qdrant — the facade's write dispatch (`store_vectors()` only knows `add`/`add_vectors`, not Weaviate's `add_objects`) and initialization (the facade never calls `connect()`/collection-selection) aren't ready for a backend that doesn't satisfy those assumptions. That's tracked in #1335 and left for its own PR. `_MIGRATE_SUPPORTED_BACKENDS` and `cli.py` are untouched here; `test_migrate_refuses_unsupported_backend_pair` still uses `faiss -> qdrant` as originally written.

Pinecone and Milvus are still unsupported for now and continue using the export/reindex workaround.

Weaviate pagination is based on:

`fetch_objects(after=<last object's UUID>)`

rather than a numeric row offset. That means there's no way to turn something like row 100000 into the correct UUID without scanning up to that point first. For that reason this implements `iter_all()` rather than trying to make `scan_vectors(offset, limit)` work, following the approach from PR #1316.

Most of the paging logic was already present in `filter_by_metadata`, including the `after` cursor loop and its `TypeError` fallbacks for older client versions, so a lot of this is reusing/extracting existing behavior.

There are a couple of intentional differences.

First, `iter_all()` uses cursor stall detection instead of keeping a `seen_ids` set.

`filter_by_metadata` is bounded by `while len(results) < limit`, and uses the seen-ID set to avoid returning duplicates. That works there because the number of results is capped.

A full collection scan has no such cap. If the backend keeps returning the same cursor, the iterator could run forever. Keeping every ID seen so far would avoid some duplication, but it also means memory use grows with the size of the collection.

Instead, `iter_all()` compares the next cursor to the current one. If it didn't advance, iteration stops. That keeps memory usage O(1) and detects the actual failure condition directly.

Second, the compatibility fallback is simpler than the one in `filter_by_metadata`.

That method has to deal with both `filters` and `after`, so it has three levels of `TypeError` fallback.

`iter_all()` only uses `after`, so the fallback is:

`after` -> offset -> plain request

It also remembers when `after` isn't supported so it doesn't retry the failing form on every page.

The initialization/unavailable guards raise rather than returning an empty iterator, same as Qdrant in PR 1.

For migration, returning no records on an error is unsafe because it's indistinguishable from a genuinely empty source. `store migrate` could otherwise finish successfully after copying zero vectors, which is the failure mode described in #1083.

This implementation assumes one unnamed vector per object, which is consistent with the rest of the Weaviate backend.

`get_vector()` and `filter_by_metadata()` already assume `obj.vector` is a single vector. Collections using named vectors can return a mapping instead, and that case is not handled here. That's an existing limitation in the module rather than something introduced by this PR, and should probably be handled separately.

**Edit:** also fixed, following review: `_extract_vector()` now unwraps weaviate-client v4's `{'default': [...]}` vector shape before converting to `np.array` — the original version would have silently produced a corrupted 0-d object array for every real (non-mocked) Weaviate collection, since the tests only exercised plain-list mocks.

An empty page in cursor mode is also no longer treated as proof the scan is done. `after` has no server-issued continuation value of its own — it's derived client-side from the last object's uuid — so an empty page gives nothing to advance it with. If Weaviate's cursor walks internal storage position rather than strict uuid order, a batch can in principle land entirely on a gap (e.g. tombstoned objects) with live data past it, the same risk already confirmed for Qdrant's scroll cursor in #1316. `iter_all()` now falls back to offset pagination once to confirm before ending the scan, since offset addresses live rows by position and has no equivalent gap.

Tests added:

* `tests/vector_store/test_weaviate_store.py`, new file. Covers cursor threading, short-page termination, the empty-page-in-cursor-mode fallback (and that it doesn't falsely end a scan with more data past the gap), not-initialized and unavailable guards, empty/dict-shaped vectors, and backend errors propagating instead of silently returning no results
* Regression test for a stalled cursor to make sure iteration terminates instead of looping forever
* Test for the offset fallback when the client rejects `after`

Local run: `test_weaviate_store.py` + `test_qdrant_store.py` + `test_vector_manager_persistent.py`, all passing.


KaifAhmad1:
Reviewed (`WeaviateStore.iter_all()` + the shared Qdrant/dispatch changes carried over from #1316's branch).

**1. Duplicate yield before the stall guard fires (`weaviate_store.py`, ~L678-716)**

The duplicate-cursor check (`next_cursor == after_cursor`) runs *after* the batch has already been yielded, not before:

```python
# records yielded here (L678-689) ...
# ... then checked for progress here (L716)
if next_cursor == after_cursor:
    raise ProcessingError("...stopped advancing...")
```

So if `fetch_objects()` ever returns the same page twice (stale/ignored `after` cursor from a proxy, client quirk), `iter_all()` yields that page's records once, yields the **same records again** on the next iteration, and only then raises. A streaming consumer (e.g. `store migrate`) that flushes incrementally will have already written the duplicate before the exception surfaces — the exact failure this guard was written to prevent.

**2. `obj.vector` is a dict for weaviate-client v4, not a bare list (~L684)**

`iter_all()`'s docstring assumes "a single unnamed vector," but `weaviate-client==4.16.2` (pinned in requirements-ci.txt) returns `.vector` as `{'default': [...]}` even for the default vector. `len(raw_vector) > 0` passes on the dict, so `np.array(raw_vector)` produces a 0-d object array wrapping the dict, not a numeric embedding — verified locally:
```
>>> np.array({'default': [0.1, 0.2, 0.3]})
array({'default': [0.1, 0.2, 0.3]}, dtype=object), shape ()
```
Every item from a real (non-mocked) Weaviate collection would carry a corrupted `vector` field. The tests only exercise this with plain-list mocks, so the real client shape isn't covered.

**3. Minor: fallback cascade duplicated**

The cursor→offset→single_page degrade logic in `iter_all()` (~L654) re-implements the same nested try/except already in `filter_by_metadata()` (~L526-553). No shared helper, so the two can drift.

**4. Qdrant still has no stall guard**

`QdrantStore.iter_all()` (~L640) has no protection against a non-advancing cursor, even though this same PR adds that exact guard for Weaviate. Worth closing the gap on the Qdrant side too (also flagged on #1316).

KaifAhmad1 (changes requested):
Reviewed (`WeaviateStore.iter_all()` + the shared Qdrant/dispatch changes carried over from #1316's branch).

**1. Duplicate yield before the stall guard fires (`weaviate_store.py`, ~L678-716)**

The duplicate-cursor check (`next_cursor == after_cursor`) runs *after* the batch has already been yielded, not before:

```python
# records yielded here (L678-689) ...
# ... then checked for progress here (L716)
if next_cursor == after_cursor:
    raise ProcessingError("...stopped advancing...")
```

So if `fetch_objects()` ever returns the same page twice (stale/ignored `after` cursor from a proxy, client quirk), `iter_all()` yields that page's records once, yields the **same records again** on the next iteration, and only then raises. A streaming consumer (e.g. `store migrate`) that flushes incrementally will have already written the duplicate before the exception surfaces — the exact failure this guard was written to prevent.

**2. `obj.vector` is a dict for weaviate-client v4, not a bare list (~L684)**

`iter_all()`'s docstring assumes "a single unnamed vector," but `weaviate-client==4.16.2` (pinned in requirements-ci.txt) returns `.vector` as `{'default': [...]}` even for the default vector. `len(raw_vector) > 0` passes on the dict, so `np.array(raw_vector)` produces a 0-d object array wrapping the dict, not a numeric embedding — verified locally:
```
>>> np.array({'default': [0.1, 0.2, 0.3]})
array({'default': [0.1, 0.2, 0.3]}, dtype=object), shape ()
```
Every item from a real (non-mocked) Weaviate collection would carry a corrupted `vector` field. The tests only exercise this with plain-list mocks, so the real client shape isn't covered.

**3. Minor: fallback cascade duplicated**

The cursor→offset→single_page degrade logic in `iter_all()` (~L654) re-implements the same nested try/except already in `filter_by_metadata()` (~L526-553). No shared helper, so the two can drift.

**4. Qdrant still has no stall guard**

`QdrantStore.iter_all()` (~L640) has no protection against a non-advancing cursor, even though this same PR adds that exact guard for Weaviate. Worth closing the gap on the Qdrant side too (also flagged on #1316).

KaifAhmad1 (changes requested):
## Review

Checked this against the existing review thread's 4 findings first, then looked for anything new.

**Findings 1-3 from the prior review are confirmed fixed** (verified against commit 78682076):
1. The duplicate-cursor stall check now runs *before* the batch is yielded, not after.
2. `_extract_vector()` correctly unwraps weaviate-client v4's `{'default': [...]}` shape before `np.array()`.
3. The cursor→offset→single_page fallback ladder is now a shared `_fetch_objects_offset_or_plain()` helper, used by both `filter_by_metadata()` and `iter_all()`.

Ran the full test suite locally: `test_weaviate_store.py` + `test_qdrant_store.py` + `test_vector_manager_persistent.py`, 68/68 passing.

**Finding 4 ("Qdrant still has no stall guard") doesn't need a fresh fix here** — `feat/vector-store-iter-all` (#1316) already fixed this on its own tip (fa967983, the "dedupe qdrant record conversion, don't abort iter_all on a live cursor with an empty page" commit). This branch is currently `BEHIND` its base and still carries the pre-fix `qdrant_store.py` (`if next_offset is None or not records: return` — no stuck-cursor protection at all). Rebasing onto #1316's latest tip should resolve this automatically; flagging so it isn't mistaken for a new bug to fix by hand.

**New finding: the same failure class that was just fixed for Qdrant is unaddressed for Weaviate itself.**

`WeaviateStore.iter_all()`'s `if not batch_objects: return` (~L622) unconditionally ends the scan on *any* empty page, in every mode, with nothing equivalent to the "keep going unless the cursor is actually stuck" logic #1316 just added for Qdrant. Reproduced concretely against this branch:

```python
store.collection.query.fetch_objects.side_effect = [
    _page([_obj("uuid-1"), _obj("uuid-2")]),  # full page, cursor -> uuid-2
    _page([]),                                 # empty page (e.g. lands on tombstoned objects)
    _page([_obj("uuid-3")]),                   # real data past the gap
]
result = [item["id"] for item in store.iter_all(batch_size=2)]
# result == ['uuid-1', 'uuid-2'] -- uuid-3 is silently dropped, no error,
# fetch_objects is never called a third time.
```

No test in `test_weaviate_store.py` covers "empty page followed by more data" — only a single-empty-page collection (`test_iter_all_empty_collection_yields_nothing`, one page, no follow-up) and the short-page/exhaustion case are covered.

I don't have a confirmed source that Weaviate's `after`-cursor pagination can actually land on an all-deleted window the way Qdrant's segment-based scroll can (the PR's own author called the analogous Qdrant case "plausible but not confirmed" too) — so this may or may not be reachable in practice. But `store migrate`'s failure mode here (#1083: silent partial copy reported as success) is exactly what this PR's docstrings say `iter_all()` exists to prevent, and there's currently no guard and no test either way. Worth either applying the same "only stop if the cursor itself stops advancing" treatment here, or explicitly confirming Weaviate can't produce this shape before merging.

**Minor:** the PR description says this "enables Weaviate as a source/destination for `store migrate`" and describes CLI wiring/tests for it, but `cli.py` isn't touched anywhere in the current diff (confirmed via `compare/main...feat/weaviate-iter-all`) — looks like leftover text from before the same CLI-wiring narrowing that was applied to #1316. Worth updating the description so it matches what's actually in scope now.

Requesting changes for the empty-page gap; happy to re-review once it's addressed and the branch is rebased.

KaifAhmad1 (approved):
## Final review

Applied the fixes myself and re-reviewed:

1. **Synced `qdrant_store.py` / `vector_store.py` / their tests to #1316's current tip** (fa967983) — this branch had forked before #1316's stall-guard fix and was carrying the pre-fix, silently-truncating version. Did a targeted 4-file content sync rather than a full branch merge, since merging pulled in a large amount of unrelated `main` drift that had landed on #1316's branch since this one diverged and had nothing to do with this fix.

2. **Fixed the empty-page gap in `WeaviateStore.iter_all()`** — the same failure class just fixed for Qdrant, applied to Weaviate. Since `after` has no server-issued continuation value (it's derived client-side from the last object's uuid), an empty page can't be advanced past the way Qdrant's `next_page_offset` can; the fix falls back to offset pagination once to confirm before treating an empty page as the end. Reproduced the original bug concretely (full page → empty page → page with real data → real data silently dropped, no error) before fixing it, and added a regression test (`test_iter_all_continues_past_empty_page_in_cursor_mode`) plus corrected two existing tests that were mocking only a single empty page (a genuinely empty collection now takes two calls to confirm, matching the added defensiveness).

3. **Updated the PR description** — it described CLI `store migrate` wiring for Weaviate that isn't actually in the diff (stale text from before the same narrowing already applied to #1316's Qdrant scope). Now accurately describes what's in scope.

Ran the full suite locally after all changes: `test_weaviate_store.py` + `test_qdrant_store.py` + `test_vector_manager_persistent.py`, 70/70 passing. All 13 CI checks green on the pushed commits.

Approving.

PR #1329, opened 2026-08-31
fix(ci): unblock py3.9 install matrix and raise Scorecard pinning/signing

KaifAhmad1:
## Summary
- **Install Matrix failure**: `pip install semantica` failed on Python 3.9 (all three OSes) because `spacy` had no upper bound, so pip resolved spacy 3.8.16 whose `thinc>=8.3.12` requirement has no cp39 wheels and no working sdist build path either. Caps `spacy<3.8.8` and adds `thinc<8.3.5` for `python_version < '3.10'`; py3.10+ stays unconstrained. Verified with `pip install --dry-run --python-version 3.9` against manylinux, win_amd64, and macosx_arm64 — all resolve to prebuilt wheels (spacy 3.8.7 + thinc 8.3.4), no source builds needed.
- **OpenSSF Scorecard — Pinned-Dependencies**: `Dockerfile`'s `FROM node:26-alpine` and `FROM python:3.13-slim` were unpinned by digest (this was the actual cause of the low score — all GitHub Actions were already SHA-pinned). Pinned both to their current digests, and pinned five previously-unversioned `pip install` tool calls in CI (`build`, `safety`, `bandit`, `semgrep`, `jq`, `pip-audit`).
- **OpenSSF Scorecard — Signed-Releases**: `attest-build-provenance` only publishes to the GH attestations API, which Scorecard's Signed-Releases check doesn't inspect (it looks for signature files attached as release assets). Added a Sigstore signing step so `dist/*.sigstore.json` bundles ship alongside the wheel/sdist on the GitHub Release.

Not addressed here (require repo-owner action, not code): Branch-Protection and Code-Review settings, CII-Best-Practices badge application, and Fuzzing (OSS-Fuzz integration). Dangerous-Workflow scored 0 on the last report but I couldn't reproduce a real script-injection/unsafe-checkout pattern in any workflow — worth re-checking the score after this lands.

## Test plan
- [ ] Install Matrix workflow passes on py3.9/3.10/3.11/3.12 across ubuntu/macos/windows
- [ ] `docker build .` succeeds with the digest-pinned base images
- [ ] Release workflow's Sigstore signing step succeeds and `.sigstore.json` files land on the next tagged release
- [ ] Re-run OpenSSF Scorecard and confirm Pinned-Dependencies / Signed-Releases scores rise

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,7 +72,7 @@ jobs:
           diff \
             <(grep -E '^[a-zA-Z0-9._-]+==' requirements-ci.txt | sed 's/ \\$//') \
             <(grep -E '^[a-zA-Z0-9._-]+==' /tmp/requirements-ci-check.txt)
-      - run: pip install build
+      - run: pip install build==1.6.0
       # wheel is build-time only (not in requirements-ci.txt) — install the
       # same pinned version [build-system] declares so --no-isolation works.
       - run: pip install wheel==0.48.0
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,7 +16,7 @@ jobs:
       cancel-in-progress: false
     permissions:
       contents: write      # for the GitHub Release
-      id-token: write      # for PyPI Trusted Publishing (OIDC) and attestation signing
+      id-token: write      # for PyPI Trusted Publishing (OIDC), attestation signing, and Sigstore
       attestations: write  # for SLSA build provenance
     # If you add another job to this workflow, give it its own explicit
     # `permissions:` block rather than relying on the workflow-level default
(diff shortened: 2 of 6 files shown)


Sameer6305 (approved):
## Final Review

Looks good overall. ✅

- Python 3.9 dependency constraints are correctly scoped.
- Docker base images and CI tools are properly pinned.
- Sigstore signing is correctly integrated.
- The changes are focused and within the PR scope.
- CI/workflow and configuration syntax was validated locally.

### Fix made during review

Qodo's Sigstore finding was valid. The artifact `inputs` were incorrectly space-separated, so I fixed them to the required multiline format for both `.whl` and `.tar.gz` artifacts.

The fix has been pushed to the PR branch.

**Final verdict: Approved.** ✅
@KaifAhmad1 ready for merge

Issue #1019, opened 2026-08-15
[FEATURE] Parametrized backend conformance suite — five store backends have zero tests, and no shared contract defines what a backend must do


pravit-amp:
## Problem Statement

This repo has multi-backend abstractions at three layers — vector stores, graph stores, triplet
stores. A change that touches only the in-memory default, or that reaches into backend-specific
internals like `self.vectors` or `self.index` instead of the public interface, passes locally and
breaks on every other backend. Catching that is currently reviewer memory, and two things make it
hard even when the reviewer remembers.

**Five backends have no dedicated test file at all:**

| Layer | Backends | Covered |
| --- | --- | --- |
| `vector_store` | faiss, milvus, pgvector, pinecone, qdrant, sqlite_vec, weaviate | faiss, pgvector, pinecone, sqlite_vec |
| `graph_store` | age, amazon_neptune, falkordb, neo4j, in-memory | age, neptune |

So `milvus`, `qdrant`, `weaviate`, `neo4j` and `falkordb` have nothing. (Triplet stores are in
good shape — anzo, blazegraph, jena, oxigraph and rdf4j all have test files.)

**There is no contract to test against.** No ABC and no `Protocol` defines what a vector store or
graph store must implement — `VectorStore` (`vector_store.py:100`) is a plain concrete class. The
repo uses `ABC` elsewhere (`version_storage.py`, `provenance/storage.py`, `managers.py`,
`progress_tracker.py`), so the convention exists; the store layer never got it.

The drift this permits is already visible in the delete surface alone:

```
pgvector_store.py:474     def delete(self, ids)
sqlite_vec_store.py:428   def delete(self, ids)
pinecone_store.py:217     def delete_vectors(...)
qdrant_store.py:607       def delete_vectors(...)
faiss_store.py            — none —
milvus_store.py           — none —
weaviate_store.py         — none —
```

Three names for one operation, and three backends where it doesn't exist — while
`VectorStore.delete_vectors()` (`vector_store.py:786`) is documented as though every backend
supports it. Nothing catches that today because nothing asserts it.

## Proposed Solution

One parametrized conformance suite that runs the same contract against every registered backend,
plus a `Protocol` that states what the contract is.

**But the design constraint matters more than the idea**, and it comes straight from review
history on this repo. Qodo has left eight findings on my test PRs, and every one is the same
failure mode: *a test that passes while the thing it claims to protect is broken.*

| PR | Finding | Shape |
| --- | --- | --- |
[...]

PR #1333, opened 2026-08-31
ci: add integration workflow with a live pgvector service

ZohaibHassan16:
Infrastructure-only change. This adds one workflow file and does not touch production code or test files.

### Why

We have added roughly 60 tests across four vector backends recently, but none of them have run against a real backend service. They are all mock-based, which means they validate our interpretation of vendor APIs rather than actual runtime behavior.

That has already bitten us. #1316 went green with 16 tests while still being broken end to end, and the issue was caught during review rather than by the suite. The blockers found since then have mostly been the same class of problem: things mocks simply cannot expose.

The good news is that pgvector already has a live integration suite in `tests/vector_store/test_pgvector_store.py`. Its docstring includes Docker setup instructions, and `docs/vector_stores/pgvector.md` already documents the `TEST_PGVECTOR_URL` environment variable.

The missing piece is CI. No existing workflow provisions PostgreSQL with pgvector, so those tests have never actually run there.

This PR does not add any new tests. It only makes the existing pgvector suite execute against a real service. I prefer that here because it verifies the infrastructure using tests that already existed, rather than tests written specifically to make the new workflow look healthy.

One thing worth noting before review: simply adding a `services:` block to `ci.yml` would not help.

`ci.yml` runs one specific pytest target:

`tests/explorer/test_explorer_deterministic_rendering_e2e.py`

It does not run the vector store suite. At that point in the job it also installs only `.[explorer]`, so `psycopg` and `pgvector` are not available anyway. Even with a PostgreSQL container running, the pgvector tests would still never execute.

#### Separate workflow

This lives in its own workflow rather than extending `ci.yml`.

`ci.yml` is a required check, and I do not think a container image pull, PostgreSQL startup, or occasional service flake should block unrelated changes. This is the same general reason `install-matrix.yml` is separate.

The workflow runs on:

* `pull_request`
* a weekly cron
* `workflow_dispatch`

It also uses the same `paths-ignore` rules as `ci.yml`, so documentation-only changes do not start a database container unnecessarily.

#### Pin the service image by digest

The pgvector image is pinned by digest rather than by tag:

`pgvector/pgvector@sha256:ccc6e83d...`

That digest resolves to the pg16 image published on 2026-08-13.

This matches the repository's general approach of pinning external CI dependencies immutably.

One caveat is that `verify-action-pins.sh` only checks `uses:` entries. It does not inspect service images, so this particular pin is not enforced automatically. I added a comment in the workflow to make that explicit rather than leaving something that can silently drift later.

#### Create the extension explicitly

`PgVectorStore._verify_pgvector_extension()` checks that the `vector` extension exists and intentionally refuses to create it itself.

Because of that, the workflow creates the extension in a separate step before running pytest.

That step also serves as a useful connectivity check. If PostgreSQL is unavailable, the job fails there clearly instead of reaching pytest and potentially producing a green run where every integration test was skipped.

For the same reason, pytest runs with `-rs`, so any skips and their reasons are visible in the workflow log.

#### Dependency installation

The workflow installs:

`pip install -e ".[vectorstore-pgvector]" pytest==9.1.1`

That matches the general install shape already used in `ci.yml` and uses the same pinned pytest version.

I considered `requirements-ci.txt`, but it is generated with `--generate-hashes`, which enables hash checking and does not work cleanly with the editable install needed here.

#### Environment variable handling

One self-review fix is included.

The extension setup step originally interpolated `${{ env.TEST_PGVECTOR_URL }}` directly into the embedded Python script.

The value is controlled by the workflow, so there was no practical injection issue here, but it is the same pattern that was flagged in the `setup-semantica` action in #1266. The script now reads the value through `os.environ` instead.

### Verification

Locally:

* the workflow YAML parses
* `verify-action-pins.sh` passes with both new pins included
* 50 pinned references are checked, all OK
* the embedded Python parses correctly with `TEST_PGVECTOR_URL` supplied
* `tests/vector_store/` is unchanged from `main`: 10 failed / 21 errors / 211 passed

That last result is expected because this PR does not change any Python code or tests.

The actual verification for this PR is its own workflow run. The important thing to check is that the pgvector tests are reported as executed rather than skipped.

And to set expectations clearly: some of those tests may fail.

They have never run against a real PostgreSQL/pgvector service in CI before. The acceptance criterion for this PR is that we make the suite run reliably, not that we hide or fix every issue it exposes in the same change.

If a test failure turns out to reveal a real backend bug or a design question, I would rather track that separately than expand this infrastructure PR.

### Follow-ups

Next I would add live coverage for Qdrant, including adapter-level round-trip tests. That should stack on #1316 because `iter_all` is not on `main` yet.

After that:

* Weaviate should be relatively straightforward: one container plus the anonymous-access environment configuration.
* Milvus needs Milvus, etcd, and MinIO, so that deserves a separate three-container workflow.
* Pinecone cannot be containerized, so live verification would require opt-in credentials. I would leave that out for now.

Related: #1265, #1019.


KaifAhmad1:
Reviewed (new `integration.yml` live-pgvector workflow + `test_pgvector_store.py` fixes). Findings, most severe first:

**1. Dead/leftover cleanup block in `test_search_empty_store` targets the wrong table via an already-closed connection (`tests/vector_store/test_pgvector_store.py` L327-344)**

The diff adds a correct `try/finally` (L307-325) that drops `empty_table` and calls `empty_store.close()`. But the *original* cleanup block from before the fix was never deleted — it's still there below the `finally`, and it:
- calls `empty_store._get_connection()` on a pool already closed one block above (`psycopg_pool.getconn()` raises `PoolClosed` immediately, re-wrapped as `ProcessingError` by `_get_connection()`), and
- targets `unique_table_name` — which is actually the *other* fixture's table (`store`/`setup_vectors`, still populated with 10 vectors in this same test) — instead of `empty_table`.

It no-ops on every run because it's wrapped in `except Exception: pass`, so it never fails loudly, but it's guaranteed-dead code that would drop the wrong table if it ever did execute. Also re-imports `psycopg_sql` a second time (L333), confirming it's a copy-paste remnant rather than intentional. Should just be deleted.

**2. Drops the repo's hash-pinned install convention (`integration.yml` ~L58)**

`pip install -e ".[vectorstore-pgvector]" pytest==9.1.1` — plain install, no `--require-hashes`. `ci.yml` explicitly documents switching to hash-verified installs to satisfy the OpenSSF Scorecard "Pinned-Dependencies" check, and `benchmark.yml`, `security.yml`, `security-scan.yml`, `release.yml` all follow that convention. This workflow reintroduces the pattern the rest of CI moved away from.

**3. No `branches:` restriction on the `pull_request` trigger (`integration.yml` L9-14)**

`ci.yml`, `codeql.yml`, `defender-for-devops.yml`, and `security.yml` all scope `pull_request:` to `branches: [main]`; this is the only workflow using just `paths-ignore:` with no `branches:` key. A PR between two feature/release branches (common during larger stacked work — like the iter_all stack this sits alongside) spins up the Postgres container and runs the full live-DB suite for a merge that never touches `main`.

**4. Vacuous assertion (`test_pgvector_store.py` L126)**

`assert all(uuid.UUID(vector_id) for vector_id in ids)` — `uuid.UUID` instances are always truthy (no `__bool__`/`__len__`), so `all(...)` can only be `True` or raise; it never evaluates `False`. If `add()` ever returned a malformed id, this fails with an unhandled `ValueError` from `uuid.UUID()` mid-generator rather than a clean `AssertionError` — the `assert`/`all()` wrapper isn't actually doing the checking, the exception path is.

**5. Minor: the "fail if suite skipped" step treats a symptom, not the cause (~L86-105)**

The 15-line junit.xml-parsing step exists because `pg_available` (the fixture) always resolves a connection failure to a silent `pytest.skip`, even when `TEST_PGVECTOR_URL` was explicitly set by CI. Raising instead of skipping when that env var is present would fix this at the fixture level for any future CI job, instead of every caller needing its own XML-scraping bandaid. Relatedly, `if: always()` on that step means it'll hit `FileNotFoundError` if an earlier step fails before pytest ever writes `junit.xml` — not a false-pass, but loses the clean diagnostic in that failure mode.

KaifAhmad1 (changes requested):
Reviewed (new `integration.yml` live-pgvector workflow + `test_pgvector_store.py` fixes). Findings, most severe first:

**1. Dead/leftover cleanup block in `test_search_empty_store` targets the wrong table via an already-closed connection (`tests/vector_store/test_pgvector_store.py` L327-344)**

The diff adds a correct `try/finally` (L307-325) that drops `empty_table` and calls `empty_store.close()`. But the *original* cleanup block from before the fix was never deleted — it's still there below the `finally`, and it:
- calls `empty_store._get_connection()` on a pool already closed one block above (`psycopg_pool.getconn()` raises `PoolClosed` immediately, re-wrapped as `ProcessingError` by `_get_connection()`), and
- targets `unique_table_name` — which is actually the *other* fixture's table (`store`/`setup_vectors`, still populated with 10 vectors in this same test) — instead of `empty_table`.

It no-ops on every run because it's wrapped in `except Exception: pass`, so it never fails loudly, but it's guaranteed-dead code that would drop the wrong table if it ever did execute. Also re-imports `psycopg_sql` a second time (L333), confirming it's a copy-paste remnant rather than intentional. Should just be deleted.

**2. Drops the repo's hash-pinned install convention (`integration.yml` ~L58)**

`pip install -e ".[vectorstore-pgvector]" pytest==9.1.1` — plain install, no `--require-hashes`. `ci.yml` explicitly documents switching to hash-verified installs to satisfy the OpenSSF Scorecard "Pinned-Dependencies" check, and `benchmark.yml`, `security.yml`, `security-scan.yml`, `release.yml` all follow that convention. This workflow reintroduces the pattern the rest of CI moved away from.

**3. No `branches:` restriction on the `pull_request` trigger (`integration.yml` L9-14)**

`ci.yml`, `codeql.yml`, `defender-for-devops.yml`, and `security.yml` all scope `pull_request:` to `branches: [main]`; this is the only workflow using just `paths-ignore:` with no `branches:` key. A PR between two feature/release branches (common during larger stacked work — like the iter_all stack this sits alongside) spins up the Postgres container and runs the full live-DB suite for a merge that never touches `main`.

**4. Vacuous assertion (`test_pgvector_store.py` L126)**

`assert all(uuid.UUID(vector_id) for vector_id in ids)` — `uuid.UUID` instances are always truthy (no `__bool__`/`__len__`), so `all(...)` can only be `True` or raise; it never evaluates `False`. If `add()` ever returned a malformed id, this fails with an unhandled `ValueError` from `uuid.UUID()` mid-generator rather than a clean `AssertionError` — the `assert`/`all()` wrapper isn't actually doing the checking, the exception path is.

**5. Minor: the "fail if suite skipped" step treats a symptom, not the cause (~L86-105)**

The 15-line junit.xml-parsing step exists because `pg_available` (the fixture) always resolves a connection failure to a silent `pytest.skip`, even when `TEST_PGVECTOR_URL` was explicitly set by CI. Raising instead of skipping when that env var is present would fix this at the fixture level for any future CI job, instead of every caller needing its own XML-scraping bandaid. Relatedly, `if: always()` on that step means it'll hit `FileNotFoundError` if an earlier step fails before pytest ever writes `junit.xml` — not a false-pass, but loses the clean diagnostic in that failure mode.

PR #1337, opened 2026-08-31
test: install extractor dispatch mocks per test, not at module scope

### LINKED ISSUE #1336: test: test_extractors_dispatch leaves MagicMocks in sys.modules, breaking 132 tests in full-suite runs
## Summary

`tests/test_extractors_dispatch.py` installs `MagicMock`s into `sys.modules` at module scope and never removes them:

```python
sys.modules['spacy'] = MagicMock()
sys.modules['semantica.semantic_extract.methods'] = MagicMock()
sys.modules['semantica.utils.logging'] = MagicMock()
sys.modules['semantica.utils.progress_tracker'] = MagicMock()
sys.modules['semantica.semantic_extract.providers'] = MagicMock()
```

pytest imports **every** test module during collection, before any test runs. So while these mocks are installed, every later test module is imported and binds them into its own globals — `from ..utils.logging import get_logger` captures a `MagicMock`, and the extractor instances built from it get mock loggers, mock method registries, and so on.

## Effect

Measured on `main` with a full-suite run, before and after fixing this one file:

| | Failed | Passed |
|---|---|---|
| `main` | **199** | 5,506 |
| with the fix | **67** | 5,638 |

**132 tests pass only because of import order.** They pass when run alone and fail when the full suite runs. A sample:

| File | alone | after `test_extractors_dispatch.py` |
|---|---|---|
| `tests/test_issue_554_fixes.py` | 0 failed | 20 failed |
| `tests/test_ner_configurations.py` | 0 failed | 8 failed |
| `tests/test_semantic_extract_deepdive_part2.py` | 2 failed | 7 failed |
| `tests/test_notebooks_verification.py` | 0 failed | 2 failed |
| `tests/test_relations_llm.py` | 0 failed | 2 failed |

## Why a teardown does not fix it

This is worth stating up front, because the obvious fix does not work. I tried three variants:

1. Restoring the five `sys.modules` entries in `tearDownModule()` — fixed 12 of the failures.
2. Also restoring the parent-package attributes (importing a submodule rebinds it on its parent) — no further gain.
3. Also dropping every `semantica.*` module imported while the mocks were live — no further gain.

All three run too late. Collection has already finished by the time any teardown fires:

```python
# in a later test module, evaluated at import (collection) time
_AT_IMPORT = isinstance(sys.modules.get("semantica.utils.logging"), MagicMock)
```

```
MOCKED_AT_COLLECTION_TIME=True
MOCKED_AT_RUN_TIME=False
```

The mocks have to never be installed at module scope in the first place.

## Suggested fix

Import the extractors for real and install the stand-in per test via `patch.dict` in `setUp`. The extractors resolve `from .methods import get_entity_method` lazily inside their methods, so the mock only needs to be in `sys.modules` while a test runs — not at import time.

The resulting file is smaller than the current one (29 insertions, 47 deletions), its own 5 tests still pass, and flake8 goes from 46 findings to 44.

`tests/semantic_extract/test_retry_logic.py` already documents this hazard for external packages — *"Leaving them in place would poison other test modules"* — and restores them after import. That works there because those imports are eager; it does not generalise to this file, whose mocks are needed at call time.

I'd like to take this on if the approach looks right — it is implemented and verified locally.

dex0shubham:
## Description

`tests/test_extractors_dispatch.py` installed `MagicMock`s into `sys.modules` at module scope and never removed them. Because pytest imports every test module during collection before running anything, those mocks were live while later modules were imported, and each bound them into its own globals — mock loggers, mock method registries, mock providers.

The result was 132 tests that pass alone and fail in a full-suite run.

## Type of Change

## Related Issues

Closes #1336

## Changes Made

- The extractors are now imported for real at module scope, with no `sys.modules` mutation there.
- The `methods` stand-in is installed per test via `patch.dict` in `setUp`, removed automatically by `addCleanup`. The extractors resolve `from .methods import get_entity_method` lazily inside their methods, so the mock only has to exist while a test runs.
- `spacy`, `utils.logging`, `utils.progress_tracker` and `semantic_extract.providers` are no longer mocked at all — the real modules import fine.

Net effect is a smaller file: 29 insertions, 47 deletions.

## Testing

- [x] Tested locally

Full suite, before and after:

| | Failed | Passed |
|---|---|---|
| `main` | 199 | 5,506 |
| this PR | 67 | 5,638 |

Each previously affected file now behaves identically alone and after this one:

```
test_issue_554_fixes.py                alone=0 failed   after=0 failed
test_ner_configurations.py             alone=0 failed   after=0 failed
test_notebooks_verification.py         alone=0 failed   after=0 failed
test_relations_llm.py                  alone=0 failed   after=0 failed
test_semantic_extract_deepdive.py      alone=1 failed   after=1 failed
test_semantic_extract_deepdive_part2.py alone=2 failed  after=2 failed
```

The residual failures in the last two are pre-existing and unrelated — identical counts alone and after.

This file's own 5 tests still pass. flake8 goes from 46 findings to 44.

## Note on approach

A `tearDownModule()` restoring the mocked entries looks like the natural fix and does not work — it fixed only 12 of the failures. Nor does additionally restoring parent-package attributes, nor dropping the modules imported while the mocks were live. All of them run after collection has already completed:

```
MOCKED_AT_COLLECTION_TIME=True
MOCKED_AT_RUN_TIME=False
```

That is why the mocks are moved into `setUp` rather than being cleaned up afterwards.


Sameer6305 (approved):
Approved.

The PR correctly addresses #1336 by removing the module-scope sys.modules mutations from tests/test_extractors_dispatch.py and scoping the required methods mock to individual tests with patch.dict and addCleanup.

I verified the effective diff is limited to this test file, with no production-code changes. The collection-time import environment is no longer polluted, and the affected test suites behave identically whether test_extractors_dispatch.py runs before them or not.

The extractor import paths were also checked to confirm that the relevant get_*_method imports are lazy and therefore compatible with the per-test mocking approach.

Validation included the focused dispatch tests, the previously affected regression suites, order-independence checks, and the full suite. The full suite completed with 6,163 passed, 73 failed, 150 skipped, and 7 xfailed; the remaining failures were investigated and determined to be pre-existing and unrelated to this PR.

No additional correctness issues were found within the scope of #1336/#1337, and no further changes are necessary.

Merge-ready.

PR #1343, opened 2026-08-31
fix(vector_store): return stored metadata in MilvusCollection.search() (#1330)

### LINKED ISSUE #1330: [BUG] Stale comment in `milvus_store` says the schema has no metadata field
There is a stale comment in `milvus_store` saying that the Milvus collection schema only stores `id` and `vector`, with no metadata field.

That is no longer true. `create_collection()` defines a metadata field at line 393:

`FieldSchema(name="metadata", dtype=DataType.JSON)`

So the "future schema migration" mentioned in the comment has already happened, but the comment and the hardcoded `{}` in `search()` were never updated.

## Why this matters

There are two separate problems here.

First, `search()` silently drops metadata that is actually stored in the collection. The other read paths in this file, including `get_vector`, `get_metadata`, and `filter_by_metadata`, request and return metadata correctly. That makes `search()` inconsistent with the rest of the module for no apparent reason.

Second, the stale comment is actively misleading.

A Qodo review on #1326 cited it as evidence that the repository "explicitly documents previously created Milvus schemas without metadata" and raised a correctness finding based on that assumption.

But this is not a documented supported configuration. It's just an outdated comment that now contradicts the actual schema definition roughly 190 lines later.

## Proposed fix

Update `search()` to include `metadata` in `output_fields` and return the stored metadata instead of a hardcoded `{}`.

At minimum, the stale comment should be corrected so it no longer describes behavior that the current schema contradicts.

This should be small and self-contained, and seems like a reasonable `good first issue`.

mhaye9545:
Closes #1330.

## Summary

`MilvusCollection.search()` contained a stale comment claiming the collection schema only stores `id` and `vector` with no metadata field. That description was accurate once, but `create_collection()` has long defined a `FieldSchema(name="metadata", dtype=DataType.JSON)` at line 393.

As a result of the stale comment, `search()` hardcoded `"metadata": {}` in every result, silently discarding metadata that was actually stored — making it inconsistent with `get_metadata()` and `filter_by_metadata()`, which both request and return metadata correctly.

### `semantica/vector_store/milvus_store.py`

**`MilvusCollection.search()`**

1. Added `output_fields=["metadata"]` to the `self.collection.search()` call so the Milvus SDK returns the stored metadata field alongside each hit.
2. Changed `"metadata": {}` → `"metadata": hit.entity.get("metadata") or {}` so callers receive the actual stored value. Falls back to `{}` for legacy records where the field is absent.
3. Updated the comment to accurately describe the current schema.

### `tests/vector_store/test_milvus_store_search_metadata.py` [new]

Four focused unit tests (no real Milvus connection needed — all collection calls are mocked):

| Test | What it checks |
|---|---|
| `test_search_returns_stored_metadata` | Result dict contains the metadata that was stored |
| `test_search_requests_metadata_output_field` | `output_fields=["metadata"]` is passed to the SDK |
| `test_search_falls_back_to_empty_dict_when_metadata_is_none` | `None` entity value yields `{}`, not `None` |
| `test_search_score_calculation` | Score formula and distance field are preserved unchanged |

## Verification

```
pytest tests/vector_store/test_milvus_store_search_metadata.py -v
4 passed in 11.59s
```


ZohaibHassan16 (changes requested):
The core fix looks right. 

There's one thing this broke though. `TestVectorStoreDeepDive::test_milvus_store` in `tests/vector_store/test_vector_store_deepdive.py` builds a Milvus mock hit with only .id and .distance set, no entity.get stub, same pattern that broke `TestMilvusSearchSchema` in `test_search_result_schema.py`. That one got the mock_hit.entity.get.return_value = None fix but this one didn't. Ran it: passes on main, fails on this branch with

`AssertionError: <MagicMock name='mock.entity.get()' id=...> is not an instance of <class 'dict'>`

Since MagicMock auto-attributes are truthy, hit.entity.get("metadata") or {} returns the mock itself instead of falling through to {}..

zeekay (approved):
The premise checks out, and the comment this replaces was the actually-wrong part. `MilvusStore.create_collection` defines the schema at milvus_store.py:388-394:

```python
FieldSchema(name="id", dtype=DataType.VARCHAR, is_primary=True, ...),
FieldSchema(name="vector", dtype=DataType.FLOAT_VECTOR, dim=dimension),
FieldSchema(name="metadata", dtype=DataType.JSON),
```

So the comment at line 211 — "Milvus collection schema stores only id+vector; no metadata field is defined in create_collection()" — has been false for as long as that schema has existed, and `search()` was returning `{}` on the strength of it. The rest of the file already reads the field (`output_fields` at lines 656, 715, 773), which is what makes `search()` the odd one out: `get_metadata()` and `filter_by_metadata()` returned real values while `search()` returned nothing.

Popping `output_fields` out of `**options` before merging is the right call rather than an afterthought — `collection.search(output_fields=..., **options)` with a caller-supplied `output_fields` would have been a `TypeError`, and that's exactly the sort of thing that only shows up for the one user who passes it.

Test runs against current `main` (merged, the branch is behind): `pytest tests/vector_store/ -m "not integration"` — 345 passed, 84 skipped, 7 xfailed, against 341 passed on `main`. The four new tests are the difference.

Two small things I checked rather than assumed:

- `list({"metadata", *caller_fields})` has nondeterministic order, but the test asserts `"metadata" in output_fields` rather than list equality, so it won't flake. Worth keeping that assertion as-is if anyone edits it later.
- The `or {}` fallback and the corresponding one-line update to `test_milvus_search_canonical_schema` keep the "legacy record with no metadata" path returning `{}` rather than `None`, so the result schema stays stable for existing callers.

zeekay (changes requested):
Correcting my own approval above — @ZohaibHassan16 is right, and I missed it because of how I ran the suite. Reproduced on this branch merged onto current `main`:

```
tests/vector_store/test_vector_store_deepdive.py::TestVectorStoreDeepDive::test_milvus_store
>       self.assertIsInstance(results[0]["metadata"], dict)
E       AssertionError: <MagicMock name='mock.entity.get()' ...> is not an instance of <class 'dict'>
1 failed, 10 deselected
```

Passes on `main`, fails here. The cause is the one they gave: the mock hit in that test sets only `.id` and `.distance`, so `hit.entity.get("metadata")` returns an auto-created `MagicMock`, which is truthy, so `or {}` never fires and the mock itself lands in the result dict. It's the same underspecified mock you already fixed in `test_search_result_schema.py` — this file just needs the same `mock_hit.entity.get.return_value = None`.

Worth knowing why I didn't catch it: `tests/vector_store/test_vector_store_deepdive.py:19` sets `pytestmark = pytest.mark.integration` for the whole module, so `pytest -m "not integration"` skips it — that's the "11 deselected" in the 345-passed run I quoted. The file only uses mocks and needs no Milvus server, so the marker is wrong, but until someone fixes that this failure is invisible to the standard local command and CI doesn't run pytest at all. If you're touching it anyway, dropping the module-level marker would stop the next person from hitting the same blind spot.

Everything else in my earlier review stands: the schema really does define `FieldSchema(name="metadata", dtype=DataType.JSON)`, the comment you're replacing was false, and popping `output_fields` out of `**options` is the right way to avoid the duplicate-keyword `TypeError`. This is a one-line test fix away from being ready.

ZohaibHassan16 (approved):
Pushed the 1-line mock fix for TestVectorStoreDeepDive::test_milvus_store and synced with main. All Milvus and vector store tests are now green. Ready to merge.

PR #1344, opened 2026-08-31
fix(vector-store): validate collection schema in MilvusStore.get_collection

### LINKED ISSUE #1331: [BUG] MilvusStore.get_collection attaches to collections without validating their schema
<html>
<body>
<h2><span>Problem</span></h2><p class="isSelectedEnd"><code dir="ltr"><span>MilvusStore.get_collection()</span></code><span> (</span><code dir="ltr"><span>semantica/vector_store/milvus_store.py:420</span></code><span>) wraps whatever </span><code dir="ltr"><span>Collection(collection_name)</span></code><span> returns without checking whether its schema is compatible with what the rest of </span><code dir="ltr"><span>MilvusStore</span></code><span> expects.</span></p><p class="isSelectedEnd"><span>The current flow is essentially:</span></p><pre dir="ltr"><code dir="ltr"><span>if not utility.has_collection(collection_name):
    raise ProcessingError(f"Collection {collection_name} does not exist")

collection = Collection(collection_name)
self.collection = MilvusCollection(collection, collection_name)</span></code></pre><p class="isSelectedEnd"><span>After that, the read paths assume the collection has the same schema created by </span><code dir="ltr"><span>create_collection()</span></code><span> around lines 388–394:</span></p><ul data-spread="false"><li><span>a VARCHAR primary key named </span><code dir="ltr"><span>id</span></code></li><li><span>a </span><code dir="ltr"><span>vector</span></code><span> field</span></li><li><span>a </span><code dir="ltr"><span>metadata</span></code><span> field</span></li></ul><p class="isSelectedEnd"><span>Those assumptions are spread across several methods:</span></p>
Line | Method | Assumption
-- | -- | --
598 | get_vector | expr=f'id == "{safe_id}"' assumes a string/VARCHAR primary key
615 | get_metadata | same assumption
660 | filter_by_metadata | uses the match-all expression "id != ''"
666 | filter_by_metadata | requests ["id", "vector", "metadata"]
730 | iter_all | uses the same expression and output fields

<p class="isSelectedEnd"><span>If </span><code dir="ltr"><span>get_collection()</span></code><span> is pointed at an existing Milvus collection with, for example, an </span><code dir="ltr"><span>INT64</span></code><span> primary key or no </span><code dir="ltr"><span>metadata</span></code><span> field, it attaches successfully.</span></p><p class="isSelectedEnd"><span>The actual failure only shows up later when one of these methods runs, usually as a Milvus expression or missing-field error. At that point the error is far removed from the real problem: the collection schema was incompatible from the start.</span></p><h2><span>Why it matters</span></h2><p class="isSelectedEnd"><span>This is especially relevant for existing collections, which is exactly the path </span><code dir="ltr"><span>get_collection()</span></code><span> is meant to support.</span></p><p class="isSelectedEnd"><code dir="ltr"><span>create_collection()</span></code><span> already delegates to </span><code dir="ltr"><span>get_collection()</span></code><span> when a collection with the requested name exists, so users pointing Semantica at a collection created elsewhere can hit this without any warning at initialization time.</span></p><p class="isSelectedEnd"><span>It matters more now that </span><code dir="ltr"><span>iter_all()</span></code><span> was added in #1326, because an existing collection can also be used as a migration source. A migration failing halfway through with an unrelated Milvus expression error is much harder to diagnose than rejecting an incompatible collection when it is first attached.</span></p><h2><span>Proposed solution</span></h2><p class="isSelectedEnd"><span>Validate the collection schema inside </span><code dir="ltr"><span>get_collection()</span></code><span> and raise a clear error if it does not match the schema Semantica currently expects.</span></p><p class="isSelectedEnd"><span>At minimum, verify that:</span></p><ul data-spread="false"><li><span>the primary key is named </span><code dir="ltr"><span>id</span></code></li><li><span>the primary key is </span><code dir="ltr"><span>VARCHAR</span></code></li><li><span>a </span><code dir="ltr"><span>vector</span></code><span> field exists</span></li><li><span>a </span><code dir="ltr"><span>metadata</span></code><span> field exists</span></li></ul><p class="isSelectedEnd"><code dir="ltr"><span>Collection.schema</span></code><span> exposes the field definitions, so this can be done once when attaching the collection rather than adding checks to every operation.</span></p><p class="isSelectedEnd"><span>Failing at the boundary would make the error much clearer and would also give us one explicit place to document the Milvus schema Semantica supports, instead of leaving that contract implicit across several methods.</span></p><h2><span>Out of scope</span></h2><p class="isSelectedEnd"><span>Making the read paths schema-adaptive is a separate problem.</span></p><p class="isSelectedEnd"><span>For example, we could theoretically derive the primary key name and type dynamically or handle collections without a metadata field, but that would require deciding whether Semantica intends to support arbitrary Milvus schemas or only the schema it creates itself.</span></p><p class="isSelectedEnd"><span>This issue is only about validating the existing contract and failing clearly when a collection does not satisfy it.</span></p><p><span>Related: #1326 added </span><code dir="ltr"><span>iter_all()</span></code><span> using the same schema assumptions as the existing read paths. A Qodo review there raised the schema coupling, which is what surfaced this issue.</span></p>
</body>
</html>

pkupt:
`get_collection()` currently wraps `Collection(name)` right after the `has_collection` guard without checking the schema. An INT64-pk or metadata-less collection attaches fine and then fails inside `get_vector`/`get_metadata` with an error that's far from the root cause.

This adds a schema check right after the attach, before the store assigns `self.collection`, so a mismatch is caught early and the error names the problem. Also adds 9 focused tests for the matching case and each rejection case.

Closes #1331.


ZohaibHassan16 (approved):
Nice, this is clean.

The validation matches `create_collection()`'s actual schema exactly:

* `id` → `VARCHAR`, `is_primary=True`, `auto_id=False`
* `vector` → `FLOAT_VECTOR`
* `metadata` → `JSON`

Every rejection case lines up with something `create_collection()` actually enforces. Nothing extra, nothing missing.

I applied the diff and ran all 9 tests. All pass.

Each rejection case is covered independently:

* wrong primary key dtype
* missing primary key
* wrong primary key name
* `auto_id` enabled
* missing vector field
* wrong vector dtype
* missing metadata field
* wrong metadata dtype

They also all assert that `store.collection` stays `None`, so an invalid collection never gets attached.

Both Qodo findings are actually fixed too. The metadata JSON-type check and the `auto_id` rejection are both in the diff, with dedicated tests for each.

I also checked the existing tests. Nothing currently calls `get_collection()` against Milvus. The deepdive test only exercises `create_collection()` with `has_collection=False`, so it never reaches this path.

The full `tests/vector_store/` run with the diff applied matches baseline exactly. Nothing new broke.

Approving.

PR #1229, opened 2026-08-27
fix(ontology): resolve endpoints in direct property inference

T1mn:
## Description
The merged relationship-endpoint fix covers the main ontology generation pipeline, but the public property inference entry point still falls back to owl:Thing when relationship endpoints are provided by entity IDs or aliases.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring

## Related Issues

Follow-up to merged PR #1170.

## Changes Made

- Extract the relationship endpoint alias and type resolution logic into a shared ontology helper.
- Apply the same resolver to PropertyGenerator and the public OntologyGenerator.infer_properties path.
- Preserve the existing high-level generation behavior and add direct-API regression coverage.

## Testing

- [x] Tested locally
- [x] Added tests for new functionality
- [ ] Package builds successfully (python -m build)

### Test Commands

```bash
PYTHONPATH=. pytest -q tests/ontology/test_ontology_relationship_endpoints.py tests/ontology/test_ontology_normalized_properties.py
python -m compileall -q semantica/ontology/relationship_utils.py semantica/ontology/ontology_generator.py semantica/ontology/property_generator.py
```

## Documentation

- [ ] Updated relevant documentation
- [ ] Added code examples if applicable
- [ ] Updated API reference if adding new APIs
- [ ] Updated cookbook if adding new examples
- [x] No documentation changes needed

## Breaking Changes

**Breaking Changes**: No

## Checklist

- [x] My code follows the project style guidelines
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] Package builds successfully

## Additional Notes

This is a focused follow-up to the merged endpoint-resolution fix. Thank you for reviewing it.

PR #1356, opened 2026-09-01
fix(ci): use pip-audit for dependency security scans

T1mn:
## Description

This standalone PR fixes a repository-level CI failure identified while validating [PR #1229](https://github.com/semantica-agi/semantica/pull/1229).

The `security-scan` job currently runs Safety against `requirements-ci.txt`. Safety 3.8.1 raises an unhandled `cuda-toolkit` exception and exits without writing `safety-report.json`, causing the job to fail before it can report dependency findings. The same failure was observed on an unrelated Dependabot security-scan run, so this is independent of the Ontology changes in #1229.

This PR switches that scan to the repository's already pinned `pip-audit` tool and keeps the existing report guard and blocking behavior for dependency vulnerabilities and Bandit HIGH findings.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Discovery context: [PR #1229](https://github.com/semantica-agi/semantica/pull/1229). This PR does not close it.

## Changes Made

- Replace Safety with the pinned `pip-audit` tool in the aggregate security scan.
- Parse the `pip-audit` JSON report and upload it under the new report name.
- Regenerate the hashed security-scan tool lockfile and update its documentation.

## Testing

- [x] Tested locally

### Test Commands

- Parsed the workflow YAML with Ruby.
- Checked the extracted workflow shell block with `bash -n`.
- Validated representative `pip-audit` JSON counting, detail rendering, and PR comment parsing.
- Rechecked the pinned tool lockfile with `uv 0.12.1` and `git diff --check`.

The full `pip-audit` invocation could not run locally because the host Python `ensurepip` aborts while `pip-audit` creates its temporary environment. GitHub Actions will provide the authoritative Ubuntu validation.

## Documentation

## Breaking Changes

**Breaking Changes**: No

## Checklist

## Additional Notes

The failure was reproduced on an unrelated Dependabot security-scan run as well as during validation of PR #1229. No Ontology runtime behavior is changed by this PR.


ZohaibHassan16 (changes requested):
This is a slightly awkward review because the bug you're fixing has already been fixed a different way on `main`.

`main` moved since this PR was opened. [[3254b9be](https://github.com/semantica-agi/semantica/commit/3254b9be804c273b6756a7987af4105804d40452)](https://github.com/semantica-agi/semantica/commit/3254b9be804c273b6756a7987af4105804d40452) landed today as part of #1157, including `fix(ci): scan declared dependencies with Safety`, which changed the command to:

```bash
safety check --file requirements-ci.txt --save-json safety-report.json || true
```

That fixes the same cuda-toolkit crash. The problem is Safety trying to inspect metadata from the installed environment, pointing it at `requirements-ci.txt` avoids that entirely. 

That's also why GitHub is showing this as `CONFLICTING` / `DIRTY`. I applied this PR's raw diff to a fresh `main` checkout and got a hunk conflict on the `safety check` line. Both changes are touching the same place for the same underlying issue.

One other thing to account for when rebasing: `main` already has a separate `security.yml` workflow running:

```bash
pip-audit -r requirements-ci.txt
```

It's path-filtered to `pyproject.toml` / `requirements-ci.txt` and blocks PRs. If this PR lands as-is, we'll be running pip-audit twice against the same dependency file , once there and once in `security-scan.yml`.

The actual implementation here looks correct, though. I installed `pip-audit==2.10.1` in a scratch venv and tested it against both a known vulnerable pin and a clean one to verify the JSON assumptions:

```json
{"dependencies": [{"name": "requests", "version": "2.25.0", "vulns": [{"id": "PYSEC-2023-74", ...}]}], "fixes": []}
```

`dependencies` contains all audited packages, not only vulnerable ones, so `.dependencies | length > 0` won't incorrectly fail a clean scan. Exit codes are also 1 when findings exist and 0 when clean, so the `|| true` plus explicit vuln-count gating behaves as expected.

I also compared this with the pre-PR JS, and the malformed-report handling is a real improvement for the Qodo finding. The old code did `(data.vulnerabilities || [])`, which meant a malformed report or missing key could silently render as `✅ No findings`. This version returns `null` and shows a warning instead. Good fix.

Small non-blocking note: that malformed-structure guard is only applied to the pip-audit section. Bandit and Semgrep still use the old unguarded pattern. Same class of issue, but I'd consider that out of scope for this PR.

So I'd rebase onto current `main`, drop whatever overlaps with the `--file` Safety fix that's already landed and remove the duplicate pip-audit run.

Once it applies cleanly, happy to take another look.

PR #1357, opened 2026-09-01
docs(cookbook): correct graph mapping and deduplication in notebook 07

### LINKED ISSUE #1287: [DOCS] Fix incorrect graph mapping and dedup output in introduction notebook 07
## Description

`cookbook/introduction/07_Building_Knowledge_Graphs.ipynb` currently reports successful graph construction by counts while constructing relationship endpoints that do not correspond to the extracted relations. Its deduplication step also reports only merged duplicate groups as the final entity set, dropping every non-duplicate entity from the displayed result.

This is a cookbook-only problem. `GraphBuilder`, `EntityResolver`, and the deduplication APIs can represent the intended result when given correctly mapped inputs; the notebook prepares and displays those inputs incorrectly.

## Reproduction

With the current sample text, the extracted mentions can be:

```text
1. 'Apple Inc.'                   ORG
2. 'Tim Cook'                     PERSON
3. 'Apple Inc. Apple Inc.'        ORG
4. 'Cupertino'                    GPE
5. 'California'                   GPE
```

The adjacent `Apple Inc.` mentions cross a sentence boundary after an abbreviation and are merged by the default spaCy model, making the example unnecessarily dependent on a sentence-boundary edge case.

The relation extractor then returns endpoints such as:

```text
'Apple Inc. Apple Inc.' --located_in--> 'Cupertino'
'Tim Cook' --works_for--> 'Apple Inc. Apple Inc.'
```

However, the notebook discards `rel.subject` and `rel.object` and fabricates endpoints from the loop index:

```python
relationships.append({
    "source": "e1",
    "target": f"e{i+1}",
    "type": rel.predicate,
    "properties": {},
})
```

The resulting graph therefore contains unrelated edges such as:

```text
e1:'Apple Inc.' --located_in--> e2:'Tim Cook'
e1:'Apple Inc.' --works_for--> e3:'Apple Inc. Apple Inc.'
```

The notebook prints only `5 entities / 2 relationships`, so the semantic corruption is hidden.

Step 2 resolves five mentions to four entities, but only prints the counts and does not show which entities were merged or how relationship endpoints should be remapped.

Step 3 then constructs its displayed result with:

```python
deduplicated_entities = [op.merged_entity for op in merge_operations]
```

One duplicate group produces one merge operation, so the notebook reports one deduplicated entity even though the complete deduplicated set contains the merged entity plus three untouched entities (four total).

## Expected behavior

The notebook should build graph edges from the actual extracted relation endpoints, use a stable example that demonstrates duplicate mentions without triggering an unrelated NER boundary failure, and distinguish merge-operation output from the complete deduplicated entity collection.

## Suggested notebook changes

- Use sample text that retains two separate `Apple Inc.` mentions while avoiding adjacent abbreviation-boundary ambiguity, for example:

  ```text
  Apple Inc. is headquartered in Cupertino, California.
  Tim Cook is the CEO of Apple Inc.
  The company is a technology company.
  ```

- Create a mapping from each extracted `Entity` object (or mention span) to its graph ID.
- Build each relationship from `rel.subject` and `rel.object`, skipping or reporting an endpoint only when it is genuinely absent from the entity mapping.
- Prefer the explicit input form `builder.build({"entities": ..., "relationships": ...})` in the tutorial.
- Print entity text/type/span, extracted relations, graph ID mappings, and human-readable graph edges instead of counts alone.
- In the resolution step, display the source entities and canonical result for each merge and explain endpoint remapping.
- In the deduplication step, add untouched entities back to the merged results (or reuse the resolver API that already returns the complete resolved collection).
- Print merge-operation count separately from final deduplicated-entity count.
- Add lightweight assertions for the expected canonical entities and edges so future notebook edits cannot silently reintroduce incorrect mappings.

## Scope

Only `cookbook/introduction/07_Building_Knowledge_Graphs.ipynb` should change. Production graph-building, entity-resolution, spaCy, and deduplication implementations are outside this issue.

## Acceptance criteria

- Every graph relationship endpoint corresponds to the actual `Relation.subject` / `Relation.object` mention mapped into the graph.
- The example produces two separate Apple mentions, then resolves them to one canonical Apple entity.
- The final resolved/deduplicated entity count is four, not one.
- The final human-readable relationships are:

  ```text
  Apple Inc. --located_in--> Cupertino
  Tim Cook --works_for--> Apple Inc.
  ```

- The notebook displays the transformed records at each stage, not only aggregate counts.
- The notebook executes from top to bottom without production-code changes.

I would like to work on this cookbook fix. Please assign it to @taoche.

taoche:
## Problem

Notebook 07 reported plausible entity and relationship counts while constructing graph edges from loop indices instead of the extracted `Relation.subject` and `Relation.object`. Its deduplication output then displayed only merge operations, which made five mentions appear to collapse to one entity instead of four canonical entities.

## What changed

- Build a stable mention-span → graph-ID mapping.
- Construct every edge from the actual extracted relation endpoints; report genuinely unmapped endpoints instead of silently mis-wiring them.
- Print entity mentions, relations, graph IDs, human-readable edges, merge provenance, and the complete resolved collection at the stage where each transformation happens.
- Remap relationship endpoints after entity resolution.
- Distinguish one merge operation from the four-entity deduplicated result.
- Install `en_core_web_sm` idempotently because spaCy language models are distributed separately and this lesson requires standalone GPE recognition.
- Assert both canonical entity identities and final relationship semantics.

## Expected result

```text
Apple Inc. --located_in--> Cupertino
Tim Cook --works_for--> Apple Inc.

5 mentions -> 4 canonical entities
1 duplicate group -> 1 merge operation -> 4 deduplicated entities
```

## Verification

Executed all instructional cells sequentially with `en_core_web_sm` 3.8.0:

- extracted five mentions and two relations;
- built two correctly mapped graph edges;
- merged only `e1` and `e5` (`Apple Inc.`);
- retained `Apple Inc.`, `Tim Cook`, `Cupertino`, and `California` as the exact canonical `(name, type)` set;
- completed all assertions without production-code changes.

This PR changes only `cookbook/introduction/07_Building_Knowledge_Graphs.ipynb`.

Closes #1287


KaifAhmad1 (changes requested):
@taoche Reviewed. The core logic fix (real relation endpoints instead of loop-index-fabricated edges, honest dedup counting) is sound — I found no bugs in that part. But I verified the new hard assertions against this repo directly, not just by reading the diff, and one blocking issue and one minor gap survive; a third issue the bot flagged does not.

**1. Blocking: the notebook can't run end-to-end without `en_core_web_sm`, and no install step is added**

Ran the notebook's actual Step 1 extraction code in an environment without `en_core_web_sm` installed (confirmed absent) and got:
```
MENTION Apple Inc. ORG 0 10
MENTION Apple Inc. ORG 77 87
MENTION Tim Cook PERSON 54 62
REL Tim Cook works_for Apple Inc.
```
No `Cupertino`/`California` GPE mentions, no `located_in` relation — the pattern-fallback NER doesn't catch them. This trips the new assertion at cell ~L158 immediately:
```python
assert ("Apple Inc.", "located_in", "Cupertino") in edges
```
and would also fail `assert len(resolved_entities) == 4` (Step 3) and `assert len(deduplicated_entities) == 4` (Step 4), since only 3 mentions (not 5) get extracted without the model.

The only install cell is `!pip install semantica` — there's no `!python -m spacy download en_core_web_sm` step, and I checked: no other cookbook notebook has one either, so this is a repo-wide gap the new hard assertions now expose for the first time. This affects the Colab badge at the top of the notebook too — Colab doesn't ship `en_core_web_sm` by default. Since the PR's own stated goal is a notebook that actually executes correctly (closing #1287, which was about corrupted/misleading output), an assertion that fails in the platform the notebook links to as its primary entry point is a real blocker, not a nitpick.

Suggested fix: add a cell (or extend the existing install cell) with `!python -m spacy download en_core_web_sm`.

**2. Minor: canonical entities are counted, not identified**

```python
assert len(resolved_entities) == 4
```
only checks the count. An incorrect resolution that happened to produce 4 entities (wrong merges, wrong names) would still pass. Consider also asserting on `resolved_names` content, e.g. that the merged entity's name is `"Apple Inc."` and that `"Tim Cook"`/`"Cupertino"`/`"California"` all survive unmerged.

**3. Not a bug: the bot's "merged_from lookup misses provenance" finding is a false positive**

Ran `EntityResolver.resolve_entities()` directly against a minimal duplicate pair:
```
e1 | merged_from top-level: ['e1', 'e5'] | metadata: {'provenance': {'merged_from': [...], 'merge_count': 2}}
```
The merged entity does carry a top-level `merged_from` key — it's set by `MergeStrategyManager.merge_entities()` (`semantica/deduplication/merge_strategy.py:334`), separate from the nested `metadata.provenance.merged_from` copy that `EntityMerger._add_provenance()` layers on top of the same object (`entity_merger.py:506`). The notebook's `entity.get("merged_from", [entity["id"]])` (Step 3) correctly finds the top-level one, so `canonical_id` is built correctly and no `KeyError` occurs. No change needed here.

KaifAhmad1 (approved):
@taoche Re-reviewed after 3ed31b9. Both requested changes are addressed, and I verified end-to-end rather than just reading the diff — actually downloaded `en_core_web_sm` and ran the full notebook logic (Steps 1-4) against the real library.

**1. spaCy model gap — fixed and verified working**

The new install cell:
```python
try:
    spacy.load("en_core_web_sm")
except OSError:
    subprocess.check_call([sys.executable, "-m", "spacy", "download", "en_core_web_sm"])
```
is idempotent and correctly uses `sys.executable`, so it targets whatever kernel is actually running the notebook (important for Colab/venvs where a bare `!python` might resolve differently). Switching `!pip install semantica` to `%pip install semantica` is also the right call — the magic form installs into the active kernel's environment rather than an ambient shell's.

I ran this for real: downloaded the model, then re-ran the notebook's exact extraction/graph/resolution code against the actual library. All assertions pass:
```
Mentions: 5
Relations: 2
edges: {('Apple Inc.', 'located_in', 'Cupertino'), ('Tim Cook', 'works_for', 'Apple Inc.')}
Step 2 asserts OK
canonical_entities: {('Tim Cook', 'PERSON'), ('Apple Inc.', 'ORG'), ('Cupertino', 'GPE'), ('California', 'GPE')}
Step 3 asserts OK — ALL PASS
```
and separately confirmed Step 4's dedup counts too (`merge_operations: 1`, `deduplicated_entities: 4`). The notebook genuinely runs top-to-bottom now, including from a clean environment without the model pre-installed — which was the actual blocker.

**2. Canonical entity identities — fixed**

```python
assert canonical_entities == {("Apple Inc.", "ORG"), ("Tim Cook", "PERSON"), ("Cupertino", "GPE"), ("California", "GPE")}
```
now pins the exact set, not just a count. Verified this passes against the real resolution output above.

**3. merged_from — correctly left unchanged**, matching what I confirmed last round (top-level key is real, bot's finding was a false positive).

No new issues introduced by this commit. Approving.

PR #1359, opened 2026-09-01
docs(cookbook): make notebook 08 a real rerunnable KG workflow

### LINKED ISSUE #1289: [DOCS] Make introduction notebook 08 a real rerunnable end-to-end workflow
## Description

`cookbook/introduction/08_Your_First_Knowledge_Graph.ipynb` is presented as an end-to-end workflow, but the outputs of its stages are not connected. As a result, the notebook reports empty parsed content, later steps can appear successful using hard-coded data, and rerunning an intermediate cell can fail because the input file has already been deleted.

This appears to be limited to the notebook; `DocumentParser.parse_document()` does return the document text.

## Reproduction

Run the notebook from top to bottom, then rerun the parsing cell.

## Current behavior

1. The parsing cell reads `parsed_document.get("content", "")`, while `DocumentParser.parse_document()` returns the text under `text` (and currently also `full_text`). The notebook therefore prints:

   ```text
   Parsed content length: 0 characters
   Preview: N/A...
   ```

2. The entity extraction step instantiates `NamedEntityRecognizer` and `NERExtractor`, but does not call either one. It uses `expected_entities`, whose character offsets also do not match the sample text.

3. The graph step instantiates `GraphBuilder`, but constructs a `networkx.DiGraph` manually from hard-coded entities and relationships.

4. The visualization step instantiates `KGVisualizer`, but does not invoke it; it only prints graph statistics.

5. The last cell deletes `sample_document.txt`. Rerunning the parser or another intermediate step afterward raises:

   ```text
   ValidationError: Document file not found: sample_document.txt
   ```

These hard-coded stages mask the earlier parsing failure, so the notebook can look successful without exercising the Semantica pipeline it is intended to teach.

## Expected behavior

Chapter 08 should be a real, rerunnable end-to-end example:

```text
sample file
  -> FileIngestor
  -> DocumentParser output
  -> entity/relation extraction
  -> GraphBuilder
  -> KGVisualizer
```

Each stage should consume the preceding stage's output, and rerunning an individual cell should not fail because a later cell removed its input.

## Suggested scope

Notebook-only changes in `cookbook/introduction/08_Your_First_Knowledge_Graph.ipynb`:

- Read parsed text from the documented parser result (`text`, with an appropriate compatibility fallback if needed).
- Replace simulated entities and relationships with actual extractor output, or label them explicitly as fixtures if simulation is intentionally retained.
- Build the graph through `GraphBuilder` rather than bypassing it with direct NetworkX calls.
- Invoke `KGVisualizer` and display the resulting visualization.
- Make setup idempotent and move deletion into an explicit optional cleanup cell, so intermediate cells are safely rerunnable.
- Print enough intermediate output to make data flow and extraction errors visible.
- Correct or remove the inaccurate hard-coded character offsets.

No production parser, extraction, graph, or visualization change should be required for this issue.

## Acceptance criteria

- A fresh top-to-bottom run completes successfully.
- The parser cell displays the actual document content.
- Downstream entities, relationships, and graph are derived from prior cell outputs rather than unrelated hard-coded replacements.
- The graph is built and visualized through the Semantica APIs introduced by the notebook.
- Rerunning the parsing and downstream cells does not produce `Document file not found`.
- The notebook clearly distinguishes real model output from any intentionally deterministic fixture data.

I would like to work on this fix. Please assign this issue to @taoche.

taoche:
## Problem

Notebook 08 was presented as an end-to-end knowledge-graph workflow, but its stages were disconnected: it read the wrong parser key, substituted hard-coded extraction fixtures, bypassed `GraphBuilder`, never called `KGVisualizer`, and deleted its own input during the normal path.

## What changed

- Read the parser's real `text` result and assert that parsing produced content.
- Feed that parsed text into `NERExtractor` and `RelationExtractor`.
- Install `en_core_web_sm` idempotently because the separately distributed model is required for standalone GPEs such as `Cupertino`.
- Map extracted mention spans to graph IDs and build edges from real relation endpoints.
- Build through `GraphBuilder` and render through `KGVisualizer.visualize_network()`.
- Print the records produced at each stage instead of masking failures with aggregate counts.
- Keep setup idempotent and move deletion to a clearly optional cleanup cell.
- Explain why repeated `Apple Inc.` mentions remain separate nodes here and link to notebook 07 for resolution.

## Verified data flow

```text
sample_document.txt
  -> FileIngestor
  -> DocumentParser["text"]
  -> NERExtractor + RelationExtractor
  -> GraphBuilder
  -> KGVisualizer
  -> knowledge_graph.html
```

Executed all instructional cells sequentially with `en_core_web_sm` 3.8.0:

- parsed 127 non-empty characters;
- extracted 8 mentions and 3 relations;
- built 8 graph nodes and these 3 real edges:

  ```text
  Apple Inc. --founded_by--> Steve Jobs
  Apple Inc. --located_in--> Cupertino
  Tim Cook --works_for--> Apple Inc.
  ```

- generated the interactive HTML visualization;
- cleanup removed only the two generated files; rerunning from setup recreates them successfully.

This PR changes only `cookbook/introduction/08_Your_First_Knowledge_Graph.ipynb`.

Closes #1289


KaifAhmad1 (changes requested):
@taoche Reviewed. Structurally this is a clean fix — real parse key, real extraction, real graph edges, correct visualizer call — but one blocking issue survives, and it's the same class of problem as #1357.

**Blocking: the `located_in` assertion fails in a fresh environment without `en_core_web_sm`**

Ran the notebook's real Step 3 extraction against the new sample text with the model absent (confirmed absent in this environment):
```
REL Apple Inc. founded_by Steve Jobs
REL Tim Cook works_for Apple Inc.
```
No `located_in` relation — the pattern-fallback NER doesn't recognize standalone `Cupertino` as a GPE, so:
```python
assert ("Apple Inc.", "located_in", "Cupertino") in edges
```
fails immediately on a fresh `pip install semantica`, which doesn't pull in `en_core_web_sm` (a separately distributed spaCy model). The `works_for` assertion would pass on its own.

This repo already documents the right install snippet elsewhere (`docs/guides/semantic-extraction.md:658`): `pip install spacy && python -m spacy download en_core_web_sm`. Worth adding that as an explicit cell here — same fix needed on #1357, so one shared pattern across both notebooks would be ideal.

**Everything else checks out — no other bugs found**

- `parsed_document.get("text", "")` — verified correct against `DocumentParser._parse_text()`'s actual return dict (`{"text": ..., "metadata": ..., "full_text": ...}`, `document_parser.py:370-378`); the old notebook's `"content"` key was indeed wrong.
- `GraphBuilder.build({"entities": ..., "relationships": ...})` — the dict-input form is valid per the method's own docstring example (`graph_builder.py:467-473`), and `merge_entities=False` by default means entities/edges pass through unchanged.
- `KGVisualizer.visualize_network(knowledge_graph, output="html", file_path=...)` — checked the implementation: despite the docstring saying it "returns a figure (if interactive) or None," the actual code always returns `fig` regardless of `output` value (`kg_visualizer.py:855-856`: "Always return the figure to allow interactive preview in notebooks"), while also writing the HTML file via `export_plotly_figure`. So the final bare `fig` cell will correctly render the interactive plot in the notebook — not a bug, just a stale docstring elsewhere in that file, pre-existing and not introduced by this PR.
- `plotly` is a core dependency in `pyproject.toml` (not an optional extra), so there's no additional install gap on that front.

Once the spaCy model install step is added (mirroring whatever fix lands on #1357), this looks ready to merge.

KaifAhmad1 (approved):
Re-reviewed after 0dbe927. The blocking issue is fixed — I didn't just read the diff, I actually ran it.

**Verification 1 — real notebook execution, real package**

Fetched the PR's actual notebook file (head `f71b711`) and ran it end-to-end with `jupyter nbconvert --execute` against a fresh kernel with `en_core_web_sm` initially absent. The install cell's `try: spacy.load(...) / except OSError: subprocess.check_call([sys.executable, "-m", "spacy", "download", "en_core_web_sm"])` is exactly the right pattern — `spacy.load()` on a missing model raises `OSError`, and `sys.executable` correctly targets the active kernel rather than some other Python on PATH. All 12 cells executed with no errors:

```
Entity mentions: 8   (Apple Inc. x3, Cupertino, California, 1976, Steve Jobs, Tim Cook)
Extracted relations:
  Apple Inc. --founded_by--> Steve Jobs
  Apple Inc. --located_in--> Cupertino
  Tim Cook --works_for--> Apple Inc.
Nodes (entities): 8
Edges (relationships): 3
```

Both assertions (`located_in`, `works_for`) pass. `knowledge_graph.html` gets written, cleanup cell removes only the generated files.

**Verification 2 — same logic, local editable install**

Independently replayed Steps 2-5 against this checkout's `semantica` (not the PyPI release) with the exact sample text from the notebook — identical output, same 3 edges, same assertions passing.

Also double-checked the `works_for` claim against the actual pattern: `semantic_extract/methods.py:1294` matches `"X is the CEO of Y"` -> `works_for`, which is why "Tim Cook is the CEO of Apple Inc." maps there and not to `ceo_of` — consistent with what the notebook asserts.

Everything from my previous pass still holds (`parsed_document["text"]` key, `GraphBuilder.build()` dict-input form, `KGVisualizer.visualize_network()` always returning `fig`). No new issues introduced by the merge commit — it only pulled in unrelated `main` history, the notebook diff is unchanged since 0dbe927.

Approving.

PR #1360, opened 2026-09-01
fix(context): dedupe heuristic causes in trace_decision_causality

### LINKED ISSUE #1358: [BUG] trace_decision_causality reports duplicate "influences" chains when decisions share multiple entities
## Bug Description

`ContextGraph.trace_decision_causality()` reports the same heuristic cause multiple times when two decisions share more than one entity. A decision that shares 2 entities with an earlier decision produces 2 identical "influences" chains, one per shared entity, instead of 1.

## Steps to Reproduce

```python

graph = ContextGraph(advanced_analytics=True)
cause = graph.record_decision(
    category="lending", scenario="earlier review", reasoning="r",
    outcome="approved", confidence=0.9,
    entities=["customer_123", "property_456"],
)
effect = graph.record_decision(
    category="risk", scenario="later review", reasoning="r",
    outcome="flagged", confidence=0.9,
    entities=["customer_123", "property_456"],
)

chains = graph.trace_decision_chain(effect)
influence_hops = [
    (hop["from"], hop["to"])
    for chain in chains
    for hop in chain["hops"]
    if hop["type"] == "influences"
]
print(influence_hops)  # [(cause, effect), (cause, effect)]  <- duplicated
```

## Expected Behavior

`influence_hops` should contain `(cause, effect)` exactly once. A potential cause found via the shared-entity heuristic is one causal link, regardless of how many entities the two decisions have in common.

## Actual Behavior

`(cause, effect)` appears once **per shared entity** (2 times in the example above). The duplicate chains also propagate to `_build_causal_chain_report()`, inflating the causal trace output.

Root cause: in `trace_decision_causality()` (`semantica/context/context_graph.py`), `potential_causes` is a list appended inside a nested loop over the current decision's entities, so the same earlier decision is appended once per shared entity.

## Environment

- **OS**: macOS
- **Python Version**: 3.12
- **Semantica Version**: main (8b125d64)

## Possible Solution

Deduplicate `potential_causes` by decision id (e.g. collect ids in a dict, which preserves insertion order). I'll open a PR for this shortly.

cxzg007:
## Description

`trace_decision_causality()` collects heuristic potential causes in a list, appending inside a nested loop over the current decision's entities. When two decisions share more than one entity (e.g. the same customer and the same property), the earlier decision is appended once per shared entity, so the causal trace reports the same "influences" chain multiple times.

This PR deduplicates `potential_causes` by decision id (a dict-as-ordered-set, preserving the original first-occurrence emission order), so each potential cause is reported exactly once regardless of entity overlap.

## Type of Change

## Related Issues

Fixes #1358

## Changes Made

- `trace_decision_causality()`: collect `potential_causes` in a dict keyed by decision id (insertion-ordered) instead of a list, with a comment explaining why
- Added regression test `test_heuristic_cause_reported_once_per_shared_entity_pair`

## Testing

- [x] Tested locally

### Test Commands

```bash
pytest tests/context/test_decision_causal_edge_regression.py  # 21 passed (fails on parent commit without the fix)
pytest tests/context/  # no new failures vs. baseline
```

## Documentation

## Breaking Changes

**Breaking Changes**: No

## Checklist

## Additional Notes

The new test fails on the parent commit (`8b125d64`) and passes with this fix. Cycle detection, `max_chains` truncation, and `explicit_cause_ids` filtering are untouched — dedup can only reduce chain count, never change the truncation marker contract.


zeekay (approved):
Correct fix, and the test earns its place. Ran it against unpatched `main` and it fails with `assert 2 == 1` on the `(cause, effect)` pair — two identical `influences` hops for two shared entities — so this is pinning a real double-report, not restating the implementation.

The dict-as-ordered-set is the right shape here. `potential_causes` is only iterated (context_graph.py:4807), and a plain `set` would have made chain ordering non-deterministic across runs, which would have been a worse bug than the one being fixed. Insertion order keeps the traversal stable.

Test runs against current `main` (cherry-picked, the branch is behind):
- `pytest tests/context/ -m "not integration"` — 804 passed, 2 skipped, 14 subtests passed
- the new test alone, without the source fix — fails as described

ZohaibHassan16 (approved):
LGTM.

PR #1363, opened 2026-09-01
fix(plugins): declare agents as an array of file paths in plugin.json

### LINKED ISSUE #1350: [BUG] Claude Code plugin cannot be installed: plugin.json declares "agents" as a string, schema requires an array of file paths
## Summary

`plugins/.claude-plugin/plugin.json` declares `"agents": "./agents"`. Claude Code's plugin schema rejects this, so **the bundled Claude Code plugin cannot be installed at all**:

```
Failed to install plugin "semantica@semantica-local": Plugin ... has an invalid manifest file
at .../.claude-plugin/plugin.json.

Validation errors: agents: Invalid input
```

## Reproduction

```bash
claude plugin marketplace add <path-to-clone>/plugins
claude plugin install semantica@semantica-local
```

Tested with Claude Code 2.1.252 on macOS (arm64), semantica v0.6.7 (tag).

## Root cause

`agents` must be an **array of `.md` file paths**. I confirmed the accepted shapes empirically with throwaway plugins:

| `agents` value | result |
|---|---|
| `"./agents"` (string, as shipped) | ❌ `agents: Invalid input` |
| `["./agents"]` (directory inside array) | ❌ `agents: Invalid input` |
| `["./agents/decision-advisor.md", ...]` (file paths) | ✅ installs |

Note `skills` is fine as a bare string (`"./skills"`) — the two fields do not accept the same shape.

## Suggested fix

```json
"skills": "./skills",
"agents": [
  "./agents/decision-advisor.md",
  "./agents/explainability.md",
  "./agents/kg-assistant.md"
]
```

## Impact

Every Claude Code user hits this on first install; the plugin has never been installable as shipped. The skills/agents content itself is fine — with only this field corrected, the plugin installs and all 17 skills + 3 agents load correctly.

7487:
## Description

`plugins/.claude-plugin/plugin.json` declares `"agents": "./agents"`. Claude Code's plugin schema rejects a bare directory string for `agents` (`Validation errors: agents: Invalid input`), so the bundled plugin cannot be installed at all. Unlike `skills`, which accepts a directory string, `agents` must be an array of `.md` file paths.

This replaces the string with the explicit list of the three agent files, exactly as suggested in the issue.

## Type of Change

## Related Issues

Fixes #1350

## Changes Made

- `plugins/.claude-plugin/plugin.json`: `"agents"` is now `["./agents/decision-advisor.md", "./agents/explainability.md", "./agents/kg-assistant.md"]`
- `tests/test_plugin_manifest.py` (new): guards the manifest shape — `agents` is a non-empty array of existing `.md` paths that stays in sync with the files in `plugins/agents/`, and the `skills` directory exists

## Testing

- [x] Tested locally

Verified with the official validator (Claude Code 2.1.231):

```console
$ claude plugin validate plugins   # before
✘ plugins[0] plugin.json → agents: Invalid input

$ claude plugin validate plugins   # after
✔ Validation passed with warnings
```

(The remaining warning — missing marketplace description — is pre-existing and out of scope.)

```bash
python3 -m unittest tests.test_plugin_manifest -v   # 3 tests, OK
```

The new test fails on the old manifest and passes with this change.

## Documentation

### DIFF
--- a/plugins/.claude-plugin/plugin.json
+++ b/plugins/.claude-plugin/plugin.json
@@ -25,5 +25,9 @@
     "mcp"
   ],
   "skills": "./skills",
-  "agents": "./agents"
+  "agents": [
+    "./agents/decision-advisor.md",
+    "./agents/explainability.md",
+    "./agents/kg-assistant.md"
+  ]
 }
--- a/tests/test_plugin_manifest.py
+++ b/tests/test_plugin_manifest.py
@@ -1,0 +1,50 @@
+"""
+Test for the Claude Code plugin manifest (Issue #1350).
+
+Claude Code's plugin schema requires "agents" to be an array of .md file
+paths (a bare directory string is rejected with "agents: Invalid input"),
+while "skills" may be a directory string. This guards the manifest shape
+so the bundled plugin stays installable.
+"""
+import json
+import unittest
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+MANIFEST = REPO_ROOT / "plugins" / ".claude-plugin" / "plugin.json"
+
+
+class TestPluginManifest(unittest.TestCase):
+    """Validate plugins/.claude-plugin/plugin.json against Claude Code's schema shape."""
+
+    @classmethod
+    def setUpClass(cls):
+        cls.manifest = json.loads(MANIFEST.read_text(encoding="utf-8"))
+        cls.plugin_root = MANIFEST.parent.parent
+
+    def test_agents_is_list_of_md_file_paths(self):
+        agents = self.manifest["agents"]
+        self.assertIsInstance(
+            agents, list,
+            'Claude Code rejects "agents" unless it is an array of .md file paths',
+        )
+        self.assertTrue(agents, "agents list should not be empty")
+        for entry in agents:
+            self.assertIsInstance(entry, str)
+            self.assertTrue(entry.endswith(".md"), f"{entry} is not a .md file path")
+            path = self.plugin_root / entry
+            self.assertTrue(path.is_file(), f"{entry} does not exist under plugins/")
+
+    def test_agents_list_covers_all_agent_files(self):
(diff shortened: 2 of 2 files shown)


KaifAhmad1 (approved):
@7487 Reviewed — small, well-scoped, and correct. I independently verified the core claims by running the actual test suite rather than just reading the diff.

**Test verified to pass with the fix and fail without it:**
```
# against this PR's plugin.json (array form):
Ran 3 tests in 0.002s — OK

# against main's plugin.json ("./agents" bare string):
FAIL: test_agents_is_list_of_md_file_paths — AssertionError: './agents' is not an instance of <class 'list'>
FAIL: test_agents_list_covers_all_agent_files
```
This matches the PR description's claim exactly. I also confirmed the three listed files (`decision-advisor.md`, `explainability.md`, `kg-assistant.md`) are exactly the contents of `plugins/agents/` — no drift, nothing missing, nothing extra. `test_agents_list_covers_all_agent_files` is a real regression guard: adding a 4th agent file later without updating the manifest will now fail CI instead of silently breaking installs again.

No bugs found. The fix is minimal and the new test isn't a rubber-stamp.

**One non-blocking note, not something to hold this up over:** six *other* plugin manifests in `plugins/` (`.cline-plugin`, `.continue-plugin`, `.cursor-plugin`, `.openclaw-plugin`, `.vscode-plugin`, `.windsurf-plugin`) declare `"agents"` the same bare-string way (`"./agents"` / `"../agents"`). I have no way to verify whether those tools' own plugin schemas are equally strict about this — they're separate products with independent validators, and `.codex-plugin/plugin.json` doesn't even have an `agents` key, so these manifests clearly aren't uniform to begin with. This PR correctly scopes itself to the one schema it has validator evidence for (Claude Code's `claude plugin validate`). Worth asking about or filing a follow-up if the other tools turn out to need the same fix, but not a reason to block this.

Approving.

PR #1366, opened 2026-09-01
fix(ci): fail closed on malformed Safety reports

T1mn:
> **Before you submit:** this follow-up is based on the review of [PR #1356](https://github.com/semantica-agi/semantica/pull/1356), which could not be reopened after its branch was rebased and force-pushed.

## Description

The Security Scan workflow already scans `requirements-ci.txt` directly. This follow-up adds an independent fail-closed check so a present but malformed Safety report cannot be rendered as a clean scan.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

- Follow-up to #1356; the overlapping pip-audit replacement was removed because the current `main` branch already contains the canonical CI fix.

## Changes Made

- Validate that `safety-report.json` contains an array-valued `vulnerabilities` field before counting findings.
- Render an explicit invalid-report warning instead of treating malformed data as no findings.
- Keep the existing Safety `--file requirements-ci.txt` scan and the separate `security.yml` pip-audit workflow unchanged.

## Testing

- [x] Tested locally

### Test Commands

```bash
ruby -e 'require "yaml"; YAML.load_file(".github/workflows/security-scan.yml")\
awk '/^          script: \\|/{found=1; next} found { if ($0 !~ /^            /) exit; sub(/^            /, ""); print }' .github/workflows/security-scan.yml | node --check /dev/stdin
```

The valid report case passes, while missing and null `vulnerabilities` fields are rejected by the `jq` guard.

## Documentation

## Breaking Changes

**Breaking Changes**: No

## Checklist

- [x] My code follows the project\s style guidelines

## Additional Notes

This PR intentionally contains only the report-structure hardening that remains independent after the maintainer\s rebase guidance. It does not replace Safety, add a second pip-audit run, or modify the scanner dependency lockfile.


ZohaibHassan16 (approved):
This looks solid. 

I also checked both Qodo findings.

For Guard 2, I tested six report shapes: valid empty, valid with findings, missing key, `null` key, string-typed, and object-typed. It accepts the two valid cases and rejects all four malformed ones.

The finding about the invalid-report warning never rendering was also correct, and this change fixes it. I pulled out `renderSection` and the Safety `parse` callback and ran them in Node with a malformed report; it now renders `Invalid report structure` as expected.

The reason that path was dead before is GitHub Actions' implicit `success()` behavior. An `if:` condition that doesn’t already use `always()` or `failure()` is effectively gated on prior steps succeeding. Before the Safety step had `continue-on-error: true`, an `exit 1` there prevented the Comment step from running at all. The warning text existed, but there was no path that could actually display it.

Using `continue-on-error: true`, then `if: always()` on Upload, Comment, and the new Enforce step is the right fix.

I traced all of the Safety step’s exit paths as well. `SAFETY_SCAN_STATUS` is initialized to `failed` before any guard runs and only changes to `passed` in the actual zero-vulnerability case. That means parse failures, malformed reports, Safety crashes, and other early exits all leave the final gate failed. The Enforce step also treats a missing status as failure, so it doesn’t accidentally fail open.

The guard numbering is consistent throughout the workflow too; I didn’t find any stale references after the renumbering.

Two small non-blocking notes:

* The `awk` command in the PR description doesn’t test what it claims. It stops at the first blank line, so it only checks the first few lines of the script. 
* With `continue-on-error` on the Safety step, that individual step will now show green/neutral in the Actions UI even when Safety fails. The actual failure signal is the Enforce Safety Gate step and the PR comment. That’s fine, but it’s worth knowing when debugging from the step icons.

Approving.

PR #1379, opened 2026-09-02
fix(context): skip redundant vector cascade when erasure's vector leg covers it

### LINKED ISSUE #1375: ErasureCoordinator._erase_memory calls batch_delete without skip_vector=True, causing a redundant best-effort vector deletion after the coordinator has already handled the vector leg
## Summary

`AgentMemory.batch_delete()` calls `delete_memory(memory_id)` for each item, and `delete_memory()` internally calls `_delete_vector_ids()` unless `skip_vector=True` is passed. The `ErasureCoordinator` always handles the vector leg itself before calling `_erase_memory()`, so by the time `batch_delete` runs, the coordinator has already attempted to delete those vector IDs through the vector leg.

This means every memory item's vectors are deletion-attempted twice:

1. By the coordinator's vector leg (reported honestly in the receipt).
2. By `delete_memory()`'s own best-effort cascade (exceptions caught and logged, not reported in the receipt).

## Impact

**No correctness bug in the receipt.** The coordinator's receipt is based on what the coordinator's vector leg reports, not on `delete_memory()`'s internal cascade. The double-delete is idempotent on all backends that support deletion (deleting a nonexistent ID is a no-op or returns success).

**Current visible symptoms:**

* Spurious log warnings on backends that return a failure indicator for deleting a nonexistent ID (the second attempt after the coordinator already deleted it).
* Unnecessary round-trip to the vector store for every memory item deleted.
* When `vector_store=False` is passed to the coordinator, `delete_memory()` still runs its own cascade against `memory.vector_store` silently. The receipt correctly shows `vectors: not_configured`, but actual vector deletions are happening unreported. Whether this is desired behavior is an open question.

## Root cause

`AgentMemory.batch_delete()` has no `skip_vector` parameter. It delegates directly to `delete_memory(memory_id)` with no way for the caller to suppress the internal vector cascade:

```python
def batch_delete(self, memory_ids: List[str]) -> int:
    deleted = 0
    for memory_id in memory_ids:
        if self.delete_memory(memory_id):   # always attempts vector deletion
            deleted += 1
    return deleted
```

## Fix

Add a `skip_vector: bool = False` keyword argument to `batch_delete()`, mirroring the existing parameter on `delete_memory()`:

```python
def batch_delete(self, memory_ids: List[str], *, skip_vector: bool = False) -> int:
    deleted = 0
    for memory_id in memory_ids:
        if self.delete_memory(memory_id, skip_vector=skip_vector):
            deleted += 1
    return deleted
```

The coordinator's `_erase_memory()` would then call `self.memory.batch_delete(memory_ids, skip_vector=True)` since it has already handled (and reported on) the vector deletion.

## Acceptance criteria

* `AgentMemory.batch_delete()` accepts `skip_vector: bool = False`
* Default behavior (`skip_vector=False`) is unchanged — existing callers not affected
* `ErasureCoordinator._erase_memory()` passes `skip_vector=True`
* No duplicate vector deletion log noise when the coordinator is used
* Test: when coordinator runs with a cooperative vector store, the vector IDs are deleted exactly once (not twice)

## Related

* Raised during PR #1027 review as a MEDIUM finding

7487:
## Description

`ErasureCoordinator`'s vector leg deletes (and reports on) every memory-owned embedding before the memory sweep runs, but `AgentMemory.batch_delete()` had no way to suppress `delete_memory()`'s own best-effort vector cascade — so every vector id was deletion-attempted twice: a redundant round-trip per item, and spurious warnings on backends that flag missing ids.

This adds a `skip_vector: bool = False` keyword to `batch_delete()` (mirroring `delete_memory()`) and has `_erase_memory()` pass it through.

One adjustment to the fix as sketched in the issue: the skip is **conditional** — `skip_vector=True` only when `coordinator.vector_store is memory.vector_store` (the default binding). With a separate coordinator store, or the vector leg disabled, `delete_memory()`'s cascade is still the only cleanup `memory.vector_store` gets; passing `skip_vector=True` unconditionally would orphan those embeddings, and `TestSeparateVectorStoreHandling::test_separate_vector_store_only_handles_coordinator_store` pins exactly that (it fails with the unconditional version).

## Type of Change

## Related Issues

Fixes #1375

## Changes Made

- `AgentMemory.batch_delete()` accepts keyword-only `skip_vector: bool = False`, forwarded to `delete_memory()`; default behavior unchanged
- `ErasureCoordinator._erase_memory()` computes `skip_vector` once (vector leg configured *and* acting on the memory-bound store) and passes it to `batch_delete()`
- Documented the extended `batch_delete` surface in the coordinator's `memory:` docstring
- Updated the two test fakes in `test_erasure_coordinator.py` to accept the new keyword
- New tests: memory-owned vector ids are deleted exactly once through the coordinator (fails on `main` with 2 delete calls); `batch_delete` still cascades to vectors by default

## Testing

- [x] Tested locally

```
$ pytest tests/context/test_erasure_coordinator.py tests/context/test_agent_memory_find_by_entity.py tests/context/test_agent_memory_markdown.py -q
113 passed, 1 skipped, 3 subtests passed
```

The new `test_memory_owned_vector_ids_are_deleted_exactly_once` fails on `main` (the store records two `delete_vectors` calls) and passes with this change.

## Documentation

- [x] No documentation changes needed (docstrings updated in place)

## Breaking Changes

**Breaking Changes**: No

Only additive: a keyword-only parameter with the old behavior as default. Custom memory-like objects passed to the coordinator now receive `skip_vector=` on `batch_delete` (noted in the coordinator docstring); a fake that doesn't accept it surfaces as a failed memory leg in the receipt rather than a silent pass.

## Additional Notes

The `vector_store=False` case from the issue's "open question" (memory's cascade still running silently) is deliberately left as-is — this PR only removes the double-delete in the configurations where the vector leg genuinely covers the same store.


zeekay (changes requested):
The reasoning here is the best part of the PR and I'd like to see it land — it just doesn't merge any more.

`mergeable: CONFLICTING`, with conflicts in both `semantica/context/agent_memory.py` and `semantica/context/erasure.py`. I tried merging onto current `main` to test it and couldn't, so nothing below is from a test run.

**What I'd keep exactly as written.** The distinction the comment draws — that store identity is not coverage — is the right one, and it's the part a naive version of this optimization would get wrong. Suppressing the memory-side cascade only when the vector leg *both* enumerated everything *and* had its delete accepted is correct; either half alone would silently drop embeddings on a GDPR path, which is the worst possible place to be clever. Threading a `complete` flag out of `_all_vector_ids` so a part-way enumeration failure disables the optimization rather than being papered over is the detail that makes it safe, and `test_rejected_vector_deletion_does_not_suppress_the_cascade` pins the case that matters.

Dropping `_vector_ids` bookkeeping in `batch_delete` even when skipping the cascade is right too, and the docstring explains why it differs from `delete_memory`'s staging use of the same flag — a stale mapping surviving into `save()`/`load()` would be a nasty one to debug.

**One thing to look at while rebasing.** `_accepts_skip_vector` probes with `inspect.signature(...).bind(...)`, and anything with a `(*args, **kwargs)` signature binds successfully:

```
MagicMock            -> binds, treated as accepting skip_vector
def wrapper(*a, **k) -> binds, treated as accepting skip_vector
```

For a `functools.wraps` decorator that's fine, since `signature` follows `__wrapped__` to the real signature. For a hand-rolled passthrough wrapper that forwards `**kwargs`, also fine. It only misfires on a wrapper that accepts `**kwargs` and drops them — the coordinator would think it suppressed the cascade when the callee ignored the keyword and cascaded anyway. That direction is harmless (a redundant delete, which is what this PR is removing) rather than dangerous, so I'd call it acceptable — but the docstring currently says unintrospectable callables get the conservative path, and `(*args, **kwargs)` is the case where that isn't true. Worth one more sentence so the next reader knows the probe's actual boundary.

Worth noting for whoever sequences these: #1073 and #1074 also touch `agent_memory.py`, so the merge order between them and this will need deciding regardless of who rebases first.

Rebase onto current `main` and I'll review it properly against a test run.

ZohaibHassan16 (approved):
Excellent implementation. Good to merge

PR #1390, opened 2026-09-02
fix(ci): migrate security-scan from Safety to pip-audit

### LINKED ISSUE #1389: security-scan: Safety's report generation crashes intermittently, independent of which package is flagged
`security-scan`'s Safety step is failing intermittently for reasons unrelated to the code being scanned.

This is separate from #1384, which covered a specific torchvision CVE. The issue here is the recurring crash in Safety itself.

## What's happening

Safety 3.8.1 sometimes throws an unhandled exception while generating its report:

```text
Unhandled exception happened: 'torchvision'
```

The failure is non-deterministic against the same dependency tree.

* Run [[33605788249](https://github.com/semantica-agi/semantica/actions/runs/33605788249)](https://github.com/semantica-agi/semantica/actions/runs/33605788249) completed normally, reported the torchvision finding, and filtered it through `IGNORED_VULN_IDS`.
* Run [[33607331537](https://github.com/semantica-agi/semantica/actions/runs/33607331537)](https://github.com/semantica-agi/semantica/actions/runs/33607331537) crashed on the same package before producing a report.
* Run [[33612894549](https://github.com/semantica-agi/semantica/actions/runs/33612894549)](https://github.com/semantica-agi/semantica/actions/runs/33612894549), from unrelated PR #1229, failed the same way on torchvision and passed on retry.

The same class of failure previously happened with `cuda-toolkit`, as documented in the history of `.github/workflows/security-scan.yml`. This doesn't appear to be specific to torchvision; Safety has now hit the same failure pattern on different packages over time.

`IGNORED_VULN_IDS` can't help with this case because that filtering happens after Safety writes its report. When Safety crashes during report generation, there is nothing to filter.

## Impact

These failures make unrelated PRs fail `security-scan` until someone manually reruns the job. The same problem has already shown up multiple times across both `cuda-toolkit` and `torchvision`.

## Proposed direction

`pip-audit` already runs in `security.yml` against the same dependency tree and hasn't shown the same reliability problem.

A recent run against the current `requirements-ci.txt`, [[33388978843](https://github.com/semantica-agi/semantica/actions/runs/33388978843)](https://github.com/semantica-agi/semantica/actions/runs/33388978843), completed successfully in 34 seconds with:

```text
No known vulnerabilities found
```

I also checked `torchvision==0.28.0` and `cuda-toolkit==13.0.3.0` directly with `pip-audit`; neither currently reports a vulnerability because those findings aren't present in OSV.

Moving this scan to `pip-audit` would remove the intermittent Safety failure and, for the current dependency set, also make the existing `IGNORED_VULN_IDS` workarounds unnecessary.

Migration is in progress.

ZohaibHassan16:
Closes #1389.

Safety's report-generation code has been failing intermittently on packages with live vulnerability matches. The same pattern previously affected `cuda-toolkit` and is now happening with `torchvision`. Runs against the same `requirements-ci.txt` can either succeed or crash, so this isn't tied to a change in the dependency tree.

`IGNORED_VULN_IDS` can't address this failure because Safety crashes before it writes a report.

This replaces the Safety step in `security-scan.yml` with `pip-audit`, which is already used successfully in `security.yml` against the same dependencies.

`security.yml` is removed as part of the same change because `security-scan.yml` now covers everything it did, with a broader trigger set and the same scheduled scan, while also running Bandit, Semgrep, and PR reporting.

`IGNORED_VULN_IDS` is now empty. `pip-audit` uses OSV, which currently doesn't contain either of the CVEs that Safety was flagging for `cuda-toolkit` and `torchvision`. I verified that directly and against a real CI run.

Local verification included:

* YAML parsing
* syntax-checking the embedded JavaScript
* testing the report-structure and vulnerability-count handling against six report shapes, including clean, vulnerable, and malformed reports
* running `verify-action-pins.sh`, which passes with 47 action references, down from 49 after removing `security.yml`


KaifAhmad1 (approved):
## Final review

Re-reviewed after commit b5199ae6. All three issues from my initial review are fixed and verified:

1. **Skipped-dependency false failure** — confirmed against `pip-audit`'s own JSON formatter source (`_format_dep`) that an unauditable dependency is reported as `{"name": ..., "skip_reason": ...}` with no `vulns` key. Guard 2, the `VULNS` count, the vulnerability-detail listing, and the PR-comment JS parser now all handle that shape instead of treating it as an invalid report and hard-failing the job. Verified the jq queries and JS `parse()` logic directly against synthetic normal/skipped/malformed report fixtures (not just read the diff) — correct behavior in all three cases, and a genuinely malformed report still correctly fails.
2. **`workflow_dispatch` restored** on `security-scan.yml` — on-demand security audits are possible again after `security.yml`'s removal.
3. **SECURITY.md updated** — no longer references the deleted `security.yml` or describes Safety as an active scanner.

Confirmed in real CI, not just locally: the post-fix "Security Scan Results" comment (10:59Z) shows `pip-audit — dependency vulnerabilities: ✅ No findings.` and `Bandit: ✅ No findings.` — the migration works end-to-end against the actual dependency tree. `verify-action-pins.sh` still passes (47/47) and all 14 required/optional checks are green, including `security-scan` itself.

Semgrep's 57 informational findings are pre-existing and non-blocking per policy (unrelated to this PR's scope) — nothing here needs to hold up the merge.

Approving.

Issue #1374, opened 2026-09-01
ErasureCoordinator: FAISS, Milvus, and Weaviate vector backends have no delete capability erasure cannot be completed on them


Sameer6305:
<html>
<body>
<html><head></head><body><h2>Summary</h2><p><code inline="">ErasureCoordinator</code> (merged in #1018) correctly detects that three vector backends expose no delete surface and reports <code inline="">status: "unsupported"</code> in the receipt rather than claiming a success it did not achieve. This means <code inline="">receipt.complete</code> is always <code inline="">False</code> when any of these backends is in use, even after a fully successful erasure of the graph and memory legs.</p><p>This issue tracks adding real delete capability to those backends so the coordinator can report <code inline="">erased</code> end-to-end.</p><h2>Affected backends</h2>
Backend | File | Current state
-- | -- | --
FAISS | semantica/vector_store/faiss_store.py | No delete or delete_vectors method. FAISSStore.count() docstring explicitly notes "FAISSStore does not implement vector deletion."
Milvus | semantica/vector_store/milvus_store.py | No delete or delete_vectors method.
Weaviate | semantica/vector_store/weaviate_store.py | No delete or delete_vectors method.

[...]

PR #1391, opened 2026-09-02
feat(milvus): add delete_vectors to MilvusStore

pkupt:
Milvus half of #1374: adds `delete_vectors(vector_ids)` to `MilvusStore` so ErasureCoordinator can erase embeddings on a Milvus backend. Weaviate follows in a separate PR, FAISS is left out for now since flat indices can't delete in place.

Ids go through the existing `_format_milvus_value` escaper before the `id == "..."` / `id in [...]` delete expression is built, and the backend's delete count is returned so "nothing matched" stays distinguishable from a failure. Deleting without an initialized collection raises ProcessingError before touching the backend.

10 tests cover both expression shapes, id escaping, empty ids, the missing-collection and backend-error paths, plus two integration tests that bind MilvusStore as a VectorStore backend and check the erasure receipt reports the vector leg erased. All of it is mock-based, and the vector_store, context and CLI suites show failure sets identical to main. I couldn't run a live Milvus locally, so the delete path is verified through mocks and the erasure integration tests; a live run would confirm the wire behavior end to end.


zeekay (approved):
Fills a real gap, and reuses the right machinery to do it.

`MilvusStore` was the only bound backend with neither `delete()` nor `delete_vectors()` — qdrant (line 664), weaviate (419) and pinecone (217, 654) all have one — so `VectorStore.delete_vectors` (vector_store.py:786-795) raised `NotImplementedError` for Milvus, and `ErasureCoordinator` had to route it to `STATUS_UNSUPPORTED`. That's the leg this closes, and returning `Dict[str, Any]` matches what weaviate already returns rather than inventing a third shape.

The escaping is the part I was most inclined to be suspicious of, and it's right: `_format_milvus_value` (milvus_store.py:56) is the same helper `filter_by_metadata` already uses, so an id containing a quote or backslash goes through one escaping path in this module, not two that can drift. `test_delete_escapes_quote_and_backslash_in_id` pins it.

Test runs against current `main` (merged, the branch is behind), unfiltered so nothing marked `integration` is skipped:
- `pytest tests/vector_store/` — 362 passed, 84 skipped, 7 xfailed, against 352 passed on `main`
- `pytest tests/context/test_erasure_coordinator.py` — 54 passed, 3 subtests

`MilvusErasureIntegrationTest` is the test I'd have asked for if it weren't already here — it binds a real `MilvusStore` behind the `VectorStore` facade and asserts the receipt comes back `erased` with `backend == "milvus"`, which is what actually proves the coordinator reaches the new method rather than just that the method exists.

Two notes, neither blocking:

- `id in [...]` is built from the full id list in one expression. Milvus caps expression length, so a large erasure could exceed it. Nothing in the tree passes enough ids for that today; batching is the eventual fix if someone hits it.
- Milvus deletes aren't immediately visible to a search under the default consistency level, and nothing in `milvus_store.py` flushes. `STATUS_ERASED` is already documented in `erasure.py:56-61` as "the store accepted the ids", not "confirmed gone", so the receipt isn't overclaiming — just worth knowing that the Milvus leg is the weakest of the backends in that respect.

Sameer6305 (approved):
### Review

Reviewed this PR against #1374 and the existing vector-store/erasure architecture.

This PR adds native `delete_vectors()` support to `MilvusStore`, using Milvus's delete expression API and returning the backend delete count. This correctly closes the Milvus gap so `ErasureCoordinator` can report the vector leg as `erased`.

The implementation has good coverage around single/multiple IDs, escaping, empty IDs, missing collections, backend errors, and erasure integration. We also independently verified the ID escaping and coordinator routing rather than relying solely on the PR description.

During review, we found a few documentation/test gaps and fixed them:

* Updated stale erasure documentation that still listed Milvus as unsupported.
* Updated the corresponding stale test documentation.
* Added a direct `VectorStore → MilvusStore.delete_vectors()` facade test.

The core implementation remains unchanged after review, and the relevant test suites pass with no regressions attributable to this PR.

**Approved.**

PR #1392, opened 2026-09-02
feat(weaviate): add delete_vectors to WeaviateStore

pkupt:
Weaviate half of #1374. Adds `delete_vectors(vector_ids)` so ErasureCoordinator can actually erase on a Weaviate backend. Milvus already landed in #1391, FAISS stays out since flat indices can't delete in place.

Ids are the object uuids `store_vectors` returns, handed one at a time to `collection.data.delete_by_id`. The count reflects objects that were really removed, since that call returns False for a missing uuid instead of raising, so the erasure receipt keeps an honest count in `backend_result`.

10 tests cover single and multi id deletes, counting that skips not-found uuids, empty ids, the missing-collection path, and two integration tests binding WeaviateStore as a backend. The vector_store suite is identical to main otherwise. I couldn't run Weaviate locally, so this is logic-level coverage; a live run would confirm the wire behavior.


ZohaibHassan16 (approved):
Finding #2 doesn't reproduce against the pinned client. I checked the actual source for `weaviate-client==4.16.2`:

```python
def delete_by_id(self, uuid: UUID) -> executor.Result[bool]:
    def resp(res: Response) -> bool:
        return res.status_code == 204
    return executor.execute(..., status_codes=_ExpectedStatusCodes(ok_in=[204, 404], ...))
```

`delete_by_id()` returns a `bool`: `204` gives `True`, `404` gives `False`, and neither case raises. So this is correct:

```python
if data.delete_by_id(vector_id):
    deleted += 1
```

Finding #1 looks resolved as well. `VectorStore.delete_vectors` checks for `backend.delete` first and falls back to `delete_vectors`. `WeaviateStore` doesn't implement `delete`, so the call does reach `delete_vectors`. `test_facade_delete_vectors_forwards_to_weaviate` covers that path.

All 11 new tests pass. The full `tests/vector_store/` suite has the same existing `11 failed / 21 errors` with and without this PR. The passing count goes from 309 to 320, which is exactly the 11 new tests, so there don't appear to be any unrelated regressions.


Approving.

PR #1394, opened 2026-09-02
fix(mcp): complete persistence and setup fixes

### LINKED ISSUE #1134: bug(mcp): progress bar writes to stdout and corrupts JSON-RPC; SEMANTICA_KG_PATH never persists; 3 broken setup steps in mcp/README.md

Sameer6305:
Closes #1134

- Fix root MCP graph loading to use load_from_file()
- Persist MCP mutations when SEMANTICA_KG_PATH is configured
- Cover persistence for both MCP server implementations
- Fix MCP installation and Claude Code setup documentation
- Fix claude mcp add invocation and document PYTHONPATH
- Add end-to-end MCP stdio JSON-RPC regression coverage

### DIFF
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -8,8 +8,9 @@ Connects Claude Code, Cursor, Windsurf, Cline, Continue, VS Code (GitHub Copilot
 ## Quick start
 
 ```bash
-# From the repo root
-pip install -e ".[mcp]"
+# From the repo root — no extra install flag needed; the root mcp/ package is
+# part of the repository and does not require an external MCP SDK.
+pip install -e .
 
 # Test the server (type a JSON-RPC request, press Enter)
 python -m mcp
@@ -89,7 +90,14 @@ python -m mcp [--debug]
 
 ## Per-tool configuration
 
-### Claude Code (`~/.claude/settings.json`)
+### Claude Code (`~/.claude.json` or `.mcp.json`)
+
+Claude Code supports two MCP configuration scopes:
+
+- **User scope** — `~/.claude.json` applies across all projects for your user account.
+- **Project scope** — `.mcp.json` in your project root applies only to that project.
+
+Both files use the same `mcpServers` structure:
 
 ```json
 {
@@ -97,15 +105,33 @@ python -m mcp [--debug]
     "semantica": {
       "command": "python",
       "args": ["-m", "mcp"],
-      "cwd": "/path/to/semantica"
+      "env": {
+        "PYTHONPATH": "/path/to/semantica"
+      }
     }
   }
 }
 ```
 
-Or use the plugin bundle:
+> **Why `PYTHONPATH`?**  The root `mcp/` package is intentionally not included in
+> the installed wheel, so `python -m mcp` only works when the repository is on
+> Python's import path.  Setting `PYTHONPATH` here ensures this works regardless
+> of the working directory Claude uses when it launches the server.
+
+Or add it via the CLI (user scope):
+
 ```bash
-claude mcp add semantica python -m mcp --cwd /path/to/semantica
+claude mcp add --scope user semantica \
+  -e PYTHONPATH=/path/to/semantica \
+  -- python -m mcp
+```
+
+Or for project scope (omit `--scope user`):
+
+```bash
+claude mcp add semantica \
+  -e PYTHONPATH=/path/to/semantica \
+  -- python -m mcp
 ```
 
 ---
@@ -216,7 +242,7 @@ Add to your Q Developer MCP config:
 
 | Variable | Default | Description |
 |---|---|---|
-| `SEMANTICA_KG_PATH` | *(in-memory)* | Path to persist/load the graph (JSON file) |
+| `SEMANTICA_KG_PATH` | *(in-memory only)* | Path to a JSON file used to **load** the graph on startup and **persist** mutations (record decisions, add entities/relationships) back to disk after each change. When unset the graph lives in memory only and is lost when the server exits. |
 
 ---
 
--- a/mcp/session.py
+++ b/mcp/session.py
@@ -33,7 +33,7 @@ def get_graph() -> Any:
         kg_path = os.environ.get("SEMANTICA_KG_PATH", "").strip()
         if kg_path and os.path.exists(kg_path):
             try:
-                _graph.load(kg_path)
+                _graph.load_from_file(kg_path)
                 log.info("Graph loaded from %s", kg_path)
             except Exception as exc:
                 log.warning("Could not load graph from %s: %s", kg_path, exc)
--- a/mcp/session.py
+++ b/mcp/session.py
@@ -16,6 +16,13 @@ log = logging.getLogger("semantica.mcp.session")
 
 _graph: Optional[Any] = None
 
+# Tracks whether the last graph initialisation successfully loaded the
+# configured SEMANTICA_KG_PATH file.  When True (or no path was configured)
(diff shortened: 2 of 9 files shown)


ZohaibHassan16 (approved):
Atomic writes look correct. The temp file is created in the destination directory, `fsync` happens before the replace, and failures clean up after themselves. That keeps `os.replace` on the same filesystem and matches the FAISS sidecar fix.

I also tested the failure paths against real files on disk. With a corrupt persistence file, `handle_add_entity` returns the persistence error and leaves both the graph and the file unchanged:

```text
result: {'error': 'Persistence blocked: ...'}
node count unchanged: True
kg_path untouched: not valid json
```

I got the same behavior when forcing `save_to_file` to fail after a mutation; the in-memory change is rolled back correctly.

I specifically checked the rollback logic for duplicate edges since that's the case most likely to expose a problem with `reversed(graph.edges)`. With an existing `a→b:knows` edge, adding a second identical edge and then rolling back removes only the new edge. The original stays in place.

Both server implementations are covered. `mcp/` and `semantica/mcp_server/` use the same persistence pattern, and `TestMCPServerMutationPersistence` exercises save/reload round trips against the second implementation directly.

The new stdio test is also exercising the real path rather than mocking it. It starts `python -m mcp`, sends JSON-RPC over stdin/stdout, and verifies that every output line is valid JSON. Both cases pass, including a `tools/call` that goes through the progress-tracker path from #1134.

The new tests are 12/12, and the full decision-persistence suite is 56/56.


Approving.

PR #1395, opened 2026-09-02
fix(context): vector_store=False must suppress AgentMemory's internal vector cascade in ErasureCoordinator

### LINKED ISSUE #1378: ErasureCoordinator with vector_store=False silently runs AgentMemory's internal vector cascade without reporting it — receipt.complete can be True while vectors survive
## Summary

When `ErasureCoordinator` is constructed with `vector_store=False`, the vector leg is disabled and the receipt reports `stores["vectors"]["status"] == "not_configured"`. `receipt.complete` is `True` because `not_configured` does not indicate residual data.

However, `AgentMemory.batch_delete()` → `delete_memory()` still runs its own best-effort vector cascade against `memory.vector_store`. If that cascade fails (e.g. the backend refuses the delete), the failure is caught, logged as a warning, and `delete_memory()` still returns `True`. The coordinator never sees this failure. `receipt.complete` reports `True` while a vector from a memory item may still be live.

## Reproduction

```python
class _RefusingStore:
    backend = "qdrant"

    def store_vectors(self, vectors, metadata=None, **options):
        return ["vec-0"]

    def delete_vectors(self, vector_ids, **options):
        return False   # refuse


memory = AgentMemory(vector_store=_RefusingStore())
memory.store("note", entities=[{"id": "e1"}], skip_graph=True)

# Disable the coordinator's vector leg explicitly
receipt = ErasureCoordinator(
    graph=graph, memory=memory, vector_store=False
).erase_entity("e1")

# This asserts True — but "vec-0" is still in _RefusingStore
assert receipt.complete   # passes — misleading
```

## The tension

The current behavior is intentional per the PR design: when `vector_store=False` the coordinator reports only on stores it was given. The question is whether `vector_store=False` should mean:

* **"I am handling vectors myself out of band"** → current behavior, `complete` can be `True` even if memory's cascade failed silently. Documented nowhere.
* **"Disable the coordinator's vector leg AND suppress memory's cascade"** → would require passing `skip_vector=True` to `batch_delete`.
* **"I know vectors will not be cleanly erased; I accept `complete=True`"** → caller's deliberate choice, but should be documented explicitly.

## What needs to happen

At minimum, document the behavior precisely in the `ErasureCoordinator` docstring so callers understand what `vector_store=False` actually guarantees.

Optionally, if the intent is to give callers a way to fully disable all vector activity (not just the coordinator's leg), the fix from Issue #XXXX (`batch_delete(skip_vector=True)`) would enable:

```python
# _erase_memory: when coordinator has no vector store, suppress memory's cascade too
skip = self.vector_store is None
removed = self.memory.batch_delete(memory_ids, skip_vector=skip)
```

## Acceptance criteria

* The behavior of `vector_store=False` is explicitly documented in the `ErasureCoordinator` class docstring, including what it does and does not guarantee about `memory.vector_store`
* Decision: does `vector_store=False` suppress `memory.vector_store` activity or not? Whatever is decided, it is implemented consistently and tested
* A test pins the chosen behavior so it cannot silently regress

## Related

* Depends on or relates to Issue #XXXX (`batch_delete skip_vector`)
* Raised during PR #1027 review as a MEDIUM finding

Harsh4r0ra:

## Description

`ErasureCoordinator(vector_store=False)` correctly disabled the coordinator's own vector leg (the receipt reported `status="not_configured"`), but `AgentMemory.batch_delete()` -> `delete_memory()` still ran its own best-effort vector delete cascade against `memory.vector_store` internally. That cascade catches any failure, logs a warning, and still returns `True`, so `receipt.complete` could read `True` while an embedding was still live in the vector store.

`vector_store=False` is meant to mean "no vector activity at all," not just "none reported by the coordinator." This PR closes that gap by threading a `skip_vector` flag from `ErasureCoordinator` down into `AgentMemory.batch_delete()` when the vector leg was explicitly disabled.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #1378
Fixes #1378

## Changes Made

- `semantica/context/agent_memory.py`: `AgentMemory.batch_delete()` gains a keyword-only `skip_vector: bool = False` parameter, threaded through to the existing `delete_memory(memory_id, skip_vector=...)` parameter (already used elsewhere in the file, just not exposed on the batch path).
- `semantica/context/erasure.py`: `ErasureCoordinator.__init__` now tracks `self._vector_leg_disabled = vector_store is False`, distinct from `vector_store` resolving to `None` because no vector store exists anywhere. `_erase_memory()` calls `batch_delete(memory_ids, skip_vector=True)` when the vector leg is explicitly disabled. Added a `Note:` block to the class docstring documenting the contract.
- `tests/context/test_erasure_coordinator.py`: rewrote `test_vector_store_false_disables_vector_leg_entirely` to assert the memory-owned vector is left untouched and the backend's delete method is never called (it previously asserted the buggy behavior). Added `test_vector_store_false_regression_refusing_backend_never_called`, reproducing the exact issue repro and pinning `delete_calls == 0`.

## Testing

- [x] Tested locally

### Test Commands

```bash
pytest tests/context/test_erasure_coordinator.py
pytest tests/context/

### DIFF
--- a/semantica/context/agent_memory.py
+++ b/semantica/context/agent_memory.py
@@ -588,16 +588,15 @@ class AgentMemory:
             return False
 
         # Remove from vector store unless a caller is staging an atomic local update.
-        if not skip_vector:
-            if self.vector_store:
-                try:
-                    vector_ids = list(self._vector_ids.get(memory_id, [])) or [
-                        memory_id
-                    ]
-                    self._delete_vector_ids(vector_ids)
-                except Exception as e:
-                    self.logger.warning(f"Failed to delete from vector store: {e}")
-            self._vector_ids.pop(memory_id, None)
+        if not skip_vector and self.vector_store:
+            try:
+                vector_ids = list(self._vector_ids.get(memory_id, [])) or [memory_id]
+                self._delete_vector_ids(vector_ids)
+            except Exception as e:
+                self.logger.warning(f"Failed to delete from vector store: {e}")
+        # Bookkeeping runs unconditionally: a skip_vector delete still removes the
+        # item, so leaving its tracked ids behind would orphan them permanently.
+        self._vector_ids.pop(memory_id, None)
 
         memory_item = self.memory_items[memory_id]
 
--- a/semantica/context/agent_memory.py
+++ b/semantica/context/agent_memory.py
@@ -1588,12 +1588,16 @@ class AgentMemory:
                     memory_ids.append(memory_id)
         return memory_ids
 
-    def batch_delete(self, memory_ids: List[str]) -> int:
+    def batch_delete(self, memory_ids: List[str], *, skip_vector: bool = False) -> int:
         """
         Batch delete.
 
         Args:
             memory_ids: List of memory IDs to delete
+            skip_vector: If True, skip each item's own vector-store cascade
+                (see ``delete_memory``). A caller that is already erasing these
+                ids' vectors itself passes this to avoid a redundant,
+                best-effort delete against the vector store.
 
         Returns:
             Number of memories deleted
@@ -1603,7 +1607,7 @@ class AgentMemory:
         """
         deleted = 0
         for memory_id in memory_ids:
-            if self.delete_memory(memory_id):
+            if self.delete_memory(memory_id, skip_vector=skip_vector):
                 deleted += 1
         return deleted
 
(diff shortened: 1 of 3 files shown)


ZohaibHassan16 (approved):
The erasure suite is 51/51. A full `tests/context/` run has the same pre-existing `27 failed / 85 errors` before and after this change, with the four new tests passing.

Approving. This is nicely scoped.

PR #1397, opened 2026-09-02
feat(ontology): add a CI-friendly quality gate

### LINKED ISSUE #1393: [FEATURE] Add a CI-friendly ontology quality gate
## Problem Statement

Generated ontologies can be structurally valid while still containing issues that are difficult to catch before export or deployment, such as orphan schema elements, unresolved relationship endpoints, invalid domain/range references, and unexpected coverage gaps. Semantica has validation and evaluation building blocks, but no single deterministic result that can be used as a release or CI gate.

## Proposed Solution

Add a focused ontology quality report and quality gate, building on the existing ontology validator and evaluator. The first milestone should:

- check unresolved relationship endpoints, orphan classes/properties, domain/range consistency, and basic coverage
- return structured issues, counts, severities, and a pass/fail result against configurable thresholds
- provide a small programmatic API that is suitable for CI
- leave dashboards, automatic fixes, large benchmark suites, and version diffs for follow-up work

## Alternatives Considered

Users can compose the existing checks themselves, but that leads to duplicated logic and inconsistent CI behavior. A full evaluation or KG quality framework would be useful, but is too broad for a first implementation.

## Use Cases

1. Check a generated ontology before export or deployment.
2. Run the same quality gate in CI for every ontology change.
3. Track quality issues consistently as the ontology evolves.

## Impact Assessment

- **Who would benefit?** All users generating, validating, or deploying ontologies.
- **Priority**: High
- **Breaking Changes**: No
- **Dependencies**: Existing ontology validator and evaluator APIs; no new runtime dependency is required for the first milestone.

## Implementation Ideas

The exact names can follow existing project conventions. An illustrative API could look like:

```python
report = ontology_quality_check(ontology, thresholds={"coverage": 0.95})
assert report.passed
```

The report should be machine-readable and easy to render in a later CLI or CI integration.

## Additional Context

This is intentionally a smaller first milestone related to the broader evaluation and KG quality directions discussed in #228 and #229. Ontology diff and migration can build on the resulting report in a separate change (#321).

## Contribution

- [x] I can help with testing

T1mn:
## Description

Adds a deterministic ontology/KG quality report and CI-friendly pass/fail gate for Issue #1393. It reuses the existing OntologyValidator, OntologyEvaluator, and GraphValidator without adding runtime dependencies.

## Type of Change

- [ ] Breaking change (bug fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Closes #1393
Fixes #

## Changes Made

- Added machine-readable quality findings with severities, metrics, statistics, and configurable thresholds.
- Added deterministic checks for ontology structure, class/property coverage, domain/range references, and KG relationship endpoints.
- Exposed the quality gate through the ontology package and `OntologyEngine`, with API documentation and examples.

## Testing

- [x] Tested locally

### Test Commands

```bash
PYTHONPATH=. pytest -q tests/ontology/test_ontology_quality_gate.py tests/ontology/test_ontology_advanced.py tests/ontology/test_ontology_classes.py tests/ontology/test_ontology_normalized_properties.py tests/ontology/test_ontology_relationship_endpoints.py tests/ontology/test_ontology_property_name_collisions.py tests/ontology/test_shacl_target_namespace.py tests/ontology/test_vocabulary.py tests/kg/test_graph_validator.py
# 63 passed, 4 skipped

uv run --no-project --with build --with 'setuptools==84.0.0' --with 'wheel==0.48.0' python -m build --no-isolation
# Successfully built the source distribution and wheel
```

## Documentation

## Breaking Changes

**Breaking Changes**: No

## Checklist

## Additional Notes

This first version reports findings only; it does not auto-fix data or add dashboard/benchmark integration.


KaifAhmad1 (changes requested):
@T1mn thanks for the follow-up round — could you take a look at these four? They're new findings from re-running the current branch, separate from the 8 already resolved earlier.

Reviewed by checking out the branch into a throwaway worktree and exercising `OntologyQualityGate` directly, not just reading the diff. Found 4 bugs, all reproduced live against this branch.

**1. Crash on `None` entities/relationships** (`quality_gate.py:328`, most severe)
`_read_graph()` correctly detects an invalid `graph_data` shape (e.g. `{'entities': None, 'relationships': []}`), records an `INVALID_GRAPH` issue, and returns `([], [])` — but `check()` calls `self._check_graph(graph, issues)` unconditionally afterward anyway. `_prepare_graph_for_validation()` then does `enumerate(graph.get('entities', []))` on the still-`None` value with no type guard and raises `TypeError: 'NoneType' object is not iterable`. For a "CI-friendly" gate whose whole point is to degrade to a reported issue instead of crashing, this defeats the purpose on exactly the malformed-input case it was built to catch.

Repro:
```python
gate.check(ontology, graph_data={"entities": None, "relationships": []})
# TypeError: 'NoneType' object is not iterable
```

**2. Content-based dedup silently undercounts real errors** (`quality_gate.py:779`)
`_add_issue()` dedups by `(code, element_id, message)`. Two genuinely distinct dangling-edge problems that happen to render identical message text collapse into a single issue, undercounting `error_count`. Confirmed: a graph with two separate relationships both dangling to the same missing target id produces 2 real `DANGLING_EDGE` findings from `GraphValidator`, but they collapse to 1 `UNRESOLVED_RELATIONSHIP_ENDPOINT` issue. With `thresholds={"max_errors": 1}`, `report.passed` comes back `True` even though there are 2 real distinct errors — the gate's own pass/fail verdict is wrong.

**3. Shared stateful evaluator leaks state across `check()` calls** (`quality_gate.py:365`, `engine.py`, `competency_questions.py`)
`OntologyEngine` builds one `OntologyEvaluator` and reuses it across calls into the quality gate. `CompetencyQuestionsManager.add_question()` never clears prior questions, so `competency_question_coverage`/`completeness_score` silently accumulate across repeated `quality_check()` calls on the same engine instance. Confirmed: calling `engine.quality_check(ontology1, competency_questions=["Q1"])` then `engine.quality_check(ontology2, competency_questions=["Q2"])` evaluates the second call against both Q1 and Q2 — `manager.questions` grows 1→2. This corrupts the "deterministic, suitable for CI" coverage metric for every call after the first on a long-lived engine.

**4. `"type": null` treated more leniently than a missing type** (`quality_gate.py:244`)
`prop.get("type", "")` returns `None` (not the `""` default) when the key exists but is `null`, so a property like `{"name": "weird", "type": None, "domain": "Person", "range": "string"}` downgrades from the ERROR-level `MISSING_PROPERTY_TYPE` to a WARNING-level `UNKNOWN_PROPERTY_TYPE` ("unknown type 'none'"). Under the default `max_errors: 0.0` threshold, an effectively typeless, JSON-null-typed property passes the gate when it should fail it the same as an omitted type would.

---

All four were reproduced directly against this branch, not inferred from reading. Given #1 is an unhandled crash and #2–4 each independently corrupt the gate's own pass/fail verdict (which is the entire point of a "CI-friendly quality gate"), requesting changes before merge.

KaifAhmad1 (approved):
Re-verified all four follow-up findings live against 103ab049 in a throwaway worktree (not just reading the diff): the None-entities crash is gone (now reports INVALID_GRAPH), duplicate dangling-edge issues are no longer deduped away (2 distinct problems -> 2 distinct issues, correct pass/fail verdict), competency-question evaluator state is isolated per call (confirmed the shared manager's question list stays empty across sequential quality_check() calls), and a null property type is now correctly treated as MISSING_PROPERTY_TYPE. Full tests/ontology/ + tests/kg/test_graph_validator.py suite passes clean (160/160). Also spot-checked engine.py, graph_validator.py, and the docs example against the real API with no issues found. LGTM.

PR #1410, opened 2026-09-03
ci: report required checks for docs-only PRs

Sameer6305:
Fix required CI checks for docs-only PRs.

- Keep build and security-scan required checks intact.
- Trigger both workflows for PRs, including docs-only changes.
- Skip expensive jobs for docs-only PRs while reporting their check status.
- Preserve full build/security scans for source or mixed changes.
- Preserve existing non-PR behavior.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,20 +29,24 @@ jobs:
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
-          # Fetch enough history to diff against the PR base.
+          # Fetch enough history to compute the merge base against the PR base.
           fetch-depth: 0
       - name: Check for source changes
         id: filter
         run: |
-          # List files changed in this PR relative to the merge base.
-          # If every changed file matches docs/** or *.md or docs_check.py,
-          # this is a docs-only PR and src=false; otherwise src=true.
+          # List files changed in this PR relative to the true merge base.
+          # Using three-dot merge-base diff so changes on the base branch that
+          # are not part of this PR do not appear in the file list.
+          # If every changed file matches docs/** or *.md (any depth) or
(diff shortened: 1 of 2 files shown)


KaifAhmad1 (approved):
Reviewed both workflow diffs for security impact — no regression found.

**Trigger safety**: still `pull_request` (not `pull_request_target`), so fork PRs keep a read-only `GITHUB_TOKEN` and no secrets regardless of the `security-scan` job's `pull-requests: write` grant (correctly called out in-file as safe for exactly this reason).

**No injection surface**: the new `changes` job interpolates `base.sha`/`head.sha` (git SHAs) into the shell script, not free-text like PR title/branch — not exploitable the way that pattern usually is.

**Least privilege intact**: `changes` defines no `permissions:` block, so it inherits the existing top-level `contents: read` in both files.

**Pinned action reused, not new**: `actions/checkout@3d3c42e...` resolves to the real `actions/checkout` v7.0.1 release; it's an existing pin reused for the new job, not a new supply-chain dependency.

**Fail-closed, verified**: stress-tested the case where `git merge-base`/`git diff` in the classifier errors. GitHub's default `bash -eo pipefail` means any git failure aborts the step immediately, `changes` reports `failure`, and `needs.changes.result != 'success'` then forces the real `build`/`security-scan` to run — it doesn't silently masquerade as docs-only. The `|| true` after `grep` is correctly scoped to grep's own no-match exit code, not masking the git calls.

**No coverage reduction vs. today**: previously docs-only PRs never triggered these workflows at all (`paths-ignore`), so `build`/`security-scan` never ran on them — they just left the required check stuck pending, which is the bug being fixed. The new classifier skips the same file set the old `paths-ignore` did; mixed/source PRs still get the full build and full pip-audit/Bandit/Semgrep scan, unchanged.

Approving — this is a narrowly-scoped CI reliability fix that doesn't weaken scan coverage or token scope.

PR #1437, opened 2026-09-03
explorer: name SHACL limit env vars in error messages

### LINKED ISSUE #1430: SHACL validation resource limits (4 env vars) are undocumented and not surfaced in limit-exceeded errors
SHACL validation in the Explorer enforces four resource limits, all read from environment variables, none documented anywhere (README/CONTRIBUTING/docs all return zero hits for these names):

https://github.com/semantica-agi/semantica/blob/main/semantica/explorer/routes/ontology.py#L119-L130

- `SEMANTICA_MAX_SHACL_TURTLE_BYTES` (default 256KB)
- `SEMANTICA_MAX_SHACL_TRIPLES` (default 1000)
- `SEMANTICA_MAX_SHACL_TIMEOUT` (default 15s)
- `SEMANTICA_MAX_SHACL_CONCURRENCY` (default 4)

Worse, when a limit trips, the error response states the number exceeded but never mentions these are configurable, e.g.:

> "SHACL graph triple count (1400) exceeds maximum allowed limit (1000)."

A developer testing SHACL validation locally against a realistically-sized ontology hits a hard wall with no pointer to the fix.

**Suggested fix:** name the relevant env var directly in each limit's error message, and document all four in the SHACL/Ontology reference docs.

pkupt:
The three limit-exceeded messages in alidate_shacl (Turtle size, triple count, timeout) didn't name the env var that controls them, so hitting a limit locally was a dead end. Each message now points to the variable to change, and the SHACL validation guide documents all four resource-limit variables with their defaults.

The existing tests already assert on these messages; I extended them to also check the env var name appears.

Closes #1430


KaifAhmad1 (approved):
Verified the three env var names in the new error messages match the actual os.environ.get() names at ontology.py:119-129, and the docs table's defaults match the code's defaults exactly. Confirmed the concurrency limit really doesn't surface in responses (bare asyncio.Semaphore, no error path). Ran the SHACL test subset on the branch: all 13 pass, including the three new env-var-name assertions. Tight, well-scoped fix — LGTM.

PR #1438, opened 2026-09-03
fix(explorer): serve entity_type from the ontology graph endpoint

### LINKED ISSUE #1433: [REFACTOR] Ontology editor node classification is duplicated across backend and frontend
## Description

The type vocabulary that classifies ontology graph nodes exists twice, in two languages:

- Backend: `_CLASS_TYPES` / `_PROPERTY_TYPES` / `_ONTOLOGY_TYPES`, `_URI_PREFIX_MAP`, and `_classify_node_type()` in `semantica/explorer/routes/ontology.py`
- Frontend: `classifyNodeType()` / `compactNodeType()` with its own `FULL_IRI_PREFIXES` table in `explorer/src/workspaces/OntologyWorkspace/ontologyEditorModel.ts`

The frontend copy exists only because `GET /api/ontology/graph` (added in #1278) returns raw node `type` strings, forcing the client to re-derive what the backend already knows. The compact-vs-full-IRI misclassification fixed during #1278 review was a direct symptom: the two copies drifted. Any future vocabulary change (e.g. supporting `skos:Concept` panels in the editor) has to be made twice and can silently diverge again.

## Proposal

Make the backend the single source of truth:

- Include a server-computed `entity_type` (`ontology` / `class` / `property` / `individual` / `concept` / `scheme` / `external`) on each node in the `/api/ontology/graph` response, using the existing `_classify_node_type()` (the `/entity/{uri}` endpoint already returns exactly this field).
- Frontend consumes `node.entity_type` and deletes `classifyNodeType` / `compactNodeType` / `FULL_IRI_PREFIXES`.
- Keep the response additive (extra field) so no client breaks.

## Acceptance criteria

- `/api/ontology/graph` nodes carry `entity_type` computed by the backend classifier.
- `ontologyEditorModel.ts` no longer contains an IRI-prefix table or type classification logic.
- Editor behavior is unchanged for both compact and full-IRI node types (existing classification matrix test moves to the backend response).

Related: #1278 (review round 1, finding 2). I'd like to work on this — please assign to @taoche.

taoche:
## Summary

Implements #1433. The node-type vocabulary existed twice — backend `_CLASS_TYPES`/`_URI_PREFIX_MAP`/`_classify_node_type()` and a frontend copy (`classifyNodeType`/`compactNodeType`/`FULL_IRI_PREFIXES`) added as a hotfix during #1278 review — and had already drifted once (the compact vs full-IRI misclassification).

- `/api/ontology/graph` nodes now carry a server-computed `entity_type`, emitted from `_classify_node_type()` verbatim so it is the same classifier **and the same wire vocabulary** as `/entity/{uri}`; the response builds copies so session-cached dicts are never mutated
- `classifyEditorNode` switches on `node.entity_type`, mapping `ontology`/`class`/`property` through and collapsing the remaining kinds (`concept`/`scheme`/`individual`/`unknown`) into read-only `"external"`. "External reference material" is what the editor *does* about a type it has no affordances for, so that reading stays in the editor rather than being baked into the wire
- `entity_type` is **required** on `OntologyGraphNode`. Optional would make a producer that forgets the field render as a graph of silently read-only nodes instead of failing `tsc`
- Frontend classification logic and its IRI-prefix table are deleted; the compact/full-IRI classification matrix test moves to the backend response (`test_ontology_graph_nodes_carry_entity_type_for_compact_and_full_iri_types`), and a new `test_graph_and_entity_endpoints_agree_on_entity_type` pins the two endpoints to the same answer for the same node

## Behavior change

Not quite "behavior unchanged", contrary to an earlier revision of this description — thanks to @zeekay for catching it.

The deleted `classifyNodeType` returned `property` for **any** type whose string contained `"Property"`, so an external node typed e.g. `ex:HasProperty` was rendered editable. The backend classifier recognizes only the exact OWL/RDFS property types, so such a node is now read-only. That is the better answer — an external term should not be editable — but it is a change, not a no-op.

Everything the editor renders for in-vocabulary nodes is identical, and nothing locally created can reach the new `default:` branch: `classifyEditorNode` is only ever fed API nodes, while the Add Class / Add Individual paths set `entityType` directly.

## Verification

- `pytest tests/explorer -m "not integration"` — 326 passed, 1 skipped (DNS-pinning case, unbindable in this environment)
- `pytest tests/explorer/test_ontology_subissue3.py` — 48 passed
- `npm run test:graph-workspace` 105/105 (108 on `main`, minus the two frontend classification tests this moves to the backend), `tsc -b` clean, `eslint` 75 problems — identical to the `main` baseline

Rebased onto `main` (`ec55385b`). #1441 landed the `_known_ontology_uris`/`_collect_core_nodes`/`_select_structure_edges` helpers where the removed `_graph_entity_type` sat; that was a placement collision only. #1439 is also in flight against `OntologyEditor.tsx`/`ontologyEditorModel.ts`.

Closes #1433

### DIFF
--- a/explorer/src/workspaces/OntologyWorkspace/OntologyEditor.tsx
+++ b/explorer/src/workspaces/OntologyWorkspace/OntologyEditor.tsx
@@ -27,7 +27,6 @@ import {
 import { loadOntologyEntityOwner, loadOntologyGraph } from "./api";
 import type { OntologyGraphEdge, OntologyGraphNode } from "./api";
 import {
-  classifyNodeType,
   inferOntologyUri,
   isEditableEntityType,
   ONTOLOGY_MINIMAP_THEME,
@@ -153,8 +152,17 @@ function nodeLabel(node: OntologyGraphNode): string {
   return trimmed.split("#").pop() || trimmed.split("/").pop() || node.id;
 }
 
+// The backend's remaining kinds (concept, scheme, individual) have no editor
+// affordances, so they collapse into the read-only "external" kind.
 function classifyEditorNode(node: OntologyGraphNode): OntologyNodeData["entityType"] {
-  return classifyNodeType(node.type);
+  switch (node.entity_type) {
+    case "ontology":
+    case "class":
+    case "property":
+      return node.entity_type;
+    default:
+      return "external";
+  }
 }
 
 function layoutEditorNodes(inputNodes: OntologyNode[]): OntologyNode[] {
--- a/explorer/src/workspaces/OntologyWorkspace/OntologyEditor.tsx
+++ b/explorer/src/workspaces/OntologyWorkspace/OntologyEditor.tsx
@@ -152,8 +152,10 @@ function nodeLabel(node: OntologyGraphNode): string {
   return trimmed.split("#").pop() || trimmed.split("/").pop() || node.id;
 }
 
-// The backend's remaining kinds (concept, scheme, individual) have no editor
-// affordances, so they collapse into the read-only "external" kind.
+// The backend's remaining kinds (concept, scheme, individual, unknown) have no
+// editor affordances, so they collapse into the read-only "external" kind. That
+// reading belongs here rather than on the wire: "external reference material"
+// is what the editor does about an unnameable type, not what the type is.
 function classifyEditorNode(node: OntologyGraphNode): OntologyNodeData["entityType"] {
   switch (node.entity_type) {
     case "ontology":
--- a/explorer/src/workspaces/OntologyWorkspace/api.ts
+++ b/explorer/src/workspaces/OntologyWorkspace/api.ts
@@ -12,7 +12,9 @@ import type {
 export type OntologyGraphNode = {
   id: string;
   type: string;
-  entity_type?: string;
+  // Required: /graph always classifies. Optional here would make a producer
+  // that forgets read as a graph of read-only nodes rather than a type error.
+  entity_type: string;
   content?: string;
   properties?: Record<string, unknown>;
 };
--- a/explorer/src/workspaces/OntologyWorkspace/api.ts
+++ b/explorer/src/workspaces/OntologyWorkspace/api.ts
@@ -12,6 +12,7 @@ import type {
 export type OntologyGraphNode = {
   id: string;
   type: string;
+  entity_type?: string;
   content?: string;
   properties?: Record<string, unknown>;
 };
--- a/explorer/src/workspaces/OntologyWorkspace/ontologyEditorModel.ts
+++ b/explorer/src/workspaces/OntologyWorkspace/ontologyEditorModel.ts
@@ -20,31 +20,6 @@ export const ONTOLOGY_MINIMAP_THEME = {
   },
 } as const;
 
-// The backend emits node types in compact (owl:Class) or full IRI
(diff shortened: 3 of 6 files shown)


zeekay (commented):
Verified against current `main` (2b866c86). Deleting the frontend copy of the vocabulary is the right move — it had already drifted once, and a second copy of a classifier is a second thing to drift. Two items before this can go in: a rebase, and one question about the value the endpoint emits.

**What I checked**

After resolving the conflict locally:

- `pytest tests/explorer -m "not integration"`: 320 passed (319 on `main` — your new case), same 3 pre-existing failures as `main`.
- `npm run test:graph-store` / `test:graph-workspace` / `test:plugin-registry`: 1 / 87 / 7 (87 = 89 on `main` minus the two frontend tests this moves to the backend). `tsc -b` clean, `eslint` totals unchanged.
- Browser check with a stubbed `/api/ontology/graph`: deep-linking to a `owl:Class` still opens "Class Details" with an editable Label field, identical to `main`. `classifyEditorNode` is only ever fed API nodes — the Add Class / Add Individual paths set `entityType` directly — so nothing locally created can fall through the new `default:` branch.

**The rebase**

One conflict, in `semantica/explorer/routes/ontology.py`. #1441 landed the `_known_ontology_uris` / `_collect_core_nodes` / `_select_structure_edges` helpers where you placed `_graph_entity_type`, so it is a placement collision, not a semantic one: keep both sides. Your change to the response construction merges cleanly on its own.

**The question: which vocabulary does `/graph` speak?**

`_graph_entity_type` maps `unknown` → `external`, but the frontend's `classifyEditorNode` already collapses everything it does not name into `"external"` via its `default:` branch — so the backend mapping changes nothing the editor sees, while making the two endpoints disagree about the same node: `/api/ontology/graph` reports `external`, `/api/ontology/entity/{uri}` reports `unknown`. The PR description says the graph endpoint uses "the same classifier and value domain as `/entity/{uri}`", and after this it does not.

Emitting `_classify_node_type(node.get("type", ""))` verbatim gives one classifier, one wire vocabulary, one place where the editor's four-kind view is derived, and identical rendered behavior. The "external reference material" reading belongs to the editor, and the editor already applies it. Worth a maintainer opinion if you disagree, but I would drop the helper.

Two smaller notes:

- `entity_type?: string` is optional on `OntologyGraphNode`, and if the field is ever missing every node silently becomes read-only rather than failing loudly. Since the backend now always sets it, making it required lets `tsc` catch a producer that forgets.
- The old `classifyNodeType` returned `property` for **any** type containing the substring "Property", so an external node typed e.g. `ex:HasProperty` was editable; the backend classifier only recognizes the exact OWL/RDFS property types, so such a node becomes read-only. I think that is the better answer — an external term should not be editable — but it is a behavior change, and "behavior unchanged" in the description is not quite right.

Also flagging the ordering: #1439 touches `OntologyEditor.tsx` and `ontologyEditorModel.ts` too and is currently conflict-free against `main`, so if it lands first this rebase grows to those files as well.

ZohaibHassan16 (approved):
LGTM.

Cleanup removing the duplicate frontend heuristic classification and serving `entity_type` directly from the backend endpoint. Parity tests look thorough and conflicts with #1439 are cleanly resolved.

PR #1441, opened 2026-09-03
perf(explorer): bound the ontology graph fetch and hydrate external nodes concurrently

### LINKED ISSUE #1436: [PERF] Ontology editor graph endpoint: bound the fetch, batch external node lookups
## Description

Two performance follow-ups on `GET /api/ontology/graph` (`get_ontology_graph` in `semantica/explorer/routes/ontology.py`), both identified during #1278 review and deferred to keep that PR scoped to correctness:

1. **The size cap is checked after an unbounded fetch.** Other `_MAX_ANALYSIS_NODES`-guarded endpoints in this file cap the fetch itself; this endpoint fetches every relevant node and edge type with `limit=2**63 - 1`, materializes and filters the result, then applies the cap afterward. The cap limits the returned payload but does not bound the work (or memory) needed to produce it, so the OOM guard the constant exists for is not actually enforced.

2. **External node lookups are serial.** Each non-core node referenced by a selected edge is hydrated with `await asyncio.to_thread(session.get_node, ...)` inside a plain loop — no batching, no `gather`. With up to 5,000 selected edges this becomes thousands of sequential thread dispatches for a single request.

## Proposal

- Pass a real limit to each per-type `session.get_nodes` / `session.get_edges` call (e.g. `_MAX_ANALYSIS_NODES + 1`) and return 413 as soon as any fetch overflows, so the cap bounds work rather than output. Keep the existing post-filter cap as the final consistency check.
- Hydrate external nodes concurrently (`asyncio.gather` over `to_thread` calls, or a bulk `get_nodes(ids=...)` if the session grows one), preserving the current deterministic output ordering.

## Acceptance criteria

- No graph/edge fetch in this endpoint uses an effectively unbounded limit.
- An over-limit ontology returns 413 without materializing the full graph first.
- External-node hydration issues lookups concurrently; response node/edge ordering stays deterministic (existing tests unchanged).

Related: #1278 (reviewer-acknowledged follow-ups). I'd like to work on this — please assign to @taoche.

taoche:
## Summary

Implements #1436 — the two performance follow-ups acknowledged during #1278 review.

**Bound the work, not just the response.** `get_ontology_graph` previously fetched every schema node and structural edge with `limit=2**63 - 1`, materialized and filtered, then checked `_MAX_ANALYSIS_NODES` afterward — the OOM guard bounded nothing. The issue sketched "paginate with page size = cap", but `paginate_nodes`/`paginate_edges` normalize and hold the entire matching set on every call before slicing, so paging would re-pay O(N) per page (O(N²/P) total) with unbounded session-side memory. Instead:

- New `GraphSession.iter_nodes`/`iter_edges` generators stream matches one at a time (id snapshot under the lock, per-node reads after; the non-atomic semantics vs `paginate_nodes` are documented)
- The endpoint filters ownership per node as it streams and raises → 413 (same detail text) the moment the ontology's **own** nodes or selected edges pass the cap, discarding foreign nodes instead of materializing them — the filter stays ahead of the cap check, so thousands of *other* ontologies' nodes still never 413 the requested one (`test_ontology_graph_ignores_unrelated_data_when_enforcing_size_limit` unchanged and green; a counting-wrapper test proves the scan early-exits at ~5k of 8,001 same-type nodes)
- The full `candidates_by_id` dict is gone; only owned nodes are retained
- The endpoint now reads as three named steps (`_known_ontology_uris`, `_collect_core_nodes`, `_select_structure_edges`)

**Concurrent hydration.** External edge endpoints are loaded in one `asyncio.gather` over `to_thread` calls (bounded by the default thread-pool), replacing the serial per-node loop; output ordering stays deterministic via the existing sorts, with a new external-target ordering test.

## Verification

- `pytest tests/explorer/` — 302 passed (1 pre-existing unrelated failure also present on main); ruff byte-identical to the main baseline

Note: #1438 adds `entity_type` to the same endpoint's response construction; whichever lands second rebases a one-line conflict.

Closes #1436


ZohaibHassan16 (approved):
Went through this end to end and ran the tests locally.

The two things Qodo called out look fine.

I don't think we need the final post-filter cap check anymore. The streaming helpers already raise as soon as the count goes over `_MAX_ANALYSIS_NODES`, so by the time we'd get to that old check, `core_nodes_by_id` and `selected_edges` are already within the limit. Adding it back would just be redundant.

The `iter_nodes`/`iter_edges` materialization issue is there but yea it's a separate change. `find_edges(edge_type=...)` still builds the full list for that edge type before filtering or checking the cap, so something like `rdf:type` can still be expensive on a big graph. Making `ContextGraph.find_edges` lazy is a bigger API/locking change and doesn't need to be part of this PR.

This still gets rid of a bunch of unnecessary work though. We're no longer normalizing every schema-typed node across every ontology or keeping the full `candidates_by_id` map around.

I also checked the concurrent external-node hydration. `get_node` only holds the lock around `find_node`, nothing stays locked across an `await`, and sorting after `gather` keeps the output order deterministic.

Traced through `normalize_node` too, and the ID filtering is the same as the old pagination path. I don't see any behavior change there.

Ran `tests/explorer/` locally. All passing.

Looks good to me. Approving.

PR #1445, opened 2026-09-03
fix(deps): unblock python 3.9 core install

### LINKED ISSUE #1347: [BUG] pip install semantica fails on Python 3.9 (all platforms) : spacy pulls in a thinc floor requiring Python >=3.10
## Problem

`pyproject.toml` says we support Python `>=3.8`, but a fresh `pip install semantica` is currently failing on Python 3.9.

The first run of `install-matrix.yml` (#1266) caught this:

```
py3.9   ubuntu-latest   -> failure
py3.9   macos-latest    -> failure
py3.9   windows-latest  -> failure
py3.10, 3.11, 3.12 (all OSes) -> success
```

Run: https://github.com/semantica-agi/semantica/actions/runs/33365156113

## Root cause

The core dependencies currently have `spacy>=3.4.0` with no upper limit.

Because of that, on a fresh install pip tries to use the newest spacy version that it can. That spacy version depends on:

```
thinc<8.4.0,>=8.3.12
```

But `thinc>=8.3.12` requires Python `>=3.10`, so on Python 3.9 the install fails with:

```
ERROR: Could not find a version that satisfies the requirement thinc<8.4.0,>=8.3.12
ERROR: No matching distribution found for thinc<8.4.0,>=8.3.12
```

So right now our package metadata says Python 3.8+ is supported, but the spacy/thinc dependency chain actually requires a newer Python version.

This means users on Python 3.9 currently can't do a fresh `pip install semantica`. Python 3.8 probably has the same problem too, but it isn't in the current matrix so I haven't verified it there.

## Proposed fix

I think there are two options:

* Add an upper bound for `spacy` so pip installs a version that still works with Python 3.9.

## Verification

After the fix, the Python 3.9 jobs in `install-matrix.yml` should pass on Ubuntu, macOS, and Windows.

I think this issue can be closed once all three of those jobs are green.

ZohaibHassan16:
Closes #1347.

The install matrix is still failing on Python 3.9 after the spaCy/thinc fix in #1329. Several other core dependencies now have minimum versions that require Python 3.10 or newer:

* `scikit-learn`
* `requests`
* `chardet`
* `grpcio`
* `pillow`
* `click`
* `onnxruntime`

This adds Python-version markers for those dependencies, following the same pattern already used for spaCy/thinc. Python 3.9 is capped at the latest compatible release for each package, while Python 3.10+ remains unconstrained.

Verified with `uv pip compile --python-version 3.9` for Linux, Windows, and macOS using the same no-extras install covered by `install-matrix.yml`. Also checked Python 3.10 and 3.12; all resolve successfully.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,11 @@ dependencies = [
   "numpy>=2.0.2",
   "pandas>=1.3.0",
   "scipy>=1.13.1",
-  "scikit-learn>=1.7.2",
+  # scikit-learn dropped Python 3.9 support at 1.7.0 (requires_python >=3.10),
+  # so an unqualified >=1.7.2 floor is unsatisfiable on 3.9. Cap 3.9 to the
+  # last 3.9-compatible release line; 3.10+ is left unconstrained.
+  "scikit-learn>=1.6.1,<1.7.0; python_version < '3.10'",
+  "scikit-learn>=1.7.2; python_version >= '3.10'",
   "umap-learn>=0.5.12",
   # thinc (spacy's core dep) dropped Python 3.9 wheels at 8.3.10, and later
   # spacy patch releases (3.8.8+) require thinc>=8.3.9-only-on-3.10+ ranges,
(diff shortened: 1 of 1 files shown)


KaifAhmad1 (approved):
Independently verified each Python-3.9 cap against live PyPI metadata (requires_python + wheel availability): scikit-learn, requests, chardet, grpcio, pillow, onnxruntime, and click all check out exactly. Consistent with the existing spaCy/thinc precedent in the same file. One minor non-blocking note: the requests comment names 2.33.0 as the cutoff, but that version was never published (releases jump 2.32.5 -> 2.34.2); the resulting <2.33.0 constraint is still correct.

PR #1455, opened 2026-09-03
refactor(context): expose public temporal normalizer

### LINKED ISSUE #1377: erasure.py imports _normalize_temporal_input from context_graph.py — a private function that creates a fragile cross-module coupling
## Summary

`semantica/context/erasure.py` imports a private function directly from another module:

```python id="9q8h0w"
# erasure.py line 42
from .context_graph import _normalize_temporal_input
```

`_normalize_temporal_input` is a module-level private function (leading underscore) in `context_graph.py`. This creates a maintenance dependency: if the function is ever renamed, moved, inlined, or made into a local closure, `erasure.py` breaks silently at import time with `ImportError`.

## Why it was done this way

The coordinator needs to resolve timestamps in exactly the same way `ContextGraph` does, so that the receipt's `erased_at` and the graph tombstone's `purged_at` always agree. Duplicating the logic would risk drift. Using the private import was the pragmatic choice at PR time; this issue tracks making it clean.

## Proposed fix

**Option A (preferred):** Promote `_normalize_temporal_input` to a public utility in `semantica/utils/` or expose it as a public function in `semantica/context/context_graph.py`:

```python id="22ksjo"
# context_graph.py
def normalize_temporal_input(value):
    """Public alias for temporal normalization used by erasure and other callers."""
    return _normalize_temporal_input(value)
```

Then `erasure.py` imports the public name:

```python id="zqf3k4"
from .context_graph import normalize_temporal_input
```

**Option B:** Move `_normalize_temporal_input` to `semantica/utils/temporal.py` and import it from there in both `context_graph.py` and `erasure.py`.

## Acceptance criteria

* `erasure.py` imports no private names (leading underscore) from any other module
* The shared normalization logic lives in one place
* No behavior change — timestamps normalized identically before and after
* Existing tests pass unchanged

## Related

* Raised during PR #1027 review as a LOW finding

Saket7002:
## Summary

- expose `normalize_temporal_input` as a public API
- update `erasure.py` to use the public helper instead of a private cross-module import
- add regression coverage for the public normalizer

## Testing

- `python -m pytest tests\context\test_context_graph_retraction.py tests\context\test_erasure_coordinator.py`
- 101 passed

Closes #1377


Sameer6305 (approved):
### Final Review — Approved ✅

Reviewed PR #1455 against issue #1377 thoroughly, including the full temporal-normalization flow and repository-wide search for similar private cross-module coupling.

### What happened in this PR

**1. Original issue**
`erasure.py` was directly importing the private `_normalize_temporal_input` from `context_graph.py`. This created fragile coupling because changes to that private implementation could break `erasure.py`.

**2. Implementation**
The PR introduced the public `normalize_temporal_input` API in `context_graph.py` while keeping the existing normalization implementation as the single source of truth.

`erasure.py` was updated to use the public helper instead of importing the private function.

**3. Behavioral compatibility**
The public helper delegates to the existing implementation, so there is no intended behavior change.

We verified normalization across the supported temporal input types, timezone-aware/naive values, valid/invalid strings, numeric timestamps, and error paths.

The erasure flow remains consistent:

`input timestamp → normalize_temporal_input() → receipt.erased_at / tombstone.purged_at`

So the timestamp consistency guarantee remains intact.

**4. Architecture review**
Option A was confirmed as the appropriate approach for this repository. Moving the helper into `utils/` would introduce an undesirable dependency direction because the existing implementation relies on private parsing functionality in `context_graph.py`.

No circular-import issue was introduced.

**5. Repository-wide review**
The repository was also searched for similar private cross-module imports and duplicated temporal-normalization logic.

Other private imports found were pre-existing and structurally different/intra-package. No additional change was required as part of this issue.

**6. Test improvements**
The original PR had limited regression coverage. Additional tests were added covering:

* Supported temporal input types
* Invalid/unsupported inputs
* Public API behavior
* Import-guard protection against reverting to the private import
* Cross-module consistency
* Timezone-related cases
* `receipt.erased_at` vs `tombstone["purged_at"]`

**7. Validation**
Final relevant test run:

`129 passed, 0 failed`

The original 101 tests also continued to pass.

### Final assessment

✅ Private cross-module import removed
✅ Shared normalization logic remains in one place
✅ No behavioral change
✅ Erasure/tombstone timestamp consistency preserved
✅ Public API properly documented
✅ Regression coverage significantly strengthened
✅ Similar patterns reviewed across the repository
✅ Relevant tests passing

**Approved — this is ready to merge.**

Closes #1377.

PR #1464, opened 2026-09-04
fix(kg): promote synthetic relation endpoints into the graph

### LINKED ISSUE #1463: [BUG] GraphBuilder drops synthetic relation endpoint entities, causing dangling edges (DANGLING_EDGE) during graph validation
## Summary

When relation extraction returns a subject or object that is not present in the NER entity list, Semantica creates a synthetic
`UNKNOWN` entity for that endpoint (`metadata={"synthetic": True}`, `confidence=0.8`).

However, `GraphBuilder` only stores the endpoint's text/ID in the relationship and never adds the synthetic entity object to the
graph's entity collection. The resulting graph contains dangling relationships and fails `GraphValidator` with `DANGLING_EDGE`.

This breaks the native end-to-end flow:

`NER extraction → relation extraction → graph building → graph validation`

**Verified against `main` @ `fa87a1a9` (v0.6.7) — reproducible with both default `GraphBuilder()` settings and
`merge_entities=False, resolve_conflicts=False`.**

## Affected components

- `semantica/semantic_extract/methods.py` — `_parse_relation_result()`
- `semantica/kg/graph_builder.py` — `_process_item()` / `build()`
- `semantica/kg/graph_validator.py` — reports the resulting `DANGLING_EDGE`

## Current behavior

In `_parse_relation_result()` (`semantica/semantic_extract/methods.py:2094`), unmatched relationship endpoints are intentionally
converted into synthetic entities so that multi-value LLM output is never silently dropped:

```python
subject_entity = match_entity(subject_text, entities)
object_entity = match_entity(object_text, entities)

if not subject_entity:
    subject_entity = Entity(
        text=subject_text, label="UNKNOWN",
        start_char=0, end_char=len(subject_text),
        confidence=0.8, metadata={"synthetic": True},
    )
if not object_entity:
    object_entity = Entity(
        text=object_text, label="UNKNOWN",
        start_char=0, end_char=len(object_text),
        confidence=0.8, metadata={"synthetic": True},
    )
```

When `GraphBuilder._process_item()` (`semantica/kg/graph_builder.py:161`) receives the resulting `Relation`, it converts the
endpoints to strings and appends only the relationship — the endpoint `Entity` objects are never routed to `all_entities`:

```python
elif hasattr(item, "subject") and hasattr(item, "predicate") and hasattr(item, "object"):
    subj = item.subject
    obj = item.object
    subj_id = getattr(subj, "id", getattr(subj, "text", str(subj))) if not isinstance(subj, str) else subj
    obj_id = getattr(obj, "id", getattr(obj, "text", str(obj))) if not isinstance(obj, str) else obj
    rel_dict = {
        "source": subj_id,
        "target": obj_id,
        "type": item.predicate,
        "confidence": getattr(item, "confidence", 1.0),
        "metadata": getattr(item, "metadata", {}),
    }
    all_relationships.append(rel_dict)
```

`build()` runs entities and relationships as two independent streams and never reconciles relationship endpoints against the entity
collection — including when `merge_entities=True` (the default): the entity resolver only operates on entities that already made it
into `all_entities`, so it cannot rescue a missing endpoint. `GraphValidator` then correctly reports `DANGLING_EDGE`, because the
relationship endpoint IDs do not exist in the graph's entity collection.

## Root cause

An interface-contract inconsistency between the extraction and graph modules:

- `semantic_extract` explicitly **creates** a synthetic entity to represent an unknown endpoint;
- `kg.GraphBuilder` **consumes only its string reference** and discards the entity object;
- the graph is therefore internally invalid by construction: every endpoint is "partially accepted".

The problem is not LLM quality. It is that the two native components disagree about what a synthetic endpoint is.

## Minimal reproduction

```python
from semantica.kg import GraphBuilder, GraphValidator
from semantica.semantic_extract import Entity, Relation

known_entity = Entity(text="Known Entity", label="CONCEPT", start_char=0, end_char=12)

synthetic_entity = Entity(
    text="Synthetic Entity", label="UNKNOWN",
    start_char=0, end_char=16,
    confidence=0.8, metadata={"synthetic": True},
)

relation = Relation(subject=known_entity, predicate="related_to", object=synthetic_entity)

graph = GraphBuilder(merge_entities=False, resolve_conflicts=False).build(
    [known_entity, relation], extract=False
)

validation = GraphValidator().validate(graph)
print(validation.to_dict())
```

Actual result — the relationship is kept but its target entity is absent:

```text
=== graph entities ===
  {'id': 'Known Entity', 'name': 'Known Entity', 'type': 'CONCEPT', 'confidence': 1.0, 'metadata': {}}
=== graph relationships ===
  {'source': 'Known Entity', 'target': 'Synthetic Entity', 'type': 'related_to', 'confidence': 1.0, 'metadata': {}}
=== validation ===
{'is_valid': False, 'issues': [{'code': 'DANGLING_EDGE',
  'message': 'Target entity ID not found or invalid: Synthetic Entity', 'severity': 'error',
  'element_id': 'Known Entity->Synthetic Entity', 'element_type': 'relationship',
  'details': {'target_id': 'Synthetic Entity'}},
  {'code': 'ORPHAN_NODES', 'message': 'Found 1 orphan nodes (no relationships).', 'severity': 'warning', ...}],
 'stats': {'total_entities': 1, 'total_relationships': 1, 'issues_found': 2, 'errors': 1}}
```

**The same failure occurs with all-default settings** (`GraphBuilder()` — `merge_entities=True`, conflict resolution enabled):
the entities collection still contains only `Known Entity` and validation still reports `errors=1` / `DANGLING_EDGE`. Default
configuration does not mitigate the defect.

## Scope notes (beyond the direct repro)

- **Triplet output is affected the same way.** `_process_item` selects the relation branch by duck typing
  (`hasattr(item, "subject") and hasattr(item, "predicate") and hasattr(item, "object")`), so triplet-extraction endpoints that do
  not exist in the entity list take the identical path.
- The dict fast path in `build()` (`semantica/kg/graph_builder.py:665-687`) normalizes `source`/`target` to strings with the same
  lossy semantics.
- Any caller that supplies relationships whose endpoints are not also present in the entity list produces an invalid graph.

## Real-world observation

In an end-to-end TXT-to-ontology run (LLM extraction), NER did not extract several phrases that the relation model later used as
endpoints. This produced four dangling edges in an otherwise successfully constructed graph, all originating from a single text
chunk:

```text
Organization A  -> Disaster Fact
Organization B  -> Disaster Fact
Loan Amount     -> Risk Resolution List
Insurance Product -> Annual Operating Income
```

The pipeline completed text processing and LLM extraction (81 merged entities, 127 relationships) but stopped at `GraphValidator`
because the missing endpoint entities were never included in the graph.

## Expected behavior

`GraphBuilder` should preserve the graph integrity invariant:

> Every relationship source and target must resolve to an entity in the graph.

At least one of the following policies should be applied consistently, controlled by configuration:

1. **Strict (default recommended by the reporter for correctness-first callers):** reject (and optionally retry-with-feedback)
   relationships whose endpoints cannot be matched to the supplied entity set, and report them explicitly instead of emitting a
   half-accepted edge.
2. **Permissive:** promote synthetic endpoint entities into the graph's entity collection, preserving their `UNKNOWN` label and
   `metadata={"synthetic": True}`.
3. **Configurable:** an explicit option, e.g. `unknown_relation_endpoint="reject" | "include" | "error"`.

Suggested implementation points:

- Apply the policy in `GraphBuilder._process_item()` (and the dict fast path), not only in the extractor — the relation branch,
  the triplet path, and dict sources all converge there, so a single fix covers every entry point.
- When promoting endpoints permissively, deduplicate by canonical entity ID (endpoint text already serves as the ID) so an
  existing entity is never duplicated when the synthetic endpoint resolves to it.
- Preserve synthetic metadata (`synthetic: True`, `confidence: 0.8`) — synthetic entities must not silently acquire the default
  `confidence=1.0` of ordinary entities.
- Run endpoint validation per extraction batch so a bad chunk is localized (in the real-world failure, all four dangling edges
  came from one chunk).

## Acceptance criteria

- [ ] A graph produced by `GraphBuilder` never contains dangling relationship endpoints unless explicitly allowed by
      configuration.
- [ ] Synthetic endpoints are either added to the graph's entity collection **or** rejected with a clear extraction/validation
      error — never silently referenced by a string only.
- [ ] Existing entities are not duplicated when a synthetic endpoint resolves to an already-present entity.
- [ ] Synthetic metadata is preserved when synthetic entities are included.
- [ ] `GraphValidator` passes for relations whose synthetic endpoints are accepted.
- [ ] The default-options configuration (`merge_entities=True`) behaves consistently — today it reproduces the bug too.
- [ ] Tests cover: one missing endpoint; two missing endpoints; endpoint deduplication; strict rejection mode; permissive
      inclusion mode; LLM relation output containing an endpoint absent from NER results; triplet endpoints with missing
      entities.

## Environment

- Semantica `0.6.7`, `main` @ `fa87a1a9` (2026-09-02)
- Python 3.11
- No external services required to reproduce (deterministic, no LLM call needed)

pkupt:
## Summary

Relation extraction synthesizes an `UNKNOWN` entity for a relationship endpoint that isn't in the NER entity list (marked `synthetic=True`, `confidence=0.8`), but `GraphBuilder._process_item` kept only the endpoint string, so the synthetic object never reached the entity collection. Any graph built from that output then failed `GraphValidator` with `DANGLING_EDGE`, breaking the native NER → relation → graph → validation flow.

This promtes synthetic endpoint entities into the entity collection in the relation branch, deduplicated by id so an existing entity is never duplicated, and keeps their `confidence=0.8` and `synthetic=True` metadata. A `unknown_relation_endpoint` option defaults to `include`; set it to `reject` to drop relationships with synthetic endpoints instead.

## Tests

- single missing endpoint promoted
- two missing endpoints promoted
- synthetic metadata and confidence preserved
- no duplicate when endpoint already present
- default settings (merge_entities=True) no longer produce DANGLING_EDGE
- reject policy drops the dangling relationship

Closes #1463


zeekay (dismissed):
The problem is real and worth fixing — #1463 is a genuine hole, and I like that you traced it to `_process_item` dropping the synthetic object rather than patching `GraphValidator` to look the other way. Two things in the implementation block it, both reproduced against your branch merged onto current `main`.

**1. A source entity carrying `metadata: None` now crashes the build.**

The `_real_ids` loop in `build()` does `_entity.get("metadata", {}).get("synthetic")`. `dict.get("metadata", {})` returns `None`, not `{}`, when the key is present with a `None` value, so the second `.get` raises. Same shape in the filter loop below it.

```python
from semantica.kg import GraphBuilder
src = {"entities": [{"id": "a", "name": "a", "metadata": None},
                    {"id": "b", "name": "b"}],
       "relationships": [{"source": "a", "target": "b", "type": "rel"}]}
GraphBuilder(resolve_conflicts=False).build(src, extract=False)
```

- on `origin/main`: `['a', 'b']`
- on this branch: `AttributeError: 'NoneType' object has no attribute 'get'`

Note that `_process_item` already guards this correctly one hunk earlier — `isinstance(getattr(endpoint, "metadata", None), dict)`. The `build()` half just needs the same care; `(_entity.get("metadata") or {})` is enough.

**2. Rebuilding `existing_ids` per relationship makes `build()` quadratic.**

The `existing_ids` set is reconstructed by walking all of `all_entities` for every relationship item, and it runs whether or not that relationship has any synthetic endpoint. With N entities followed by N relations that's O(N²).

Measured, `Entity`/`Relation` objects, `resolve_conflicts=False`, no synthetic endpoints anywhere:

| items | `origin/main` | this branch |
|---|---|---|
| 3,000 | 0.007 s | 2.311 s |
| 6,000 | 0.022 s | 9.040 s |

Doubling the input quadruples the time here and roughly doubles it on `main` — 410x at 6k, and it gets worse from there. Two fixes, either works: hoist the set out of `_process_item` and maintain it as entities are appended, or at minimum wrap the whole block in `if synthetic_endpoints:` so builds with no synthetic endpoints pay nothing. The first is better, since a build that *is* full of synthetic endpoints is exactly the case this PR is for.

**Two smaller things, worth settling before this lands:**

*The `unknown_relation_endpoint` option is a new public knob with no home.* It isn't in `semantica/kg/config.py` and isn't mentioned in `semantica/kg/kg_usage.md`, so `kg_config.get_method_config("build")` doesn't know it exists and nothing documents that `"reject"` is the alternative to the default. Per the per-module convention in CLAUDE.md that's where module configuration lives.

Related: the nested lookup

```python
cfg = self.config
nested_cfg = cfg.get("config")
if isinstance(nested_cfg, dict):
    cfg = nested_cfg
```

is true to how `orchestrator.py:164` calls `GraphBuilder(config=self.config.get("kg", {}))`, but it makes this one option the only one in the class that reads through that nesting — `entity_resolution` and `conflict_detection` (graph_builder.py:120, 137) both read `kwargs` directly and are silently ignored on the orchestrator path. Normalizing once in `__init__` (fold a `config=` dict into `self.config`) fixes all three and leaves one way to read an option instead of two.

*The HuggingFace path tags everything.* In `extract_triplets_huggingface`, `hf_entities = kwargs.get("entities") or []` is always empty — your own comment says so — which makes `match_entity(endpoint_text, [])` a filter that never filters, so every REBEL triplet gets both endpoints in `synthetic_endpoints` and both get promoted as `UNKNOWN`. If that's the intent, say it directly instead of routing it through a comparison that can't do anything; if it isn't, the extractor needs the real entity list threaded in. As written, a reader has to work out that the machinery is inert.

For what it's worth the rest holds up: `pytest tests/kg tests/semantic_extract -m "not integration"` is 522 + 156 passed on the branch, and the new test file covers the promotion, dedup and reject cases well. Fix the `None` crash and the quadratic scan and I think this is close.

Sameer6305 (dismissed):
Approved — PR #1464 is merge-ready.

I reviewed this fix in multiple stages, including an independent final audit of the resulting branch.

What was addressed in this PR:

1. Root cause fixed
   - Relation extraction can create synthetic UNKNOWN entities for endpoints that are absent from NER.
   - GraphBuilder previously kept only the endpoint ID/string and discarded the synthetic Entity object.
   - This produced relationships pointing at entities that were not present in the graph and consequently DANGLING_EDGE validation failures.
   - GraphBuilder now promotes accepted synthetic relation endpoints into the graph's entity collection.

2. Synthetic endpoint policy added
   - Added `unknown_relation_endpoint` configuration.
   - Default behavior is `include`, preserving the synthetic endpoint as an UNKNOWN entity.
   - `reject` drops relationships containing synthetic endpoints instead of creating dangling edges.
   - Synthetic confidence (`0.8`) and `synthetic=True` metadata are preserved.

3. Deduplication and entity identity handling
   - Synthetic endpoints are deduplicated against existing entities.
   - Both `id` and `entity_id` are accounted for.
   - Existing real entities win over previously promoted synthetic entities.
   - Shared synthetic endpoints do not create duplicate entities.
   - Unhashable/None identifier edge cases were also verified.

4. Extraction paths covered
   - LLM Relation extraction
   - LLM Triplet extraction
   - HuggingFace REBEL triplets
   - Pattern/rules triplets
   - Direct Relation objects
   - Direct Triplet objects
   - The HuggingFace path now receives the actual extracted entity list when determining unresolved endpoints.

5. Configuration integration
   - `unknown_relation_endpoint` is now part of the KG build configuration and documented.
   - GraphBuilder now normalizes nested `config=` input consistently, which also avoids silently dropping the existing entity-resolution/conflict-detection configuration on orchestrator-style calls.

6. Regression fixes from review
   - Fixed the `metadata=None` crash in the build/reconciliation logic.
   - Avoided the O(N²) regression caused by rebuilding the existing-ID set for every relationship.
   - The scan is gated so normal builds without synthetic endpoints retain the fast path.

7. Tests
   - Added coverage for missing source/target endpoints, both endpoints missing, deduplication, entity_id handling, synthetic metadata/confidence, reject mode, default configuration, LLM relation output, HuggingFace triplets, orchestrator configuration, and metadata=None.
   - Final relevant suite:
     732 passed, 11 skipped.

I also performed a final independent audit of the branch after the review changes were made. The end-to-end graph integrity invariant was verified: accepted relationships have resolvable endpoints and GraphValidator does not report DANGLING_EDGE.

The only additional test failure encountered was an unrelated pre-existing `pyoxigraph` dependency issue in an exporter test outside the files touched by this PR; it reproduces on origin/main as well.

The final branch is clean and the reviewed changes have been pushed to:
fix/1463-dangling-synthetic-endpoint

No remaining correctness concerns found.

✅ Approved.

PR #1464 is merge-ready.

Sameer6305 (approved):
Re-reviewed and approved.

The remaining Qodo observability finding is now resolved: reject-mode drops are emitted at WARNING level and exposed via `graph["metadata"]["rejected_relationships"]`, with regression tests covering the count, logging, and counter reset.

The final branch was re-audited with 736 tests passing (11 skipped), `git diff --check` clean, and no remaining correctness concerns.

✅ PR #1464 is merge-ready. Approved for merge.

PR #1473, opened 2026-09-04
fix: correct production env var to SEMANTICA_API_KEY (fixes #1429)

### LINKED ISSUE #1429: README production deployment guidance references nonexistent SEMANTICA_SECRET_KEY (should be SEMANTICA_API_KEY)
README's production deployment guidance says to set an environment variable that doesn't exist anywhere in the codebase:

> "For production deployments... Set `SEMANTICA_SECRET_KEY`, configure a persistent LPG graph store..."

https://github.com/semantica-agi/semantica/blob/main/README.md#L1524

`SEMANTICA_SECRET_KEY` appears nowhere in `semantica/` (grep confirms zero hits outside this one README line). The actual variable the Explorer auth code reads is `SEMANTICA_API_KEY`, with `SEMANTICA_ALLOW_ANONYMOUS=true` as the explicit opt-out:

https://github.com/semantica-agi/semantica/blob/main/semantica/explorer/dependencies.py#L28

A developer following this instruction to secure a production Explorer deployment sets a variable that's silently ignored: the server then either 503s every route (safe but confusing, no explanation of why) or runs anonymous, with nothing telling them `SEMANTICA_SECRET_KEY` did nothing.

**Fix:** update the README line to `SEMANTICA_API_KEY`.

v01dst:
## Problem
Fixes #1429 — README production guidance tells users to set SEMANTICA_SECRET_KEY, which exists nowhere in the codebase.

## Root cause
README.md:1546 references SEMANTICA_SECRET_KEY. The Explorer auth code (semantica/explorer/dependencies.py:30) reads SEMANTICA_API_KEY, with SEMANTICA_ALLOW_ANONYMOUS=true as opt-out. A deploy following the README sets a silently ignored variable, then hits 503s or unintended anonymous mode.

## Fix
README.md: SEMANTICA_SECRET_KEY to SEMANTICA_API_KEY on the production line. 1-line docs-only diff.

## Verification
grep SEMANTICA_SECRET_KEY README.md shows 0 hits after fix. No code touched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1543,7 +1543,7 @@ pip install semantica[watch]                # Directory file watcher
 pip install semantica[explorer]             # Knowledge Explorer dashboard
 ```
 
-For production deployments, use Docker or Kubernetes rather than a local `pip install`. Set `SEMANTICA_SECRET_KEY`, configure a persistent LPG graph store (Neo4j / FalkorDB / Apache AGE / AWS Neptune) and/or RDF triple store (Blazegraph / Apache Jena / Eclipse RDF4J), and point the vector store at a hosted backend (Qdrant / Pinecone). See [ARCHITECTURE.md](ARCHITECTURE.md) for the full deployment topology.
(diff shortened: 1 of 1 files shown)


ZohaibHassan16 (approved):
LGTM.

PR #1474, opened 2026-09-05
fix(scripts): broaden Windows Mintlify noise filter to EBUSY (#1427)

### LINKED ISSUE #1427: docs_check.py Windows Mintlify-export noise filter only matches EPERM, not EBUSY
`docs_check.py`'s Mintlify export check only treats the failure as Windows temp-cleanup noise (safe to skip) when the combined stdout/stderr contains the literal string `"EPERM"`:

https://github.com/semantica-agi/semantica/blob/main/docs_check.py#L248-L253

In practice, npm's post-command cleanup on Windows also fails with `EBUSY` (`resource busy or locked`), not just `EPERM`. When that happens, the check falls through to the real-failure branch and reports a "failure" whose printed output is just npm deprecation warnings, not any actual Mintlify error, since Mintlify itself already finished before cleanup ran. Repro on Windows: run `python docs_check.py` a few times in a row (or after other recent `npx mintlify` invocations); it intermittently reports `FAIL Mintlify export builds without errors` with no real error in the output, then passes on a retry with no code changes in between.

This makes local Windows runs unreliable for exactly the check that's supposed to catch real doc build breakage, and the failure output gives no signal for whether it's noise or a genuine problem.

**Suggested fix:** broaden the noise-detection check to also match `EBUSY` (and ideally match on the *absence* of a real Mintlify error signature rather than the *presence* of specific npm cleanup error codes, since new npm versions may raise others).

ALDRIN121:
## Summary

Fixes #1427.

`docs_check.py`'s Mintlify export check only treats a Windows failure as temp-cleanup noise (safe to skip) when the combined output contains the literal `EPERM`. In practice, npm's post-command cleanup on Windows also fails with `EBUSY` (`resource busy or locked`), which currently falls through to the real-failure branch and reports a false `FAIL Mintlify export builds without errors` with no actual Mintlify error in the output.

## Change

Broaden the noise filter to also match `EBUSY` alongside `EPERM`, while still requiring the absence of the real Mintlify error signature (`could not be generated`). The change remains Windows-only (`sys.platform == "win32"`), and non-Windows behavior is unchanged.

## Testing

No dedicated test file exists for `docs_check.py`. Verified the modified branch logic in isolation (EPERM/EBUSY → skip on win32; EBUSY + `could not be generated` → fail; non-win32 → fail regardless).


zeekay (approved):
Verified against current `main`, and the fix is doing real work.

I exercised the actual check function in `docs_check.py` with `subprocess.run` stubbed and `sys.platform` swapped, over five inputs:

| platform | output | before | after |
| --- | --- | --- | --- |
| win32 | `EBUSY: resource busy or locked, rmdir` | FAIL | pass |
| win32 | `EPERM: operation not permitted` | pass | pass |
| win32 | `EBUSY` + `erro page could not be generated` | FAIL | FAIL |
| darwin | `EBUSY: ...` | FAIL | FAIL |
| win32 | `erro something else entirely` | FAIL | FAIL |

So the only behaviour that changes is the one the issue describes: EBUSY cleanup noise on Windows stops being reported as a docs build failure. The real Mintlify error signature still fails, non-Windows still fails, and unrelated Windows failures still fail — the filter stays fail-closed on anything it does not recognise.

`python docs_check.py` on `main` + this patch: all nine structural checks pass (the Mintlify export check fails locally only because node 26 is newer than Mintlify supports, which is unrelated). Changed lines are under 88 columns and match the continuation style of the surrounding block.

Re the issue's alternative suggestion (key off the absence of a real error rather than the presence of specific npm codes): keeping the error-code allowlist is the safer call — inverting it would swallow every unrecognised Windows failure. Adding codes as they show up is the right trade.

Sameer6305 (approved):
LGTM. ✅

Reviewed the implementation and surrounding `docs_check.py` logic against `origin/main`.

* Correctly handles both `EPERM` and `EBUSY` as Windows cleanup noise.
* Still fails when the real Mintlify error signature (`could not be generated`) is present.
* Non-Windows behavior remains unchanged.
* Scope is minimal and directly addresses #1427.
* No unrelated changes or refactoring needed.
* Verified the relevant logic paths and basic formatting checks.

No changes requested. **Approved — merge as-is.** 🚀

PR #1486, opened 2026-09-06
feat(ingest): support line-delimited JSON files

sunchaokun:
## Summary

- Add `jsonl` and `ndjson` to Semantica's supported document formats.
- Preserve line-delimited JSON content through `FileIngestor`.

This supports common bulk-data exports such as Synthea FHIR Bulk NDJSON without adding healthcare-specific logic to the core ingestor.

## Validation

- `python -m pytest -q tests/ingest/test_file_ingestor.py tests/test_import.py tests/test_cli_foundation.py` — 57 passed.
- Synthea FHIR Bulk sample — 19 NDJSON files, 7,887 valid JSON lines, 0 invalid lines.
- `git diff --check` passed.

## Scope

This change adds generic line-delimited JSON file support only. FHIR resource parsing and healthcare-specific mapping are intentionally out of scope.

No production or patient data is included.

### DIFF
--- a/semantica/utils/constants.py
+++ b/semantica/utils/constants.py
@@ -54,6 +54,8 @@ SUPPORTED_DOCUMENT_FORMATS = [
     "html",
     "xml",
     "json",
+    "jsonl",
+    "ndjson",
     "csv",
     "xlsx",
     "pptx",
--- a/tests/ingest/test_file_ingestor.py
+++ b/tests/ingest/test_file_ingestor.py
@@ -122,6 +122,33 @@ def test_type_detector_extended() -> None:
     assert detected_unknown == "unknown"
 
 
+@pytest.mark.parametrize("suffix", [".jsonl", ".ndjson"])
+def test_type_detector_supports_line_delimited_json(suffix: str) -> None:
+    """Line-delimited JSON is a common interchange format for bulk data."""
+
+    detector = FileTypeDetector()
+
+    assert detector.detect_type(f"resources{suffix}") in {"jsonl", "ndjson"}
+    assert detector.is_supported(suffix.lstrip("."))
+
+
+def test_ingest_ndjson_file_preserves_content(tmp_path: Path) -> None:
+    """FHIR Bulk/NDJSON resources can be ingested as a supported file."""
+
+    path = tmp_path / "Patient.000.ndjson"
+    path.write_text(
+        '{"resourceType":"Patient","id":"synthetic-1"}\n'
+        '{"resourceType":"Patient","id":"synthetic-2"}\n',
+        encoding="utf-8",
+    )
+
+    result = FileIngestor().ingest_file(path)
+
+    assert result.file_type == "ndjson"
+    assert result.metadata["is_supported"] is True
+    assert result.text.count('"resourceType":"Patient"') == 2
+
+
 # --- CloudStorageIngestor Tests ---
 @patch("boto3.client")
 def test_cloud_storage_s3(mock_boto: MagicMock) -> None:


ZohaibHassan16 (changes requested):
Thanks for putting this together. There are few CLI spots that also should be included in this PR. Couple of these:

i) In `semantica/cli.py`, `INGEST_FORMATS` is a hardcoded list around line 1447 so passing `--format ndjson` crashes Click with an invalid error. Also `semantica watch` 's default `--patterns` (line 1021) doesn't have `*.jsonl, *.ndjson`, so the watcher just ignores them. Please add both formats to those two spots

ii) Also Python's `mimetypes` doesn't know about these extensions so `file_obj.mime_type` ends up as `None`. Quick `mimetype.add_type("application/x-ndjson", ...) in `FileTypeDetector.__init__` would fix that.

iii) For the tests, `assert detector.detect_type(...) in {"jsonl, "ndjson"}` is a bit too loose, let's just assert `==suffix.lstrip("."). And please parametrize the file ingest test so `.jsonl` actually gets tested too alongside `.ndjson`.

zeekay (approved):
Reviewed and tested against current `main` (the branch is behind, so I cherry-picked it onto `origin/main` and ran it there).

Checked the blast radius first, because adding an extension to `SUPPORTED_DOCUMENT_FORMATS` looks like it could change parsing behaviour. It doesn't: the constant is read only by `FileTypeDetector`, and `is_supported` is used only inside `file_ingestor.py` — for the "Unsupported file type" warning and the `metadata["is_supported"]` flag. `StructuredDataParser._detect_format` still maps `.json` only, so `.jsonl`/`.ndjson` keep falling through to content sniffing exactly as before. Nothing downstream starts trying `json.loads` on a whole line-delimited file.

Test run (`pytest tests/ingest/ -m "not integration"`):
- on `origin/main`: 6 failed, 289 passed
- with this change: 6 failed, 292 passed

The 6 failures are pre-existing and unrelated (`test_web_ingestor.py` / `TestWebIngestor`, `TestFeedIngestor`). The 3 new passes are the tests added here.

One small note for a future pass, not worth holding this up: in `test_type_detector_supports_line_delimited_json`, the first assertion passes with or without the change — `detect_type` returns the extension verbatim for any suffix. The `is_supported` assertions are the ones doing the work.

zeekay:
Following up on my approval, because @ZohaibHassan16's review raises a gap I checked and can confirm — worth folding in before this merges.

Verified on current `main`:

- `semantica/cli.py:1468` — `_INGEST_FORMATS = ["pdf", "docx", "csv", "excel", "html", "json", "parquet", "xml", "rdf"]`, used as the `click.Choice` for `ingest --format` at line 1500. `--format ndjson` is rejected as an invalid choice.
- `semantica/cli.py:1021` — `watch --patterns` defaults to `*.pdf,*.docx,*.txt,*.csv,*.json`, so the watcher skips `.jsonl`/`.ndjson` files.

So the library now calls these files supported while the two CLI surfaces that gate on a format list still don't. Either extend both lists here, or say in the PR description that the CLI surface is deliberately a follow-up — but the two shouldn't silently disagree.

Two smaller ones from that review that I'd weigh differently:

- The `mimetypes` point is real (`file_obj.mime_type` comes back `None` for `.ndjson`), but `mimetypes.add_type` mutates a process-global registry from a constructor. If it goes in, `semantica/utils/constants.py` next to the format list is a better home than `FileTypeDetector.__init__`.
- Tightening `detect_type(...) == suffix.lstrip(".")` and parametrizing the ingest test over both suffixes — agreed, I flagged the same loose assertion independently.

My approval stands on the constants change itself, which I tested and found correct and contained. The CLI question is a scope call for a maintainer.

ZohaibHassan16 (approved):
Merging with a commit of mine to address the review I posted earlier.

PR #1508, opened 2026-09-07
Fix/parse and qdrant vector store

yanyushuai:
## Description

Three fixes that came out of running PDF → knowledge-graph flows end to end, plus a round of fixes for the Qodo review findings.

### 1. Parser dispatch recursed infinitely (`semantica/parse/methods.py`)

The six built-in parser dispatchers (`parse_method`, `chunk_method`, `extract_method`, `transform_method`, `validate_method`, `export_method`) registered themselves as their own `default` custom method. Any call with `method="default"` dispatched to itself until `RecursionError`, instead of falling through to the built-in implementation. The self-registration is removed — built-in defaults now run directly, and user-registered overrides still take precedence.

### 2. The Qdrant backend was unusable through `VectorStore` (`semantica/vector_store/`)

- `store_vectors()` now dispatches to `QdrantStore.insert_vectors()` (it only looked for `add`/`add_vectors`, so qdrant raised `NotImplementedError`).
- `search_vectors()` uses `client.query_points()` on current qdrant-client releases, falling back to the legacy `client.search()` on old clients (`search` was deprecated in 1.10 and removed later).
- Lazy collection init: insert/search auto-create or attach the configured collection from the observed vector dimension (mirrors `FAISSStore`), so no explicit `create_collection()` is needed.
- Plain-text documents passed to `store()` are kept in the vector payload (`{"document": ...}`) instead of being silently dropped.

**Qodo review fixes (latest commit):**
- `store_vectors()` returned `insert_vectors()`'s upsert status dict although the facade promises a `List[str]` of vector IDs (decision storage indexes the result at position 0). It now returns the generated or caller-supplied IDs after a successful insert.
- A non-empty `ids` list shorter than the vector batch was silently truncated by `zip(vectors, ids)` while the completion message still reported the full batch as inserted. A count mismatch now raises `ValidationError` before any write.
- `_ensure_default_collection()` looked up the legacy `collection` config key, so the documented `collection_name="..."` option was ignored and lazy init always fell back to `semantica_default`. `collection_name` now wins; `collection` still works as an alias.
- The new `parse-pdf` extra was missing from both aggregate `all` bundles, so `semantica[all]` still installed without `pdfplumber` and the default `PDFParser` raised `ProcessingError` on first use. `parse-pdf` is now in both.

### 3. Dependencies were unsatisfiable on Python 3.9 (`pyproject.toml`)

Several floors referenced versions that had dropped Python 3.9 (`requires_python >= 3.10`), making plain `pip install semantica[...]` fail resolution on 3.9:

- `snowflake-connector-python`, `pyarrow` (also `ingest-parquet`/`ingest-arrow`), and `pre-commit` are capped below their 3.10-only lines on 3.9; 3.10+ keeps the newer floors.
- `docling`, `litellm`, `crewai`, `google-adk` have no un-yanked 3.9-compatible release at all and are gated with `python_version >= "3.10"` markers.
- Supported floor raised to `>= 3.9.2` (matching what the constrained deps actually resolve on).
- New `parse-pdf` extra (`pdfplumber`) so the default PDF parser's dependency is installable explicitly.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Qodo review thread on this PR (4 findings — all addressed in the latest commits).

## Changes Made

- `semantica/parse/methods.py` — stop the built-in dispatchers from registering themselves as the `default` custom method (infinite recursion).
- `semantica/vector_store/vector_store.py` — dispatch `store_vectors()` to `QdrantStore.insert_vectors()`, return the stored IDs, keep plain-text documents in the payload.
- `semantica/vector_store/qdrant_store.py` — `query_points` search with legacy fallback, lazy `_ensure_default_collection()` honoring `collection_name`, ids/vectors count validation.
- `pyproject.toml` — Python 3.9-satisfiable version constraints, `parse-pdf` extra, `parse-pdf` included in both `all` bundles.
- `requirements-ci.txt` — regenerated with `uv pip compile pyproject.toml --python-version 3.11 --python-platform linux --extra all --generate-hashes` so the CI lockfile check includes the new `pdfplumber`/`pdfminer-six` pins (this was the only `build` check failure on the previous push; no other version changed).
- Tests — new regression tests (`tests/vector_store/test_qdrant_facade_writes.py`, `tests/test_pyproject_extras.py`), removed the now-stale strict xfail for qdrant write dispatch, and migrated the qdrant search mocks from the removed `client.search` to `client.query_points` (3 tests were failing on the branch).

## Testing

- [x] Tested locally

### Test Commands

```bash
pytest tests/vector_store/ tests/parse/ tests/test_pyproject_extras.py -q
# vector-store + extras: 363 passed, 84 skipped, 6 xfailed
# parse:                 30 passed, 20 subtests passed
python -m build --wheel   # semantica-0.6.8-py3-none-any.whl
```

## Documentation

## Breaking Changes

**Breaking Changes**: No

`qdrant_store.search_vectors()` result rows still use the canonical schema; the only behavioral change is that `VectorStore.store_vectors()` on qdrant now actually returns IDs (it previously raised `NotImplementedError`).

## Checklist

## Additional Notes

The strict-xfail contract test `test_store_vectors_dispatch_resolves[qdrant]` XPASSed once the write dispatch landed — per that file's convention, the stale marker was removed (`_NO_WRITE_DISPATCH` now only holds pinecone/weaviate).


KaifAhmad1 (approved):
Reviewed in depth, resolved the merge conflicts against `main`, and fixed a follow-up CI failure. All checks are now green and this is ready to merge.

**Findings from review:**

1. **Parser recursion fix (`semantica/parse/methods.py`)** — correct root-cause fix. Note: the crash this describes can no longer reproduce on current `main`, since #1528 (merged after this branch was cut) already added `if custom_method and custom_method != parse_document:` guards to all six dispatchers as a band-aid. This PR removes the self-registration entirely instead, which is the cleaner fix (no per-dispatcher guard to remember for future dispatchers) — worth taking even though the acute bug is already contained.

2. **Qdrant `VectorStore` facade dispatch** — genuinely useful, not overlapping with #1530. Verified against current `main`: `store_vectors()` still only dispatches to `add`/`add_vectors`, so it really does raise `NotImplementedError` for Qdrant today. The `insert_vectors` dispatch, ID-count validation, lazy `_ensure_default_collection()`, and plain-text-document payload retention are all correct — traced the call signatures against the actual `QdrantStore` methods.

3. **Qdrant search path** — overlapped with #1530 (merged separately), which already replaced `.search()` with `.query_points()` unconditionally (version floor is now `qdrant-client>=1.10.0`), making this PR's `hasattr()`-gated dual-path fallback redundant. Resolved by keeping `main`'s simpler, already-tested version.

4. **Python 3.9 dependency gating** — spot-checked rather than taken on faith: downloaded `litellm==1.83.9`'s wheel metadata directly from PyPI and confirmed it pins `python-dotenv==1.0.1`, which does conflict with this repo's core `python-dotenv>=1.2.1` floor, and that `litellm==1.83.10` genuinely requires `Requires-Python: >=3.10`. The stated reasoning is accurate; didn't re-verify every other package individually but the pattern holds up.

**Follow-up work pushed to this PR:**
- Resolved the `pyproject.toml` and `semantica/vector_store/qdrant_store.py` merge conflicts against `main` (commit `6031240`) — kept `main`'s `documents`/`all` extras structure and the already-merged `query_points`-only search path, layered this PR's `parse-pdf` extra and Qdrant facade fixes on top. Verified with the PR's own stated test command post-merge (507 passed, only 3 failures from qdrant-client not being installed in my local environment — unrelated to the merge).
- Fixed a CI failure the merge surfaced (commit `822248e`): `test_parse_methods_dynamic_default_resolution` asserted `"default"` resolves through the registry to `parse_document`, which was only true because of the self-registration this PR removes. Updated it to assert the correct post-fix behavior (`"default"` is unregistered, falls through to the built-in path).

All 12 CI checks pass on `822248e8` and the PR is mergeable. Approving.

PR #1488, opened 2026-09-06
test(gemini): regression tests for per-call model override

sakshi04-ui:
## Description

The Gemini provider's optional per-call `model=` argument was reported as behaving
inconsistently between `generate()` and `generate_structured()`. Investigation showed the
two methods are **already aligned** on current `main` — both honor a per-call `model`
override identically across the new `google.genai` client path and the legacy
`google.generativeai` path (the latter via `_legacy_client_for()`, added in #1262).

What was missing was **test coverage** locking that behavior in. This PR adds regression
tests only — no production code changes.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring 

## Related Issues

Closes #1268 

## Changes Made

- Add `tests/semantic_extract/test_gemini_model_override.py` covering `GeminiProvider` directly:
  - **New `google.genai` path** (`_use_new_genai=True`): `generate()` and `generate_structured()` both forward `model=` to `client.models.generate_content(...)`, and both fall back to the construction-time model when `model=` is omitted; a consistency check asserts both methods resolve to the same model string.
  - **Legacy `google.generativeai` path** (`_use_new_genai=False`): both methods resolve a per-call override through `_legacy_client_for()` (building a new `GenerativeModel`) and leave the construction-time client untouched; without an override both reuse the construction client; `generate_structured()` still forwards `generation_config` (temperature etc.) on this path.
- Add `test_per_call_model_override_forwarded_for_both_methods` to `tests/test_llm_gemini.py`, asserting the `semantica.llms.Gemini` wrapper passes `model=` through for both methods.
- Tests follow existing conventions: `patch.object(GeminiProvider, "_init_client", ...)` + injected mock client (as in `test_provider_limits.py`), pytest-style classes, `monkeypatch.setitem(sys.modules, ...)` to stub the optional SDK (as in `test_retry_logic.py`).

## Testing

- [x] Tested locally
- [x] Added tests for new functionality

### Test Commands

```bash
pytest tests/semantic_extract/test_gemini_model_override.py tests/test_llm_gemini.py
# 19 passed

pytest tests/semantic_extract/ tests/test_llm_*.py
# 207 passed, 11 skipped

black --check tests/semantic_extract/test_gemini_model_override.py tests/test_llm_gemini.py
[...]

Issue #1268, opened 2026-08-30
Gemini provider: make per-call model override behavior consistent across generate methods


Sameer6305:
The per-call `model=` behavior has already been resolved by #1262.

#1262 aligned the model-selection logic between `generate()` and `generate_structured()` across both supported Gemini SDK paths (`google.genai` and legacy `google.generativeai`).

The remaining work is to add regression tests covering per-call `model=` overrides for both generation methods across both SDK paths, so this behavior is protected against future regressions.

sakshi04-ui:
Hey @Sameer6305
I can take this one! I'll align the per-call `model` override and generation configuration logic between `generate()` and `generate_structured()` across both `google.genai` and `google.generativeai` SDK paths, and add regression tests for both methods.

sakshi04-ui:
Hi @Sameer6305 — I looked into reproducing this and I believe it's **already resolved by #1262**.

Your review on that PR flagged this exact point ("Non-blocking — Gemini legacy per-call model handling"), and @ZohaibHassan16 fixed it in commit `2f8a98e1` ("fix(llms): address review findings"), squash-merged as `ea0dd17b`. That commit:

- added `GeminiProvider._legacy_client_for()` plus a `_legacy_model_cache` (the per-call model-instance caching you suggested), and
- routed **both** `generate()` and `generate_structured()` through that same helper on the legacy `google.generativeai` path (previously `generate()` reused the init-time client while `generate_structured()` had its own inline handling).

On current `main`, `generate(model=...)` and `generate_structured(model=...)` now use identical model-selection logic across both SDK paths (`google.genai` and legacy `google.generativeai`). I verified all four combinations honor the per-call override with a small probe.

The one thing #1262 did **not** add is the regression tests this issue asks for — there's currently no `GeminiProvider`-level test asserting a per-call `model=` override for either method. I'd like to take that: a PR adding tests for both methods × both SDK paths so this can't silently regress, then we close this out. Does that work?

Sameer6305:
@sakshi04-ui that works and thanks for reviewing main i will review it from my end and update the issue, till then you can raise the PR.

Sameer6305:
Hi @sakshi04-ui, I’ve updated the issue to reflect that the implementation is already resolved by #1262 and that the remaining work is adding regression tests.

You can go ahead and raise the PR for the test coverage. Once it’s up, I’ll review it from my end.

sakshi04-ui:
[...]

PR #1512, opened 2026-09-07
test(gemini): cover legacy-SDK per-model instance cache

### LINKED ISSUE #1269: Gemini legacy provider: avoid unnecessary GenerativeModel recreation for per-call model overrides
### Description

The legacy Gemini SDK path can create a new `GenerativeModel` instance when a different model is supplied for an individual request.

This is functionally acceptable, but repeated requests using the same alternate model can result in unnecessary model-object creation.

### Suggested improvement

Consider caching/reusing model instances per model name where appropriate, while keeping credential isolation and provider-instance behavior safe.

For example, a provider could maintain model instances keyed by model name rather than recreating the same model object for every request.

### Note

This is primarily a performance/implementation improvement and does not block the current provider-wrapper PR.

### Priority

Non-blocking / follow-up.

sakshi04-ui:
## Description

The legacy `google.generativeai` fallback path in `GeminiProvider` was reported (issue #1269) as potentially rebuilding a `GenerativeModel` object on every request that supplies a non-default `model=` override.

Investigation showed the caching is **already implemented** on `main` — it landed alongside the Gemini wrapper in #1262 (commit `ea0dd17b`). `GeminiProvider._legacy_client_for()` keeps a per-instance cache keyed by model name and re-asserts `genai.configure(api_key=...)` with the instance's own key before each use (the legacy SDK holds the API key as module-level state).

What was missing is **test coverage locking that behavior in**. PR #1488 (issue #1268) added `test_gemini_model_override.py`, but its fake `google.generativeai` module memoizes `GenerativeModel` internally, so those tests verify *which* model a per-call override resolves to — not *that the provider caches the instance*. This PR adds regression tests targeting that gap. No production code changes.

## Type of Change

- [x] Code refactoring 

## Related Issues

Closes #1269

## Changes Made

- Add `tests/semantic_extract/test_gemini_legacy_model_cache.py` covering `GeminiProvider._legacy_client_for()` on the legacy `google.generativeai` path:
  - **`TestLegacyModelCacheReuse`** — a repeated per-call `model=` override across `generate()` and `generate_structured()` builds exactly one `GenerativeModel`; `_legacy_client_for()` cache hits return the same object; distinct override models are cached under separate keys; the construction-time model is never cached or rebuilt (the `requested_model == self.model` short-circuit).
  - **`TestLegacyModelCacheIsolation`** — `genai.configure()` is re-asserted with this provider's own key before every call, including cache hits; two `GeminiProvider` instances keep separate caches and separate module-level keys (no cross-instance credential bleed).
- The stub `google.generativeai` module in this file deliberately does **not** memoize `GenerativeModel` (unlike the fixture in `test_gemini_model_override.py`), so the constructor call count reflects the provider's own caching rather than the fake's.
- Tests follow existing conventions: `patch.object(GeminiProvider, "_init_client", ...)` + injected mock client (as in `test_provider_limits.py`), pytest-style classes, `monkeypatch.setitem(sys.modules, "google.generativeai", ...)` to stub the optional SDK (as in `test_gemini_model_override.py`).
- Note: `test_default_model_is_not_cached_or_rebuilt` lightly overlaps `test_without_override_uses_construction_client` from #1488, but adds an explicit assertion on the `_legacy_model_cache` dict.

## Testing

- [x] Tested locally

### Test Commands

```bash
pytest tests/semantic_extract/test_gemini_legacy_model_cache.py tests/semantic_extract/test_gemini_model_override.py tests/test_llm_gemini.py -q
# 25 passed

pytest tests/semantic_extract/ -q
# 211 passed, 11 skipped

black --check tests/semantic_extract/test_gemini_legacy_model_cache.py
isort --check --profile black tests/semantic_extract/test_gemini_legacy_model_cache.py
flake8 --max-line-length=88 --extend-ignore=E203,W503 tests/semantic_extract/test_gemini_legacy_model_cache.py
# all clean
```

## Documentation

## Breaking Changes

**Breaking Changes**: No

## Checklist

- [x] My code follows the project's style guidelines (Black / isort / flake8 clean)


KaifAhmad1 (approved):
Reviewed the diff and the underlying implementation (`GeminiProvider._legacy_client_for()` in `semantica/semantic_extract/providers.py`). Test-only change, no production code touched.

Verified locally:
- All 6 new tests pass.
- Mutated `_legacy_client_for()` to remove the per-model cache (always rebuild) — 4 of 6 tests immediately failed on the expected assertions, confirming these are real regression guards, not tautological mock checks.
- black / isort / flake8 clean as claimed.

The fixture correctly avoids the gap in `test_gemini_model_override.py`'s fake SDK (which memoizes `GenerativeModel` internally, masking whether the provider itself caches). Approving.

PR #1528, opened 2026-09-07
feat(deps): slim core dependencies and move 22 heavy packages to opti…

### LINKED ISSUE #1513: Slim core dependencies: move ~22 heavy packages to optional extras
## Goals — in priority order

1. **Don't break anyone.** Every existing call path either keeps working unchanged or fails with one clear, actionable line telling the user exactly which extra to install — never a bare `ModuleNotFoundError`, never a silent no-op, never a surprise at runtime in production. `semantica[all]` and pinning `<0.7.0` are both permanent, supported escape hatches for anyone who wants the old bundled behavior back. This ships behind a version bump and a migration note, not silently.
2. **Make Semantica dramatically lighter to install — this is a distribution lever, not just tidiness.** A `pip install semantica` that pulls torch + CUDA wheels, opencv, spacy, and librosa today is itself an adoption barrier: it's a slower, heavier, more failure-prone install for the majority of users who came for the graph/provenance/reasoning layer and never touch the ML/CV stack. It's also what stands between Semantica and being welcomed as someone else's *optional* dependency — no maintainer adds an optional integration that drags in gigabytes of transitive weight. A 22-package core install is a strictly better front door: faster `pip install`, fewer CI/Docker build failures downstream (see the wheel-availability issue that broke the Docker build in #1290/#6146 — that risk *shrinks* every time we depend on fewer heavy, fast-moving packages by default), and a much easier "yes" for any upstream project (LangChain, LlamaIndex, Qdrant, ...) asked to list Semantica as an optional extra. Lighter install → higher completed-install rate → more of the funnel described in the distribution playbook actually converts.

Both goals are satisfied by the same change: nothing here trades safety for size, or size for safety.

## What it looks like today vs. after

**Today (v0.6.x):**
```
$ pip install semantica
# pulls in, among others: torch, transformers, sentence-transformers, spacy,
# thinc, matplotlib, seaborn, plotly, ipywidgets, umap-learn, librosa,
# opencv-python, faiss-cpu, fastembed, onnxruntime, tokenizers, python-docx,
# openpyxl, lxml, beautifulsoup4, GitPython, gensim, + 22 lighter packages
# = 44 direct dependencies, multiple GB on disk (torch alone is ~800MB+ with
# its CUDA wheels on Linux), for a package whose own docstring's quickstart
# is a knowledge-graph/provenance example that touches none of them.
```

**After this change**, verified with a real clean-venv install (`pip install -e .`, no extras):
```
$ pip list
numpy  pandas  scipy  scikit-learn  rdflib  networkx  requests  chardet
protobuf  grpcio  pillow  pydantic  click  rich  tqdm  pyyaml  toml
python-dotenv  loguru  structlog  httpx  pyarrow
# 22 direct dependencies. import semantica still works immediately.
```

**And touching a moved feature fails cleanly instead of silently working or crashing raw** — this is the actual, tested behavior, not a hypothetical:
```
>>> import semantica                          # works, no extras needed
>>> from semantica.vector_store import FAISSStore
>>> FAISSStore(dimension=3).create_index()     # faiss-cpu not installed
ProcessingError: FAISS is not available. Install it with: pip install
'semantica[vectorstore-faiss]' (CPU) or 'semantica[gpu]' (GPU)

>>> from semantica.visualization import EmbeddingVisualizer  # matplotlib/plotly not installed
>>> EmbeddingVisualizer().visualize_2d_projection(x)
ProcessingError: Plotly is required for embedding visualization. Install
with: pip install 'semantica[viz]'
```
No bare `ModuleNotFoundError` three frames deep, no silent no-op — one line telling you exactly which extra to install. This is the mechanism that makes goal #1 possible at all.

## Why this is safe (the audit)

Every one of the ~22 packages falls into one of three buckets, confirmed by grepping the actual codebase, not assumed:

1. **Already guarded, just mis-classified as core.** `torch`, `faiss`, `sentence_transformers`, `fastembed`, `umap`, `plotly`, `gensim` are all lazily imported inside a `try/except` or a function body with their own `"pip install X"` error already. `spacy` goes through a shared `safe_import()` helper. `GitPython` goes through `ingest/__init__.py`'s lazy-attribute map, which already has a dedicated friendly error for it. None of these needed a single line of new guard code — just a `pyproject.toml` change.
2. **Eagerly imported but completely dead code.** `seaborn`, `ipywidgets`, `librosa`, `opencv-python`, `onnxruntime`, `tokenizers`, and `thinc` (as a direct import) have **zero references anywhere in the repository** — not `semantica/`, not `integrations/`, not `tests/`. They're installed for no reason today, so removing them from core changes nothing at runtime for anyone.
3. **Eagerly imported, needed a small guard.** `matplotlib` in `visualization/embedding_visualizer.py` and `visualization/ontology_visualizer.py` — turned out to be dead imports in both files too (never actually called), so wrapping them in the same `try/except None` pattern their neighboring `plotly`/`umap`/`graphviz` imports already use is a zero-risk fix. `python-docx`/`openpyxl`/`lxml`/`beautifulsoup4` have real (but narrow) usage in specific `parse`/`ingest` modules.

## Proposed extras mapping

| Move from core to | Extra | Notes |
| --- | --- | --- |
| `torch`, `transformers` | `models-huggingface` (exists) | Already duplicated there — just delete from core |
| `sentence-transformers`, `fastembed`, `onnxruntime`, `tokenizers` | `embeddings-local` (new) | The local/offline embedding pipeline |
| `spacy` | `nlp-spacy` (new) | `thinc` drops entirely — transitive via spacy already |
| `matplotlib`, `seaborn`, `plotly`, `ipywidgets`, `umap-learn` | `viz` (expand existing) | None of `visualization/` loads until touched via the lazy proxy |
| `librosa`, `opencv-python` | `media` (new) | Zero references anywhere in the repo today |
| `faiss-cpu` | `vectorstore-faiss` (new) | The one vector backend that wasn't already extras-gated like its siblings (`vectorstore-qdrant`, `-pinecone`, `-weaviate`, `-milvus`) |
| `python-docx`, `openpyxl`, `lxml`, `beautifulsoup4` | `documents` (new) | |
| `GitPython` | `ingest-git` (new) | |
| `gensim` | `graph-embeddings` (new) | Node2Vec; already duplicated in `split-topic` for BERTopic's unrelated use, which stays |

## Non-breaking checklist — must all be true before merge

- [ ] Bump to a new minor version (`0.7.0`) with a clear CHANGELOG/README migration note — `semantica<0.7.0` stays the escape hatch, `semantica[all]` restores the old bundled behavior
- [ ] Add the `try/except` guard to the two dead-but-unguarded `matplotlib` imports (see above) so removing the package doesn't turn a dead import into a hard `ModuleNotFoundError` on module load
- [ ] Sweep existing `"pip install X"` error strings (~15 of them) to reference the new `semantica[extra]` form where it improves discoverability — a few are done as part of this audit, most aren't yet
- [ ] CI matrix: run tests against core-only *and* `semantica[all]`, not just the latter, so a silently-required import can't slip back in
- [ ] Re-run the clean-venv verification above after the real change lands (it was done against a local draft, not yet against a merged version)
- [ ] Confirm no existing test asserts on an error-message string this changes (checked once already — `tests/kg/test_node_embeddings.py` and `tests/test_cli_commands.py` both match on substrings that survive this change, but re-check against the final diff)

I've done the audit and a local draft of the `pyproject.toml`/guard changes above — including the verified clean-venv output — but nothing is pushed yet. Happy to hand off the specifics to whoever picks this up, or open the PR myself once there's agreement on the extras names and the version bump.

ZohaibHassan16:
Closes #1513

### Summary

Reduces Semantica's direct core dependencies from 44 to 22, moving heavy, slow-moving, or specialized dependencies (e.g. PyTorch, spaCy, FAISS, document parsers, visualization libraries, and computer vision tools) into modular optional extras. 

A base `pip install semantica` is now lightweight, fast, and safe to import immediately. Full backward compatibility is preserved: `pip install "semantica[all]"` continues to install the complete bundled suite, and pinning `semantica<0.7.0` remains a permanent escape hatch.

### Key Changes

1. **Slim Core Dependencies (44 → 22)**:
   - Kept in core: `numpy`, `pandas`, `scipy`, `scikit-learn`, `rdflib`, `networkx`, `requests`, `chardet`, `protobuf`, `grpcio`, `pillow`, `pydantic`, `click`, `rich`, `tqdm`, `pyyaml`, `toml`, `python-dotenv`, `loguru`, `structlog`, `httpx`, `pyarrow`.

2. **Modular Optional Extras Defined**:
   - `models-huggingface`: `torch`, `transformers`
   - `embeddings-local`: `sentence-transformers`, `fastembed`, `onnxruntime`, `tokenizers`
   - `nlp-spacy`: `spacy`, `thinc`
   - `viz`: expanded to include `matplotlib`, `seaborn`, `plotly`, `ipywidgets`, `umap-learn`
   - `media`: `librosa`, `opencv-python`
   - `vectorstore-faiss`: `faiss-cpu` (also included in `vectorstore-all`)
   - `documents`: `python-docx`, `openpyxl`, `lxml`, `beautifulsoup4`
   - `ingest-git`: `GitPython`
   - `graph-embeddings`: `gensim` (also included in `graph-all`)
   - `all`: updated to include all newly modularized extras.

3. **Safe Lazy Construction & Fallbacks**:
   - `DOCXParser`, `ExcelParser`, `HTMLParser`, and `XMLParser` can be instantiated without error on `__init__()`, checking dependencies only when `.parse()` is invoked.
   - `XMLParser` automatically falls back to standard library `xml.etree` (`_parse_with_etree`) when `lxml` is absent.
   - Guarded top-level `matplotlib` imports in `EmbeddingVisualizer` and `OntologyVisualizer` with `try...except (ImportError, OSError)`.
   - Guarded `lxml` in `PublicAPIIngestor` and `RepoIngestor` imports.
   - Standardized user-facing error messages across CLI, vector stores, and parsers to provide exact install commands (`pip install 'semantica[<extra>]'`).

4. **Version Bump & Documentation**:
   - Bumped version to `0.7.0` across `pyproject.toml`, `semantica/__init__.py`, and `CITATION.cff`.
   - Added migration guide and extras catalog to `README.md` and `CHANGELOG.md`.
   - Recompiled CI lockfiles (`requirements-ci.txt`, `.github/requirements/base-deps.txt`, `.github/requirements/explorer-extra-py311.txt`, `.github/requirements/explorer-extra-py313.txt`).

### Verification

- Added comprehensive test suite `tests/test_issue_1513_slim_core.py` (11/11 passed):
  - Validates exact 22 core dependencies count in `pyproject.toml`.
  - Verifies all optional extras and their inclusion in `all`.
  - Verifies parser lazy construction and informative error messages on parse.
  - Verifies `XMLParser` standard library fallback.
  - Verifies missing-extra actionable hints across FAISS, spaCy, Node2Vec, and visualizers.
- Ran parser construction tests: `pytest tests/parse/test_parser_construction.py` (20 subtests passed).
- Ran CLI doctor tests: `pytest tests/test_cli_commands.py -k "test_doctor"` (passed).
- Ran parser and visualizer suites: `tests/parse/` (30 passed), `tests/visualization/` (58 passed).


KaifAhmad1 (changes requested):
@ZohaibHassan16 Thanks for the thorough audit and test coverage on this — the core approach (lazy construction, guarded imports, `semantica[all]`/`<0.7.0` escape hatches) is sound. Found two correctness issues that should be fixed before merge, plus one related consistency issue worth folding into the same pass.

## Blocking

**1. `XMLIngestor` raises the wrong exception type** ([`semantica/ingest/xml_ingestor.py:78`](https://github.com/semantica-agi/semantica/blob/slim-core-dependencies/semantica/ingest/xml_ingestor.py#L78))
Raises `ProcessingError` when lxml is missing, but every sibling guard (`RepoIngestor`, and the whole `_OPTIONAL_DEPENDENCY_MESSAGES` scheme in `ingest/__init__.py`) uses `ImportError` for a missing optional dependency. `ProcessingError` subclasses `SemanticaError`, not `ImportError`, so code written against the established `try: ... except ImportError: <fallback>` pattern won't catch this — the exception propagates uncaught instead of triggering the intended fallback.

**2. Missing-dependency detection for git/lxml silently breaks** ([`semantica/ingest/__init__.py:299`](https://github.com/semantica-agi/semantica/blob/slim-core-dependencies/semantica/ingest/__init__.py#L299), [`repo_ingestor.py:47`](https://github.com/semantica-agi/semantica/blob/slim-core-dependencies/semantica/ingest/repo_ingestor.py#L47), [`xml_ingestor.py:27`](https://github.com/semantica-agi/semantica/blob/slim-core-dependencies/semantica/ingest/xml_ingestor.py#L27))
Now that `repo_ingestor.py`/`xml_ingestor.py` guard `import git` / `from lxml import etree` internally (setting them to `None` instead of letting `ModuleNotFoundError` escape), `importlib.import_module(".repo_ingestor"/".xml_ingestor")` in `ingest/__init__.py`'s `__getattr__` always succeeds now — so the `except ModuleNotFoundError` branch that raises a friendly `ImportError(message)` for missing git/lxml is dead code for these two modules. A common probe pattern like:
```python
try:
    from semantica.ingest import RepoIngestor
    HAS_GIT = True
except ImportError:
    HAS_GIT = False
```
now reports `HAS_GIT = True` even without GitPython installed — the import always succeeds — and only crashes later, deep inside `if HAS_GIT` logic, when `RepoIngestor()` is actually constructed. This moves the failure from import-time to construct-time, which is a real behavior change from today.

## Worth fixing in the same pass

**3. Inconsistent exception tuples for optional imports** — the new guards in `repo_ingestor.py:47`, `xml_ingestor.py:27`, `docx_parser.py:37`, `excel_parser.py:36`, `html_parser.py:36`, and `public_api_ingestor.py:31` use `except (ImportError, ModuleNotFoundError)`, which is redundant (`ModuleNotFoundError` is already an `ImportError` subclass) and — unlike ~90 existing guards elsewhere in this codebase, including this same PR's own `ontology_visualizer.py`/`embedding_visualizer.py` — omits `OSError`. That's the established convention here specifically to catch native-extension/DLL load failures (e.g. a broken libxml2 binding on Windows), so as written, that failure mode would produce a raw traceback instead of the intended `pip install 'semantica[documents]'` hint.

## Non-blocking, optional follow-up
- `semantica/parse/methods.py`: the recursion fix patches 9 dispatchers with individual exclusion tuples rather than removing the root-cause self-registration at lines ~796-801 — fine as-is, but fragile for future dispatchers added to the registry.
- ~8 new guards hand-roll `try/except: X = None` instead of reusing the existing `safe_import()` helper in `semantica/utils/helpers.py` already used elsewhere for this.
- Minor dead code: `ingest/methods.py:877`'s old `_is_missing_dependency(exc, "git")` guard is now unreachable given the new guard at ~890; `embedding_visualizer.py:627`'s `opts.pop("n_components")` is unreachable given the call-site strips.

Happy to re-review once #1 and #2 are addressed.

KaifAhmad1 (approved):
LGTM

PR #1530, opened 2026-09-07
fix: update Qdrant client compatibility

Sameer6305:
## Summary

Update the Qdrant vector store integration for compatibility with current `qdrant-client` releases.

## Changes

- Replace the removed `QdrantClient.search()` API with `query_points()`.
- Preserve Semantica's existing `search_vectors()` result structure and metadata handling.
- Update `get_stats()` to use `points_count` as the fallback for the removed `vectors_count` field.
- Update the minimum `qdrant-client` version to `>=1.10.0`.
- Add focused regression tests covering Qdrant search and collection statistics compatibility.

## Validation

- 49 relevant Qdrant/vector-store tests passed.
- Runtime integration flow verified against `qdrant-client 1.19.0`.
- Verified connect → create collection → insert vectors → search.
- Verified metadata filtering, collection statistics, retrieval, pagination, and deletion.
- `git diff --check` passes.

## Context

This was identified while validating the Qdrant integration example for upstream Qdrant framework documentation. The existing QdrantStore API is retained; this PR updates its implementation to remain compatible with current `qdrant-client` versions.


KaifAhmad1 (approved):
Reviewed and verified against the actual `qdrant-client` library, not just the mocked tests — installed 1.19.0 in isolation to check:

- `CollectionInfo` genuinely drops the `vectors_count` field in 1.19.0 (confirmed via `model_fields`), so the `getattr(collection_info, "vectors_count", fallback)` pattern matches real behavior, not just the `MagicMock(spec=...)` doubles.
- `CollectionParams.vectors` is really `Union[VectorParams, Dict[str, VectorParams], None]`, so the `isinstance(vectors_cfg, VectorParams)` check correctly separates single-vector collections (safe to fall back to `points_count`) from named/multi-vector ones (correctly returns `None` instead of a misleading total — this is the "operators see false vector totals" fix, genuinely resolved).
- `QdrantClient.query_points()`'s real signature accepts every kwarg used here, and returns `QueryResponse.points: List[ScoredPoint]` — matches what the new code assumes.
- The `QDRANT_AVAILABLE and isinstance(...)` short-circuit avoids crashing when qdrant-client isn't installed (`VectorParams = None` in that branch).

19 new tests cover the API call shape, filter forwarding, all three score-normalization regions, empty results, error wrapping, and all four `vectors_count` fallback branches. Ran them locally both with and without qdrant-client installed — pass in both cases.

Also resolved the merge conflict in `pyproject.toml` (main had moved after #1528/#1529 merged) and pushed `5606767a` — kept both the new `vectorstore-faiss` line from main and this PR's `qdrant-client>=1.10.0` bump. Re-ran the full Qdrant test suite post-merge (20/20 passed) and confirmed all CI checks are green on the merge commit.

Approving.

PR #1532, opened 2026-09-08
fix(explorer): place static causal-distance route before dynamic decision_id route

### LINKED ISSUE #1531: GET /api/decisions/causal-distance returns 404 because of route ordering
I noticed that the `GET /api/decisions/causal-distance` endpoint currently can't be reached.

In `semantica/explorer/routes/decisions.py`, the dynamic route:

```text
/api/decisions/{decision_id}
```

is registered before the static:

```text
/api/decisions/causal-distance
```

route. Because of Starlette's route matching order, a request to `causal-distance` gets treated as a `decision_id`.

### What happens

For example:

```bash
curl 'http://127.0.0.1:8000/api/decisions/causal-distance?source=A&target=B'
```

returns:

```json
{"detail":"Decision 'causal-distance' not found"}
```

with HTTP `404`.

This is a bit misleading because `causal-distance` is supposed to be an actual endpoint, not a decision ID.

The endpoint is also documented in:

* `docs/reference/explorer.md`
* `docs/reference/distance.md`
* the FR-8 changelog entry

### What I expected

The request should reach `causal_distance()` and return a `CausalDistanceReport`.

For invalid source/target IDs, a validation error such as `422` would be fine, but the request shouldn't end up as:

```text
Decision 'causal-distance' not found
```

### How to reproduce

1. Start the Explorer backend.
2. Run:

```bash
curl -s -w '\n%{http_code}\n' \
  'http://127.0.0.1:8000/api/decisions/causal-distance?source=x&target=y'
```

3. The response is a `404` saying that `causal-distance` is not a decision.

This makes a documented Decision Intelligence endpoint effectively unusable.

There also doesn't appear to be a frontend caller for this endpoint currently, so the issue isn't being caught by the existing frontend flow.

nilaymallikk:
Closes #1531.

## Problem
`GET /api/decisions/causal-distance` was unreachable. The static route was defined after `GET /api/decisions/{decision_id}` in `semantica/explorer/routes/decisions.py`, so Starlette bound `decision_id="causal-distance"` and returned `404 Decision 'causal-distance' not found`.

## Repro
```bash
curl 'http://127.0.0.1:8000/api/decisions/causal-distance?source=x&target=y'
# before: {"detail":"Decision 'causal-distance' not found"} (404)
# after:  200 with CausalDistanceReport (unreachable-distance report for unknown ids)
```

## Changes
- Moved the `/causal-distance` handler above `/{decision_id}` plus a comment guard against reordering regressions.
- New `tests/explorer/test_decisions_causal_distance_route.py` (3 tests): unreachable-ids report, 1-hop path between linked decisions, unknown-id 404 unchanged.

## Testing
- New tests: 3 passed (2 of 3 failed pre-fix, confirming the bug).
- Existing: `test_decision_route_timestamp.py` + `test_explorer_api.py` — 129 passed.


ZohaibHassan16 (approved):
LGTM. Thanks for the quick fix and the solid regression tests.

Good to merge.

Issue #1510, opened 2026-09-07
[FEATURE] Schema-guided extraction: validate/condition extraction against a domain ontology, and bootstrap a draft schema from data


Besokus:
### Problem Statement

Semantica can *induce* an ontology from extracted `{entities, relationships}` (`OntologyGenerator`), but there is no supported way to go the other direction: **reuse an ontology as an extraction schema** to constrain what gets extracted, or **learn such a schema from a sample of your own data** so future extraction stays on-domain. As a result, extraction output can contain concepts/relations outside the intended domain vocabulary (noise, drift), and there's no clean path from "unconstrained extraction" to a domain-conforming graph. The existing `ExtractionValidator` only filters by *confidence* — it never checks whether an entity/relation belongs to a domain vocabulary or satisfies a predicate's `domain`/`range`.

### Proposed Solution

A generic, domain-agnostic **schema-guided extraction** capability — two complementary parts plus one optional enhancement:

1. **`SchemaValidator` (deterministic)** — constrain extraction output (`entities`/`relations`) to a domain ontology: drop concepts outside the vocabulary, predicates not in the schema, and relations violating `domain`/`range`. A **different axis** from confidence-based `ExtractionValidator` — the two compose.
2. **`bootstrap_schema`** — induce a **draft** `Ontology` (concepts + object properties with domain/range) from a sample of `{entities, relationships}`, serialized to **TTL for a human to review, edit, and commit**. Never auto-applies.
3. **(optional) schema-conditioned prompting** — inject the allowed concepts/predicates into the LLM prompt to raise precision, layered on the existing extractor.

Together they close the loop: *unconstrained extraction → frequency-thresholded induction (draft) → human review → schema-gated / schema-conditioned extraction → schema-conforming graph.* This is the classic **OBIE + ontology-learning + middle-out** pattern.

### Alternatives Considered

- **A dedicated `DomainTemplate` type** — rejected: an OWL `Ontology` already expresses concepts + object properties with `domain`/`range`; a parallel type duplicates representation. Instead, expose a thin `ExtractionSchema` *view* over `Ontology`.
- **Auto-applying the induced schema (no human step)** — rejected: OntoClean/UFO meta-properties (rigidity, identity, unity) are modeling judgments that cannot be inferred from frequency; ontology-learning literature is explicit that automatic induction yields *proposals*, not final ontologies. Emit a draft for human ratification.
[...]

PR #1537, opened 2026-09-08
feat(ontology): bootstrap a draft domain schema from extracted data (PR 2/3)

pkupt:
PR 2/3 for #1510, on top of #1527.

`bootstrap_schema()` induces a draft ontology from a sample of extracted `{entities, relationships}`. The intended loop: extract a sample, bootstrap a draft, ratify it by hand, then hand the ratified schema to `SchemaValidator` to gate future extraction. Where #1527 constrains extraction against an existing ontology, this induces that ontology from data in the first place.

It is a thin wrapper over the existing `OntologyGenerator` pipeline rather than a parallel implementation: induction, domain/range inference and TTL export all stay in the pipeline. What the wrapper adds is the `min_occurrences` frequency gate, threaded through both class and predicate inference, and the draft contract: results always come back `draft: True` with a TTL for review and are never auto-applied, since frequency alone yields proposals, not final ontologies.

One wiring detail worth knowing: `min_occurrences` is passed both to the `OntologyGenerator` constructor and to `generate_ontology()`, because the class gate reads instance config while the predicate gate reads per-call options.

12 new tests in `tests/test_bootstrap_schema.py` cover the draft contract, class/property induction with domain/range, the gate dropping sparse types and sparse predicates, TTL emission, and empty input. Local run together with the schema validator suite: 29 passed.


ZohaibHassan16 (approved):
LGTM. Ready to merge

PR #1538, opened 2026-09-08
fix(cli): fail when kg build produces an empty knowledge graph

### LINKED ISSUE #1352: [BUG] kg build reports success while logging "No entities or relationships found" — finds 0 entities where extract finds 7, persists nothing
## Summary

`semantica kg build --source <file>` prints `✓ Knowledge base built — 1 source(s) processed` while logging `WARNING - No entities or relationships found in results`, and writes nothing to the configured Neo4j backend. The exit status and the success message do not reflect the warning.

Crucially, **the input is not the problem**: `semantica extract` finds 7 entities in the same file.

## Reproduction

```bash
export GRAPH_STORE_DEFAULT_BACKEND=neo4j
export GRAPH_STORE_NEO4J_URI=bolt://127.0.0.1:7689
export GRAPH_STORE_NEO4J_USER=neo4j
export GRAPH_STORE_NEO4J_PASSWORD=<redacted>

cat > /tmp/neutral.txt <<'TXT'
Ada Lovelace worked with Charles Babbage in London during 1843.
Acme Corporation opened an office in Berlin in March 2020.
TXT

# CONTROL — the extract command finds entities in this exact file:
semantica extract /tmp/neutral.txt --mode ner --method ml --output /tmp/out.json
# -> "Extracted 7 entities using ml"

# count before: MATCH (n) RETURN count(n) -> 6

semantica kg build --source /tmp/neutral.txt
# WARNING - No entities or relationships found in results
# WARNING - No text content found in results for embedding
# CORE:
# PIPELINE:
#  ✓ Knowledge base built — 1 source(s) processed.  1.0s

# count after:  MATCH (n) RETURN count(n) -> 6    <-- unchanged
```

## Expected

Either the build extracts the same entities `semantica extract` finds and persists them, or it fails loudly. A warning that "no entities were found" should not be followed by a success message claiming the knowledge base was built.

## Analysis

The extraction inside the `kg build` pipeline does not appear to be the extraction the `extract` command runs — `extract --mode ner --method ml` returns 7 entities with `metadata.extraction_method == "ml"` (spaCy `en_core_web_sm` installed), while the build pipeline reports zero on identical input.

Two separable problems:
1. The build pipeline finds no entities where the extractor does.
2. Even given the warning, the command reports success and exits 0.

(2) is arguably the more dangerous half: it turns a total failure into a green tick.

## Environment

semantica 0.6.7 (tag), Python 3.12.14, macOS arm64, Neo4j 2026.07.1 Community, neo4j driver 6.2.0, spaCy `en_core_web_sm` 3.8.0 installed.

dex0shubham:
## Description

`semantica kg build --source <file>` printed `✓ Knowledge base built — 1 source(s) processed` and exited 0 while writing nothing, having logged `No entities or relationships found in results` immediately beforehand.

The source count was the only thing checked. With no pipeline configured the default pipeline is a single step with no handler, so the executor returns its input unchanged (`semantica/pipeline/execution_engine.py:692-695`) — every source is counted as "processed" while the graph stays empty.

This makes that state visible instead of reporting success. It does not change what the default pipeline does; that is the separate half of #1352.

## Type of Change

## Related Issues

Closes #1352

## Changes Made

- `kg build` now inspects the resulting graph, not just the source count. An empty graph after processing raises a `ClickException` naming the likely cause and pointing at `semantica extract` to check the input.
- The success message reports entity and relationship counts, so "1 source processed" can no longer stand in for "something was built".

## Testing

- [x] Tested locally

Before:

```
$ semantica kg build --source neutral.txt
WARNING - No entities or relationships found in results
 ✓ Knowledge base built — 1 source(s) processed.
$ echo $?
0
```

After:

```
$ semantica kg build --source neutral.txt
Error: 1 source(s) processed but the knowledge graph is empty — no entities or
relationships were extracted. This usually means no pipeline was configured, so
the sources were read but never parsed or extracted. Supply a pipeline with
--config, or use 'semantica extract' to check the sources yield entities.
$ echo $?
1
```

`python -m pytest tests/test_cli_commands.py -q`: 290 passed, 2 skipped. flake8 on `semantica/cli.py` unchanged from main.

## Scope

This is the "fail loudly" half. Making the default pipeline actually ingest/parse/extract is the larger behaviour decision raised on #1352 and is not attempted here — happy to follow up once you have picked a direction there.

### DIFF
--- a/semantica/cli.py
+++ b/semantica/cli.py
@@ -328,8 +328,30 @@ def _run_build(cli_ctx: CLIContext, sources: Sequence[str]) -> None:
 
     stats = result.get("statistics", {}) if isinstance(result, dict) else {}
     processed = stats.get("sources_processed")
+
+    # A source count alone does not mean anything was built: with no pipeline
+    # configured the default pipeline passes its input through untouched, so
+    # every source is "processed" and the graph stays empty. Reporting success
+    # there hid the failure completely (#1352).
+    graph = result.get("knowledge_graph", {}) if isinstance(result, dict) else {}
+    entity_count = len(graph.get("entities") or [])
+    relationship_count = len(graph.get("relationships") or [])
+
+    if processed and not entity_count and not relationship_count:
+        raise click.ClickException(
+            f"{processed} source(s) processed but the knowledge graph is empty "
+            "— no entities or relationships were extracted. This usually means "
+            "no pipeline was configured, so the sources were read but never "
+            "parsed or extracted. Supply a pipeline with --config, or use "
+            "'semantica extract' to check the sources yield entities."
+        )
+
     if processed is not None:
-        _ok(cli_ctx, f"Knowledge base built — {processed} source(s) processed.")
+        _ok(
+            cli_ctx,
+            f"Knowledge base built — {processed} source(s) processed, "
+            f"{entity_count} entities, {relationship_count} relationships.",
+        )
     else:
         _ok(cli_ctx, "Knowledge base build completed.")
 


ZohaibHassan16 (changes requested):
Thanks for putting this together.

Some issues: `test_build_result_with_stats_shows_source_count` in `tests/test_cli_foundation.py` is failing with this. The mock there returns `{"statistics": {"sources_processed": 3}}` without a `knowledge_graph` key, so it hits the new check and exits 1 instead of 0. we probably only want to trigger the empty check if `"knowledge_graph"` is actually in `result`.

Also noticed a bug with multiple sources around line 326 of `cli.py`: the progress loop overwrites `result` on each file (`result = framework.build_knowledge_base(sources=[src])`), so `result` at the end is just whatever the last source returned. If someone passes 3 files and the first two extract 50 entities but the last one has 0, the whole command blows up with exit 1. and when it does succeed, it only prints the counts from the last file. We'll need to aggregate the counts across the loop.

Mind fixing those and adding a couple of tests in `tests/test_cli_foundation.py` to cover the new exit 1 and count output?

ZohaibHassan16 (approved):
LGTM. Ready to merge

Issue #1481, opened 2026-09-05
[BUG] CLI reports 'Graph Store: memory' then fails demanding Neo4j — _get_graph_store defaults differently from the other two call sites


dex0shubham:
## Summary

On a plain `pip install semantica`, the CLI reports that it is using the in-memory graph store and then fails demanding Neo4j. Both happen in the same session, with no configuration and no environment variables set:

```
$ semantica
  Graph Store      memory

$ semantica decision list
Neo4j is not available. Install it with: pip install neo4j
```

`neo4j` ships only in the optional `graph-neo4j` extra, so this is the out-of-the-box experience for anyone following the README quickstart (`pip install semantica`).

## Cause: one value, two different defaults

`semantica/cli.py` resolves the graph-store backend in three places. Two default to `memory`, one defaults to `neo4j`:

```python
# cli.py:449 — status display
cli_ctx.store_backend or cfg.get("graph_db", {}).get("backend", "memory")

# cli.py:835
backend = cli_ctx.store_backend or cfg.get("graph_db", {}).get("backend", "memory")

# cli.py:1154 — _get_graph_store(), the one that actually builds the store
backend = cli_ctx.store_backend or graph_db.pop("backend", "neo4j")
```

So the status panel truthfully reports what two of the three call sites would use, while the site that constructs the real `GraphStore` picks a different backend. `_get_graph_store` looks like the outlier rather than the intended default.

## Affected commands

Verified on `main` (`30fc6447`), no env vars set:

```
decision list                  Neo4j is not available
decision record --title Probe  Neo4j is not available
export --format json           Neo4j is not available
```

`kg stats` and others that do not go through `_get_graph_store` are unaffected.

## The Python API does not have this problem

Worth noting because it rules out "decisions need a database". The README quickstart works verbatim on the same install:

```python
from semantica.context import ContextGraph

graph = ContextGraph(advanced_analytics=True)
decision_id = graph.record_decision(
    category="vendor_selection",
    scenario="Choose cloud provider for HIPAA workload",
    reasoning="AWS offers BAA",
    outcome="selected_aws",
    confidence=0.93,
)
similar = graph.find_similar_decisions("cloud vendor", max_results=5)
```

Both calls succeed in memory. It is only the CLI that insists on Neo4j, and only via that one call site.

## Which direction would you prefer?

[...]

PR #1539, opened 2026-09-08
fix(cli): report the graph backend actually used, not a nonexistent memory one

dex0shubham:
## Description

The CLI reported a graph-store backend that `GraphStore` cannot construct, and `doctor` returned a passing check for it:

```
$ semantica
  Graph Store      memory

$ semantica doctor
  Graph store      ✓   memory (always available)

$ semantica decision list
Neo4j is not available. Install it with: pip install neo4j
```

`GraphStore._initialize_store_backend()` only branches on `neo4j`, `falkordb`, `neptune` and `age`; `GraphStore(backend="memory")` raises `ValidationError: Unknown backend: memory`. The two reporting sites got away with claiming `memory` because they special-cased it and never built a store, while `_get_graph_store()` defaulted to `neo4j` and produced the error above.

This aligns the reporting with the backend that is actually used. It deliberately does **not** change `_get_graph_store()`'s default, so no existing setup changes behaviour.

## Type of Change

## Related Issues

Refs #1481

## Changes Made

- Status panel reports the backend that will actually be used (`neo4j` when unset) rather than `memory`.
- `doctor` no longer short-circuits on `memory` with a green "always available". It probes the resolved backend and reports the real result, so an unusable graph store is surfaced as a failure instead of a pass.

## Testing

- [x] Tested locally

After:

```
$ semantica
  Graph Store      neo4j

$ semantica doctor
  Graph store      ✗   Neo4j is not available
```

`python -m pytest tests/test_cli_commands.py -q`: 290 passed, 2 skipped. flake8 on `semantica/cli.py` unchanged from main.

## Scope

Deliberately partial, per the discussion on #1481. @KaifAhmad1's preference there is a real `MemoryGraphStore` with `_get_graph_store()` defaulting to it, which is the fix that makes all three call sites and the Python API genuinely agree. That needs a decision first — the backend contract includes Cypher (`DecisionQuery` issues `MATCH (d:Decision {category: $category}) RETURN d`, with 16 Cypher statements in that file and 72 `execute_query` call sites across `context/` and `kg/`), so a memory backend needs either an embedded Cypher engine or a subset interpreter.

This PR is the stopgap described in that thread: it stops `doctor` giving a green check for something no command can use, without pre-empting that decision. #1481 should stay open for the real backend.

### DIFF
--- a/semantica/cli.py
+++ b/semantica/cli.py
@@ -465,7 +465,7 @@ def _show_startup(cli_ctx: CLIContext) -> None:
         return
     cfg = cli_ctx.config.to_dict()
     graph_store = (
-        cli_ctx.store_backend or cfg.get("graph_db", {}).get("backend", "memory")
+        cli_ctx.store_backend or cfg.get("graph_db", {}).get("backend", "neo4j")
     )
     vector_store = (
         cli_ctx.vector_store_backend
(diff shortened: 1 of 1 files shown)


ZohaibHassan16 (approved):
LGTM, regarding security check, that is phantom CI failure that doesn't have anything to do with this PR or the repo for that matter. There is some sort of niche bug in accelerate library by Hugging Face

PR #1541, opened 2026-09-09
fix(explorer): end the temporal scrubber at now instead of inventing 2030

### LINKED ISSUE #1536: [BUG] Temporal scrubber invents a 2030 upper bound and starts the playhead in the future when /temporal/bounds returns max: null
## Bug Description

When `GET /api/temporal/bounds` returns `max: null` — which is the normal result for a graph whose nodes carry `valid_from` instants and no `valid_until` — `TimelinePanel` substitutes a hardcoded `2030-01-01` upper bound and then places the playhead at the **midpoint** of `[min, fake max]`. The scrubber therefore opens at a timestamp that is years in the future, and the workspace's first `/api/temporal/snapshot` request is issued for that future time.

`explorer/src/workspaces/GraphWorkspace/TimelinePanel.tsx` (blob `6535c2f3f654a7c158b79ec929781ba5afb7af10`, identical on `main` and tag `v0.6.7`):

- L15 — `const DEFAULT_MAX_DATE = new Date("2030-01-01T00:00:00Z");`
- L76 — `const maxBound = useMemo(() => safeDate(maxDate, DEFAULT_MAX_DATE), [maxDate]);`
- L77 — `const defaultTime = useMemo(() => new Date(Math.round((minBound.getTime() + maxBound.getTime()) / 2)), [maxBound, minBound]);`

`session.get_temporal_bounds()` deliberately leaves `max` open when no node carries `valid_until` (`semantica/explorer/session.py`), so this is not a malformed-data case — it is the expected response shape for event-instant graphs.

## Steps to Reproduce

1. Build a graph whose nodes carry only `valid_from` (e.g. every node's validity comes from an `occurred_at`-style instant). `GET /api/temporal/bounds` returns `{"min": "2026-01-...", "max": null}`.
2. Open the Knowledge Explorer graph workspace with the temporal panel visible.
3. Read the scrubber header and the playhead label.

## Expected Behavior

- The header reflects the real data range; with `max: null` the graph is only known up to **now**, so the window should end at the present, not at an arbitrary future year.
- The playhead starts at "now" (clamped into `[min, max]`), so the initial snapshot describes the current state.

## Actual Behavior

- Header reads `Temporal Scrubber · 2026-2030`; the 2030 is pure fiction — no data supports it.
- Playhead defaults to ~2028, and `GraphWorkspace.tsx` (L1532) fetches `/api/temporal/snapshot?at=<future timestamp>`, so the active-node count shown on load is a projection of a future time rather than the present.
- Three further settings are tuned for the fictional ~60-year default window and degrade on real ranges: `zoomMin` is 1 year (L94), `timeAxis: { scale: "year", step: 5 }` (L98) renders one or two labels when the true span is months, and `PLAY_STEP_MONTHS = 6` (L18) gives ~2 frames per loop on a one-year span.

## Environment

- `explorer` at `main` (file unchanged since `v0.6.7`); Python package `semantica` 0.6.8
- Reproduced in Chrome on macOS

## Proposed Fix

We are running this patch in a downstream copy and it resolves all four points:

1. Drop `DEFAULT_MAX_DATE`; fall back to a `now` captured once per mount (memoised, so the value is stable across renders and does not retrigger the timeline effect).
2. `defaultTime` becomes `now` clamped into `[minBound, maxBound]` instead of the midpoint.
3. Lower `zoomMin` to one day and stop pinning `timeAxis`/`format`, letting vis-timeline pick the granularity for the real span.
4. Replace `PLAY_STEP_MONTHS` with a span-proportional step, `max(oneDay, span / 60)`, so a play-through is ~60 frames at any range.

Happy to open the PR — **feel free to assign this issue to us.**

taoche:
## Description

`TimelinePanel` substituted a hardcoded `2030-01-01` upper bound whenever `GET /api/temporal/bounds` reported `max: null`, then placed the playhead at the **midpoint** of that fabricated window. `session.get_temporal_bounds()` deliberately leaves `max` open for any graph whose nodes carry `valid_from` instants and no `valid_until`, so this was the normal response shape rather than malformed data. Two visible consequences: the header advertised `Temporal Scrubber · 2026-2030` when no data supported 2030, and the workspace's first `/api/temporal/snapshot` request asked about a timestamp years in the future, so the active-node count on load described a projection rather than the present.

The fallback is now a `now` captured once per mount, and `defaultTime` is that same `now` clamped into `[minBound, maxBound]`. Three settings that were tuned for the fictional ~60-year default window follow from that change, as noted in the issue.

## Type of Change

## Related Issues

Closes #1536

## Changes Made

- `maxDate` falls back to `now` (memoised once per mount, so re-renders keep a stable reference and don't retrigger the timeline effect) instead of `DEFAULT_MAX_DATE`, which is removed.
- `defaultTime` is `now` clamped into the range, replacing the midpoint calculation that produced the future start time.
- `zoomMin` drops from one year to one day, and the `timeAxis: { scale: "year", step: 5 }` / `format` pinning is removed so vis-timeline picks a granularity appropriate to the real span — the old settings rendered one or two labels when the true range was months.
- The fixed `PLAY_STEP_MONTHS = 6` becomes `max(oneDay, span / 60)`, keeping a play-through at roughly 60 frames whether the graph spans months or a decade (it was ~2 frames on a one-year span).
- Bound and step arithmetic moves into `temporalScrubberBounds.ts` so it can be asserted without mounting vis-timeline, mirroring the existing `temporalLifecyclePredicates.ts` split.

Ordering guard worth flagging for review: when `min` is itself in the future and `max` is open, `maxBound` is clamped up to `minBound` so vis-timeline never receives `min > max`. There is a test for it.

## Testing

- [x] Tested locally

### Test Commands

```bash
cd explorer
npm run test:graph-workspace          # new file registered in this script
node --import tsx --test tests/temporalScrubberBounds.test.ts   # 11/11 pass
npx tsc -b                            # clean
npx eslint src/workspaces/GraphWorkspace/TimelinePanel.tsx \
  src/workspaces/GraphWorkspace/temporalScrubberBounds.ts \
  tests/temporalScrubberBounds.test.ts   # clean
```

The 11 new cases cover: open `max` ending at now, playhead starting at now, playhead explicitly *not* at the midpoint, a supplied `max` still being honoured, clamping for ranges that end before / start after now, the `min > max` ordering guard, malformed date fallbacks, and the ~60-frame play step at one-year, decade, and sub-day spans.

Two pre-existing `.tsx` suites in `test:graph-workspace` (`explorerCapabilities`, `markdownEditorInteraction`) failed in my sandbox with `ERR_MODULE_NOT_FOUND: jsdom` because of a stale local `node_modules`; they are untouched by this change and unrelated to it.

## Documentation

## Breaking Changes

**Breaking Changes**: No — the component's props are unchanged, and a graph that does report `valid_until` keeps exactly its previous window.


Sameer6305 (approved):
Reviewed the implementation and tests against #1536. The open-ended `max: null` case now correctly ends at `now`, the playhead is clamped to the current range, the timeline bounds remain ordered, and the proportional play-step behavior is covered by tests.

No blocking issues found. The remaining `2030` fallback in the overlay is pre-existing and cosmetic, so it doesn't block this PR.

**LGTM — approving.**

PR #1553, opened 2026-09-10
feat(llms): standardize generate_typed() across all provider wrappers

### LINKED ISSUE #1271: Improve generate_typed() API consistency across LLM provider wrappers
### Description

The LLM wrappers are not completely consistent in how `generate_typed()` is exposed and typed across providers.

The underlying `BaseProvider` already provides typed-generation functionality, but the public wrapper APIs do not all present it with the same explicit interface/documentation.

### Suggested improvement

Review the existing LLM wrappers and standardize:
- `generate_typed()` availability
- method signatures
- type annotations
- documentation
- test coverage

This would make the public `semantica.llms` API more predictable regardless of provider.

### Note

This appears to be broader API consistency work rather than a blocker specific to PR #1262.

### Priority

Non-blocking / follow-up.

sakshi04-ui:
## Description

The `semantica.llms` provider wrappers were inconsistent in how they exposed typed generation. `BaseProvider` already implements `generate_typed()`, and the newer wrappers (`Anthropic`, `Gemini`, `Ollama`, `DeepSeek`, `Novita`) surfaced it — but `Groq`, `OpenAI` and `HuggingFaceLLM` didn't expose it at all, and `LiteLLM` (which doesn't sit on `BaseProvider`) had no path to it.

This PR makes all nine wrappers present the same generation interface: `generate()`, `generate_structured()`, `generate_typed()`, `is_available()` — so the public API is predictable regardless of provider.

## Type of Change

- [x] Documentation update

## Related Issues

Closes #1271 

## Changes Made

- **Groq / OpenAI / HuggingFaceLLM**: add `generate_typed()` forwarding to the underlying provider, matching the existing five wrappers verbatim (signature, docstring, `is_available()` guard that raises `ProcessingError`).
- **LiteLLM**: implement `generate_typed()` directly — uses `instructor` when installed, otherwise a validate-and-retry loop on top of `generate_structured()` that feeds the Pydantic validation error back into the prompt and raises `ProcessingError` after `max_retries`.
- **Docs**: document the shared `generate` / `generate_structured` / `generate_typed` / `is_available` interface in the `semantica.llms` module docstring and the LLM integrations guide (previously `generate_typed()` was only shown for Anthropic).
- **Tests**: new wrapper test files for Groq, OpenAI, HuggingFaceLLM and LiteLLM (the four that had no dedicated coverage) — construction, delegation, arg
  forwarding, and the unavailable-provider error path.

## Testing

- [x] Tested locally

### Test Commands

```bash
pytest tests/test_llm_groq.py tests/test_llm_openai.py \
       tests/test_llm_huggingface.py tests/test_llm_litellm.py
# 75 passed (full tests/test_llm_*.py suite)

### DIFF
--- a/docs/guides/llm-integrations.md
+++ b/docs/guides/llm-integrations.md
@@ -3,7 +3,7 @@ title: "LLM Integrations"
 description: "Connect Semantica to Groq, OpenAI, Anthropic, HuggingFace, Novita AI, and 100+ LLM providers through a unified interface."
 ---
 
-Semantica exposes a unified provider interface — a single `.generate()` method — across Groq, OpenAI, Anthropic Claude, HuggingFace, Novita AI, and 100+ providers via LiteLLM. Use it when you need to swap providers for latency, accuracy, cost, or data-residency reasons without touching application code.
+Semantica exposes a unified provider interface — the same `.generate()`, `.generate_structured()` and `.generate_typed()` methods — across Groq, OpenAI, Anthropic Claude, HuggingFace, Novita AI, and 100+ providers via LiteLLM. Use it when you need to swap providers for latency, accuracy, cost, or data-residency reasons without touching application code.
 
 ## What Are LLM Integrations?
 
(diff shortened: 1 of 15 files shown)


Sameer6305 (dismissed):
Thanks for addressing the previous findings. The implementation looks good overall.

I have a couple of small follow-ups I'd like to see in this PR before approval:

1. Please add a test covering the LiteLLM `instructor` initialization failure → manual fallback → successful generation path. The implementation handles this correctly, but we should have a regression test for it.

2. Please update at least one wrapper delegation test to pass an additional kwarg (e.g. `temperature=0.5`) and assert that it is forwarded to the underlying provider. This will explicitly verify the `**kwargs` contract.

One additional consideration: the LiteLLM manual fallback retry loop currently doesn't sleep between retries, whereas the existing `BaseProvider` fallback does. Please check whether we should preserve the same retry/backoff behavior here. If not, I'm fine treating that as a follow-up.

The remaining observations are non-blocking and don't need separate issues/PRs.

Once the above small items are addressed, I'm happy to approve.

sakshi04-ui:
@Sameer6305 
Addressed all three follow-ups.

- **Instructor init-failure regression test.** Added `test_generate_typed_falls_back_when_instructor_init_fails`: instructor is importable but `from_litellm()` raises — asserts it falls through to the manual loop and still returns a valid result, rather than propagating the init error.
- **`**kwargs` contract test.** Extended Groq's `generate_typed` delegation test to pass `temperature=0.5` alongside `max_retries`, asserting both reach the underlying provider unchanged.
- **Retry backoff.** LiteLLM's manual fallback loop now sleeps 1s between non-final attempts, matching `BaseProvider`'s fallback behavior instead of hammering the API immediately. The retry test mocks `time.sleep` and asserts it fires once, so the suite stays fast.

LLM wrapper suite: 78 passing.

Sameer6305 (approved):
Looks good to me. All the requested changes have been addressed, and the tests are passing. This PR is ready to merge. Approved.

PR #1556, opened 2026-09-10
feat(context): filter retrieved context by active fact and source support

### LINKED ISSUE #1555: [FEATURE] Filter retrieved context against current truth-maintenance facts and source supports
## Problem Statement

`TruthMaintenanceSession` can remove a conclusion when it loses all support, but that does not invalidate a copy already stored as vector content or agent memory. A subsequent retrieval can still supply the obsolete conclusion to an agent.

Example:

1. Source `employment-v1` supports `Employed(Alice)`.
2. A rule derives `Eligible(Alice)`.
3. An application stores “Alice is eligible” as retrievable context.
4. The source is withdrawn; the session removes both facts.
5. The vector store still returns the cached sentence with a high relevance score.

Similarity is not a check of whether the sentence's supporting facts still hold. Removing graph records alone also does not remove copies in other stores.

There is a second, distinct case: another source keeps `Eligible(Alice)` supported, but a cached explanation still says “according to employment-v1.” The general conclusion can remain usable while that source-specific explanation must be excluded.

**Dependency:** the truth-maintenance session API proposed in #1542, currently under review in PR #1544. This issue builds on that API and does not assume anything beyond it.

Relevant code:

- `semantica/reasoning/truth_maintenance.py` (PR #1544): `facts`, `version`, `explain()` and `apply()` expose current logical state, without a retrieval integration.
- `semantica/context/context_retriever.py`: `retrieve()` gathers vector, graph and memory candidates, then merges/ranks them and selects top results.
- `_rank_and_merge()` merges metadata and, for non-graph results, currently deduplicates using the first 100 content characters. Applying a validity check only after this merge can lose dependency information.
- `TemporalGraphRetriever` filters temporal graph attachments. It does not connect stored text to source-support identities from the truth-maintenance session or rewrite stale prose.

This is an additive, explicitly enabled retrieval contract, not a change to the default behavior of existing RAG calls.

## Proposed Solution

Add an opt-in `TruthMaintenanceContextFilter` to `ContextRetriever.retrieve()`.

The caller annotates each managed context item with the facts and, optionally, source supports that its content requires. Retrieval checks these annotations against one immutable session snapshot **before ranking, merging or final top-k selection**.

### Proposed public use

```python
from copy import deepcopy
from semantica.context import ContextRetriever, TruthMaintenanceContextFilter
from semantica.reasoning import FactSupport, Rule, TruthMaintenanceSession

session = TruthMaintenanceSession(rules=[
    Rule("eligibility", "Eligibility", ["Employed(?x)"], "Eligible(?x)"),
])
session.apply(assertions=[FactSupport("employment-v1", "Employed(Alice)")])

# Attach this JSON-compatible metadata when storing the cached sentence.
metadata = {
    "truth_maintenance": {
        "schema_version": 1,
        "session_id": "employment-session-1",
        "required_facts": ["Eligible(Alice)"],
        "required_support_ids": [],
    }
}

gate = TruthMaintenanceContextFilter(session, session_id="employment-session-1")

class CachedVectorStore:
    def __init__(self):
        self.rows = [{
            "id": "cached-eligibility",
            "content": "Alice is eligible",
            "score": 0.9,
            "metadata": metadata,
        }]

    def search(self, *, query, limit):
        return deepcopy(self.rows[:limit])

store = CachedVectorStore()
retriever = ContextRetriever(vector_store=store, use_graph_expansion=False)
assert len(retriever.retrieve("eligibility", truth_filter=gate)) == 1

session.apply(retractions=["employment-v1"])
assert retriever.retrieve("eligibility", truth_filter=gate) == []
assert store.rows[0]["content"] == "Alice is eligible"
```

The example runs as written against a deterministic fake store — no external database, network access, or model download is involved.

### Dependency semantics

For an item `c` and snapshot `s`:

```text
usable(c, s) =
    annotation is valid and belongs to this session
    AND every required fact is in s.facts
    AND every required support ID is active in s
    AND every attached entity/relationship passes the same check
```

- Requirements within one item are **AND** requirements.
- Alternative derivations remain the session's responsibility. A conclusion-only item depends on the conclusion, not on every original supporting source.
- An explanation quoting a specific source additionally lists that source's support ID.
- Source withdrawal means “not currently supported,” not “proven false.”
- Dependencies must cover the entire returned text. The filter cannot discover omitted claims or check whether prose is entailed by the annotations.
- `session_id` is a caller-managed identity for one logical session incarnation, distinct from `support_id`. Use a new identity for an unrelated/recreated session. It is not authentication or a tenant-authorization mechanism.

### Scope and behavior

1. Add `TruthMaintenanceSession.snapshot()` returning a frozen `TruthMaintenanceSnapshot(version, facts, active_supports)`; collections and support records are immutable and detached.
2. Add the filter and a versioned, JSON-compatible dependency annotation format. Reuse the session's fact canonicalization; do not introduce an extraction model or another rule language.
3. Enable filtering per call with `retrieve(..., truth_filter=gate)`. No filter means existing behavior.
4. When enabled, reject unannotated, malformed, unsupported-version, wrong-session, or unsupported candidates by excluding the whole item. No permissive mode.
5. Graph bundles require a valid root annotation and valid annotations in every attached entity/relationship's `metadata`. If any attachment cannot be checked, exclude the whole bundle, including its prose. Do not just remove graph fields and leave stale text.
6. Validate raw candidates from all three sources before `_rank_and_merge()`. In this opt-in path, disable the existing lossy merge/deduplication; retain scoring and sorting. Duplicate valid results are allowed.
7. Add `metadata["truth_maintenance_validation"] = {"session_id": ..., "version": ...}` to detached accepted results. This is the version actually checked, not a timestamp or a new fact assertion.
8. Capture the snapshot at the start of the call and check the session version before returning. A change during retrieval raises `ProcessingError`, with no partial result or automatic retry. Access remains caller-serialized; this is not a thread-safety guarantee.
9. Existing source candidate limits remain bounded. Filtering can return fewer than `max_results`; do not refill with unchecked items or promise exhaustive top-k recall.
10. Support the direct `ContextRetriever.retrieve()` entry point and aliases that already delegate to it. In this PR, explicitly reject non-null `truth_filter` on `ContextRetriever.query_with_reasoning()`, `AgentContext.retrieve()` and `AgentContext.query_with_reasoning()` rather than silently ignoring it or promising coverage of their other paths.

The last boundary is intentional: `query_with_reasoning()` can construct additional graph paths after retrieval, and `AgentContext` has a separate memory-only route. Those require a later integration contract. No global validation mode or implicit filtering is introduced here.

The annotation must contain exactly the four fields shown above. Both requirement fields are lists of strings, not strings or null; at least one list must be nonempty. Duplicate requirements are harmless, fact strings are canonicalized, and support IDs must be nonempty with no surrounding whitespace.

## Acceptance Criteria

- [ ] A cached derived conclusion disappears from checked retrieval after its final supporting source is withdrawn, while the backing store still contains and returns the item.
- [ ] Removing only one of several supports/derivations keeps a conclusion-only item.
- [ ] Source-specific context disappears when its named support is withdrawn, even if its conclusion remains supported.
- [ ] Replacing a source in one batch, with no net fact change, retains the general conclusion and excludes the old citation.
- [ ] Multi-fact annotations require all facts; reactivating a support restores eligible candidates on the next call.
- [ ] Unknown facts/supports, wrong session IDs, empty dependency lists, invalid schemas and missing annotations are excluded.
- [ ] A graph bundle containing an uncheckable attachment is excluded in full; a fully annotated valid bundle remains usable.
- [ ] Invalid high-scoring candidates are removed before score normalization and final top-k; valid lower-scoring candidates already in the fetched pool remain eligible.
- [ ] Two candidates sharing a content prefix or graph ID cannot lose dependency metadata through the legacy merge in filtered mode.
- [ ] All results in a successful call bear the same checked version. Support-only changes during the call also cause `ProcessingError`.
- [ ] A failed session update or no-op leaves previously valid retrieval behavior intact.
- [ ] Filtering/ranking does not mutate store-owned candidate metadata, attachments or scores.
- [ ] Default calls preserve their current merge, scoring and retrieval behavior. The three explicitly unsupported entry points fail before retrieval or LLM invocation if passed a non-null filter.
- [ ] A deterministic integration test captures the context supplied to a fake agent/LLM before and after withdrawal, without an external database, network access or model download.

## Explicit Non-goals

- Automatically extracting memories from conversations or verifying natural-language claims.
- Updating `ContextGraph` nodes/edges from maintenance deltas, persisting an audit log, or changing graph retraction semantics.
- Deleting/re-embedding vectors, repairing all memory copies, or cross-store transactions.
- Time-driven expiry, historical `at_time` queries, interval propagation or bitemporal reasoning.
- Recursive rules, changes to the session's matching/retraction algorithms, new inference engines or confidence propagation.
- Full `AgentContext`, framework-adapter, decision-query or built-in reasoning-path integration.
- New LLM calls, runtime/test dependencies, background workers, automatic retries or concurrency guarantees.
- Evaluating the correctness of a model's answer after it receives checked context.

## Alternatives Considered

- **Physically delete all copies on every delta:** useful eventually, but requires store-specific identifiers, failure recovery and synchronization. Read-time validation establishes a smaller observable contract first.
- **Wrap the final top-k list:** smaller code change, but valid candidates may already have been discarded and metadata may already have been overwritten during merging.
- **Project all facts into a temporal ContextGraph now:** advances the larger roadmap, but combines graph identity, reactivation, history and failure atomicity with retrieval integration. Defer to a separate PR.

## Use Cases

1. An agent stops quoting an eligibility decision after its supporting employment fact is withdrawn.
2. Corrected documents replace old citations without unnecessarily suppressing conclusions that still have evidence.
3. Applications obtain checked context from vector or memory stores while physical cleanup is handled separately.

## Impact Assessment

- **Who benefits:** developers of changing, rule-backed agent knowledge and RAG context.
- **Priority:** Medium; an observable integration on top of the session API.
- **Breaking changes:** None for existing calls; new filtering behavior is opt-in and strict.
- **Dependencies:** the PR1 public session API (#1542 / PR #1544); Python >=3.9.2 and existing project dependencies.
- **Cost:** One snapshot copy per checked retrieval plus dependency checks and candidate copies. Do not claim cost proportional only to changed facts or an automatic latency improvement.

## Additional Context

#1542 / PR #1544 maintain the supported fact set. This issue checks explicitly annotated context against that set. Temporal support activation/expiry and graph/agent integrations are subsequent work, not implied by this issue.

I have a working implementation for this scope and would like to be assigned to this issue following CONTRIBUTING.md.

## Contribution

- [x] I'm willing to help implement this feature.
- [x] I can help with documentation.
- [x] I can help with testing.

cxzg007:
## Summary

Closes #1555.

Cached RAG content can outlive the evidence that originally supported it. This change adds opt-in dependency validation to `ContextRetriever.retrieve()`, so source withdrawal excludes unsupported conclusions and source-specific explanations before ranking, even while their vector or memory records remain stored.

Validation uses one immutable truth-maintenance snapshot per call. Filtered retrieval excludes unannotated bundles, preserves dependency metadata by skipping legacy merging, and rejects a session-version change before return. Full `AgentContext`, post-retrieval reasoning paths, temporal expiry and physical store cleanup remain outside this change.

What is validated is the usability of explicitly annotated candidates under the current rules and evidence — not the correctness of a model's answer, and not any automatic synchronization of memories across stores.

## Dependency

Stacked on #1544 (head `feat/truth-maintenance`). Only the five commits after `6b9ff17e` are new:

- `ac2e600d` feat(reasoning): expose immutable truth maintenance snapshots
- `e0397091` feat(context): validate context dependencies against truth snapshots
- `06c428eb` feat(context): filter grounded candidates before ranking
- `316dcc03` fix(context): reject unsupported truth-filter entry points
- `45aae98d` docs(context): demonstrate support-aware agent context retrieval

After #1544 merges, this branch will be rebased onto main; the resulting diff is otherwise unchanged.

## How it works

- `TruthMaintenanceSession.snapshot()` returns a frozen `TruthMaintenanceSnapshot(version, facts, active_supports)`; later `apply()` calls cannot mutate a snapshot already in hand.
- `TruthMaintenanceContextFilter(session, session_id=...)` binds one session incarnation. Stored items declare dependencies in JSON-compatible metadata:

```json
{
  "truth_maintenance": {
    "schema_version": 1,
    "session_id": "employment-session-1",
    "required_facts": ["Eligible(Alice)"],
    "required_support_ids": []
  }
}
```

- `retrieve(..., truth_filter=gate)`:
  - validates raw candidates from the vector, graph and memory sources **before** `_rank_and_merge()`;
  - disables the legacy lossy merge/deduplication in this opt-in path (duplicate valid results are allowed), so shared content prefixes or graph IDs cannot strip dependency metadata;
  - excludes any unannotated, malformed, wrong-session or ungrounded candidate entirely — including graph bundles whose root is valid but any attached entity/relationship is not, so stale prose never survives without its attachments;
  - stamps accepted results (deep copies; store-owned records are never mutated) with the actually-checked `session_id` and version;
  - re-checks the session version before returning; a mid-call change raises `ProcessingError` with no partial result.
- `ContextRetriever.query_with_reasoning()`, `AgentContext.retrieve()` and `AgentContext.query_with_reasoning()` raise `ValidationError` early when passed a non-null `truth_filter`, instead of silently ignoring it or implying coverage they do not have.

## Compatibility

- Calls without `truth_filter` keep their current merge, scoring and top-k behavior.
- No new runtime or test dependencies; the filtered path performs no LLM calls and creates no external clients.

## Evidence

- `pytest tests/reasoning/test_truth_maintenance.py tests/reasoning/test_truth_maintenance_differential.py tests/reasoning/test_truth_maintenance_snapshot.py -q` → **74 passed** (includes 8 new snapshot-isolation tests).
- `pytest tests/context/test_truth_maintenance_filter.py tests/context/test_grounded_context_retrieval.py tests/context/test_temporal_retriever.py -q` → **119 passed** (includes 78 new tests: 38 filter/entry-point rejection, 34 retrieval integration, 6 graph-retrieval alignment and malformed-metadata hardening).
- The integration suite includes a deterministic recorded-prompt test with a fake LLM and fake vector store: after withdrawing the final supporting source, the agent-visible prompt no longer contains the cached conclusion, while the backing store still returns the row unchanged.
- `pytest tests/context/test_agent_context_smoke.py` → 3 failures in this checkout, pre-existing on the base and unrelated to this change: verified by stashing this branch's changes and re-running the same file with identical failures.
- `git diff --check` → clean. The documentation example in `docs/reference/context.md` was executed directly as written (self-contained fake store, no network access or model download).

New tests total: 86.

## Limitations

- Opt-in and strict: unannotated or malformed candidates are excluded in filtered mode; there is no permissive fallback in this change.
- Dependencies must cover the entire returned text; the filter cannot discover omitted claims or check entailment of the prose.
- Caller-serialized access; no thread-safety guarantee.
- Physical deletion or re-embedding of stored copies, temporal expiry, full `AgentContext` and post-retrieval reasoning integration are follow-ups.

## Documentation

- `docs/reference/context.md`: new "Support-Aware Retrieval (`truth_filter`)" section with a self-contained runnable example, manual annotation semantics, source-specific citations, graph bundles, and the explicitly unsupported entry points.
- `docs/reference/truth_maintenance.md`: snapshot API and per-call snapshot copy cost, cross-linked to the context page.


ZohaibHassan16:
Hi @cxzg007,

I just merged #1544 

Could you please rebase this one to current main?

cxzg007:
Hi @ZohaibHassan16, rebased onto the latest main (`92c2578a`) and pushed the updated branch (`13df8cde`). The diff now excludes the already-merged #1544 changes, and GitHub reports no merge conflicts.

I resolved the retrieval-mode conflicts while preserving main's behavior. Non-local modes now explicitly reject `truth_filter` instead of silently bypassing validation, with regression tests added.

Validation: 413 related tests passed, and the standalone documentation example ran successfully. The PR description now reflects the rebase and documents the optional-dependency limitations reproduced on main. CI is running. Ready for another review!

ZohaibHassan16 (approved):
Good to merge, thanks for the implementation

PR #1564, opened 2026-09-10
fix(vector-store): train PQ indexes in FAISSIndexBuilder

### LINKED ISSUE #1560: [BUG] FAISSIndexBuilder.train_index silently skips PQ indexes
## Bug Description

Calling `FAISSIndexBuilder.train_index()` on a PQ index returns without an error, but the index stays untrained. Adding vectors then fails with `is_trained failed`.

Calling FAISS's `train()` directly on the same index with the same data works.

## Steps to Reproduce

1. Create a PQ index with `builder.build_index("pq", m=2, bits=2)`.
2. Call `builder.train_index(index, vectors)`.
3. Add vectors to the index. See the example below.

## Expected Behavior

After training, the PQ index should accept vectors.

## Actual Behavior

`index.index.is_trained` is still `False`, and `add_vectors()` raises a `RuntimeError`.

## Environment

- OS: Windows 11
- Python: 3.11.15
- Semantica: 0.7.0, built from commit `9d93a808406887b99a4eed9220494485f9bddafa`
- faiss-cpu: 1.12.0
- NumPy: 2.2.6
- Installed from source with the `vectorstore-faiss` extra in a fresh virtual environment.

Also reproduced with Python 3.13.14, faiss-cpu 1.15.0 and NumPy 2.5.3.

## Code Example

```python
import faiss
import numpy as np
from semantica.vector_store.faiss_store import FAISSIndexBuilder

faiss.omp_set_num_threads(1)
vectors = np.random.default_rng(42).random((256, 4), dtype=np.float32)
builder = FAISSIndexBuilder(dimension=4)
index = builder.build_index("pq", m=2, bits=2)
builder.train_index(index, vectors)
print("trained through builder:", index.index.is_trained)
try:
    index.add_vectors(vectors[:2], ids=["a", "b"])
except RuntimeError as exc:
    print("insertion failed:", exc)

# Training the same index directly works.
index.index.train(vectors)
index.add_vectors(vectors[:2], ids=["a", "b"])
print("native training:", index.index.is_trained, "count:", index.index.ntotal)
```

## Error Messages / Logs

Output, with the FAISS build path shortened:

```text
trained through builder: False
insertion failed: ... Error: 'is_trained' failed
native training: True count: 2
```

## Additional Context

The [training method](https://github.com/semantica-agi/semantica/blob/9d93a808406887b99a4eed9220494485f9bddafa/semantica/vector_store/faiss_store.py#L513-L518) returns early unless the index is an `IndexIVFFlat`. This skips `IndexPQ`, which also needs training.

## Possible Solution

Include PQ in the training path and add a test that trains a PQ index, inserts vectors and runs a search.

I'd like to fix this and add a regression test. Could you assign it to me?

Evanwang-3:
## Description

`train_index()` skipped PQ indexes, leaving them untrained and causing the next `add_vectors()` call to fail.

This adds `IndexPQ` alongside `IndexIVFFlat` in the training check. The regression test covers training, inserting vectors and searching the PQ index.

## Type of Change

- [x] Bug fix

## Related Issues

Closes #1560

## Testing

- [x] Regression test fails before the fix and passes afterward.
- [x] FAISS index and search schema tests: 32 passed, 11 deselected.
- [x] Source distribution and wheel built successfully from commit `33f8719c` with `python -m build --installer uv`.

Tests were run on Windows using a fresh `--basetemp` directory to avoid a permission error during pytest's default temporary-directory cleanup.

## Documentation

- [x] No documentation changes needed.

## Breaking Changes

No.

### DIFF
--- a/semantica/vector_store/faiss_store.py
+++ b/semantica/vector_store/faiss_store.py
@@ -514,9 +514,9 @@ class FAISSIndexBuilder:
         return FAISSIndex(index, self.dimension, index_type)
 
     def train_index(self, index: FAISSIndex, training_vectors: np.ndarray):
-        """Train index on sample vectors."""
-        if not isinstance(index.index, faiss.IndexIVFFlat):
-            return  # Only IVF indices need training
+        """Train IVF and PQ indexes on sample vectors."""
+        if not isinstance(index.index, (faiss.IndexIVFFlat, faiss.IndexPQ)):
+            return
 
         index.index.train(training_vectors.astype(np.float32))
 
--- a/tests/vector_store/test_faiss_index.py
+++ b/tests/vector_store/test_faiss_index.py
@@ -9,6 +9,7 @@ import pytest
 from semantica.utils.exceptions import ProcessingError
 from semantica.vector_store.faiss_store import (
     FAISSIndex,
+    FAISSIndexBuilder,
     FAISSStore,
     _metadata_path,
 )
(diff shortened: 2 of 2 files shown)


KaifAhmad1 (approved):
Verified the fix: `train_index()` now includes `faiss.IndexPQ` alongside `IndexIVFFlat`, matching the codebase's 4 supported index types (flat/ivf/hnsw/pq) — flat and hnsw don't need training, ivf and pq do, so this closes the gap completely. Confirmed the new regression test fails on `main` (is_trained stays False, add_vectors raises) and passes on this branch. Minimal, correctly scoped fix.

Issue #1557, opened 2026-09-10
[FEATURE] context: trust tiers for graph facts and tier-based retrieval filtering


pkupt:
## Problem Statement

Retrieval results carry a relevance score but no evidence-quality signal. `RetrievedContext` gives the agent text plus a similarity score, so a fact corroborated by three ingested sources and a singleton extracted at 0.6 confidence arrive looking exactly the same. The signals to tell them apart exist inside semantica: extraction confidence flows into graph node metadata (`graph_builder` persists `item.confidence`, default 1.0), and `kg.ProvenanceTracker` knows how many sources back each entity (`get_all_sources`). None of that reaches the retrieval boundary.

There is also a defaulting problem. The extraction schemas default `confidence` to 0.9 and `graph_builder` falls back to 1.0, so "no confidence signal" is indistinguishable from "high confidence" once it is stored. A consumer that trusts the raw field over-trusts absent data. #1285 makes the same point from the extraction side, moving unavailable entity confidence to `None`; this issue consumes that clearer signal at the retrieval boundary. If #1285 lands first, the material here is a direct consumer of it, not a duplicate.

## Proposed Solution

A trust tier layer on top of the existing signals, additive and computed at retrieval time:

- `TrustTier` enum: `gold`, `silver`, `bronze`, `quarantine`.
- A tier calculator that combines corroboration count (from `ProvenanceTracker.get_all_sources`) as the primary signal and extraction confidence as the secondary one. Corroboration leads because of the defaulting problem above: confidence alone cannot distinguish "0.9 because it was earned" from "0.9 because that is the schema default". Missing confidence degrades the tier rather than defaulting high.
- `filter_by_tier(min_tier, ...)` on `ContextRetriever`, alongside the existing `filter_by_score` / `filter_by_type` / `filter_by_date`.
- The tier attached to each result's `RetrievedContext.metadata`, so downstream consumers see it even without filtering.
- An optional `min_tier` option on the `query_graph` MCP tool's search mode, following the pattern of the existing optional `node_type` filter.

Computed for retrieved nodes only, at query time. Nothing is persisted, so tiers stay fresh as source sets change, and storage layouts are untouched.

## Alternatives Considered

- Persisting tiers at write time: rejected, a tier stored when the fact lands goes stale as corroboration and source credibility evolve. Nothing is persisted, so tiers stay fresh as source sets change.
[...]

PR #1583, opened 2026-09-11
feat(context): add trust tier calculation for graph facts

pkupt:
## Description

Phase 1 of #1557. Adds the evidence-quality grading primitives and nothing else: `TrustTier` and `TierCalculator` in a new `semantica/context/tiers.py`. No existing module imports them yet, so retrieval behaviour is unchanged.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Related Issues

Related to #1557. Phase 2 (`filter_by_tier` wiring) and phase 3 (MCP parameter, docs, changelog) come as separate PRs, so this one does not close the issue.

## Changes Made

- `TrustTier`: quarantine / bronze / silver / gold, ordered through `rank` and compared with `meets(min_tier)`; `downgrade()` steps one tier down for the phase 2 filter.
- `TierCalculator.calculate(corroboration_count, confidence)`: 2+ sources is gold, 1 source is silver, 0 sources with usable confidence is bronze, otherwise quarantine. Confidence below `min_confidence` (0.7) or missing pulls the result down one tier.
- `treat_as_missing`, default `1.0`: confidence values in this set count as absent rather than as measurements.
- 49 unit tests over ordering, threshold edges, degradation, placeholder confidence, messy input and config validation.

## Testing

```bash
pytest tests/context/test_tiers.py -q     # 49 passed
python -m doctest semantica/context/tiers.py
python -m flake8 --max-line-length=88 --extend-ignore=E203,W503 semantica/context/tiers.py
```

- [x] Tested locally

## Documentation

- [x] No documentation changes needed in this PR

The module docstring carries runnable doctests. User-facing docs and the changelog entry belong to phase 3.

## Breaking Changes

**Breaking Changes**: No

## Additional Notes

Points worth flagging for the review.

The issue names `kg.ProvenanceTracker.get_all_sources` for the corroboration count. That class is deprecated and raises a `DeprecationWarning` on call; `ProvenanceManager.get_all_sources` is its replacement, and that is what phase 2 will call. Neither is called from this PR.

Missing confidence degrades the tier, so a fact with 2+ sources and no usable confidence lands on silver. That follows the issue line "Missing confidence degrades the tier rather than defaulting high". `missing_confidence_degrades=False` keeps the corroboration-based tier for anyone who reads it the other way.

`treat_as_missing` exists because `semantic_extract.types.Entity` defaults `confidence` to `1.0`, and `graph_builder` keeps `1.0` for objects without the attribute. A fact nothing scored is therefore stored exactly like a fact scored perfectly, and a degradation rule that only recognised `None` would never fire on the extraction path. Treating `1.0` as absent fixes that. Pass an empty iterable to switch the substitution off, or include `0.9` to extend it to the schema default; `0.9` stays out by default because it would downgrade nearly every LLM-extracted fact.

One caveat for phase 2: `get_all_sources` returns one entry per entity on the upstream lineage chain, not one per document. A parent and its child extracted from the same document look exactly like two genuine sources. Counting unique `source` values before grading is what keeps a single document from pushing a fact to gold.


pkupt:
Merged main into the branch (92c2578a), no conflicts. The diff is unchanged: `semantica/context/tiers.py` and its tests. 49 tier tests pass locally, and CI is green again.

ZohaibHassan16 (approved):
Merging, thanks

PR #1586, opened 2026-09-11
fix(explorer): map enrich dedup pairs to the Entity Resolution shape

nilaymallikk:
POST /api/enrich/dedup serialized DuplicateCandidate fields verbatim (entity1/entity2/similarity_score) while the Entity Resolution tab parses entity_a/entity_b/similarity|score, so every pair rendered with empty ids and a 0% score bar and merge POSTed empty ids. Map candidates to the frontend-facing keys, keeping the legacy keys for compatibility.

### Test Commands

```bash
# Build the package
pip install build
python -m build

# Optional: Run your own tests
pytest tests/

# Optional: Format code
black semantica/
isort semantica/
```


Sameer6305 (changes requested):
@nilaymallikk Thanks for the fix! The implementation looks good overall. I just have two small test coverage requests before we can merge:

1. **Add a merge round-trip test**
   Since the reported issue includes merge requests being sent with empty IDs, could you please add a test that:

   * calls the dedup endpoint,
   * extracts `entity_a.id` / `entity_b.id`,
   * uses those IDs in the `/api/enrich/merge` request, and
   * verifies the merge succeeds (HTTP 200).

   This will ensure the full flow is covered and prevent regressions in the actual failure scenario.

2. **Strengthen the similarity assertions**
   In `test_dedup_pairs_use_frontend_shape`, could you explicitly assert that:

   ```python
   assert "similarity" in item
   assert item["similarity"] == item["similarity_score"]
   ```

   The current `> 0` check could still pass if `similarity` is missing and the fallback `score` is used.

Once these are covered, the PR should be good to go. Thanks!

Sameer6305 (approved):
Thanks for addressing the requested test coverage.

I’ve completed the final review of the PR. The wire-shape fix correctly maps `entity_a`, `entity_b`, and `similarity` while preserving the legacy fields for compatibility. The new merge round-trip test covers the reported dedup → merge regression, and the strengthened similarity assertions verify the frontend-facing contract.

The reported `test_ontology_subissue3.py` failures are also pre-existing and unrelated to this change.

**Approved — ready to merge.** 🚀

PR #1591, opened 2026-09-11
feat(semantic_extract): pluggable persistent backend for ExtractionCache

### LINKED ISSUE #1581: [FEATURE] Persistent, pluggable backend for ExtractionCache (survive process restarts)
## Problem Statement

`ExtractionCache` (`semantica/semantic_extract/cache.py`) already content-addresses extraction results: a SHA-256 key over `text` + parameters (including `provider`, `model`, and generation params such as `temperature`/`seed`), across the `entities`, `relations`, and `triplets` namespaces, with sensitive keys filtered out. It works well *within a process*.

But it is a purely **in-memory** `OrderedDict` LRU with a TTL (default 1h). It does **not survive a process restart**, and there is no on-disk/persistent cache anywhere in the package. So every fresh process — a CI job, a notebook kernel restart, a batch worker, or an extraction subprocess — re-pays *every* LLM extraction call for inputs it already extracted in a previous run. For LLM-backed extraction, those calls are the dominant cost, so the cache's value is lost exactly across the boundary (a new run) where it would help most.

## Proposed Solution

Introduce a small `CacheBackend` protocol (`get` / `set` / `clear` / `stats`) behind `ExtractionCache`, and:

1. Ship the **current in-memory LRU + TTL as the default backend** — zero behavior change for existing users.
2. Add an optional **persistent backend using only the Python standard library** (`sqlite3`, or a JSON/file store), keyed by the *existing* SHA-256 cache key.
3. Make it selectable via the existing config surface, e.g. `cache_backend: "memory" | "sqlite"`, `cache_path: <file>`.

Namespaces, TTL semantics, LRU eviction, and the sensitive-key filtering all stay intact — only the storage layer becomes pluggable.

## Alternatives Considered

- **External cache deps (`diskcache`, `redis`)** — rejected: adds a runtime dependency and/or infra. `sqlite3`/`json` from the stdlib covers the need with **zero new deps** (consistent with keeping `requirements-ci.txt` and the supply-chain surface stable).
- **Just raise `ttl` / `max_size`** — doesn't help: an in-memory cache dies with the process regardless of TTL.
- **Cache at the provider layer instead** — the extraction-level key already captures the semantically relevant axes (provider/model/generation params) and namespaces; a provider-level cache would duplicate that and miss the namespace structure.

## Use Cases

Describe specific use cases where this feature would be helpful:

1. CI extracts the same corpus on every run — with a persistent cache, unchanged inputs are free after the first run.
2. Iterative notebook/dev work — restart the kernel without re-paying LLM extraction.
3. Batch / multi-process workers sharing a single cache file.

## Impact Assessment

- **Who would benefit?** Anyone doing LLM-backed extraction repeatedly — CI, batch jobs, notebooks, and incremental pipelines.
- **Priority**: Medium.
- **Breaking Changes**: No — additive; the default remains today's in-memory cache.
- **Dependencies**: None new (stdlib `sqlite3` / `json`). Composes with, and is a building block under, the source-hash incremental re-extraction primitive proposed in the companion issue (result-level layer here; version/source-level layer there).

## Implementation Ideas

```python
# semantica/semantic_extract/cache.py
from typing import Protocol, Any, Optional

class CacheBackend(Protocol):
    def get(self, namespace: str, key: str) -> Optional[Any]: ...
    def set(self, namespace: str, key: str, value: Any, ttl: Optional[int]) -> None: ...
    def clear(self, namespace: Optional[str] = None) -> None: ...
    def stats(self) -> dict: ...

class InMemoryBackend:      # current OrderedDict LRU + TTL (default)
    ...

class SqliteBackend:        # stdlib sqlite3; one table (namespace, key, value, expires_at)
    ...

class ExtractionCache:
    def __init__(self, max_size=1000, ttl=3600, backend: "CacheBackend | None" = None):
        self._backend = backend or InMemoryBackend(max_size=max_size, ttl=ttl)
    # _generate_key(), get(), set() unchanged — they delegate to self._backend
```

## Additional Context

- **Related in-repo:** `ExtractionCache` (`semantica/semantic_extract/cache.py`); `config.py` (`enable_cache` / `cache_size` / `cache_ttl`); wiring in `semantic_extract/methods.py` (`_result_cache`).
- **Companion issue:** content-hash-driven incremental re-extraction & carry-forward (source/version-level; this issue is the result-level layer it builds on). 

## Contribution

- [x] I can help with testing

---

Besokus:
## Summary

Closes #1581.

`ExtractionCache` already content-addresses results (SHA-256 over text + params, incl. `provider`/`model`/generation params, with sensitive keys filtered), but it is a purely **in-memory** LRU + TTL — it does not survive a process restart. So every fresh process (CI job, notebook kernel, batch worker, extraction subprocess) re-pays *every* LLM extraction call it already made in a previous run.

This PR makes cache storage pluggable and adds an optional persistent backend, with **no behavior change by default**.

## Changes

- **`CacheBackend` abstraction** — `ExtractionCache` delegates storage to a backend, keeping stable key derivation and the public `get`/`set`/`clear`/`get_stats` API in one place. Backends only ever see the already-hashed key, never raw text or params.
- **`InMemoryBackend` (default)** — the historical `OrderedDict` LRU + TTL, behavior-identical; `._caches`/`._locks` are proxied for backward compatibility.
- **`SqliteCacheBackend`** — persistent, stdlib `sqlite3`; results survive restarts. TTL + LRU (by last access) mirror the in-memory backend. Pluggable serializer (`pickle` by default; pass e.g. `json` to avoid pickle — point the DB at a trusted, local path).
- **Config** — `cache_backend` (`"memory"` | `"sqlite"`) + `cache_path`; construction failure degrades gracefully to in-memory.
- New exports: `CacheBackend`, `InMemoryBackend`, `SqliteCacheBackend`.

## Design notes

- **Backward compatible**: default is the unchanged in-memory behavior; persistence is opt-in.
- **Zero new runtime dependencies** (stdlib `sqlite3`).
- Key policy stays centralized in `ExtractionCache` (incl. sensitive-key filtering); a model/provider change already varies the key, so a persistent cache stays correct across model upgrades.

## Tests

- New `tests/semantic_extract/test_cache_backends.py` (19 tests): in-memory parity (TTL, LRU, sensitive-key filtering, stats, disabled), sqlite round-trip, **persistence across a simulated restart**, TTL expiry, LRU bound, clear, custom JSON serializer, and graceful handling of unserializable/corrupt payloads.
- Existing cache-touching tests (`test_performance`, `test_groq_real_world`) and `test_schema_validator` pass unchanged.


Sameer6305 (dismissed):
@Besokus Thanks for the implementation. The overall `CacheBackend` design and SQLite persistence approach look good, but I found a few issues that should be addressed before merge.

### 🔴 Required changes

**1. Fix SQLite transaction/rollback handling**

`_safe_rollback()` can run after `self._lock` has been released. Since the same SQLite connection is shared, another thread can start/commit work while the failed transaction is still active. This can cause operations from a failed request to be committed by a later request.

Please keep rollback inside the same lock/exception scope as the failed SQLite operation.

The expired-entry cleanup path in `get()` should also explicitly rollback when its `DELETE` fails, so the connection is never left with an open failed transaction.

**2. Make SQLite cache-file creation atomic and secure**

The current flow checks the path and then calls `sqlite3.connect()`, followed by `chmod(0600)`. This leaves both:

* a TOCTOU/symlink race between validation and opening the file
* a brief permission window before `0600` is applied

Please create/open the cache file atomically with restrictive permissions from the start, and reject unsafe symlink/path cases. This is especially important because the default serializer is `pickle`; if an attacker can influence the cache DB, persisted bytes can reach `pickle.loads()`.

### 🟡 Required documentation/tests

Please add/update the documentation to clearly state the persistent cache trust model:

* `cache_path` / `SEMANTICA_CACHE_PATH` must point to a trusted, user-private location.
* The SQLite cache may contain sensitive extraction results.
* The default `pickle` serializer must only be used with a trusted cache file.

Please also add regression tests covering the transaction/rollback failures and unsafe cache-file/symlink/permission cases.

Finally, please update **`CHANGELOG.md`** with the new persistent cache backend, configuration options, and any relevant compatibility/security notes.

### 🟢 Nice-to-have, not blockers

The following can be addressed separately if preferred:

* validate deserialized cache values
* broaden sensitive-key filtering
* document the cache-key change/invalidation behavior
* use `>=` for `ttl=0`
* add an SQLite index for LRU eviction

**Request changes:** the required fixes above should be addressed before merge. After that, I’m happy to re-review.

Sameer6305 (approved):
Thanks for addressing the requested changes. I completed the final re-review, and the transaction handling, secure SQLite file creation, rollback paths, tests, documentation, and changelog all look good.

No remaining merge blockers from my side.

**Approved. ✅**

PR #1593, opened 2026-09-11
fix: resolve 23 pre-existing test failures across pipeline, Snowflake, Salesforce, and CLI

Sameer6305:
### Summary

This PR fixes 23 pre-existing test failures that were discovered during the Amazon Redshift PR review. The failing test count drops from **73 → 50**. No production logic is changed — all fixes are either corrections to test assertions that didn't match the actual implementation, or test isolation fixes where one test was polluting module state for later tests.

The remaining 50 failures are all environment-bound and cannot be fixed in code: missing optional dependencies (`fastapi`, `pyoxigraph`, `defusedxml`), tests making real network calls, a FAISS SIMD hardware constraint, a flaky Jena concurrency test, and stale top-level test copies whose structured counterparts under `tests/semantic_extract/` pass cleanly.

---

#### `semantica/pipeline/pipeline_builder.py`
- `add_step()` return-type annotation corrected to `PipelineStep` (was accidentally changed to `PipelineBuilder`; the implementation was always returning `PipelineStep`).
- `build()` gains an optional `validate=False` parameter so tests that need to construct intentionally invalid pipelines to exercise the validator can do so without triggering the built-in pre-build validation guard.
- `build_pipeline()` internal step-variable reference aligned with the correct `PipelineStep` return of `add_step()`.

#### `tests/test_pipeline_orchestration.py` (14 tests fixed)
- All `builder.add_step(...).build()` fluent-chain calls separated into two statements, because `add_step()` returns `PipelineStep`, not the builder.
- Validator tests pass `validate=False` to `build()` so they can construct invalid pipelines for external validation testing.
- `test_execution_engine_failure`: assertion changed from `result.metrics['steps_failed']` to `result.errors` — the execution engine returns a metrics-less `ExecutionResult` on the exception path.
- `test_pipeline_validator_missing_dependency`: assertion relaxed to case-insensitive `"missing"` substring match, aligning with the actual error message text.

#### `tests/test_snowflake_ingestor.py` (1 test fixed)
- `test_ingest_table_with_limit`: `SnowflakeIngestor` uses DB-API bind parameters (`LIMIT %s`), not literal interpolation. Assertion updated to check for the `LIMIT` keyword and verify `100` appears in the bound params tuple.

#### `tests/test_cli_commands.py` (1 test fixed)
- `test_migrate_refuses_unsupported_backend_pair`: Rich word-wraps the error message across terminal lines, breaking a single-string `in` check. Assertion now checks each of the three backend names individually, making it robust to line-wrapping.

#### `tests/test_issue_1513_slim_core.py` (1 test fixed)
- `test_salesforce_ingestor_package_import_missing_hint`: the test patched `SALESFORCE_AVAILABLE=False` and popped the `SalesforceIngestor` cache entry but did not clean up, leaving `_SalesforceAuthenticationFailed=None` in the module namespace for later tests in other files. Added post-test cache cleanup.

#### `tests/test_salesforce_ingestor.py` (6 tests fixed via autouse fixture)
- `_mock_simple_salesforce_if_needed` autouse fixture: when `simple_salesforce` is not installed the module-level sentinels (`_SalesforceAuthenticationFailed`, `_SalesforceError`, `SALESFORCE_AVAILABLE`) are `None`/`False` at import time. The fixture now also patches those sentinels inside the `sys.modules` mock context, so tests that import `_SalesforceAuthenticationFailed` directly from the module get a real callable stub class instead of `None`.

---

### Test results

| Metric | Before | After |
|---|---|---|
| Total failures | 73 | 50 |
| Fixed by this PR | — | **23** |
| Remaining (env/infra) | — | 50 |

### DIFF
--- a/semantica/pipeline/pipeline_builder.py
+++ b/semantica/pipeline/pipeline_builder.py
@@ -227,10 +227,10 @@ class PipelineBuilder:
 
         try:
             # Validate pipeline structure (skip when caller opts out)
-            self.progress_tracker.update_tracking(
-                tracking_id, message="Validating pipeline structure..."
-            )
             if validate:
+                self.progress_tracker.update_tracking(
+                    tracking_id, message="Validating pipeline structure..."
+                )
                 validation_result = self.validator.validate_pipeline(self)
                 if not validation_result.valid:
                     errors = validation_result.errors
--- a/semantica/pipeline/pipeline_builder.py
+++ b/semantica/pipeline/pipeline_builder.py
@@ -205,12 +205,16 @@ class PipelineBuilder:
         self.pipeline_config["parallelism"] = level
         return self
 
-    def build(self, name: str = "default_pipeline") -> Pipeline:
+    def build(self, name: str = "default_pipeline", validate: bool = True) -> Pipeline:
         """
         Build pipeline from configuration.
 
         Args:
             name: Pipeline name
+            validate: Whether to validate the pipeline structure before
+                building.  Set to ``False`` when you need a pipeline object
+                to pass to an external validator (e.g. to test validation
+                logic separately).
 
         Returns:
             Built pipeline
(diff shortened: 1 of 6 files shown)


ZohaibHassan16 (changes requested):
Just two quick things:

1. Wrap the `yield` in `test_salesforce_ingestor.py` (and `test_issue_1513_slim_core.py`) in a `try...finally` so the teardown dict pop still runs if a test fails mid-run, otherwise the mocks stay leaked in `semantica.ingest`.
2. In `test_semantica_ingest_imports_cleanly_without_lib`, add a `pytest.raises(ImportError, match=r"semantica\[db-salesforce\]")` check when accessing `pkg.SalesforceIngestor` so we verify the missing-dep error actually fires instead of just checking `__all__`.
3. Tiny nit: move `update_tracking("Validating pipeline structure...")` in `pipeline_builder.py:226` inside `if validate:`.

Push those tweaks and I'll merge this.

ZohaibHassan16 (approved):
LGTM

PR #1598, opened 2026-09-11
fix(context): forget(days_old=) deleted recent memories instead of stale ones

### LINKED ISSUE #1597: AgentContext.forget(days_old=) deleted my recent memories and kept the old ones
I was using `forget(days_old=90)` to clean up memories older than 90 days, and I noticed my recent memories were disappearing while very old ones stayed.

What I did:

```python
from datetime import datetime, timedelta

old_id = ctx._memory.store("old", timestamp=datetime.now() - timedelta(days=100))
new_id = ctx._memory.store("new", timestamp=datetime.now())

ctx.forget(days_old=90)
```

I expected the 100-day-old memory to be deleted and the fresh one to be kept. Instead it was the other way around the new memory was gone and the old one was still there.

nilaymallikk:
Closes #1597.

## Problem

`AgentContext.forget(days_old=N)` is documented as deleting memories older than N days, but it filtered on `start_date` — a lower bound in `_matches_filters` — so `clear_memory` purged memories *newer* than the cutoff and kept the stale ones.

## Changes

- `semantica/context/agent_context.py`: filter on `end_date` instead of `start_date`, with an aware-UTC cutoff matching the store's aware stamps. `clear()` delegates to `forget()` and is covered by the same fix.

## Testing

- New `tests/context/test_agent_context_forget.py`: old-deleted/new-kept round trip plus an all-recent boundary case. Both fail on the unfixed code, pass with the fix.
- Full `tests/context/` suite: 813 passed, 2 skipped.

### DIFF
--- a/semantica/context/agent_context.py
+++ b/semantica/context/agent_context.py
@@ -688,10 +688,10 @@ class AgentContext:
         if user_id:
             filter_dict["user_id"] = user_id
         if days_old:
-            from datetime import datetime, timedelta
+            from datetime import datetime, timedelta, timezone
 
-            filter_dict["start_date"] = (
-                datetime.now() - timedelta(days=days_old)
+            filter_dict["end_date"] = (
+                datetime.now(timezone.utc) - timedelta(days=days_old)
             ).isoformat()
         filter_dict.update(filters)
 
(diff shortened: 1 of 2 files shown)


Sameer6305 (approved):
LGTM. The filter direction and UTC-aware cutoff correctly fix forget(days_old=), and the relevant tests fail on the pre-fix implementation and pass with the changes. Existing context tests also pass.

PR #1603, opened 2026-09-13
fix(mcp): resolve nested metadata filtering in handle_query_decisions (#1247)

### LINKED ISSUE #1247: mcp query_decisions: category/outcome filters always drop every result (find_nodes() nests properties, filter checks top level)
**Version:** v0.6.7 (`ecb33a5b7d1c`)

**File:** `mcp/tools/decisions.py`, `handle_query_decisions()`, lines 63-68

```python
nodes = graph.find_nodes(node_type="decision")
decisions = list(nodes)[:limit * 5]
if category:
    decisions = [d for d in decisions if d.get("category") == category]   # always None
if outcome_filter:
    decisions = [d for d in decisions if d.get("outcome") == outcome_filter]  # always None
```

`ContextGraph` nodes serialize as `{"id": ..., "type": ..., "properties": {...}}` (see `ContextNode.to_dict()`, `semantica/context/context_graph.py:407-416`) — custom fields like `category`/`outcome` live under `properties`, not at the top level. So `d.get("category")` is always `None`, and any `query_decisions` call with a `category` or `outcome` filter returns zero results even when matching decisions exist.

**Repro:** record a decision with `category="foo"`, then call `query_decisions` with `{"category": "foo"}` → `{"decisions": [], "count": 0}`. Calling it with no filter returns the decision fine, confirming the filter itself is the bug.

**Suggested fix:**
```python
if category:
    decisions = [d for d in decisions if d.get("properties", {}).get("category") == category]
if outcome_filter:
    decisions = [d for d in decisions if d.get("properties", {}).get("outcome") == outcome_filter]
```

Found while installing Semantica locally as an MCP server for Claude Code — happy to submit a PR once assigned per CONTRIBUTING.md.

JohnnyWilson16:
## Description

Fixes decision query filtering in `handle_query_decisions` when filtering by `category` and `outcome`.

In `ContextGraph.find_nodes()`, node attributes and properties are merged into a nested `metadata` dictionary (`semantica/context/context_graph.py:2364`). `handle_query_decisions()` previously inspected `d.get("category")` and `d.get("outcome")` at the root dictionary level, causing all category and outcome filters to evaluate to `None` and silently drop 100% of candidate decisions.

## Type of Change

- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement

## Related Issues

Fixes #1247

## Changes Made

- Added `_get_decision_field()` helper to extract attributes across nested `metadata`, `properties`, and root keys.
- Updated `handle_query_decisions()` to use `_get_decision_field()` for `category` and `outcome` filtering.
- Forwarded `category` filter to `graph.find_similar_decisions()` when a natural language query is provided.
- Added unit test suite in `tests/test_mcp_decisions_query.py` covering field extraction, nested metadata filtering, intersection queries, and exception handling.

## Testing

- [x] Tested locally with pytest
- [x] Code style checks pass (`ruff check`, `ruff format`)

```bash
python3 -m pytest tests/test_mcp_decisions_query.py -v
# ============================== 7 passed in 0.07s ===============================

ruff check semantica_mcp/mcp/tools/decisions.py tests/test_mcp_decisions_query.py
# All checks passed!
```

## Checklist

- [x] I have added tests that prove my fix is effective


KaifAhmad1 (approved):
## Review

Verified this against the actual `ContextGraph` code (not just the diff) by pulling the branch into a local worktree and reproducing both the original bug and the follow-up issues.

**Original bug (#1247) — fixed correctly.** `find_nodes()` merges node `properties`/`metadata` into a single `metadata` dict on each returned node (confirmed in `_add_decision_to_graph` / `find_nodes`), and `_get_decision_field()` correctly reads `category`/`outcome` from there. Confirmed via `git show` that decision nodes really do store these fields as properties, so the non-query filter path is sound.

**Two issues flagged by the automated review were real, and are now fixed in the latest commit (`3ea7b64`):**

1. Outcome filtering on the natural-language query path was checking the wrong dict — `find_similar_decisions()`/`find_precedents_by_scenario()` return each match wrapped as `{"decision": {...}, "similarity": ...}`, but `_get_decision_field()` was called on the wrapper, not the nested decision. Any `query` + `outcome` combination silently returned zero results. Reproduced locally before the fix:
   ```
   handle_query_decisions({"query": "migrate db to postgres", "outcome": "approved", "limit": 5})
   → {'decisions': [], 'count': 0}   # despite a matching decision existing
   ```
   Fixed by unwrapping via `d.get("decision", d)` before field extraction.
2. The response `limit` was passed straight into `find_similar_decisions(max_results=limit)`, which sorts/truncates before returning — so the outcome filter, applied afterward, could drop valid matches ranked just below the cutoff. Fixed by over-fetching (`limit * 5`) whenever an outcome filter is present, then filtering and slicing to `limit`.
3. The original test suite mocked `find_similar_decisions` with the flat `find_nodes()` shape rather than the real wrapper shape, so it gave false confidence — it passed while the actual bug was present. Updated the mocks to the real shape and added a regression test (`test_query_outcome_filter_does_not_drop_lower_ranked_matches`) that fails against the pre-fix logic.

**Confirmed clean:**
- `pytest tests/test_mcp_decisions_query.py -v` → 8/8 passing
- `ruff check` / `ruff format --check` → clean
- All required CI checks (build, CodeQL, security-scan, GitGuardian, MSDO) are green on the latest commit

Nice catch on the original issue, and thanks for the responsiveness on the follow-up. Approving.

KaifAhmad1 (approved):
Approving — verified the follow-up fix (commit 3ea7b64) resolves both issues flagged in review (outcome-filter wrapper unwrapping and premature limit truncation), tests pass (8/8), ruff clean, and all required CI checks are green.

Issue #1478, opened 2026-09-05
[FEATURE] Implement true reasoning-engine dispatch for semantica reason run


Sameer6305:
## Summary

`semantica reason run --engine <engine>` currently supports only the engines that can be executed through `Reasoner.infer_facts()` (`rete` / `forward-chain`).

The other supported engine values are currently rejected with a clear message rather than being incorrectly reported as executed.

Implement true engine dispatch so `reason run` can execute each supported reasoning engine through its corresponding API.

## Current Behavior

`reason run` currently routes supported forward-chaining modes through:

```python
Reasoner.infer_facts(facts, rules)
```

Other engine values are rejected because the dedicated engines have different APIs and input requirements.

For example:

* SPARQL/Datalog use query/rule-oriented inputs
* abductive reasoning uses observations
* deductive reasoning uses `Premise` objects
* other engines have their own execution models

## Expected Behavior

`semantica reason run --engine <engine>` should:

1. Select the requested reasoning engine.
2. Convert/load the required inputs into the format expected by that engine.
3. Execute the corresponding implementation.
4. Normalize the result into the CLI's output format where possible.
5. Provide clear validation/errors when required inputs are missing or incompatible.

The existing `rete` / `forward-chain` behavior through `Reasoner.infer_facts()` should remain unchanged.

## Scope

Investigate and implement dispatch for the remaining reasoning engines supported by the CLI.

Pay particular attention to the different input models and APIs of each engine rather than forcing them through `Reasoner.infer_facts()`.

Add appropriate unit and CLI regression tests for each supported engine.

## Related

Follow-up to #1354 and PR #1372.

PR #1372 intentionally keeps this out of scope because its goal is to fix the broken CLI dispatch while keeping `reason run` behavior honest rather than claiming unsupported engines were executed.

sakshi04-ui:
I'd like to take this one (follow-up to #1354 / #1372). Could you assign it to me @Sameer6305  @KaifAhmad1 ?

**Where things stand:** after #1372, `reason_run` (`semantica/cli.py` L2271) only dispatches
`rete` / `forward-chain` through `Reasoner.infer_facts()`; `deductive`, `abductive`,
`datalog`, `sparql`, and `graph` raise `"not wired to 'reason run' yet"`. Goal here is
real per-engine dispatch without forcing everything through `infer_facts()`.

**Proposed approach:**

1. Add a dispatch table in `reason_run` keyed by `--engine`, one small adapter function
[...]

Issue #1354, opened 2026-09-01
[BUG] semantica reason run fails with 'Reasoner' object has no attribute 'run' (and store connect has a sibling dispatch error)


hanifjaafar:
## Summary

`semantica reason run` fails immediately:

```
'Reasoner' object has no attribute 'run'
```

The CLI dispatches to a method that does not exist on `Reasoner`.

## Reproduction

```bash
export GRAPH_STORE_DEFAULT_BACKEND=neo4j
export GRAPH_STORE_NEO4J_URI=bolt://127.0.0.1:7689
export GRAPH_STORE_NEO4J_USER=neo4j
export GRAPH_STORE_NEO4J_PASSWORD=<redacted>

semantica reason run
# -> Semantica error: 'Reasoner' object has no attribute 'run'
```

`semantica reason --help` lists `run` as a documented subcommand ("Execute a reasoning engine against the knowledge graph").

## Analysis

`Reasoner` exposes `infer_facts` (the v0.6.7 changelog notes a `reasoning_provenance.py` import fix that renamed `ReasoningEngine`/`infer` to `Reasoner`/`infer_facts`). The CLI's `reason run` path appears to have been missed by that rename.

## Related

`semantica store connect --backend neo4j` also fails on a dispatch error in the same family:

```
get_graph_store_method() missing 1 required positional argument: 'method_name'
```

Notably it fails **identically** with correct credentials, deliberately wrong credentials, and no configuration at all — so it never reaches a connection attempt, and cannot be used to diagnose connectivity. Happy to split that into its own issue if preferred.

## Environment

semantica 0.6.7 (tag), Python 3.12.14, macOS arm64, Neo4j 2026.07.1 Community, neo4j driver 6.2.0.

7487:
I'd like to take this on. Both dispatch failures trace to the CLI layer: `reason run` calls `Reasoner.run()`, which doesn't exist (the facade's API is `infer_facts(facts, rules)`), and `store connect` calls `get_graph_store_method(backend)` — the `(task, method_name)` method registry, not a backend factory — so it raises a TypeError before any connection attempt, which is why it fails identically regardless of credentials. I have a fix with regression tests ready and will open a PR for triage.

PR #1631, opened 2026-09-15
feat(cli): dispatch reason run to real datalog and graph reasoning APIs

sakshi04-ui:
## Description

Follow-up to #1478 (itself a follow-up to #1354 / #1372). `reason run --engine <engine>` currently only executes `rete`/`forward-chain` through `Reasoner.infer_facts()`; every other engine value fails with "not wired to 'reason run' yet." This PR wires two of the remaining engines — `datalog` and `graph` — to their real APIs. `deductive`/`abductive` are a separate PR (splitting per the issue discussion, both linked to #1478); `sparql` is flagged separately since it has no query-execution path to dispatch to at all.

## Type of Change

## Related Issues

Related to #1478
(Partial — this PR covers `datalog`/`graph` only; `deductive`/`abductive` land in a follow-up PR linked to the same issue.)

## Changes Made

- `reason run --engine datalog`: builds a `DatalogReasoner`, loads `--rules` as Datalog Horn clauses (`head(X) :- body(X).`), lowercases graph-store facts so they read as ground constants (Datalog treats a leading-uppercase argument as a variable), calls `derive_all()`, and reports only newly-derived facts.
- `reason run --engine graph`: adds a required `--query "<question>"` option, converts the graph store into the `{entities, relationships}` shape `GraphReasoner.reason()` expects, and returns its LLM-generated answer as a new `answer` field (this engine does NL reasoning over the graph, not fact derivation, so it doesn't fit the `inferred_facts` shape).
- `rete`/`forward-chain` path is unchanged.
- Added `_graph_store_as_context()` and `_run_reasoning_with_status()` helpers in `semantica/cli.py` to support the above without duplicating the graph-store-read logic or the spinner/quiet-mode branching three times over.

## Testing

- [x] Tested locally

### Test Commands

```bash
# Build the package
pip install build
python -m build

# Run the reasoning-specific tests
pytest tests/test_cli_commands.py -k "Reason or reason"

# Run the full CLI test suites
pytest tests/test_cli_commands.py tests/test_cli_foundation.py

# Run the other CI-required checks touched by this area (docs check, slim-core import)
pytest tests/test_docs_check.py tests/test_issue_1513_slim_core.py
```

Results:
- `pytest tests/test_cli_commands.py -k "Reason or reason"` → 24 passed (18 existing + 6 new)
- `pytest tests/test_cli_commands.py tests/test_cli_foundation.py` → 340 passed, 2 skipped (no regressions)
- `pytest tests/test_docs_check.py tests/test_issue_1513_slim_core.py` → 27 passed
- `python -m build` → builds cleanly; verified the built wheel installs in a fresh venv and `semantica reason run --help` shows the new `--query` option correctly

New tests added: datalog success (facts count + inferred-only facts, confirms base facts aren't echoed back as "inferred"), datalog rejecting rete-style `IF/THEN` rules cleanly (no traceback), datalog with no `--rules`, graph missing `--query`, graph dispatching through a faked `GraphReasoner`, and `deductive`/`abductive` still cleanly rejected (unchanged behavior, regression-guarded).

## Documentation

## Breaking Changes

**Breaking Changes**: No

`rete`/`forward-chain` behavior is byte-for-byte unchanged. `sparql`/`deductive`/`abductive` remain rejected exactly as before, with the same error/hint text.

## Checklist

## Additional Notes

`sparql` is intentionally **not** included here or in the planned `deductive`/`abductive` follow-up. While implementing this, I checked `SPARQLReasoner` directly: `execute_query()` raises `NotImplementedError` unconditionally (#1087 — "no triplet-store execution path exists yet"), and there is no `.query()` method on the class at all. The `reason query` command that the current rejection message points SPARQL users to calls a method that doesn't exist, so it's non-functional for real use today. This isn't a CLI-wiring gap like the other engines — it's a missing engine capability (no triplet-store query execution exists anywhere in the codebase) — so it doesn't belong in either of these two PRs. Flagged on #1478 ahead of this PR; happy to open it as its own issue once both PRs land.


Sameer6305 (changes requested):
Thanks for addressing the previous review findings. I re-checked the implementation and the current scope looks correct. I have two small test-coverage changes I'd like before merging.

### 1. Cover Graph relationship source/target resolution

**File:** `tests/test_cli_commands.py`
**Test:** `test_run_graph_dispatches_to_graph_reasoner`

The test currently verifies the relationship type and properties, but it doesn't assert that `start_node_id` / `end_node_id` are resolved to the expected node names.

Since `_graph_store_as_context()` contains new ID → name resolution logic, please add assertions such as:

```python
assert graph["relationships"][0]["source"] == "Alice"
assert graph["relationships"][0]["target"] == "Bob"
```

This ensures a regression cannot silently pass integer IDs into `GraphReasoner` and degrade the graph context sent to the LLM.

### 2. Add regression coverage for GraphReasoner generation failures

**File:** `tests/test_cli_commands.py`
**Test area:** `TestReason`

The new CLI code handles both GraphReasoner error paths:

```python
answer.startswith("Error: LLM provider not initialized")
answer.startswith("Error during reasoning:")
```

The existing test covers the first case, but there is no test for the `"Error during reasoning:"` path.

Please add a regression test where the fake `GraphReasoner.reason()` returns:

```text
Error during reasoning: connection timed out
```

and assert that:

```python
assert result.exit_code != 0
assert "Error during reasoning" in result.output
assert "Traceback" not in result.output
```

This protects the newly added behavior from regressing to an exit-0 result that incorrectly treats an engine failure as a successful `answer`.

### Scope

I don't think any production-code changes are required beyond the current implementation. In particular, I would keep:

* `deductive` / `abductive` deferred to the follow-up PR.
* `sparql` deferred because there is currently no underlying query-execution API.
* The existing `click.Choice` behavior unchanged.
* The current Datalog fact normalization unchanged.
* The existing `rete` / `forward-chain` path unchanged.

Once these two tests are added and the relevant suites remain green, this should be good to merge.

sakshi04-ui:
> Thanks for addressing the previous review findings. I re-checked the implementation and the current scope looks correct. I have two small test-coverage changes I'd like before merging.
> 
> ### 1. Cover Graph relationship source/target resolution
> **File:** `tests/test_cli_commands.py` **Test:** `test_run_graph_dispatches_to_graph_reasoner`
> 
> The test currently verifies the relationship type and properties, but it doesn't assert that `start_node_id` / `end_node_id` are resolved to the expected node names.
> 
> Since `_graph_store_as_context()` contains new ID → name resolution logic, please add assertions such as:
> 
> ```python
> assert graph["relationships"][0]["source"] == "Alice"
> assert graph["relationships"][0]["target"] == "Bob"
> ```
> 
> This ensures a regression cannot silently pass integer IDs into `GraphReasoner` and degrade the graph context sent to the LLM.
> 
> ### 2. Add regression coverage for GraphReasoner generation failures
> **File:** `tests/test_cli_commands.py` **Test area:** `TestReason`
> 
> The new CLI code handles both GraphReasoner error paths:
> 
> ```python
> answer.startswith("Error: LLM provider not initialized")
> answer.startswith("Error during reasoning:")
> ```
> 
> The existing test covers the first case, but there is no test for the `"Error during reasoning:"` path.
> 
> Please add a regression test where the fake `GraphReasoner.reason()` returns:
> 
> ```
> Error during reasoning: connection timed out
> ```
> 
> and assert that:
> 
> ```python
> assert result.exit_code != 0
> assert "Error during reasoning" in result.output
> assert "Traceback" not in result.output
> ```
> 
> This protects the newly added behavior from regressing to an exit-0 result that incorrectly treats an engine failure as a successful `answer`.
> 
> ### Scope
> I don't think any production-code changes are required beyond the current implementation. In particular, I would keep:
> 
> * `deductive` / `abductive` deferred to the follow-up PR.
> * `sparql` deferred because there is currently no underlying query-execution API.
> * The existing `click.Choice` behavior unchanged.
> * The current Datalog fact normalization unchanged.
> * The existing `rete` / `forward-chain` path unchanged.
> 
> Once these two tests are added and the relevant suites remain green, this should be good to merge.

Both added in 2efb4a8 — relationship source/target assertions in `test_run_graph_dispatches_to_graph_reasoner`, and the new `test_run_graph_surfaces_generation_failure_as_error` test for the second error path. No production code changes, as agreed. Full suite: 343 passed, 2 skipped.

Sameer6305:
@sakshi04-ui can you please fix the conflict .
i have reviewed the PR and its ready to merge.

sakshi04-ui:
Hey @Sameer6305 — before force-resolving this, wanted to flag some context on why it's conflicting.

PR 2 (deductive/abductive, #1632) was branched from this PR's branch and got merged into `main` first via squash-merge. Since PR 2's branch had all of this PR's commits as ancestors, `main` already contains every fix from this PR 

A normal merge of this branch into `main` would reintroduce already-superseded code and resurrect test methods that were intentionally replaced during PR 2's work (e.g. the old "deductive/abductive rejected" test, which is now false since both are wired up) — git's merge sees them as unrelated additions since the branches diverged around them, not as a deliberate replacement.

The only thing in this branch that isn't already in `main` is the last commit (`2efb4a8e` — the two tests you requested: relationship source/target resolution and the second `GraphReasoner` error-path test).

Would it make sense to cherry-pick just that one commit onto `main` instead of merging the whole branch? I tested it locally and it applies cleanly with no conflicts (suite green, 42 passed) — that would get you the two requested tests without dragging in the merge noise. If that sounds right, this PR could probably close without a merge once the cherry-pick lands. Let me know if you'd rather handle it differently.

Sameer6305:
@sakshi04-ui thanks for clarifying it i think it would be better to use cherry pick so we can land only intended changes into main.
Can you raise different PR for those cherry pick changes??
I will review and merge that PR.
You can close this PR.

sakshi04-ui:
@Sameer6305 
Closing this PR. Everything in it is already in main from #1632. #1660 adds the two tests that were asked for in review.

Sameer6305:
> @Sameer6305 Closing this PR. Everything in it is already in main from #1632. #1660 adds the two tests that were asked for in review.

Thanks @sakshi04-ui will review that PR soon

Issue #1386, opened 2026-09-02
[FEATURE] Advanced Semantic Layer: ERP-to-Knowledge-Graph pipeline (schema mapping → conflict resolution → ontology → SKOS taxonomy → SHACL → RDF → SPARQL)


KaifAhmad1:
## Problem Statement

Enterprise data — the thing most organizations actually want a knowledge graph *of* — lives in ERP/warehouse systems as rows and columns: Snowflake, Databricks, SAP, Salesforce. Semantica already ships ingestors for all four (`SnowflakeIngestor`, `DatabricksIngestor`, `SAPIngestor`, `SalesforceIngestor`, plus the generic `DBIngestor`, whose `get_database_schema()` even surfaces `foreign_keys`) — but nothing between "rows I pulled from a table" and `GraphBuilder.build(sources=[{"entities": [...], "relationships": [...]}])` turns a primary key into an entity id, a foreign key into a typed relationship, or a junction table into a relationship instead of a spurious third entity. Today that mapping has to be hand-written per table.

That gap is exactly what `cookbook/advanced/13_Manual_Ontology_Snowflake_Mapping.ipynb` papered over — hardcoded column→entity mapping, no cross-source conflict handling, no SHACL gate before data reached the store. It was removed from the cookbook today (commit `1d18755a`) rather than left as the reference example for this workflow.

The second half of the gap: real ERP data disagrees with itself. The same customer's address, credit limit, or account owner is rarely identical between a CRM extract in Snowflake and the master-data table in Databricks/SAP. `semantica/conflicts` (`ConflictDetector`, `ConflictResolver`) already implements exactly the resolution strategies this needs (voting, credibility-weighted, most-recent, etc.), but there's no cookbook example wiring it into an ingestion pipeline before the data becomes graph nodes — so today those contradictions just silently land in the graph as duplicate or overwritten facts.

[...]

PR #1640, opened 2026-09-16
feat(kg): add RelationalSchemaMapper for row-and-column sources

costajohnt:
Part 1 of #1386 (the core addition); the cookbook notebook is a separate PR once the mapper shape is agreed.

## What

`semantica.kg.RelationalSchemaMapper` turns rows from `DBIngestor`, `SnowflakeIngestor`, `DatabricksIngestor`, `PandasIngestor`, a pandas `DataFrame` or plain row dicts, plus a small schema spec, into the `{"entities": [...], "relationships": [...]}` shape `GraphBuilder` and `OntologyGenerator` consume:

- `entity_tables`: `{table: {"pk": column or [columns], "type": ClassName, "name": column (optional)}}`. One row becomes one entity; the primary key becomes the id, the other columns become properties (flat on the entity, which is where `ConflictDetector.detect_value_conflicts` reads them).
- `foreign_keys`: `[{"table", "column", "references": (table, column), "predicate"}]`. One foreign key becomes one typed relationship; `predicate` defaults to the referenced table's name. A null foreign key produces no relationship.
- `schema`: the dict from `DBIngestor.get_database_schema()`. Its `foreign_keys` (SQLAlchemy inspector dicts) are used for columns no explicit key covers. Note the ingestor does not record which table a constraint belongs to, so the mapper reads `table_name` when present and otherwise infers the owner from which mapped table has the column (never reading the referenced table's own key column as a self-reference). Adding `table_name` in `DBIngestor` would make this exact; I left the ingestor alone here to keep the PR focused.
- A table that is not an entity table and is held together by exactly two foreign keys is a junction table: each row becomes one relationship (extra columns become relationship properties), never a phantom entity. Any other unknown table raises.
- Every entity carries `source` flat (the key `ConflictDetector` reads) and both entities and relationships carry `metadata: {"source", "table"}`; `metadata` is a control field `OntologyGenerator`'s property inference skips, so the tags do not become datatype properties. Relationships keep `source` for their subject id.
- A column whose name clashes with a key the mapper sets or with an alias `GraphBuilder` promotes (`id`, `type`, `name`, `source`, `target`, `object`, `subject`, `*_id`, `metadata`) raises instead of being overwritten; upper-case ERP columns like `TYPE` are fine.
- Junction relationships run from the first foreign key listed to the second (explicit keys before schema-derived ones); the predicate is the second key's, then the first's, then the table name. Composite keys are joined with `|` with the separator escaped.
- Keys are normalised so the same row keeps one id whichever path it came in by: a pandas integer column with a gap is read back as float64, and `1.0` still maps to `"<Type>:1"`; NaN, NaT and NA become `None`.

Entity ids are `"<Type>:<pk>"`, so the same customer ingested from two systems of record maps to one id and the two copies group together for conflict detection (the shape the issue's example relies on).

## Tests

`tests/kg/test_schema_mapper.py`, 30 tests: entity shape and source tags, shared ids across sources, explicit foreign keys, junction tables (with and without a predicate), schema-derived foreign keys with and without an owning table, explicit keys overriding schema ones, the four ingestor result shapes plus a DataFrame, pandas gaps (NaN, NaT) and float-widened integer keys, reserved-column rejection, junction predicate precedence and direction, junction edge cases (null key, three keys, schema-only), a schema self-reference and the documented column-name inference limit, composite keys with the separator inside a value, rows without a key, unknown tables, config validation, that `metadata` is skipped by `PropertyGenerator`, and an end-to-end `GraphBuilder().build(sources=[mapped])`.

`pytest tests/kg` otherwise unchanged (the one pre-existing failure there needs the optional `gensim` extra). black and isort clean.

Closes nothing on its own; relates to #1386.


costajohnt:
Pushed 32c18ba for the bot findings 1, 3, 4, 5 and 7. Two left alone on purpose: without primary key metadata the mapper cannot tell a junction table from an ordinary two-key table, so the two-key heuristic stays as documented (happy to tighten it once the schema dict carries per-table primary keys), and the flat `source` is where `ConflictDetector` reads provenance and `test_source_field_still_inferred_as_business_attribute` from #1420 pins it as a business attribute, so it stays out of the control fields.

ZohaibHassan16:
Thank you, can you please resolve conflicts?

costajohnt:
Done, merged main in. The only conflict was CHANGELOG.md: since 0.7.0 shipped without this, I moved the entry under [Unreleased]. Schema mapper tests still pass (34/34).

ZohaibHassan16 (approved):
LGTM, merging

Issue #519, opened 2026-04-30
[FEATURE] Visual Ontology Editor, Drafts, Versions, and Change Proposals


KaifAhmad1:
## Subissue 2: Visual Ontology Editor, Drafts, Versions, and Change Proposals

- Parent: [Ontology Hub - Semantica Knowledge Explorer](./ontology-hub-parent-feature-request.md)

## Summary

- Add visual ontology authoring and governed schema-change workflows.
- Let users edit ontology classes, properties, individuals, restrictions, axioms, and metadata without hand-writing OWL or Turtle.
- Stage all edits as drafts, then submit them as proposals with impact analysis and SHACL pre-validation.
- Let reviewers approve, request changes, reject, comment, and publish approved changes.

## What This Improves

- Editing:
  - Current: ontology changes require technical OWL or Turtle work.
  - Improved: visual canvas supports schema editing in the Explorer.
- Safety:
  - Current: there is no clear staged-edit model in the UI.
  - Improved: edits become pending diffs and drafts before proposal.
- Review:
  - Current: no in-app proposal or review workflow.
  - Improved: proposals include impact analysis, SHACL pre-validation, inline comments, and review actions.
- Versioning:
  - Current: version history is not visible to Explorer users.
  - Improved: timeline shows drafts, published versions, authors, dates, and summaries.
- Publishing:
  - Current: publishing is not tied to review state.
  - Improved: approved proposals publish through `VersionManager.create_version()`.

## Scope

- Included:
  - Visual Ontology Editor tab.
  - `@xyflow/react` canvas.
  - Class and property context menus.
  - Selected element detail panel.
  - Debounced staged diffs.
  - Draft endpoint integration.
  - Version timeline.
  - Proposal modal.
  - Impact analysis.
  - SHACL pre-validation.
  - Proposal review panel.
  - Inline comments.
  - Approve, Request Changes, Reject, and Publish actions.
- Not included:
  - Registry and loader. Covered by Subissue 1.
  - Health, SHACL Studio, and alignments. Covered by Subissue 3.
  - Real-time collaborative editing.

## User Stories

- As an ontology author, I want to edit classes and properties visually without writing Turtle.
- As a schema owner, I want to submit a proposal, see impact analysis and SHACL pre-validation, and get review before publish.
- As a reviewer, I want a side-by-side diff and inline comments before approving or rejecting a proposal.

## Editor Requirements

- Use `@xyflow/react`.
- Render classes as nodes.
[...]

PR #1643, opened 2026-09-18
docs(ontology): document the draft, proposal and version flow

pkupt:
## Description

`docs/guides/ontology.md` covers generating, validating, growing and exporting an ontology, but says nothing about changing one through the draft → proposal → approval → publish flow. Those endpoints have been in the tree since #519; the only trace of them in `docs/` is a single row in the endpoint table in `reference/explorer.md`, so the reader has to open `semantica/explorer/routes/ontology.py` to find out how a schema change is supposed to be staged.

This adds a section covering the state machine, the four-step flow with request and response shapes, what `submit_proposal` checks, what `publish` applies and in what order, and version listing and comparison. Two entries are added to the Common Pitfalls section.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Related Issues

Related issue: #519 (the proposal/approval workflow this documents)

Closes #
Fixes #

## Changes Made

- New section "Drafting, Reviewing, and Publishing Ontology Changes" in `docs/guides/ontology.md`: the five states and what moves between them; the four steps with curl; the two advisory blocks (`impact_analysis`, `shacl_validation`) and when the SHACL check reports `skipped` or `error`; the order `publish` works in, and that it applies additions only; version listing and pairwise comparison.
- Two additions to Common Pitfalls in the same file: the flow holds its state in application memory, and a `#` in an ontology URI must be percent-encoded when that URI is a path segment.
- `docs/reference/explorer.md`: the ontology table listed `/api/ontology/proposals` as `GET/POST`. Only `GET` is mounted at that path; submission is `POST /api/ontology/propose`. Corrected, and a row for `PATCH /api/ontology/draft` added, since the flow starts there.

## Testing

- [x] Tested locally
- [ ] Package builds successfully (python -m build)

### Test Commands

    python docs_check.py

Every request and response in the new section was run against the app through `TestClient` before being written down: staging a draft, submitting it, listing and reading it, commenting, approving, publishing, having a second publish on an already-published proposal refused with `400`, listing versions, comparing two versions, and confirming a publish lands the added class as a graph node. The percent-encoding pitfall was checked both ways: encoded finds the draft, unencoded answers `200` with `[]`. The `PATCH`-not-`POST` claim on the draft endpoint is pinned by asserting `405` on a `POST` to the same path.

Docs-only change; no code paths are touched.

## Breaking Changes

No.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have updated the relevant documentation

## Additional Notes

`docs_check.py` passes its content checks locally. The last check, the Mintlify export, was not run here: it needs `npx mintlify@4.2.632` against a Node runtime Mintlify supports, which this environment does not have.

The nine checks that do run all pass:

- docs.json validity
- nav pages on disk
- Card hrefs
- stale repo URLs
- reference frontmatter
- banned class names
- Python 3.9+ syntax in code blocks
- module coverage in `index.md`
- JSX balance


KaifAhmad1 (approved):
Verified every claim against semantica/explorer/routes/ontology.py directly: all endpoint paths/methods, the state machine and the 400 gate on non-approved publish, the create_version-before-graph-write ordering, the additions-only publish behavior, the impact_analysis/shacl_validation response shapes, VersionCompareResponse fields, the in-memory app.state storage (no persistence, not shared across workers), and the percent-encoding silent-200-empty-list failure mode. The explorer.md table correction (GET-only /proposals, separate POST /propose) also matches the actual routes and is consistent with existing prose elsewhere in that file. CI is green. Approving.

PR #1655, opened 2026-09-18
fix(mcp): serialize Decision objects returned by get_causal_chain

### LINKED ISSUE #1651: [BUG] MCP get_causal_chain fails with "Object of type Decision is not JSON serializable" on any non-empty chain
Tested on `main` @ 7996d1ab (source install). Affects both MCP servers.

## Summary

`CausalChainAnalyzer.get_causal_chain()` returns `Decision` dataclasses. Both `get_causal_chain` tool handlers return that list unchanged. The tools/call dispatcher then `json.dumps` the result outside the handler's own `try/except`, so every chain that actually reaches a decision fails:

| Server | Handler returns the raw list | Serialized at | What the client gets |
| --- | --- | --- | --- |
| `semantica-mcp` (`semantica/mcp_server`) | [`__init__.py#L278`](https://github.com/semantica-agi/semantica/blob/7996d1ab4bd0c96ea7e3d91a4ce70fe5881d224b/semantica/mcp_server/__init__.py#L278) | [`#L912`](https://github.com/semantica-agi/semantica/blob/7996d1ab4bd0c96ea7e3d91a4ce70fe5881d224b/semantica/mcp_server/__init__.py#L912) | `-32603 Object of type Decision is not JSON serializable` |
| `semantica_mcp` package | [`tools/decisions.py#L264`](https://github.com/semantica-agi/semantica/blob/7996d1ab4bd0c96ea7e3d91a4ce70fe5881d224b/semantica_mcp/mcp/tools/decisions.py#L264) | [`server.py#L126`](https://github.com/semantica-agi/semantica/blob/7996d1ab4bd0c96ea7e3d91a4ce70fe5881d224b/semantica_mcp/mcp/server.py#L126) | `-32603 Method 'tools/call' failed (TypeError)` |

An empty chain serializes as `[]`. The tool therefore looks healthy on any graph without causal links, and breaks the first time it follows one.

The existing tests in `tests/test_mcp_decisions_causal_chain.py` (added for #781) mock the analyzer so it returns string chains. That is why they could not see this.

## Reproduce

```python
import json
from semantica import mcp_server

g = ContextGraph()
a = g.record_decision(category="c", scenario="Adopt HIPAA program", reasoning="r", outcome="approved", confidence=0.9)
b = g.record_decision(category="c", scenario="Require BAA", reasoning="r", outcome="approved", confidence=0.9)
g.add_causal_relationship(a, b, "CAUSED")
mcp_server._graph = g

print(mcp_server._handle({"jsonrpc": "2.0", "id": 1, "method": "tools/call",
                          "params": {"name": "get_causal_chain",
                                     "arguments": {"decision_id": b, "direction": "upstream"}}}))
# {'jsonrpc': '2.0', 'id': 1, 'error': {'code': -32603,
#   'message': 'Object of type Decision is not JSON serializable'}}
```

The same thing happens over stdio against the installed `semantica-mcp` with `SEMANTICA_KG_PATH` set. Record decisions in one process, link them with `add_relationship(type="CAUSED")` in another, then call `get_causal_chain` from a third.

## Fix

`Decision` already has `to_dict()`. I have a PR ready that converts each entry with it in both handlers. Entries without `to_dict()` are passed through unchanged, so the fallback-backend path and its tests stay as they are.

gyroscope1110:
Closes #1651

Both MCP servers returned the `Decision` objects from `CausalChainAnalyzer.get_causal_chain()` unchanged. The tools/call handler then failed to `json.dumps` them, so every non-empty chain came back as JSON-RPC `-32603`. An empty chain serializes fine, which is why this went unnoticed.

## Change

Each entry is now converted with the existing `Decision.to_dict()` in:

- `semantica/mcp_server/__init__.py` (the `semantica-mcp` console script)
- `semantica_mcp/mcp/tools/decisions.py`

Entries without `to_dict()` pass through unchanged, so the fallback-backend path keeps its current behaviour and tests.

## Tests

The tests use a real `ContextGraph` (A -CAUSED-> B -CAUSED-> C) rather than a mocked analyzer, and go through each server's tools/call handler, so they exercise the `json.dumps` that was failing:

- `tests/test_mcp_server_causal_chain.py` (new) covers upstream, downstream, entries being plain dicts, and an unlinked decision still returning `[]`.
- `tests/test_mcp_decisions_causal_chain.py` gains `TestCausalChainSerializationWithRealGraph`. The existing tests there mock the analyzer with string chains, which is why they couldn't catch this.

On `main` the five new non-empty-chain tests fail with `Object of type Decision is not JSON serializable`, and they pass on this branch. `pytest tests/test_mcp_*.py` gives 151 passed and 4 skipped (145 existing plus 6 new).


gyroscope1110:
**On the Qodo findings.** Both are valid. `to_dict()` copies `metadata` as-is, and `record_decision` accepts arbitrary values, so a chained decision with a datetime in its metadata still failed with `Object of type datetime is not JSON serializable`.

Fixed in 73698660. Both servers now go through the existing `serialize_decision()` (`default=str`) rather than calling `to_dict()` directly. Each test suite gains a chained decision with a datetime and a set in its metadata, and those tests fail against the previous commit. `pytest tests/test_mcp_*.py` gives 153 passed and 4 skipped.

Sameer6305 (approved):
Thanks for the solid contribution, @gyroscope1110! I reviewed the implementation and the follow-up changes, including the serialization edge cases raised by Qodo.

The PR is solid: it fixes the original `Decision` serialization issue across both MCP servers, handles non-JSON metadata through the existing `serialize_decision()` path, and adds meaningful end-to-end regression coverage with real `ContextGraph` objects.

Everything looks good from my side. **Approved — thanks for the thorough fix and tests!**

PR #1656, opened 2026-09-18
fix: align generate_structured return annotations

### LINKED ISSUE #1270: Align generate_structured() return annotations across LLM providers
### Description

Some LLM provider wrappers annotate `generate_structured()` as returning a `Dict`, while the underlying structured parsing behavior can produce broader JSON-compatible structures, including a top-level list.

This makes the public type annotations narrower than the actual supported runtime result.

### Suggested improvement

Review the return annotations across the LLM providers and align them with the actual structured-output contract.

In particular, consider using a broader JSON-compatible return type where both objects and arrays are supported.

### Why

This would make the public API typing more accurate and consistent across providers.

### Priority

Non-blocking / follow-up.

Sameer6305:
## Summary

Fixes #1270 by aligning `generate_structured()` return annotations with the actual JSON-compatible runtime behavior.

### Changes

* Updated affected backend providers from `-> dict` to `-> Union[dict, list]`.
* Fixed the `HuggingFaceLLMProvider` JSON fallback to handle top-level arrays.
* Fixed the `LiteLLM` fallback to handle top-level JSON arrays.
* Updated the test `MockProvider` annotation.
* Added coverage for list responses and array fallback parsing.
* Verified that the public LLM wrappers were already correctly annotated.

### Testing

* Targeted tests: **97 passed**
* Syntax checks: **all changed files passed**
* Full non-integration suite: **7062 passed, 50 failed**
* The baseline had **7050 passed, 51 failed**, so the remaining failures are pre-existing/unrelated.

Closes #1270.


Sameer6305:
@KaifAhmad1 @ZohaibHassan16 ready for review.

ZohaibHassan16 (approved):
LGTM. Merging it

PR #1662, opened 2026-09-18
test: guard lxml, GitPython and google-adk imports so collection succeeds on a clean [dev] install

### LINKED ISSUE #1661: test: pytest tests/ aborts collection on a clean [dev] install — 3 modules import lxml / GitPython / google-adk unguarded
## Summary

On a fresh `pip install -e ".[dev]"`, `pytest tests/` aborts during collection and runs **zero** of the 7,649 tests. Three test modules import an optional extra at module scope with no skip guard: `lxml` (`documents` extra), `GitPython` (`ingest-git` extra) and `google-adk` (`google-adk` extra).

This is the same shape as #1167 (fastapi), which #1232 fixed for 10 modules. Two of the three new cases were introduced when #1513 moved `lxml` and `GitPython` out of the core dependency set on 7 Sep; the third came in with the ADK integration (#1312).

## Reproduction

On `de2e200e` (main, 18 Sep), Python 3.12, clean venv:

```
$ pip install -e ".[dev]"
$ python -m pytest tests/ -q -p no:randomly
...
ERROR tests/ingest/test_notebook_02.py
ERROR tests/ingest/test_xml_ingestor.py
ERROR tests/integrations/google_adk/test_shared_graph.py
!!!!!!!!!!!!!!!!!!! Interrupted: 3 errors during collection !!!!!!!!!!!!!!!!!!!!
18 skipped, 3 errors in 5.62s
```

Each error is an `ImportError` at import time:

```
tests/ingest/test_notebook_02.py:9: in <module>
semantica/ingest/__init__.py:361: in __getattr__
E   ImportError: Repository ingestion requires optional dependency 'GitPython'. ...

tests/ingest/test_xml_ingestor.py:5: in <module>
semantica/ingest/__init__.py:367: in __getattr__
E   ImportError: XML ingestion requires optional dependency 'lxml'. ...

tests/integrations/google_adk/test_shared_graph.py:16: in <module>
E   ModuleNotFoundError: No module named 'google.adk'
```

## Cause

`semantica/ingest/__init__.py` lazy-loads `RepoIngestor` and `XMLIngestor` through `__getattr__`, which deliberately raises `ImportError` when the backing package is missing (`semantica/ingest/__init__.py:357-367`). That is the intended behaviour for library users, but two test modules import those names unconditionally at module scope:

- `tests/ingest/test_xml_ingestor.py:5` — `from semantica.ingest import (XMLIngestionData, XMLIngestor, ...)`
- `tests/ingest/test_notebook_02.py:15` — `RepoIngestor, CodeExtractor, GitAnalyzer,` inside the big `from semantica.ingest import (...)` block. Only `test_06_repo_ingestion` uses them; the other nine tests in the file are unrelated.

`tests/integrations/google_adk/test_shared_graph.py` is different: it already has the `require_adk` autouse fixture that every sibling file in that directory uses (`pytest.importorskip("google.adk")` unless the test name contains `missing_adk`), but line 16 does `from google.adk.events import Event` at module scope, which runs before any fixture can skip anything. The other four files in `tests/integrations/google_adk/` don't have a module-level `google.adk` import and collect fine.

## Why it's unambiguously a bug

- `#1513`'s stated goal is that the core install works without these packages, and CI's "Verify core-only package importability and slim behavior" step (`.github/workflows/ci.yml:117`) tests exactly that — but only for `tests/test_issue_1513_slim_core.py`, not the suite as a whole.
- The sibling tests already encode the intended idiom: `tests/ingest/test_parquet_ingestor.py:5` uses `pa = pytest.importorskip("pyarrow")` above its `semantica.ingest` import, and the four other `google_adk` test files use the `require_adk` fixture without a module-level ADK import.
- CI never sees this because it installs `requirements-ci.txt`, which pins `lxml==6.1.2`, `gitpython==3.1.61` and `google-adk==2.7.0` alongside everything else.

## Blast radius

**7,649 tests can't run** on a clean `[dev]` install — pytest's default is to stop after collection errors. With the three modules guarded, collection succeeds and the suite runs to completion (7,304 passed / 274 skipped / 228 failed / 140 errors in 70s on my machine — the failures are the fail-not-abort class, mostly other optional extras, and out of scope here).

## Suggested fix

Match the existing idioms, one file each:

1. `test_xml_ingestor.py` — every test needs lxml, so `pytest.importorskip("lxml")` above the imports with `# noqa: E402` on the imports that follow (the `test_parquet_ingestor.py` pattern).
2. `test_notebook_02.py` — drop `RepoIngestor, CodeExtractor, GitAnalyzer` from the module import (`GitAnalyzer` is unused) and import the two used names inside `test_06_repo_ingestion` after `pytest.importorskip("git")`, so the other nine tests keep running.
3. `test_shared_graph.py` — move `from google.adk.events import Event` into `test_shared_graph_event_and_knowledge_nodes`, the only test that uses it; the existing `require_adk` fixture then does its job.

Verified both ways: with `lxml`, `GitPython` and `google-adk==2.7.0` installed the affected tests pass (14 passed); without them they skip (7 skipped) and the rest of the suite collects.

Once collection works, `test_notebook_02.py` also surfaces 3 pre-existing failures (`boto3` and `sqlalchemy` missing — also moved to extras by #1513 — and a `MagicMock` type error in `test_03_web_ingestion`). Those are a different class of problem (fail, not abort) and I'd keep them out of this fix.

I'd like to take this on if the approach looks right — the change is ready.

dex0shubham:
Fixes #1661

## Problem

On a fresh `pip install -e ".[dev]"`, `pytest tests/` stops with `Interrupted: 3 errors during collection` and runs none of the 7,649 tests. Three test modules import an optional extra at module scope with no skip guard. Two of them (`lxml`, `GitPython`) became optional in #1513; the third (`google-adk`) always was. CI installs `requirements-ci.txt` with every extra pinned, so it never sees this. Full write-up with tracebacks in #1661.

## Change

One file each, matching the idiom already next to it:

- **`tests/ingest/test_xml_ingestor.py`** — every test needs lxml, so `pytest.importorskip("lxml")` above the imports, `# noqa: E402` on the three that follow. Same shape as `test_parquet_ingestor.py`.
- **`tests/ingest/test_notebook_02.py`** — only `test_06_repo_ingestion` uses `RepoIngestor`/`CodeExtractor` (`GitAnalyzer` was imported but unused). Import the two names inside that test after `pytest.importorskip("git")`, so the other nine tests in the file keep running.
- **`tests/integrations/google_adk/test_shared_graph.py`** — the file already has the `require_adk` autouse fixture its four siblings use, but a module-level `from google.adk.events import Event` ran before any fixture could skip. Moved the import into the one test that uses it; the fixture now does its job.

No source changes.

## Verification

Clean venv, Python 3.12, `main` at `de2e200e`.

Without the extras (the failing state on `main`):

```
$ python -m pytest tests/ -q -p no:randomly --co | tail -1
7649 tests collected in 3.47s

$ python -m pytest tests/ingest/test_xml_ingestor.py tests/integrations/google_adk/test_shared_graph.py \
    "tests/ingest/test_notebook_02.py::TestNotebook02DataIngestion::test_06_repo_ingestion" -q -p no:randomly
7 skipped in 0.13s
```

With `lxml`, `GitPython` and `google-adk==2.7.0` installed, to prove the guards don't hide anything:

```
14 passed in 2.66s
```

Full suite now runs to completion: 7,304 passed / 274 skipped / 228 failed / 140 errors. The failures are pre-existing and a different class (fail, not abort) — mostly other optional extras such as agno — and are out of scope here.

flake8 (`--max-line-length=88 --extend-ignore=E203,W503`) finding counts per file, `main` → this branch: 40 → 39, 0 → 0, 8 → 7.

## Note

Once `test_notebook_02.py` collects, it surfaces 3 pre-existing failures (`boto3`, `sqlalchemy` — also moved to extras in #1513 — and a `MagicMock` `TypeError` in `test_03_web_ingestion`). Left out deliberately; happy to take them as a follow-up.


KaifAhmad1 (approved):
Approving. Thanks for the clear write-up and the minimal, well-scoped fix.

**Verified**
- Reproduced the collection abort on `main` (`ModuleNotFoundError: google.adk` at `test_shared_graph.py:16` interrupts the whole session).
- On this branch, with lxml and GitPython shadowed to simulate a clean `[dev]` install, collection succeeds and each guard skips exactly what it should: the XML module, the five adk tests, and `test_06_repo_ingestion` only. With the extras present, the touched tests run normally, so nothing is hidden.
- `test_notebook_02.py`: `RepoIngestor` and `CodeExtractor` are used only inside `test_06_repo_ingestion`, and `GitAnalyzer` was unused. The remaining top-level imports resolve lazily through `semantica.ingest.__getattr__`, so they don't need their extras at collection time.
- `test_xml_ingestor.py`: the `importorskip` plus `# noqa: E402` form matches `test_parquet_ingestor.py`. No test in that file covers the missing-lxml path, so the module-level skip hides no coverage. That path is covered in `test_optional_imports.py`.
- `test_shared_graph.py`: moving `Event` into the one test that uses it lets the existing `require_adk` autouse fixture skip first. It was the only unguarded `google.*` import in that directory.
- CI is green (CodeQL neutral), no source changes.

**Non-blocking follow-up**
Now that `test_notebook_02.py` collects, it surfaces pre-existing failures that need other extras. I saw `test_08_database_ingestion` fail on missing sqlalchemy and `test_03_web_ingestion` fail as well. You mentioned boto3 too. Please take those as the separate PR you offered, with the same `importorskip` pattern, so a `[dev]`-only run is fully clean.

PR #1665, opened 2026-09-19
fix: prevent stale provider reuse when API credentials change

### LINKED ISSUE #1664: fix: prevent stale provider reuse when API credentials change
## Problem

`ProviderPool` caches providers using a key derived from the arguments passed to `create_provider()`. However, when an API key is supplied through an environment variable, that effective credential is resolved inside the provider constructor and is **not included in the pool cache key**.

This can cause a cached provider created with one credential to be reused after the environment credential changes.

For example:

1. `OPENAI_API_KEY` is set to credential A.
2. A provider is created without an explicit `api_key`.
3. The provider is cached under a key based only on arguments such as `model="gpt-4"`.
4. `OPENAI_API_KEY` changes to credential B.
5. The same provider configuration is requested again.
6. `ProviderPool` returns the cached provider that still uses credential A.

This creates incorrect credential isolation and can result in requests being made with a stale credential.

## Expected behavior

Provider cache identity should account for the **effective API credential** used by the provider.

Changing the environment-provided credential should not cause a provider initialized with the previous credential to be silently reused.

At the same time:

* Explicit `api_key` values must continue to produce isolated providers.
* Identical provider configuration and credentials should continue to reuse the cached provider.
* `use_pool=False` must continue to bypass the pool.
* API credentials must not be exposed in logs, error messages, or other unnecessary identifiers.

## Scope

Investigate and fix the provider-pool cache identity so that credential changes cannot result in stale provider reuse.

Add regression coverage for:

* environment API key changing from A → B;
* explicit API key A vs B remaining isolated;
* identical configuration continuing to reuse the same provider;
* `use_pool=False` continuing to bypass caching.

Keep the change minimal and preserve the existing public API.

Sameer6305:
## Summary

Fixes two ProviderPool correctness issues:

- Prevents stale provider reuse when environment-resolved API credentials change.
- Makes provider creation thread-safe to prevent duplicate instances during concurrent first-time access.

## Root Cause

`ProviderPool` previously generated cache keys from the raw kwargs passed by callers. When an API key was resolved from configuration/environment inside the provider constructor, that effective credential was absent from the cache key.

As a result, changing an environment API key could still return a provider initialized with the previous credential.

Provider creation was also unsynchronized, allowing concurrent callers to construct multiple instances for the same cache key.

## Changes

- Include the effective API credential in the provider pool cache identity.
- Preserve explicit `api_key` behavior.
- Keep non-API-key providers unchanged.
- Add locking around the provider check/create/store sequence.
- Synchronize `clear()` with provider creation.
- Preserve `use_pool=False` behavior.
- Add regression and concurrency coverage.

## Testing

- 20/20 new ProviderPool tests passed
- 8/8 existing performance tests passed
- 328 semantic-extract tests passed
- 11 tests skipped, matching baseline
- Concurrency test repeated 10 times with no failures
- No new failures compared with baseline

## Scope

No public API changes.


Sameer6305:
@KaifAhmad1 @ZohaibHassan16 ready for review.

ZohaibHassan16 (approved):
Good work, merging

PR #1666, opened 2026-09-19
Run the open_ontologies integration tests in CI (follow-up to #1166)

fabio-rovai:
#1166 merged a few minutes before I could push this onto it, so it comes as a follow-up. It is the CI half of that PR, plus one stale sentence.

### The twelve tests merged in #1166 do not run anywhere

`requirements-ci.txt` is generated by `uv pip compile pyproject.toml --extra all --generate-hashes`, and on the maintainer advice in #1166 the `open-ontologies` extra was deliberately kept **out of `all`**. So `open-ontologies-lite` is never installed by any job. There is also no `pytest tests/integrations/open_ontologies/` step in any workflow. `ci.yml` has exactly four pytest invocations, at lines 122, 123, 141 and 182, and none of them reaches that directory.

The tests are therefore not skipped, they are never collected, and the green checks on #1166 carried no information about the integration they were meant to cover. That is an awkward thing to leave on main for a feature whose entire argument is that a generator cannot check its own output.

### What this does

It follows the `explorer-extra` precedent already in `ci.yml` rather than moving the extra into `all`, which is the thing that was not wanted for dependency-audit reasons: a pinned, hashed requirements file plus a build step that installs it and runs the tests.

Resolved with `--constraint requirements-ci.txt`, so it shares 44 packages with the existing CI environment at identical versions, with zero conflicts, and **adds exactly one package**, `open-ontologies-lite`. It is the last step in the `build` job, so it cannot perturb anything ahead of it.

### The stale sentence

The test module docstring still said the documented path emits `<Acme Corp>`. That stopped being true when `RDFExporter` began percent-encoding a bare entity id into the semantica namespace: the canonical path now emits `<https://semantica.dev/ns#Acme%20Corp>`, which is a valid IRI. That change is exactly why the `UNSOUND` fixture was moved to an id that survives normalisation, so the prose above the fixture now describes what the fixture actually does.

### Verification

12 tests pass with the extra installed. The gate is also still falsifiable, which is the only thing that shows it can refuse: flipping the `fallback_on_custom_error` default at `semantica/export/methods.py:225` fails exactly one test, `test_registration_routes_export_rdf_through_the_gate`, logging the original #1108 symptom `falling back to default because fallback_on_custom_error was set`. Nothing else moves.

Worth knowing for anyone installing the extra locally: `open-ontologies-lite` 0.4.0 declared `mcp` as a core dependency, which pulls a starlette major that `fastapi>=0.109.2` rejects, so installing it uninstalls fastapi and every `semantica/explorer` import then fails. That is fixed on the library side in 0.5.0, which is why the floor is where it is.


fabio-rovai:
The `verify` check here is failing for a reason unrelated to this PR, and it will fail on every PR opened from now until someone re-pins.

`github/codeql-action` is pinned to `b96794f0`, a commit dated 9 Sep, and the `v4` tag has since been moved upstream. The pins that trip the check live in `codeql.yml`, `container-scan.yml`, `scorecard.yml` and `defender-for-devops.yml`. This PR touches three files, `ci.yml`, `.github/requirements/open-ontologies-extra.txt` and the integration test module, and adds no `uses:` line at all, so none of the flagged references are mine.

The last successful run of that workflow on main was 14 Sep, which brackets when the tag moved.

Worth knowing before re-pinning: the tag is moving actively rather than once. The job log at 11:28Z resolved `v4` to `1c5b6756`, and querying the ref a few minutes later returned `d8073367`. Pinning to whatever `v4` points at today will drift again, so it may be worth pinning to an immutable release SHA and updating the comment deliberately, which is what the check is asking for anyway.

Happy to send that as a separate PR if it is useful, since it is unrelated to this change.

ZohaibHassan16:
Thank you for opening this, can you please fix Qodo's findings @fabio-rovai

fabio-rovai:
@ZohaibHassan16 both addressed in 43db7a19. Both findings were right, and the second one is the more serious of the two.

**1. Stale packages (freshness).** The `open-ontologies` extra is deliberately outside `all`, so the existing "Verify requirements-ci.txt is up to date" step cannot see `.github/requirements/open-ontologies-extra.txt` at all. A `pyproject.toml` change would have left the integration tests running against a stale dependency graph with nothing to report it. Added the same constrained re-resolve and version-line diff for that file, with the flags matching the header of the committed lock, `--python-platform linux` included, so the two resolutions cannot legitimately differ and turn the gate into noise.

Checked in both directions rather than only the happy one: the committed file re-resolves today to an identical set of 45 pins, and mutating a single pin (`urllib3==2.7.0` to `2.6.0`) makes the diff fail as it should. A check that has never been shown to fail is not yet a check.

**2. Security scan bypass.** Correct, and it was a real hole: CI installs and imports `open-ontologies-lite`, but `pip-audit` only ever saw `requirements-ci.txt`, so a disclosed vulnerability in that package or in anything unique to its resolution could not fail the gate. Both locks are now audited and merged into the single `pip-audit-report.json` that every step downstream reads, so the validation guards, the allowlist, the count, the PR comment and the artifact upload are all untouched.

The merge deduplicates on lowercased name plus version and unions `vulns` by id. Both graphs resolve against the same constraint file and share most of their packages, so without that a single finding in a shared package would have been counted twice by the gate and listed twice in the PR comment, which is the "duplicating report data" case. An entry that pip-audit could audit is preferred over the same package reported with a `skip_reason`, so merging can never downgrade a real result into an unaudited one.

Evidence, since a merge step sitting in front of a security gate should not be taken on trust:

* Lifted the `jq` program verbatim out of the workflow and ran it over fixtures. A shared finding is counted once and not twice, an open-ontologies-only finding does reach the gate and can fail it, the `IGNORED_VULN_IDS` allowlist still applies to the merged report, and Guard 2's shape check still passes.
* Ran `pip-audit` against the hashed lock for real. It reads the file directly, resolves all 45 packages, reports no skips and no findings, so this adds coverage without a false failure and without needing the extra installed in that job.
* The 12 integration tests still pass, and flipping the `fallback_on_custom_error` default in `semantica/export/methods.py` fails exactly one of them, `test_registration_routes_export_rdf_through_the_gate`, logging the original #1108 symptom. That is the check that the gate this PR switches on is actually load-bearing.

One thing I deliberately did not widen this PR to cover: `explorer-extra-py311.txt` and `explorer-extra-py313.txt` have both of the same gaps. They are installed and exercised by CI, they are not in the freshness check, and they are not in the pip-audit scope. That is pre-existing rather than introduced here, so I have left it alone. Happy to send it as its own PR if you want it closed.

The `verify` check is still failing for the `github/codeql-action` v4 pin drift described above, which is unrelated to this change and will fail on every PR until someone re-pins. That one is also ready to go as a separate PR whenever it is useful.

ZohaibHassan16 (approved):
Thanks for a good follow up, merging it

